PR #462, opened 2018-11-15
Adding banner to link to service workies game

Meggin:
Rob, not sure if a banner will work in markdown file.

Should I try staging it and then revert commit if it looks terrible?

Seems like a good way to call it out though-- the service workies game, but only in the one guide for now.


Meggin:
How's this?

robdodson:
@Meggin do you think we could add an image as well?

Meggin:
So images made it into the repo-- and I can try and add some to this, but it's tricky to know what it'll look like without being able to stage. Would it be possible to merge this PR, and I will follow up with a separate one that's just about trying to add images? Might make sense to do this one in stages.

robdodson:
If you want to add an image I can stage this PR for you.

Meggin:
The pull request has to merge first, right? I can stage in devsite, but
only after commit.

I want to play with appearances a bit.

Maybe not a big deal. I can push something and we can iterate.

robdodson:
nope it doesn't have to be merged. I can pull the PR and deploy it to staging. If you want to add your images then we can stage and you can mess around with the presentation in devtools.

robdodson:
@Meggin friendly ping on this. If you want to add a link to an image in your markdown I can stage this PR.

PR #478, opened 2018-11-20
Remove in-page anchor for now.

kosamari:
how we parse markdown file seems to not include in-page anchor tag. Removing it for now. (see screenshot below, `{: #clickjacking }` should not be visible to readers)

CC: @robdodson  this is how I put anchor in `developer.google.com/web` to do "jump to __" link. (I don't know how it's called) Would be great if we can do the same for web.dev

https://web.dev/secure/same-origin-policy
same_origin_policy_ _ _web_dev

### DIFF
--- a/content/secure/same-origin-policy/index.md
+++ b/content/secure/same-origin-policy/index.md
@@ -19,7 +19,7 @@ expose everything on a user's browser.
 
 The same-origin policy prevents this from happening by blocking read access to
 resources loaded from a different origin. "But wait", you say. "I load images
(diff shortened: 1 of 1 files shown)


robdodson:
Ah yeah I think that's using Jekyll's liquid syntax (maybe?) and we're not using Jekyll at all, just straight markdown.

I think our parser generates an ID for headings so you can still do: https://web.dev/secure/same-origin-policy#what-is-permitted-and-what-is-blocked-
but it doesn't insert a little button to link to them or anything.

robdodson:

PR #482, opened 2018-11-20
measure: add info on what lh is

ebidel:


content/measure.html
```diff
@@ -59,7 +59,7 @@ <h1 class="hero-title">Measure</h1>
               <h2>Audit your web page</h2>
             </div>
             <p>
-              Provide any URL, and Lighthouse will run a series of audits generating a report of how well the page did.
+              Provide any URL, and web.dev will run a series of audits using <a href="https://developers.google.com/web/tools/lighthouse/" class="gc-analytics-event" data-category="web.dev" data-label="measure, Lighthouse link" data-action="click">Lighthouse</a>. We'll generate a report of how well you did.
```
TimvdLippe: Should this be a `class="external" target="_blank"` like the other audit guide links that link to `developers.google.com`?

content/measure.html
```diff
@@ -85,7 +85,14 @@ <h2>Learn how to fix it</h2>
 
   <section class="lh-categories-section">
     <div class="container">
-      <h2>What Lighthouse looks for:</h2>
+      <h2>What is Lighthouse?</h2>
+      <div class="what-is-lighthouse">
+        Lighthouse is an open-source, automated tool for improving the quality of your web apps.
+        It is integrated directly into the Chrome DevTools Audits panel, but can also be
+        run from the command line, as a <a href="https://www.npmjs.com/package/lighthouse" class="gc-analytics-event" data-label="measure, Lighthouse module lnk" data-action="click">Node module</a>,
```
TimvdLippe: Same comment on the links in this paragraph.

content/measure.html
ebidel: We tried that but there were too many external link icons.

PR #485, opened 2018-11-21
Give guides a meta description. Fixes #470.

### LINKED ISSUE #470: Content pages are missing their meta description
For example https://web.dev/fast/discover-performance-opportunities-with-lighthouse does not appear to have a description field so it doesn't have a descriptor in social sharing cards:

image

robdodson:
- Renames the `blurb` field in path pages to `description`.
- Makes guide/codelab pages look for a `description` YAML field.
- Adds a description to two sample guides to demonstrate.


samthor:
I'd just sanity-check `git grep blurb` that we haven't left any other references. Otherwise LGTM 👍

robdodson:
Checked and it didn't seem like there were any.

PR #494, opened 2018-11-28
Add meta descriptions to all files.

robdodson:


robdodson:
Adds meta description to all guides. Also wraps a few of bits to fit in the 80 char limit.

PR #496, opened 2018-11-30
Add travis.

robdodson:
Check that PRs don't break the build command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+node_js:
+  - "10.14.1"
 sudo: false
 dist: trusty
 cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: node_js
+node_js:
+  - "10.14.1"
```
ebidel: ">= 10"?

ebidel: To match package.json

robdodson: Sam had a thing that only worked on node v10.11 for a bit... let's see if that was fixed. I'll try v10

robdodson: using "10" seems to work. I don't think travis supports >= so i'll just go with 10 for now.

PR #499, opened 2018-11-30
Add a11y vid

robdodson:


robdodson:
I sent in a CL to add support for the `rel` param to the devsite-yt player.

PR #501, opened 2018-11-30
Re-doing service workies section

Meggin:
It's likely I will need to play with image appearance a good bit, but starting somewhere.


robdodson:
cool if we close https://github.com/GoogleChrome/web.dev/pull/462 ?

Meggin:
I think this might look OK. Not as bad as I thought in first go.

PR #505, opened 2018-12-02
Accessibility content review

kjue25:
Sweep of accessibility content (8 guides total)

### DIFF
--- a/content/accessible/control-focus-with-tabindex/index.md
+++ b/content/accessible/control-focus-with-tabindex/index.md
@@ -37,7 +37,7 @@ to improve the focusability of those controls.
 <div class="aside warning">
   If you don't see a focus indicator at all, it may be hidden by your
   CSS. Check for any styles that mention <code>:focus { outline: none; }</code>.
-  You can learn how to fix this in our guide on [styling focus](https://web.dev/accessible/style-focus).
+  You can learn how to fix this in our guide on <a href="https://web.dev/accessible/style-focus">styling focus</a>.
 </div>
 
 ## Insert an element into the tab order
@@ -134,7 +134,7 @@ method on it.
   Curious what those role="" attributes are for? They let you change the
   semantics of an element so it will be announced properly by a screen reader.
   You can learn more about them in our guide on
-  [screen reader basics](https://web.dev/accessible/semantics-and-screen-readers)</a>.
+  <a href="https://web.dev/accessible/semantics-and-screen-readers">screen reader basics</a>.
 </div>
 
 ## Keyboard access recipes
--- a/content/accessible/control-focus-with-tabindex/index.md
+++ b/content/accessible/control-focus-with-tabindex/index.md
@@ -47,8 +47,7 @@ Insert an element into the natural tab order using `tabindex="0"`. For example:
 <div tabindex="0">Focus me with the TAB key</div>
 ```
 
-The element comes into focus when the `TAB` key is pressed or when its `focus()`
-method is called.
+To focus an element, press the `TAB` key or call the element's `focus()` method.
 
 <div class="glitch-embed-wrap" style="height: 346px; width: 100%;">
   <iframe
--- a/content/accessible/control-focus-with-tabindex/index.md
+++ b/content/accessible/control-focus-with-tabindex/index.md
@@ -15,14 +15,14 @@ wf_blink_components: Blink>Accessibility
 # Control focus with tabindex
 
 Native HTML elements such as `<button>` or `<input>` have keyboard accessibility
-built-in for free. If you're building _custom_ interactive components, use
-`tabindex` to ensure that they're keyboard accessible.
+built-in for free. If you're building _custom_ interactive components, use the
+`tabindex` attribute to ensure that they're keyboard accessible.
 
 <div class="aside note">
   Whenever possible, use a native HTML element rather than building your
   own custom version. &lt;button&gt;, for example, is very easy to style and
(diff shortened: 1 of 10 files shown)


content/accessible/use-semantic-html/index.md
```diff
@@ -16,7 +16,7 @@ By using the correct semantic HTML elements you may be able to meet most or all
 of your keyboard access needs. That means less time fiddling with `tabindex`,
 and more happy users!
 
-## Keyboard support for free (and better on mobile)
+## Keyboard support for free (and better mobile experiences)
```
kjue25: Changed this because I wasn't 100 percent sure what was meant by "better on mobile" (but it felt like the parenthetical was also an object of "for" in the prepositional phrase).  Let me know if this isn't correct!

content/accessible/use-semantic-html/index.md
```diff
@@ -84,7 +85,7 @@ and `SPACE` to attempt to click on them.
 </div>
```
kjue25: The Glitch embed above is nice, but as a user, I'm more interested in seeing the code differences between the two (rather than just a button that works and a button that doesn't).  Or maybe add more explicit instructions to also try clicking them manually to see that the div button actually does work in some ways?

content/accessible/use-semantic-html/index.md
```diff
@@ -114,7 +115,7 @@ Some built-in elements, in particular `<input>`s, can be difficult to style.
 With a bit of clever CSS you may be able to work around some of these
 limitations. The (hilariously named) [WTFForms](http://wtfforms.com/) project
```
kjue25: What does "limitations" refer to here?  Maybe an example here would be helpful.

content/accessible/labels-and-text-alternatives/index.md
```diff
@@ -176,12 +179,12 @@ Similarly, any `frame` or `iframe` elements should have a `title` attribute.
 While an `iframe` may contain its own internal `title` element, a screen
```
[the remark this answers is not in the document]
robdodson: The iframe links to another html file, so you end up with something like this:

<iframe src="rob.com">
  #document
   <html>
   <head>
     <title>Welcome to Rob's site</title>

content/accessible/use-semantic-html/index.md
robdodson: lgtm

content/accessible/headings-and-landmarks/index.md
```diff
@@ -135,7 +135,8 @@ results of the "The page contains a heading, skip link, or landmark region"
 audit.
 
 <div class="note">
```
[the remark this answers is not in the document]
robdodson: It's not a huge difference but without `.aside` the link styles will be a little off. We should use `class="aside note"` everywhere.

kjue25:
I signed it!

content/accessible/control-focus-with-tabindex/index.md
```diff
@@ -47,8 +47,8 @@ Insert an element into the natural tab order using `tabindex="0"`. For example:
 <div tabindex="0">Focus me with the TAB key</div>
 ```
 
-The element can be focused by pressing the `TAB` key and by calling its
-`focus()` method.
+The element comes into focus when the `TAB` key is pressed or when its `focus()`
+method is called.
```
robdodson: @Meggin which of these two treatments do you prefer?

Live example: https://web.dev/accessible/control-focus-with-tabindex#insert-an-element-into-the-tab-order

I was wondering if it should be phrased like, "Press the TAB key to focus the element. Alternatively, call the element's focus() method"

content/accessible/labels-and-text-alternatives/index.md
```diff
@@ -52,13 +52,14 @@ DevTools Accessibility Reference
 </a>.
 </div>
 
-Whether you're looking at an `img` with `alt` text, or an `input` with a
-`label`, you'll soon discover that all of these things result in the same
-outcome: giving an element its accessible name.
+Whether you're looking at an `img` with `alt` text or an `input` with a
+`label`, all of these attributes result in the same outcome: giving an
```
robdodson:
```suggestion
`label`, all of these scenarios result in the same outcome: giving an 
```

suggest maybe using "scenario" because `label` is a tag not an attribute

content/accessible/what-is-accessibility/index.md
```diff
@@ -44,34 +44,34 @@ categories:
...
 descriptions of content and controls.
 
 ### Motor/dexterity
 
 Motor and dexterity impairments may affect a user's ability to use a mouse,
 touchscreen, or other pointing device. Some users may rely on alternative input
-devices to access content, such as a keyboard, head or eye tracking software,
-switch devices, sip and puff devices, or voice access.
+devices to access content. These devices might include a keyboard, head- or 
+eye-tracking software, switch devices, sip and puff devices, or voice access.
```
robdodson:
```suggestion
eye-tracking software, switch devices, sip-and-puff devices, or voice access.
```

content/accessible/control-focus-with-tabindex/index.md
Meggin: Hey, I like the second version-- active voice, 'Press the TAB key...'

kjue25: Updated!

robdodson:
Nice work, thank you @kjue25 !

PR #517, opened 2018-12-10
Remove our hardcoded twitter card paths

robdodson:
These should come from devsite now.

Not ready for merge yet. I need to test devsite's CL against this PR.


robdodson:
Verified devsite is properly adding these now.

Issue #460, opened 2018-11-15
Add author bio


robdodson:
Here's the design from our mocks. This should go on the guide pages.

![image](https://user-images.githubusercontent.com/1066253/48575060-ca000380-e8c5-11e8-938e-f54ca456d4db.png)

robdodson:
@TimvdLippe maybe this would be a good next thing to work on?

TimvdLippe:
Yes, will do!

robdodson:
This is fixed and live. Great work @TimvdLippe!!!

PR #525, opened 2018-12-17
Add author information to guide generator

### LINKED ISSUE #538: Add JS linting for build scripts
Should check for things like 2 space indent, etc.

TimvdLippe:
This is part of #460. It still requires styling in devsite as well as the relevant information of authors in this repository.
In terms of configuration, we require all guides to add an object with the relevant links and image link.
Then we need a separate folder with all profile images of the authors, or a generic one if they don't want to upload that.
For now, we ignore all author configuration that is just a string.

An alternative would be to have a separate configuration file where we store this information and then keep the author
configuration as a string, which is an index of that configuration. If we have several guides of the same author,
this would probably save us some copy-pasting. However, I am punting that to after the styling, to keep PRs easy.

Fixes #538


TimvdLippe:
Styles added in cl/225811439

TimvdLippe:
image

content/discoverable/fix-http-status-codes/index.md
```diff
@@ -1,7 +1,12 @@
 ---
 page_type: guide
 title: Fix unsuccessful HTTP status codes
-author: ekharvey
+author:
```
TimvdLippe: Need to revert this, but it gives you an impression of what information is required.

ebidel:
Nice! Definitely think we should setup a contributors.yaml now and reference author ids in the markup instead of hardcoding the profile metadata in articles. That'll save us down the road from doing lots of copy and pastes.

Here's the /web code:
https://github.com/google/WebFundamentals/blob/master/src/data/_contributors.yaml
https://github.com/google/WebFundamentals/blob/623a4c65da0b54730529211a509b2f7c8b0c512b/gulp-tasks/wfContributors.js

robdodson:
+1 to Eric's point. Doing a lookup in contributors.yaml sounds good.

TimvdLippe:
Do I just copy that file into this repository? Do I copy these images as well? Or do we extract this logic in a separate module that is used by both WebFundamentals and web.dev?

robdodson:
Good questions.

> Do I just copy that file into this repository?

We could probably copy over contributors.yaml but we should probably remove anyone who isn't a Googler because they haven't contributed to this project yet.

> Do I copy these images as well?

For the Googlers, yes.

> Or do we extract this logic in a separate module that is used by both WebFundamentals and web.dev?

I haven't read through it yet but I assume our build process and the web fundamentals one might be different. It may be easier to do our own variation of their wfContributors.js. @ebidel wdyt?

TimvdLippe:
@robdodson Fixed. PTAL!

lib/config/contributors.js
```diff
@@ -0,0 +1,49 @@
...
+    const hasPictureFile = await fs.exists(path.resolve(CONTRIBUTORS_PICTURES_LOCATION, author + '.jpg'));
+
+    authorInformation.picture = hasPictureFile ? author : 'no-picture';
+
+    return authorInformation;
+}
+
+module.exports = {
+    loadContributorFromConfiguration,
+};
```
ebidel: newline

lib/config/contributors.js
```diff
@@ -0,0 +1,49 @@
...
+const path = require('path');
+
+const CONTRIBUTORS_CONFIG_LOCATION = path.resolve(__dirname, '..', '..', 'contributors', '_contributors.yaml');
+const CONTRIBUTORS_PICTURES_LOCATION = path.resolve(__dirname, '..', '..', 'content', 'images', 'contributors');
+
+let authorsConfig;
+
+async function loadContributorFromConfiguration(author) {
+    if (!authorsConfig) {
+        const contributorsFile = await fs.readFile(CONTRIBUTORS_CONFIG_LOCATION, 'utf8');
```
ebidel: 2 space indents

lib/deps/fsp.js
```diff
@@ -24,11 +24,13 @@ const mkdirp = require('mkdirp');
  * Promisifed versions of Node's filesystem methods.
  */
 module.exports = {
+  ...fs,
```
ebidel: what's this for?

lib/config/contributors.js
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+'use strict';
+
+const YAML = require('yaml').default;
+const fs = require('../deps/fsp.js');
+const path = require('path');
+
+const CONTRIBUTORS_CONFIG_LOCATION = path.resolve(__dirname, '..', '..', 'contributors', '_contributors.yaml');
```
robdodson: Do you need to separate all of these or is it possible to do:

```js
path.resolve(__dirname, '../../contributors/_contributors.yaml')
```

lib/config/contributors.js
robdodson: let's see if we can add some linting to enforce this.

robdodson: https://github.com/GoogleChrome/web.dev/issues/538

lib/config/contributors.js
```diff
@@ -0,0 +1,49 @@
...
+const CONTRIBUTORS_PICTURES_LOCATION = path.resolve(__dirname, '..', '..', 'content', 'images', 'contributors');
+
+let authorsConfig;
+
+async function loadContributorFromConfiguration(author) {
+    if (!authorsConfig) {
```
robdodson: Do you need this check? Will this always be undefined when this function is called?

lib/config/contributors.js
TimvdLippe: That would not work on Windows. This is the OS cross-compatible way of doing directory traversal.

lib/config/contributors.js
TimvdLippe: It's to lazily compute the map, as `fs.readFile` is async. The assignment is on line 31.

lib/deps/fsp.js
TimvdLippe: Initially for debugging, as this now exposes all of `fs` apis in `fsp.js`. E.g. some of the methods are patched to be promisified, but you can still call all other methods. Will revert for now, because I am no longer using any non-promisified version.

lib/config/contributors.js
TimvdLippe: Yeeee for writing code on my Mac and not on my linux system 🎉 😢 Updated my VSCode to 2 spaces and fixed indentation here.

lib/config/contributors.js
TimvdLippe: Done

lib/config/contributors.js
robdodson: ah good point!

lib/config/contributors.js
robdodson: ah got it. I didn't realize this function was being called over and over and preserving the state.

templates/guide.html
```diff
@@ -38,15 +38,15 @@
           <div class="web-author-information__details">
             <span class="web-author-information__written-by">Written by</span>
             <span>{{author.name.given}} {{author.name.family}}</span>
-            <div class="web-author-information__links"></div>
+            <div class="web-author-information__links">
               {{#if author.glitch}}
-              <a class="web-author-information__link" href="{{author.glitch}}">Glitch</a>
+              <div class="web-author-information__link"><a href="{{author.glitch}}">Glitch</a></div>
```
robdodson: Do you have a screenshot of the before and after? Just curious why you needed the extra div.

TimvdLippe: When hovering over a link, the dot would also receive a big (ugly) underline. This was because the link was display flex. Not making it flex meant that the text was not centered. The eventual solution was to make a wrapping div which is flex and give that the dot.

TimvdLippe:
@robdodson I had to make several updates to the linter and the gulpfile to make Travis happy. Not sure why this wasn't an issue before.

Therefore, I added `npm run format` which runs Prettier to fix the formatting. I also fixed `npm run build` to not fail with a stream error on recent versions of Node. PTAL

robdodson:

PR #539, opened 2018-12-21
Add eslint

robdodson:
- Checks can be run with `yarn lint`.
- Fixes all lint errors. There were a lot :P
- Add VS Code workspace settings. If you're using VS Code, you can fix lint errors with the Prettier extension and the included settings.
- Add a git pre-push check to verify lint rules are passing.
- Add linting step to travis builds.
- Add .editorconfig to enforce consistent spaces and line endings.
- Also, removes package-lock.json. Going all in on yarn!


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "prettier.eslintIntegration": true
+}
```
ebidel: newline

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "web.dev",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
ebidel: big time stuff

gulpfile.js
```diff
@@ -36,4 +37,17 @@ gulp.task('watch', gulp.series('build', () => {
   }, gulp.series('build'));
 }));
 
+gulp.task('lint', () => {
+  return gulp.src(['{lib,server}/**/*.js'])
```
ebidel: server/index.js didn't have any updates after linting? I'm surprised.

robdodson:
Longer explanation of the linting features:

**Check if there's a linter error**
`yarn lint` will run eslint in gulp and log errors to the console. It will exit(1) if there are any failures.

**See and fix errors while you work**
If you would like to actually fix the lint failures in the project as you code you can use Prettier. This step is optional, but it's nicer than running `yarn lint` over and over and fixing everything by hand.

I've setup ESLint to use the Prettier plugin, so if you have VS Code you can install the [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) and [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extensions, then format your document with option+shift+f. You can also add a user setting to have it format on save if you want. `"editor.formatOnSave":true`.

I've checked in the VS Code workspace settings so if you already have prettier and eslint extensions installed, formatting the document with the keyboard shortcut should "just work".

**Don't wait on Travis to tell you about a lint failure**
The husky pre-push hook will run `yarn lint` and block the push if there are lint failures. This is nicer than pushing, submitting a PR, and waiting on Travis to tell you that you forgot a comma.

Finally, as a last check, Travis _will_ run `lint` and `build` to verify everything is A-OK before something lands.

gulpfile.js
robdodson: aaaaah.... they're `.mjs`.

let's see...

ebidel: ah yea :)

TimvdLippe:
I don't run Yarn on corp and would like to stick to npm for that. Will see if I run into lock issues.

PR #551, opened 2019-01-09
contributors.js: use correct filename for missing author image

ebidel:
The file name is: https://github.com/GoogleChrome/web.dev/blob/master/content/images/contributors/no-photo.jpg


TimvdLippe:
Oh whoops. Thanks for the fix :smile:

PR #552, opened 2019-01-09
preview server: image tests

ebidel:


package.json
```diff
@@ -42,6 +44,7 @@
     "gulp-eslint": "^5.0.0",
     "gulp-watch": "^5.0.1",
     "husky": "^1.2.1",
+    "mocha": "^5.2.0",
```
ebidel: not tied to mocha...

server/test/test.js
```diff
@@ -0,0 +1,97 @@
...
+        assert.ok(false, `${resp.url()} was missing.`);
+      }
+    }
+
+    assert.ok(true);
+  }
+
+  it('renders local images on /', async function() {
+    await verifyImagesLoad(`${SERVER}/`);
+  }).timeout(10 * 10000);
```
robdodson: I think you can do `this.timeout(10 * 10000);` in the `describe()` block and it will apply to all of these tests.

server/test/test.js
```diff
@@ -0,0 +1,97 @@
...
+
+  it('renders local images on /measure', async function() {
+    await verifyImagesLoad(`${SERVER}/measure`);
+  }).timeout(10 * 10000);
+
+  it('renders local images on page 1', async function() {
```
robdodson: Might want to test a page like `/fast`. I think these have a header image.

server/test/test.js
```diff
@@ -0,0 +1,97 @@
...
+    await verifyImagesLoad(`${SERVER}/installable/add-manifest`);
+  }).timeout(10 * 10000);
+
+  it('renders local images on page 2', async function() {
```
robdodson: Maybe it could also say "renders images on /fast" and then for the guides it could say "renders images on guide 1", etc. Just to create a bit of distinction between path and guide pages.

server/test/test.js
```diff
@@ -0,0 +1,97 @@
...
+  }).timeout(10 * 10000);
+
+  it('renders local images on page 3', async function() {
+    await verifyImagesLoad(`${SERVER}/fast/use-imagemin-to-compress-images`);
+  }).timeout(10 * 10000);
+
+  it('renders local images on page 4', async function() {
+    await verifyImagesLoad(`${SERVER}/reliable/runtime-caching-with-workbox`);
+  }).timeout(10 * 10000);
+});
```
robdodson: Need a test for /about

server/test/test.js
```diff
@@ -0,0 +1,97 @@
+/* eslint-env node, mocha */
```
robdodson: What do you think about moving this into `test/images.js` and requiring it in `test/test.js`. In case at some point we want to test other parts of the site?

server/test/test.js
ebidel: Ugh. Guess I can write a traditional `function` instead of an arrow. Done.

server/test/test.js
ebidel: Done

server/test/test.js
ebidel: Done

server/test/test.js
ebidel: done

server/test/test.js
ebidel: yea cool.

ebidel:
PTAL

robdodson:
shipit!

PR #563, opened 2019-01-14
Add the ability to mark a guide as "draft"

### LINKED ISSUE #419: Add publish status to guide yaml
Sometimes it's nice to be able to check a guide or codelab into the repo without it actually being shipped. Or sometimes something may already be published and you'd like to unpublish it without removing it from the repo. I'm currently in that position with a few of the codelabs that still need more work and might not make the launch cutoff.

Post-launch it might be worth adding a `status: {draft|published}` field to the YAML front matter.

TimvdLippe:
If a guide has `web_status: draft`, do not generate a guide file.

image

Fixes #419

### DIFF
--- a/lib/config/generators.js
+++ b/lib/config/generators.js
@@ -25,6 +25,8 @@ const templates = require('../templates.js');
 const markdown = require('../deps/markdown.js');
 const flags = require('../flags.js');
 const {loadContributorFromConfiguration} = require('./contributors.js');
+const log = require('fancy-log');
+const colors = require('ansi-colors');
 
 /**
  * Returns default DevSite 2 meta content for this path.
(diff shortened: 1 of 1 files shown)


robdodson:

robdodson:
Nice work!

Testing this out I came across a few issues.

A link to the guide still shows up on the /fast page because it's in Fast's [guides.yaml](https://github.com/GoogleChrome/web.dev/blob/757114543f26926fdf6ef14ec8075db46725ce71/content/fast/guides.yaml#L43). I'm wondering if the task that builds that table of contents should check the published status on each doc.

I also noticed that it doesn't work for codelabs, so I wasn't able to unpublish `fast/codelab-avoid-invisible-text.md`.

TimvdLippe:
Ah there is some duplication in terms of guide config processing. Thinking of refactoring that...

TimvdLippe:
After a quick discussion we came to the conclusion the generator requires a rewrite. Therefore I am closing this PR for now and work on that instead.

PR #564, opened 2019-01-14
Update about.html

Meggin:
Improving about page to better align with Google voice.

### DIFF
--- a/content/about.html
+++ b/content/about.html
@@ -22,7 +22,7 @@
             </h1>
             <div class="masthead-left__body">
               <p>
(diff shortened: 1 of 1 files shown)


content/about.html
```diff
@@ -22,7 +22,7 @@ <h1 class="masthead-left__title">
             </h1>
             <div class="masthead-left__body">
               <p>
-                web.dev is the ultimate resource for developers of all
+                web.dev is a resource for developers of all
```
Meggin: I don't that's right?

Issue #88, opened 2018-11-01
Let users know that applications take some time to reload with webpack


housseindjirdeh:
With a working Glitch environment, any update to a file that should be edited reloads the application quickly:

![glitch1](https://user-images.githubusercontent.com/12476932/47868429-6522b480-ddc1-11e8-9647-9427e4aead2e.gif)

If webpack is used, the build process takes a bit longer (~5-10s), and the only indication that the application is in the process of reloading is the tiny orb thing in the Status bar

![glitch2](https://user-images.githubusercontent.com/12476932/47868618-da8e8500-ddc1-11e8-85ef-abf1ef3c44b2.gif)

Readers new to the environment may not realize this, so we probably need to mention this in the user setup instructions of the codelab.

Codelabs that use webpack:

* Reduce JavaScript payloads with code-splitting
* Reduce network payloads using text compression
* Remove unused code
* Serve modern code to modern browsers for faster page loads

robdodson:
Hey @housseindjirdeh, friendly ping on this, do you think you'll have time to work on this this quarter? Happy to reassign if you're too busy.

housseindjirdeh:
Will get to it this week!

PR #567, opened 2019-01-16
Adds note on webpack build times

housseindjirdeh:
For #88 

Just added a single `aside` after the `Measure` section for each codelab with webpack. Happy to move/adjust if anyone things I should 🕺


robdodson:
giphy

PR #569, opened 2019-01-17
[WIP] Rewrite generator

TimvdLippe:
Rewrite the generator to make changes in the future easier. This prepares us for staging guides as well as easily introducing new translations.

The generator is based on 3 components:

1. First all metadata is read in `readers.ts`. This file will generate the appropriate interim data-format. The format is specified in `file-types.ts`
2. Then `index.ts` does the wiring and and invokes the `writer`. (We can add the logger here)
3. Lastly `writers.ts` writes from the data format to the build output.

Please note that this is the MVP which produces (almost) equivalent files. I would first like to iterate on the design and then start working on tidying up the code.

### DIFF
--- a/generator/contributors.ts
+++ b/generator/contributors.ts
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+import * as yaml from 'js-yaml';
+import * as fs from './fsp.js';
+import * as path from 'path';
+
+const CONTRIBUTORS_CONFIG_LOCATION = path.resolve(
+    __dirname,
+    '..',
+    '..',
+    'contributors',
+    '_contributors.yaml',
+);
+const CONTRIBUTORS_PICTURES_LOCATION = path.resolve(
+    __dirname,
+    '..',
+    '..',
+    'content',
+    'images',
+    'contributors',
+);
+
+export interface Author {
+  picture: string;
+}
+
+let authorsConfig: {[author: string]: Author};
+
+/**
+ * Load the contributor information for an author of a codelab.
+ * This function will lazily parse the config, if it hadn't done so before.
+ *
+ * @param {string} author Author name to search for in `_contributors.yaml`
+ */
+export async function loadContributorFromConfiguration(author: string):
+    Promise<Author> {
+  // Load and parse file once if we hadn't done so already.
+  if (!authorsConfig) {
+    const contributorsFile = await fs.readFile(
+        CONTRIBUTORS_CONFIG_LOCATION,
+        'utf8',
+    );
+    authorsConfig = yaml.load(contributorsFile);
+  }
+
+  const authorInformation = authorsConfig[author];
+
+  if (!authorInformation) {
+    throw new Error(`Could not find author information for ${author}!`);
+  }
+
+  const hasPictureFile = await fs.exists(
+      path.resolve(CONTRIBUTORS_PICTURES_LOCATION, author + '.jpg'),
+  );
+
+  authorInformation.picture = hasPictureFile ? author : 'no-photo';
+
+  return authorInformation;
+}
--- a/generator/index.ts
+++ b/generator/index.ts
@@ -1,0 +1,56 @@
+import * as path from 'path';
+
+import {FileData} from './file-types.js';
+import * as fs from './fsp.js';
+import {readLearningPath, readTopLevelFile} from './readers.js';
+
+const TOP_LEVEL_DIRECTORY = path.resolve(__dirname, '..', '..');
+// const BUILD_OUTPUT = path.resolve(TOP_LEVEL_DIRECTORY, 'build_output');
+const CONTENT = path.resolve(TOP_LEVEL_DIRECTORY, 'content');
+
+async function readContentDirectory(): Promise<FileData[]> {
+  const contentFilesOrDirectories =
+      await fs.readdir(CONTENT, {withFileTypes: true});
+
+  if (typeof contentFilesOrDirectories[0] === 'string') {
+    throw new Error(`You need at least Node 10.10.0 for this script`);
+  }
+
+  return Promise.all(contentFilesOrDirectories.map(async (fileOrDirectory) => {
+    const resolvedName = path.resolve(CONTENT, fileOrDirectory.name);
+
+    if (fileOrDirectory.isDirectory()) {
+      // Don't try to parse the images folder as a learning path
+      if (await fs.exists(path.resolve(resolvedName, 'guides.yaml'))) {
+        return readLearningPath(resolvedName, fileOrDirectory.name);
+      }
+
+      return {name: fileOrDirectory.name};
+    }
+
+    return readTopLevelFile(resolvedName, fileOrDirectory.name);
+  }));
+}
+
+const shouldPreserveFileForBuildOutput = (_file: FileData) => {
+  return true;
+};
+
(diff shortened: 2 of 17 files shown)


robdodson:

generator/readers.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  return {attributes, body};
+}
+
+function isCodelab(fileName: string) {
+  return fileName.startsWith('codelab-') && fileName.endsWith('.md');
+}
+
+async function readCodelab(
```
robdodson: should this have a declared return type?

generator/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const rootCards: RootCards = {paths: []};
+  const allGuides: SerializedGuideJson[] = [];
+
+  for (const file of files) {
+    if ('body' in file) {
+      await writeTopLevelFile(BUILD_OUTPUT, file);
+    } else if ('topics' in file) {
+      await writeLearningPath(BUILD_OUTPUT, file);
+      rootCards.paths.push({...file, href: `/${file.name}`});
```
robdodson: I was wondering if this section and the block below that builds `allGuides` should be handled by readers as well. It feels odd for `index.ts` to own this logic.

generator/writers.ts
```diff
@@ -0,0 +1,92 @@
...
+    const codeLabBody = CODELAB_TEMPLATE(
+        {main: markdown(codelab.body), glitch, title, relatedGuide: guide});
+
+    await fs.writeFile(
+        path.resolve(guideDirectory, codelab.name + '.html'),
+        DEVSITE_TEMPLATE({title, meta: {description}, body: codeLabBody}));
+  }
+}
+
+export async function writeLearningPath(directory: string, file: LearningPath) {
```
robdodson: Suggest renaming the first argument to `outputDirectoryPath`

generator/writers.ts
```diff
@@ -0,0 +1,92 @@
...
+  const {name, title, description} = file;
+
+  await fs.writeFile(
+      path.resolve(directory, name + '.html'),
+      DEVSITE_TEMPLATE(
+          {title, meta: {description}, body: PATH_TEMPLATE(file)}));
+
+  const contentDirectoryName = path.resolve(directory, name);
```
robdodson: suggest renaming to `contentDirectoryPath`

generator/index.ts
```diff
@@ -0,0 +1,86 @@
+import * as path from 'path';
+
+import {FileData, RootCards, SerializedGuideJson} from './file-types.js';
+import * as fs from './fsp.js';
+import {readLearningPath, readTopLevelFile} from './readers.js';
+import {writeLearningPath, writeRootCards, writeSerializedGuideJson, writeTopLevelFile} from './writers.js';
+
+const TOP_LEVEL_DIRECTORY = path.resolve(__dirname, '..', '..');
+const BUILD_OUTPUT = path.resolve(TOP_LEVEL_DIRECTORY, 'build_output', 'en');
```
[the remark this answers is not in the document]
robdodson: also assuming the name will change to `build` instead of `build_output` before landing?

generator/contributors.ts
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+import * as yaml from 'js-yaml';
+import * as fs from './fsp.js';
```
robdodson: I think I mentioned this to sam at the time...

would it make more sense to use `fs-extra` here?
https://github.com/jprichardson/node-fs-extra

generator/writers.ts
```diff
@@ -0,0 +1,92 @@
...
+const GUIDE_TEMPLATE = readTemplate('guide');
+const CODELAB_TEMPLATE = readTemplate('glitch');
+
+export async function writeTopLevelFile(directory: string, file: TopLevelFile) {
+  await fs.writeFile(path.resolve(directory, file.name), file.body);
+}
+
+async function writeSingleGuide(
+    directory: string, learningPath: LearningPath, topic: PathTopic,
+    guide: GuideHTMLFileWithMetadata) {
```
robdodson: I was wondering why this type name includes `HTML`. Isn't it a JS object containing markdown?

generator/writers.ts
```diff
@@ -0,0 +1,92 @@
...
+  const guideDirectory = path.resolve(directory, guide.name);
```
robdodson: guideDirectoryPath

generator/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+      return {name: fileOrDirectory.name};
+    }
+
+    return readTopLevelFile(resolvedName, fileOrDirectory.name);
+  }));
+}
+
+function shouldPreserveFileForBuildOutput(_file: FileData) {
```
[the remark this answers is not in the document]
TimvdLippe: TS doesn't compile if you have an unused variable (really annoying behavior!). If a variable starts with an underscore, the compiler ignores that specific unused variable.

generator/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  rootCards.paths.sort((one, other) => one.order - other.order);
+
+  await writeRootCards(BUILD_OUTPUT, rootCards);
+  await writeSerializedGuideJson(BUILD_OUTPUT, allGuides);
+}
+
+async function main() {
+  const files = await readContentDirectory();
+  const filteredFiles = files.filter(shouldPreserveFileForBuildOutput);
```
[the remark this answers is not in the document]
TimvdLippe: It was mostly a left-over of me trying to model the "this is a draft guide, ignore it"-feature. I should probably put that in the reader though.

generator/readers.ts
```diff
@@ -0,0 +1,150 @@
...
+const CODELAB_REQUIRED_ATTRIBUTES = [...REQUIRED_ATTRIBUTES, 'glitch'];
+const GUIDE_CONFIGURATION_REQUIRED_ATTRIBUTES =
+    ['title', 'description', 'overview', 'topics', 'order'];
+
+async function readYamlAndAssertAttributes(
+    requiredAttributes: string[], fileName: string) {
+  const {attributes, body} =
+      yamlFrontMatter(await fs.readFile(fileName, 'utf8'));
+
+  for (const requiredAttribute of requiredAttributes) {
```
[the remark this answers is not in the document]
TimvdLippe: It normally does, but because we are dynamically reading here, TS has no clue what you read from disk. Therefore this function exists to assert it is in the correct shape, thus afterwards we do adhere to the interface (and TS will check for that).

generator/readers.ts
```diff
@@ -0,0 +1,150 @@
...
+          requiredAttribute}" was not specified for "guides.yaml" in "${
+          directoryName}"`);
+    }
+  }
+
+  return guideConfiguration;
+}
+
+export async function readLearningPath(
+    directoryName: string, learningPathName: string): Promise<LearningPath> {
```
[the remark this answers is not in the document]
TimvdLippe: Yeah these names are more of a remnant of the old naming. We should fix this, good point!

generator/readers.ts
```diff
@@ -0,0 +1,150 @@
...
+      guideContentFiles.filter(file => !isCodelab(file) && file !== 'index.md')
+          .map(artifact => ({
+                 source: path.resolve(guideFileName, artifact),
+                 fileName: artifact
+               }));
+
+  const href = `/${path.basename(directoryName)}/${guideName}`;
+
+  const codelabs = await Promise.all(codelabFiles.map(codelabFile => {
+    const codeLabName = codelabFile.substring(0, codelabFile.length - 3);
```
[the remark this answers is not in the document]
TimvdLippe: TIL, awesome!

generator/index.ts
TimvdLippe: Sounds like a plan.

generator/contributors.ts
TimvdLippe: I am mostly indifferent about that. I just was hesitant to include yet another dependency for something we can do.

generator/writers.ts
TimvdLippe: It is, but it is based on the HTML file that we read. Mostly for me to keep track of what the original value was. In general, naming is bonkers atm and I should fix that first.

robdodson:
Hey @TimvdLippe is it ok if I close this PR in favor of Ewa's eleventy work?

TimvdLippe:
Yes let's do that.

PR #570, opened 2019-01-20
Fix profile image.

robdodson:
Quick fix for Milica's profile photo on the site. I'd like to follow up with another PR that allows or flags when images end in `jpeg` instead of `jpg`.


TimvdLippe:
(The generator should simply fail fast if it can't detect an image. I prefer if authors explicitly said "no image", so that we can catch these issues at build time rather than in prod)

robdodson:
Yeah that sounds good.

PR #572, opened 2019-01-22
Fix .babelrc targets to use array

dangdennis:


robdodson:
This looks good. I think the problem with the previous example was the string was split by a `,` but it looks like it has to be one continuous string. Using an array of strings seems a bit nicer in my opinion because it's harder to make this mistake.

robdodson:
@dangdennis 	are you able to sign the CLA?

dangdennis:
I signed the CLA. Thank you!

PR #573, opened 2019-01-22
Enforce that the author image actually exists

TimvdLippe:
If authors do not want their picture shown on the website,
they can set `picture: no-photo` in the `_contributors.yaml`.

This would prevent issues like #570

### DIFF
--- a/lib/config/contributors.js
+++ b/lib/config/contributors.js
@@ -64,7 +64,17 @@ async function loadContributorFromConfiguration(author) {
     path.resolve(CONTRIBUTORS_PICTURES_LOCATION, author + '.jpg'),
   );
 
-  authorInformation.picture = hasPictureFile ? author : 'no-photo';
+  if (authorInformation.picture !== 'no-photo') {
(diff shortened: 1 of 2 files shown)


lib/config/contributors.js
```diff
@@ -64,7 +64,17 @@ async function loadContributorFromConfiguration(author) {
     path.resolve(CONTRIBUTORS_PICTURES_LOCATION, author + '.jpg'),
   );
 
-  authorInformation.picture = hasPictureFile ? author : 'no-photo';
+  if (authorInformation.picture !== 'no-photo') {
+    if (!hasPictureFile) {
+      throw new Error(
+        `Could not find author picture for "${author}".` +
```
robdodson: Maybe it should also say "Did you use .jpeg instead of .jpg"? That seems to be what happened in the previous issue.

Alternatively, it could be tolerant and accept either.

TimvdLippe: Done

PR #602, opened 2019-02-08
Correct script attributes

dangdennis:
`type="module"` are actually appended to `.mjs` files, and `type="nomodule"` are appended to `.js` files.


samthor:
So you've fixed one typo here but it's not `type="nomodule"`, it's definitely the `nomodule` attribute.

dangdennis:
thanks for reviewing. My mistake on the nomodule attribute.

PR #603, opened 2019-02-08
add simple gauge from internal

samthor:


src/components/web_lighthouse_gauge/web_lighthouse_gauge.scss
```diff
@@ -0,0 +1,66 @@
+@import '../../styles/settings/colors';
+
+
```
robdodson: Remove extra space

src/components/web_lighthouse_gauge/web_lighthouse_gauge.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+@customElement('web-lighthouse-gauge')
+export class WebProgressBar extends BaseElement {
+  @property({type: Number}) score = 1;
+
+  // eslint-disable-next-line require-jsdoc
+  render() {
+    const score = this.score;
+
+    this.setAttribute('role', 'progressbar');
```
robdodson: Do you know if LitElement has a lifecycle callback like connectedCallback where these could be set? Or is it recommended to do that work in render? Just to avoid calling setAttribute more than once.

src/components/web_lighthouse_gauge/web_lighthouse_gauge.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.setAttribute('aria-valuemin', '0');
+    this.setAttribute('aria-valuemax', '100');
+    this.setAttribute('aria-valuenow', `${score * 100}`);
```
robdodson: I think this is the only value that would need to be updated on render.

src/components/web_progress_bar/web_progress_bar.ts
```diff
@@ -16,7 +16,7 @@ export class WebProgressBar extends BaseElement {
   render() {
     return html`
       <div class="web-progress-bar-wrapper">
-        <div class="web-progress-bar-indeterminate"></div>
+        <div class="web-progress-bar-indeterminate">isAwesome=${this.isAwesome}</div>
```
robdodson: If you want to delete this and the underlying property I think that's ok. I put it in there initially as an example of how to use the decorators.

src/sandbox/elements.html
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
```
robdodson: small nit, would you mind renaming this file to 'components.html'?

src/components/web_lighthouse_gauge/web_lighthouse_gauge.scss
```diff
@@ -0,0 +1,66 @@
...
+    color: currentColor;
+    display: flex;
+    font: inherit;
+    justify-content: center;
+    left: 0;
+    position: absolute;
+    right: 0;
+    top: 0;
+  }
+}
```
robdodson: Mind adding a new line? I need to fix them elsewhere as well.

src/scripts/main.ts
```diff
@@ -1,5 +1,6 @@
 import '../styles/main.scss';
 import '../components/web_progress_bar/web_progress_bar.ts';
+import '../components/web_lighthouse_gauge/web_lighthouse_gauge.ts';
 
 /* eslint-disable-next-line */
 console.log('hello world');
```
robdodson: new line

src/components/web_progress_bar/web_progress_bar.ts
samthor: I mostly did this to test that it was working. Reverted now.

src/components/web_lighthouse_gauge/web_lighthouse_gauge.ts
samthor: I think we can probably put these into `firstUpdated` below.

src/components/web_lighthouse_gauge/web_lighthouse_gauge.ts
samthor: Ack

PR #604, opened 2019-02-08
V2 moar styles

robdodson:


src/styles/generic/type.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @include bp(lg) {
+    @include font($BODY);
+    @include create_margins($BODY);
+    @include create_paragraph_spacing($PARAGRAPH_SPACING);
+  }
+}
+
+// hr {
```
samthor: remove if disused?

src/styles/tools/_mixins.scss
```diff
@@ -69,3 +76,14 @@
...
   }
 }
+
+// Create selectors to allow paragraphs to evenly separate.
+// ---
+// @param {number} $font-size: A font-size
+// @content: A set of selectors and modifiers for altering the visual layout
+//  of paragraph elements
+@mixin create_paragraph_spacing($margin) {
+  & + & {
```
samthor: what does `& + &` do again? worth a comment?

src/styles/generic/type.scss
robdodson: oh I plan to use it. I just commented it out because I'm hitting a stopping point for the night.

src/styles/tools/_mixins.scss
robdodson: Added a comment. It's a mixin that will generate sibling selectors based on its parent selector.

PR #610, opened 2019-02-08
Add admin owners to everything.

robdodson:
GitHub Codeowners does not allow super users to approve PRs on subdirectories if those subdirectories have owners. Which seems like a really crappy design decision.

Posted a comment here:
https://github.community/t5/How-to-use-Git-and-GitHub/CODEOWNER-Required-Reviews/m-p/15422/highlight/false#M4786

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,19 +6,19 @@
 * @robdodson @samthor
 
 # Guide collections
-/content/ @Meggin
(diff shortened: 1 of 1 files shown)


robdodson:
Yeah it's _almost_ a good system.

There's a flag that makes it so repo admins are exempt from the rules. Which is kind of a bummer but I may just use that instead.

PR #616, opened 2019-02-14
 Add topics to path yaml

devnook:


robdodson:

_includes/guide.njk
```diff
@@ -50,35 +50,43 @@ layout: layout
   </div>
 </main>
 
-{# <div class="cta-next-guide">
+{% set next = false %}
+{% for item in collections[tags[0]] %}
```
robdodson: Would this work if we had more than one tag? Or is this a temporary solution to test the idea?

robdodson: I'm not sure if this will work because there's no way to ensure an order to the items in the collection. I think they are ordered by their author date.

devnook: This was the reason why I asked if we allow more than one collection. I'm not sure we'll need this code at all, if we scrape the "next guide" link, but wanted to show it here to discuss exactly this. 

As for ordering, there is a way to order the collection other than by date, but its tricky. So first we should answer the question - do we allow post to belong to multiple freeform collections and then we don't have linear navigation in any of them, or we allow 1 "path" post (and then it can be ordered arbitrarily)?

robdodson: I think we'll want to only allow a post to live in one collection.

Sorry I meant, if we have a collection of posts tagged 'accessible' and a different collection of posts tagged 'speed', won't `tags[0]` always equal 'accessible' and thus never show posts from 'speed'?

robdodson: I think my concern about ordering is it would probably require a post having an `order` field in its YAML, and at that point I feel like the benefits of using a tag for the collection start to go down because you can't quickly get a look at the table of contents for the whole collection.

I think this is why I ended up using a directory data file instead to organize the table of contents for a particular collection because it consolidated everything into one place.

devnook:
In current change set:

1. All "paths" are defined in _data (each in its own file). "path" = a collection that has explicit order and TOC with topics. It also includes "path" metadata like owners list.

2. You can move to next guide within path, once at the end of path you can get back to path home page. 

3. I moved some common fn to helpers file.

4. Reading next guide title from its slug is still slow (requires disk read).

Let me know what you think of this approach?

.eleventy.js
```diff
@@ -1,22 +1,38 @@
...
 module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(pluginSyntaxHighlight);
   eleventyConfig.addPassthroughCopy("css");
 
   eleventyConfig.addShortcode("getTitleBySlug", function(inputPath, slug) {
-    // TODO(robdodson): This may be expensive and slow. See if there's a way
-    // to do this without hitting the disk, or at least do it async.
-    // Maybe this should be a macro
-    const file = fs.readFileSync(
+    return helpers.getTitleBySlug(
```
robdodson: I was wondering if another way to do this would be to make sure every article has the tag `post`. Then you might be able to loop through the `post` collection and find the matching article? This might also let you get rid of the `basePath` property.

templateHelpers.js
```diff
@@ -0,0 +1,27 @@
...
+    // to do this without hitting the disk, or at least do it async.
+    // Maybe this should be a macro
+    const file = fs.readFileSync(
+      filePath,
+      "utf8"
+    );
+    return yamlFont.loadFront(file).title;
+  },
+
+  getCollectionName: function (page) {
```
robdodson:
```suggestion
  getCollectionName: function(page) {
```

templateHelpers.js
```diff
@@ -0,0 +1,27 @@
...
+    const urlParts = page.url.split('/');
+    if (urlParts.length > 5 && urlParts[3] !== page.fileSlug) {
+      return urlParts[3];
+    }
+    return '';
+  },
+
+  getGuidesFromTopics: topics => topics.reduce(
```
robdodson: Do you prefer fat arrow or `function` keyword style? I think either is fine but it may be good to stick with one throughout. I think this file mixes both.

content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -10,7 +10,7 @@ author: robdodson
 web_lighthouse:
   - tabindex
   - focusable-controls
-wf_blink_components: Blink>Accessibility
+wf_blink_components: Blink>Accessibilitygit stash pop
```
robdodson: typo

_data/paths/accessible.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  'basePath': 'content/en/accessible',
```
robdodson: I don't think you need to stringify the keys because it's a JS object.

_data/paths/accessible.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
robdodson: Did you take a look at putting this data inside of `content/en/accessible/accessible.json` to create a [directory data file](https://www.11ty.io/docs/data-template-dir/) instead? I think it would then get inherited into each page's data object along with the page's YAML.

This might make it easier because then you wouldn't need to look up the collection that a particular page is in when you're generating the next page link. You would know that `topics` is always the topics specific to `accessible`.

_data/paths/accessible.js
```diff
@@ -0,0 +1,18 @@
...
+  'basePath': 'content/en/accessible',
+  'owners': [],
```
robdodson: Was this meant for code owners? If so, we may be able to rely on [GitHub CODEOWNERS](https://help.github.com/articles/about-code-owners/) instead. I've [already started using it a bit on the site](https://github.com/GoogleChrome/web.dev/blob/master/CODEOWNERS).

.eleventy.js
devnook: It's good idea with the collections. I'd like to keep the basePath though, as this allows us to cater for cases where guides have the same slug (e.g. accessibility/intro.md and security/intro.md).

templateHelpers.js
devnook: Good point, added to lint

templateHelpers.js
devnook: done

content/en/accessible/control-focus-with-tabindex/index.md
devnook: done

_data/paths/accessible.js
devnook: Yes, I tried that first actually. The problem is that directory data file if visible only to templates that are children of that directory. If we want a guide to belong to 2 paths / collections in future, it might be more beneficial to keep this data independent of actual directory structure. This was the reason why I decided to put it in global data, just in separately named files.

wdyt?

_data/paths/accessible.js
devnook: Sure, that works.

devnook:
PTAL

.eleventy.js
robdodson: I wonder if we would be able to get that same information using the collection file's `inputPath`?
https://www.11ty.io/docs/collections/#collection-item-data-structure

_data/paths/accessible.js
robdodson: Yep that makes sense. I think I'm leaning toward saying that a guide should only be able to live inside of one collection. I'm mostly thinking from a user experience perspective it might be strange to see the same guides popping up over and over but in different places.

_includes/guide.njk
```diff
@@ -50,35 +52,25 @@ layout: layout
...
-{# <div class="cta-next-guide">
+<div class="cta-next-guide">
   <div class="cta-next-guide__wrapper">
-    {% if next %}
-    <div class="cta-next-guide__heading">
-      Next Guide
-    </div>
-    <a class="cta-next-guide__link gc-analytics-event"
+    {% set nextUrl = macros.nextGuide(page, paths).trim() %}
+    {% if nextUrl !== '' %}
```
robdodson: I noticed when I got to the end of the guides it still showed a next guide but it was pointing at the path landing page.

image

_data/paths/accessible.js
devnook: Even if we limit ourselves to one collection, "upgrading" a post to a collection guide would be easier if it's not tied to directory structure.

_includes/guide.njk
devnook: In the current version on web.dev "Accessible to all" is both a) repeated twice b). Links to the same destination as "Return to all guides". Maybe it would be best to not have it in the last guide in a series at all, and leave only "return to all guides" link?

_data/paths/accessible.js
robdodson: Ah yeah that's a good point. Ok let's keep the _data approach.

robdodson:

PR #620, opened 2019-02-20
Correct `cache-control` directive in example response

### LINKED ISSUE #609: Correct the sentence on 'no-cache' - stating it means: don't cache
https://web.dev/secure/cross-origin-resource-sharing#how-does-a-resource-request-work-on-the-web

> Sample Response header
> 
> Content-Encoding: gzip
> Cache-Control: no-cache 
> 
> Above is same as saying "Data is encoded with gzip. **Do not cache this please**"

That statement is incorrect. [`no-cache`](https://tools.ietf.org/html/rfc7234#section-5.2.2.2) tells the client to revalidate with the server before serving the (cached) file, responding with a 304 if nothing changed. 

The explanation should be corrected, unless the intention was to say [`no-store`](https://tools.ietf.org/html/rfc7234#section-5.2.2.3) - then change to that directive?

Malvoz:
Resolves #609.

### DIFF
--- a/content/secure/cross-origin-resource-sharing/index.md
+++ b/content/secure/cross-origin-resource-sharing/index.md
@@ -67,7 +67,7 @@ Above is equivalent to saying "I want to receive HTML in response. Here is cooki
 
 <strong>Sample Response header</strong>
 <pre>Content-Encoding: gzip
-Cache-Control: no-cache </pre>
+Cache-Control: no-store </pre>
 Above is same as saying "Data is encoded with gzip. Do not cache this please"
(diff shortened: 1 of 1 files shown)


Malvoz:
[Signed CLA]

robdodson:
I think `no-store` seems like the correct text to use here. Because the PR has been around for a while and might require some weird rebasing I just fixed the issue on master. Thank you for pointing it out @Malvoz!

Malvoz:
Thanks! :+1:

PR #686, opened 2019-03-04
Implement /learn page with path cards

devnook:


_data/paths/accessible.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   basePath: 'content/en/accessible/',
+  disabled: false,
```
robdodson: We could consider removing this. I don't know if we want to do "Coming Soon" stuff anymore or if it would make more sense to just publish the collections when they're ready.

robdodson:
Going to close this since we're working on it in the frontend repo.

PR #712, opened 2019-03-09
Fix missing glitch

robdodson:
Not sure what happened but I noticed this glitch started 404ing. Looks like the name was updated maybe? This PR should fix it.

Glitch is here:
https://glitch.com/~fix-sneaky-404


robdodson:
All good. Thank you!

PR #760, opened 2019-04-01
Add LH audit to responsive images guide

robdodson:


robdodson:
@khempenius just an fyi that I added this.

PR #838, opened 2019-04-29
Added Uploadcare to the list of CDNs in Other options

### LINKED ISSUE #837: Add Uploadcare CDN to the list of CDNs in Other options
https://web.dev/serve-responsive-images#other-options

[Uploadcare](https://uploadcare.com) is an alternative to Cloudinary with a great focus on images. Worth mentioning.

rsedykh:
Fixes #837.

- Add Uploadcare to the list of CDNs in Other options (https://web.dev/serve-responsive-images#other-options).

### DIFF
--- a/src/site/content/en/fast/serve-responsive-images/index.md
+++ b/src/site/content/en/fast/serve-responsive-images/index.md
@@ -72,12 +72,12 @@ tiny bit more HTML.
 
 ### Other options
 
-Image services like [Thumbor](https://github.com/thumbor/thumbor) (open-source)
-and [Cloudinary](https://cloudinary.com/) are also worth checking out. Image
-services provide responsive images (and image manipulation) on-demand. Thumbor
-is setup by installing it on a server; Cloudinary takes care of these details
-for you and requires no server setup. Both are easy ways to create responsive
-images.
+Image services like [Thumbor](https://github.com/thumbor/thumbor) (open-source),
+[Cloudinary](https://cloudinary.com/) and [Uploadcare](https://uploadcare.com/)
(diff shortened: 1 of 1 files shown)


rsedykh:
Hi @robdodson, should I change something or it's ok and I just wait?

robdodson:
Hi @rsedykh, sorry for the slow reply, I've been out on vacation.

We discussed this change internally and I think we're going to hold off on it for now. We tried to put together a bit of an explanation in the second section here: https://github.com/GoogleChrome/web.dev/wiki/Writing-style#content-style

Happy to discuss further but in general it's difficult for us to have an exhaustive list of services so we're tending toward those that we have prior experience working with and that have a large community around them already.

rsedykh:
@robdodson it was my turn to be out on vacation. :-)

> Guides and code labs should use well-known and well-established tooling.

What're your criteria for well-known and well-established?

- We've been serving adaptive media for seven years already for customers including Mozilla, Facebook, Webflow. Right now we have tens of thousands of active customers, tens of millions uploads per month, hundreds of millions of objects stored, hundreds of terabytes delivered per month.
- We developed and open-sourced the fast and reliable image resizing algorithm (https://news.ycombinator.com/item?id=14712146, 1142 stars / 1252 forks right now) to help spread the word about Responsive Images and leverage the inequality in implemented techs between smaller and larger companies.

Also, if not Uploadcare, isn't service like imgix.com not well-known and well-established enough to provide an alternative to Cloudinary? My intent is the same as yours — promoting democracy on the web and diversifying the set of provided solutions to improve the user experience on the internet. If not Uploadcare, then why not at least anybody else?

PR #842, opened 2019-04-29
Proposed edits to load-fast-enough-for-pwa

mfriesenhahn:

- Fixed broken links
- Edits for concision.


robdodson:

robdodson:
I manually merged this in but now I'm wondering if that was a mistake because I noticed it removed a lot of stuff. @Meggin should I revert back to the previous content?

Meggin:
Yeah, I think this should be rolled back. I wound up adding images and more to the guide. We could re-apply a lot of the good suggestions too-- but for sure, don't want to lose the lighthouse report, for example.

PR #874, opened 2019-05-03
Improve the language on the measure page.

robdodson:
Fixes #755.

- Update the language in the measure page based on #755.

This does not include all of the changes in #755 because we'd need to write some updated CSS for those and unfortunately we missed the cut for I/O. But this at least explains how the Lighthouse scoring works and updates all of the existing static text.

I removed all of the links to various Lighthouse formats because I didn't think it really helped much and we're already linking to the Lighthouse landing page.

### DIFF
--- a/src/site/content/en/measure/index.njk
+++ b/src/site/content/en/measure/index.njk
@@ -2,8 +2,8 @@
 layout: layout
 title: Measure
 description: |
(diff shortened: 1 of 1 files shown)


robdodson:
@meggynw fyi. Sorry for taking _forever_ to get to this.

PR #880, opened 2019-05-04
Add style overrides.

robdodson:
Adds a new include which can be injected into every page. This include has an style element which we can use to sneak in last minute style changes while we wait on devsite releases.


robdodson:
I'm going to go ahead and merge this but wanted to mention it as an FYI in case you see any last minute style quirks. We should have the power to sneak in short term fixes.

Meggin:
Cheeky, Rob.

PR #892, opened 2019-05-07
Rephrase portals intro and remove extra paragraph.

robdodson:
@uskay @Meggin I removed the paragraph referring to the manga viewer in the video and updated the intro paragraph because it felt a little too short.

What do y'all think?


robdodson:
I'm going to chance it and land this :)

PR #897, opened 2019-05-07
Fix factual error in introduction 

jpmedley:
I removed the unused hero image while I was here. Don't let that fool. The textual change is extremely important.


robdodson:
Happy to merge the text change, but we still use that hero image (we swapped back to it instead of using the surma one). I can manually merge this so you don't have to mess with PR because I assume you're at I/O :)

robdodson:
Should be fixed on master now.

PR #921, opened 2019-05-21
Fix hero image.

jpmedley:
@robdodson When we swapped the hero image on this article the old one got left behind. At some point I tried to remove the old image and deleted the wrong one. This corrects that mistake.


robdodson:
oh I think this deletes the wrong image. John preferred the one with the app screenshots.

jpmedley:
John,

Is that true?
 816-678-7195

robdodson:
@jpmedley I don't think John is on the thread but he and I spoke over chat during I/O. He preferred the screenshots of the app because our image compression was being too aggressive on the shot of Surma.

jpmedley:
I remember the discussion now.

PR #942, opened 2019-05-28
Rev eleventy to 0.8.3 and rev syntax highlighter.

robdodson:
- Rev eleventy to 0.8.3
- Rev elventy-syntax-highliter to 2.0.3
- Should fix issues with multiline comments in code blocks.

cc @tomayac


tomayac:
Woohoo, thanks, @robdodson!

PR #983, opened 2019-06-06
Remove glitches directory.

robdodson:
I think in theory we like the idea of backing stuff up to GitHub and treating it as the source of truth but in practice I think folks find it easier to just update their Glitch on the website.

- Remove `glitches/` directory.


robdodson:
Tagged a number of folks in just in case they have feelings about this. If nobody says anything I'll assume folks are ok with the change :D

PR #990, opened 2019-06-07
Add Instruction component.

robdodson:
Adds a new {% Instruction %} shortcode which can be used
to DRY up common lists of instructions.

Fixes #964
Supersedes #984 

Changes proposed in this pull request:

- Adds new Instruction short code.
- Adds sandbox examples.
- Externalizes filter functions from Aside shortcode so others can use them.

Note I kept the switch/case from the previous component because if someone types "devtools-tabz" I want to throw an error so they catch the typo. Also TIL `substr` isn't supported in newer versions of Node!

PR #991, opened 2019-06-07
Instruction fixes

robdodson:
Depends on (and rebased against) #990 

- Fixes docs to use new Instruction component.


TimvdLippe:
I tried to change the base, but the diff still seems incorrect. It does not seem to include the changes from #990. Was the rebase successful?

mfriesenhahn:
This PR now only updates posts so that they use the Instruction widget. I did a fairly thorough search for Glitch instructions since they're the most likely to change. The search for DevTools instructions was more superficial, and I've almost certainly missed various cases. But to put this PR to bed and keep its scope from getting totally out of control, I'm leaving it here.

@kaycebasques, if you could review, that'd be great.

kaycebasques:
What a great example of "leave the campsite better than you found it"... thanks everyone!

PR #1015, opened 2019-06-18
Don't increment numbered lists when there are nested lists.

robdodson:
Fixes an issue where nested ordered lists were forcing the outer list to increment.
See example on: https://web.dev/value-of-speed/

- Only increment lists when the immediate child is an li, ignoring nested sublists.


robdodson:
Going to go ahead and land this so it can go out in today's publish.

PR #1017, opened 2019-06-19
Update documentation for `<a>` tag

shreykejriwal:
If clicking on an element adds new content in the web page, we should still use `<button>` instead. As native `<a>` has properties like opening the link in new window or tab , this feature won't be helpful for loading the new content unless we change the url of the page.

Fixes # .

- 
- 
-

### DIFF
--- a/src/site/content/en/accessible/use-semantic-html/index.md
+++ b/src/site/content/en/accessible/use-semantic-html/index.md
@@ -100,8 +100,7 @@ should you choose?
 
 - If clicking on the element will perform an _action_ on the page, use
   `<button>`.
-- If clicking on the element will _navigate_ the user to a new page (or load new
-  content in a single-page web app) then use `<a>`.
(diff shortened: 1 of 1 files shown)


robdodson:
> If clicking on an element adds new content in the web page, we should still use <button> instead. As native <a> has properties like opening the link in new window or tab , this feature won't be helpful for loading the new content unless we change the url of the page.

I think I can specify a bit more. I'm specifically talking about when folks are using anchors to do routing in a SPA. So in this case it would update the URL (and load new content).

robdodson:
Created a bug to track this #1020. @shreykejriwal would you be interested in updating your PR to keep the original language but clarify that you can use an anchor if you're doing routing in a SPA.

Maybe something like:

If clicking on the element will _navigate_ the user to a new page then use `<a>`. This includes if you're navigating within a SPA and using client-side routing to update the URL.

robdodson:
Fixed in #1020

PR #1021, opened 2019-06-20
`prefers-color-scheme` article for web.dev

tomayac:
This pull request adds a **new**:
- [x] Guide(s) 📘
- [ ] Codelab(s) 🔬

Have you read the wiki on writing [a guide](https://github.com/GoogleChrome/web.dev/wiki/How-to-write-a-post) or [a codelab](https://github.com/GoogleChrome/web.dev/wiki/How-to-write-a-codelab)?
- [x] Done! 🍰

Does your content tell a user how to fix a Lighthouse audit?
Have you added [the Lighthouse audit ID to your YAML](https://github.com/GoogleChrome/web.dev/wiki/How-to-write-a-guide#lighthouse-audits)?
- [x] My changes do not mention Lighthouse audits. 🙅‍♂️
- [ ] I have added the Lighthouse audit IDs to my content. 👍

Does your content reference any Glitch projects?
Have you read the wiki on [working with Glitch](https://github.com/GoogleChrome/web.dev/wiki/Working-with-Glitch)?
Have you added the Glitch project to [our team page](https://glitch.com/@webdev)?
- [ ] My changes do not contain any Glitches. 🙅‍♂️
- [x] I have read the wiki on working with Glitch. 🍰
- [x] I have added my Glitches to the team page. 👏


src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,725 @@
...
+with the space-separated value `light dark`.
+This tells the browser which color themes my app supports
+and allows it to activate special variants of the user agent stylesheet.
+This is useful to, for example, let the browser render form fields
+with a dark background and light text, adjust the scrollbars,
+or to enable a theme-aware highlight color.
+The exact details of `color-scheme` are specified in
+[CSS Color Adjustment Module Level&nbsp;1](https://drafts.csswg.org/css-color-adjust-1/).
+
+{% Aside 'note' %}
```
rowan-m: Not sure about the look of the two `aside`s following each other. Perhaps the content of the first one could be moved inline to the preceding text.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,725 @@
...
+---
+
+## Dark mode before *Dark Mode*
+
+<figure class="w-figure w-figure--inline-right">
+  <img style="height:175px; width:auto;" src="green-screen.jpg" alt="Green screen computer monitor" intrinsicsize="640x480">
+  <figcaption class="w-figcaption">Fig. — Green screen (<a href="https://commons.wikimedia.org/wiki/File:Compaq_Portable_and_Wordperfect.JPG">Source</a>)</figcaption>
+</figure>
+
+We have gone full circle with dark mode.
```
rowan-m: May want to consider compressing the intro, this article is easily double the length of most of the others.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,725 @@
...
+[Chrome’s](https://chromium.googlesource.com/chromium/blink/+/master/Source/core/css/html.css).
+
+<figure class="w-figure w-figure--inline-left">
+  <img style="height:175px; width:auto;" src="smartphone-in-bed.jpg" alt="Smartphone used while lying in bed" intrinsicsize="500x334">
+  <figcaption class="w-figcaption">Fig. — Smartphone used in bed (<a href="https://unsplash.com/photos/W39xsPWZgA4">Source</a>)</figcaption>
+</figure>
+
+Nowadays, a lot of content consumption and creation has shifted to mobile devices
+that use backlit Liquid Crystal Displays (LCD)
+or energy-saving Active-Matrix Organic Light-Emitting Diode (AMOLED) screens,
```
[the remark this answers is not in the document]
tomayac: Thanks, addressed in https://github.com/GoogleChrome/web.dev/pull/1021/commits/ec82763326e16821c52d1c2b7b20977e09356b67.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,725 @@
...
+A small detail you may have noticed is that I don’t use pure white.
+Instead, to prevent glowing and bleeding against the surrounding dark content,
+I choose a slightly darker white, `rgb(250, 250, 250)` or similar works well.
+
+## Full example
+
+In the following [Glitch](https://dark-mode-baseline.glitch.me/) embed,
+you can see the complete example that puts the concepts from above into practice.
+
+<div style="height: 900px; width: 100%;">
```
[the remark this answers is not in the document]
tomayac: You are actually not supposed (yet) to use the toggle button, it's more an easter egg that gets explained further down, when I introduce the custom element. I have clarified the text now in https://github.com/GoogleChrome/web.dev/pull/1021/commits/640aababe02b42fc04e47de5a9498dbaf72d4fb2.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,725 @@
...
+### Accessibility
+
+However, there’re also people who actually *need* dark mode or other accessibility tools,
+for example, users with low vision.
+The earliest occurrence of such an accessibility tool I could find is
+[System&nbsp;7](https://en.wikipedia.org/wiki/System_7)’s *CloseView* feature that had a toggle for
+*Black on White* and *White on Black*, which arguably can be called dark mode.
+While System&nbsp;7 supported color, the default user interface was still black-and-white.
+
+{% Aside 'note' %}
```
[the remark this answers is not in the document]
tomayac: I respectfully disagree here, this is the first time we had an accessibility setting in the browser that catered for people's a11y needs, just like today for people who *need* dark mode.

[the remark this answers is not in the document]
tomayac: Wait, are you referring to the "play with System 7 in the Internet Archive" side note? I could be convinced to remove that.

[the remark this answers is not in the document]
tomayac: Removed the Internet Archive aside in https://github.com/GoogleChrome/web.dev/pull/1021/commits/73edd13d82997519504aa88fd39540070b2c0821.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: Ungrouped the two by reflowing a bit in https://github.com/GoogleChrome/web.dev/pull/1021/commits/73edd13d82997519504aa88fd39540070b2c0821.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: Fully agree it's long, but I think it's necessary because else people don't understand why it's such a big deal. But as a compromise, I have added a "skip (Flash) intro" link right at the beginning in https://github.com/GoogleChrome/web.dev/pull/1021/commits/078e53c8b0d5eb16e14480d9167052ba2bcbb5ac.

robdodson:
I'm not sure if the hero image quite makes sense at first glance. In particular, I didn't understand the turtle or why light.css was highlighted. Also since it was taken against a white background with a drop shadow it looks oddly cut off on the bottom left and right edges.

image

A few options:

- Use something abstract ([example](https://unsplash.com/photos/YeUVDKZWSZ4))
- Maybe a side-by-side comparison of two pages (one in dark mode and one without)?
- Omit the hero entirely

src/site/content/en/blog/prefers-color-scheme/index.md
robdodson: I might propose inverting this a bit. Putting the content about the media query at the top, and making the second half of the post about the backstory. Another option would be to split it into two posts. One on the media query and another on the history.

We usually recommend that articles be between 1000 to 1500 words but this one is ~3400. We would like to stick to this policy because there are only a small group of folks doing content reviews and it's hard for them to scale when content is really lengthy. What has tended to happen in the past is really long articles get stuck in the content review stages for a long time and we want to avoid that.

tomayac:
> I'm not sure if the hero image quite makes sense at first glance.

Fair remark. I have just added a [new hero image](https://github.com/GoogleChrome/web.dev/pull/1021/commits/108545d23a99086dbbad00da794533e4c38c6e50) that looks way better.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: > I might propose inverting this a bit. Putting the content about the media query at the top, and making the second half of the post about the backstory. Another option would be to split it into two posts. One on the media query and another on the history.

I don't think this would flow right if inverted. Note that there is a [skip intro](https://github.com/GoogleChrome/web.dev/blob/108545d23a99086dbbad00da794533e4c38c6e50/src/site/content/en/blog/prefers-color-scheme/index.md#dark-mode-before-dark-mode) link right at the top. 

> We usually recommend that articles be between 1000 to 1500 words but this one is ~3400. We would like to stick to this policy because there are only a small group of folks doing content reviews and it's hard for them to scale when content is really lengthy. What has tended to happen in the past is really long articles get stuck in the content review stages for a long time and we want to avoid that.

@robdodson If the concern here is review constraints, please note that this has already been [reviewed](https://docs.google.com/document/d/1SfFa2PqVS4PzvoYgD7yUntiFBkyE_jdxdVJ1ehkiKck/edit?usp=sharing) by lzbylut@ and @rowan-m.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -811,6 +809,7 @@ Rune is also a co-editor of the [CSS Color Adjustment Module Level&nbsp;1](https
 I would like to 🙏 thank [Lukasz Zbylut](https://www.linkedin.com/in/lukasz-zbylut/)
 and [Rowan Merewood](https://twitter.com/rowan_m)
 for their thorough reviews of this article.
+The [loading strategy](#loading-strategy) is the brainchild of [Jake Archibald](https://twitter.com/jaffathecake).
```
tomayac: @jakearchibald Thank you!

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -456,7 +456,7 @@ so that they never compete with resources that are needed by the site right now.
   <figcaption class="w-figcaption">Fig. — Site in default light mode on a browser that doesn’t support <code>prefers-color-scheme</code> loads the dark mode CSS with lowest priority.</figcaption>
 </figure>
 
-### Finding out if dark mode is supported
+### Finding out if dark mode is supported by the browser
```
tomayac: @emilio via https://github.com/filamentgroup/loadCSS/issues/304#issuecomment-505229524:
> @tomayac fwiw, that post contains an error in the "Finding out if dark mode is supported" section. (prefers-color-scheme) is supposed not to match if there's no preference for light or dark, so the right way to test whether it's supported is not window.matchMedia('(prefers-color-scheme)').matches but window.matchMedia('(prefers-color-scheme)').matches || window.matchMedia('(prefers-color-scheme: no-preference)').matches (or something like that). See w3c/csswg-drafts#3857 for more fun stuff related to that.

Thanks for the pointer, but this is actually really about finding out whether the `prefers-color-scheme` is supported (i.e., understood) by the browser at all. So the current check is correct. I have appended "by the browser" to hopefully make this clearer.

chiragd:
Non-owner LGTM

src/site/content/en/blog/prefers-color-scheme/index.md
emilio: My point is that the browser may understand the `prefers-color-scheme` media feature while returning false for `matchMedia('(prefers-color-scheme)').matches`. The way to test whether the media query is correctly parsed is either what I mentioned above or `machMedia('(prefers-color-scheme)').media != 'not all'`.

tomayac: Oooh, I see what you mean now. The boolean context… Thanks a lot! 

[L4](https://www.w3.org/TR/mediaqueries-4/#mq-boolean-context) had this:
*When written like this, the media feature is evaluated in a boolean context. If the feature would be true for any value other than the number 0, a <dimension> with the value 0, or the keyword none, the media feature evaluates to true. Otherwise, it evaluates to false.*

The **addition** in [L5](https://drafts.csswg.org/mediaqueries-5/#boolean-context) is important:
*When written like this, the media feature is evaluated in a boolean context. If the feature would be true for any value other than the number 0, a <dimension> with the value 0, the keyword none, **or a value explicitly defined by that media feature to evaluate as false in a boolean context**, the media feature evaluates to true. Otherwise, it evaluates to false.*

And `no-preference` is [well defined](https://drafts.csswg.org/mediaqueries-5/#valdef-media-prefers-color-scheme-no-preference) as such a value. I will correct the article and credit you :-) Thanks again!

src/site/content/en/blog/prefers-color-scheme/index.md
robdodson: I guess there are two concerns:

1. Content that's >1000 words is harder for users to read quickly. Often times folks will open an article and quickly scroll to the bottom to see if they have time to read it all. If it looks like a lot the worry is someone might skip it.
2. It adds to the workload for reviewers. I think if other folks have reviewed the content then that's definitely helpful, but that won't always be the case so we want to encourage folks to stick to those lengths when they can.

In the future I might suggest splitting the post into two parts if the history is important but takes a lot of time to explain. Having a multi-part post series on a topic seems cool to me.

In this instance I think we can move ahead with this PR because it's already really close to the finish line.

tomayac: > In the future I might suggest splitting the post into two parts if the history is important but takes a lot of time to explain. Having a multi-part post series on a topic seems cool to me.

Duly noted for the next time. On the plus side, this will be a one-stop shop for all things dark mode. 

> In this instance I think we can move ahead with this PR because it's already really close to the finish line.

Great! If any of the owners LGTMs, I can merge tomorrow my local time and we can go live later that say.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,828 @@
...
+
+Operating systems that support a dark mode or dark theme
+typically have an option to activate it somewhere in the settings.
+On macOS&nbsp;X, it’s in the system preference’s *General* section and called *Appearance*,
+and on Windows&nbsp;10, it’s in the *Colors* section and called *Choose your color*.
+For Android&nbsp;Q, you can find it under *Display* as a *Dark Theme* toggle switch,
+and on iOS&nbsp;13, you can change the *Appearance* in the *Display &amp; Brightness*
+section of the settings.
+
+<figure>
```
robdodson: We've discouraged folks from doing custom CSS on the site because it will likely break as we rev the design in the future.

In this case the screenshots don't seem critical so it might be worth replacing this with a single screenshot as an example.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,828 @@
+---
+title: Hello Darkness, My Old Friend
+subhead: Overhyped or necessity? Learn everything about dark mode and how to support it to benefit your users!
+authors:
+  - thomassteiner
+date: 2019-06-25
+hero: hero.jpg
```
robdodson: Do you have a high res version of this image?
I think we need it at 1600 x 800 and a 2x version. We haven't been using srcset for these images but we should start so having the 2x version handy would be good.

tomayac: I have just added the 1600x800 version in https://github.com/GoogleChrome/web.dev/pull/1021/commits/e82cdb6493ed94a7523f3dd4de57cb6c411e2c19.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: I have removed the screenshots and just left one image in https://github.com/GoogleChrome/web.dev/pull/1021/commits/9cd638b5c40b640fa70b2dec87964148d895e779.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,786 @@
...
+  --accent-background-color: rgb(0, 0, 238);
+  --accent-color: rgb(250, 250, 250);
+}
+```
+
+### Full example
+
+In the following [Glitch](https://dark-mode-baseline.glitch.me/) embed,
+you can see the complete example that puts the concepts from above into practice.
+Try toggling dark mode in your particular [operating system’s settings](#activating-dark-mode-in-the-operating-system)
```
robdodson: I initially did this on my mac and didn't see any change. It was only after reading further down that I saw that it's only in Chrome 76. I would suggest pulling that paragraph up before this one and possibly using the feature detection snippet in the glitch itself to show some text to the user saying if their browser either supports or does not support dark mode.

src/site/content/en/blog/prefers-color-scheme/index.md
```diff
@@ -0,0 +1,786 @@
...
+has shown that the majority of the surveyed people
+prefer slightly less vibrant and brilliant images when dark mode is active.
+I refer to this as *re-colorization*.
+
+<figure>
+  <div style="width: 100%">
+    <div style="display: inline-block;
+      width: 45%;
+      height: 100px;
+      background-image: url(dark.png);
```
robdodson: I've spent maybe 30 minutes trying to track down these images. I can see them showing up when I preview the site but there are separate files in this directory also called dark.png and light.png and they're different from these images.

image

I'm going to change this so it uses [our columns component](https://web.dev/handbook/web-dev-components/#columns) instead of inline styles. But I think you'll need to swap these images back in.

robdodson: OH! I just looked at the styles. It's using CSS to crop one of the existing images.

Instead I'd suggest using actual screenshots. I'll put the `w-columns` code in place and you can swap out the images.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: I have moved the compatibility paragraph up now. https://github.com/GoogleChrome/web.dev/pull/1021/commits/361af599746d6b691d1e0cd24048bae28f6952b5

tomayac: …and added the feature detection script to the Glitches.

src/site/content/en/blog/prefers-color-scheme/index.md
tomayac: I'm now using the `w-columns` code universally.
https://github.com/GoogleChrome/web.dev/pull/1021/commits/a32246b9621447b0a607db076c8c3c0b5a3c9423 and https://github.com/GoogleChrome/web.dev/pull/1021/commits/164566e5545a30ad15a83a267ad7350d21fd266e.

PR #989, opened 2019-06-07
"Bringing service workers to Google Search" case study

jeffposnick:
Fixes #982

The text content of this post has already been approved internally by the relevant product teams.

**Please hold off on merging** for the time being as we get final sign off on the formatting and metadata, and as @harleenkbatra08 is working on a potential replacement hero image. I wanted to get this staged in the meantime.

PR #1029, opened 2019-06-22
Restore the edits to the Google Search case study

jeffposnick:
Hey @robdodson—I think something happened in the course of [force-pushing](https://github.com/GoogleChrome/web.dev/pull/989#commits-pushed-f4cb485) the last-minute changes to the Google Search SW case study. I just noticed that the text matches my original language, before all of @Meggin's (excellent) edits, so I think you accidentally used my original commit as the base.

Fortunately, I still had a copy of the text with all of the edits in a local branch, and I've created this PR to get the edits reinstated.


robdodson:
oh shoot i'm so sorry about that!

PR #1040, opened 2019-06-25
Add getting started angular article

mgechev:
Introduce the first article from the Angular optimization collection. Minor updates from the original draft:

- Dropped the Glitch instructions until we make them work
- Restored the CLI section and added details on how to proceed with already existing projects from GitHub
- Added "What's next?" without links so far

Header image from [here](https://pixabay.com/photos/painter-paint-cans-brush-paintbrush-1246619/).


mfriesenhahn:
Rather than commenting line-by-line, which can get annoying, I tried just pushing a new commit for this review. The main thing: there were two collection overviews, one on the landing page and one in the Get Started post. I wasn't sure which was desired, but I don't think you need both, so I moved the one from Get Started to the landing page. Please fix it if that's not what you want.

Otherwise, just minor stuff.

mfriesenhahn:
Oh, also: I removed the bit about needing to clone the project and install dependencies in other posts. That's the kind of thing that's hard to remember across contexts--I think it should probably go in the posts where readers need to do it.

mgechev:
I was thinking of this article as the "Introduction" article for the collection.

Thanks for all the fixes!

Looks like the only missing piece before we merge is manually updating the CLA?

mfriesenhahn:
> I was thinking of this article as the "Introduction" article for the collection.

@robdodson, what do you think: should the collection overview statement be on both the landing page and the Getting Started post?

> Looks like the only missing piece before we merge is manually updating the CLA?

I think it's me that made it unhappy. Rob will need to appease it.

robdodson:
> Dropped the Glitch instructions until we make them work
Hey @mgechev, what kind of issues are you hitting with Glitch? Is there anything I can take a look at to help debug? If we switched to an embedded StackBlitz would that resolve things?

robdodson:
> @robdodson, what do you think: should the collection overview statement be on both the landing page and the Getting Started post?

I think just having it once makes sense. So +1 to the current version.

mfriesenhahn:
LGTM! Thanks for all the great work on this and the other posts in the Angular collection, @mgechev! I think it's going to be one of the tightest collections on the site.

mgechev:
The Angular CLI examples are taking minutes to load. We started discussion in the Glitch slack, not sure there's much we can do.

@housseindjirdeh is also talking to the Glitch team.

StackBlitz could be helpful!

robdodson:
I'd love if we could use Glitch but it sounds like it's becoming a real barrier so maybe it would make sense to give StackBlitz a shot. That may be more familiar to the Angular audience anyway. If you want to try that in one of the articles we can see how it goes?

mgechev:
Sounds good!

PR #1049, opened 2019-06-26
Fix all instances of "resilient/resilience"

mfriesenhahn:

- Consistently use "reliable/reliability" rather than "resilient/resilience" across the site.


robdodson:
@jeffposnick just wanted to double check you're cool with this change?

jeffposnick:
👍 Thanks for making this change!

PR #1079, opened 2019-06-28
Add referenced SVG tip to prefers-color-scheme article

tomayac:
`currentColor` is also supported in `<svg><use href="…"></use></svg>` apart from inline SVG. This patch adds this bit of information. Thanks, @xeenon!


tomayac:
Added `updated` field, merging.

PR #1132, opened 2019-07-11
Added a dot at the end of the sentence

kkislay20:
I noticed this sentence is missing a dot.


robdodson:
Hey @kkislay20, the sentence already has a dot but it appears inside of the quotation marks which, I believe, is grammatically correct. Thank you for taking a look though!

PR #1150, opened 2019-07-16
Clarify a versus button usage.

### LINKED ISSUE #1020: Clarify `<a>` vs `<button>` language in
**Describe the bug**
See #1017 

https://web.dev/use-semantic-html
The text in this article is a little unclear when to use an anchor or a button. We mention adding new content to the page is allowed if you're using an anchor, but we should be more specific and say if you're using an anchor to navigate within a SPA and you use routing to update the URL then you should use an anchor.

robdodson:
Fixes #1020 .

- Update the text to clarify when to use an anchor or a button.

### DIFF
--- a/src/site/content/en/accessible/use-semantic-html/index.md
+++ b/src/site/content/en/accessible/use-semantic-html/index.md
@@ -101,7 +101,7 @@ should you choose?
 - If clicking on the element will perform an _action_ on the page, use
   `<button>`.
 - If clicking on the element will _navigate_ the user to a new page then use
-  `<a>`. This includes single page web apps which load new content and update
+  `<a>`. This includes single page web apps that load new content and update
   the URL using the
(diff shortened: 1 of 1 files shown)


src/site/content/en/accessible/use-semantic-html/index.md
```diff
@@ -100,8 +100,10 @@ should you choose?
 
 - If clicking on the element will perform an _action_ on the page, use
   `<button>`.
-- If clicking on the element will _navigate_ the user to a new page (or load new
-  content in a single-page web app) then use `<a>`.
+- If clicking on the element will _navigate_ the user to a new page then use
+  `<a>`. This includes single page web apps which load new content and update
```
mfriesenhahn:
```suggestion
  `<a>`. This includes single page web apps that load new content and update
```

PR #1154, opened 2019-07-17
Change verb to correct grammar

### LINKED ISSUE #1047: Browser inconsistencies with explanation of :focus
**Describe the bug**
The [Style focus article](https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/accessible/style-focus/index.md) explains that *"Regardless of whether you use a mouse to click on it or a keyboard to tab to it, **the button's focus indicator will *always* look the same.**"*

However, this is not the case for all browsers, as we can see below.

* Clicking with the mouse in Chrome:
Chrome behavior
The focus indicator is shown, as explained in the article.

* Clicking with the mouse in Safari:
Safari behavior

* Clicking with the mouse in Firefox:
Firefox behavior
In both Safari and Firefox, the—according to the article, expected—orange focus indicator is not shown.

**Expected behavior**
Add a note about browser inconsistencies.

alexandersandberg:
Fixes grammar issue in 2e2ef6204ff7416433ac5ebfd2aeae547a87e17b.

The updated information to fix #1047 looks great, but I noticed a tiny grammar issue that this PR fixes. 🙂 

> "Because the behavior of focus ~~be~~ **is** inconsistent, ..."

---

Here's a reference from Grammarly:
grammarly-help

### DIFF
--- a/src/site/content/en/accessible/style-focus/index.md
+++ b/src/site/content/en/accessible/style-focus/index.md
@@ -76,7 +76,7 @@ for a summary of which browsers and operating systems will apply focus to
 `<button>` elements.
 {% endAside %}
 
-Because the behavior of focus be inconsistent, it may require a bit of testing
+Because the behavior of focus is inconsistent, it may require a bit of testing
 on different devices to ensure your focus styles are acceptable to your users.
 
 ## Use `:focus-visible` to selectively show a focus indicator


alexandersandberg:
I signed it!

robdodson:
thank you!

PR #1159, opened 2019-07-18
Add note about raw and verbatim tags to handbook.

robdodson:

- Adds additional info to the handbook about escaping code.

### DIFF
--- a/src/site/content/en/handbook/markup-code/index.md
+++ b/src/site/content/en/handbook/markup-code/index.md
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 </label>
 ```
 
+## Escaping code
+If you need to use double `{` braces in your code examples you will need to
+escape them:
+
+```html
+{% raw %}{% raw %}{% verbatim %}{{ item.num | calculate }}{% endverbatim %}{% endraw %}{% endraw %}
+```
+
+The `{% raw %}{% verbatim %}{% endraw %}` tags will be visible when you stage
+locally but not in production.
 
 ## Coding Style
 Indent using two spaces.
--- a/src/site/content/en/handbook/markup-code/index.md
+++ b/src/site/content/en/handbook/markup-code/index.md
@@ -80,7 +80,6 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 
 ## Escaping code
 If you need to use double `{` braces in your code examples you will need to
-escape them:
 
 ```html
 {% raw %}{% raw %}{% verbatim %}{{ item.num | calculate }}{% endverbatim %}{% endraw %}{% endraw %}
--- a/src/site/content/en/handbook/markup-code/index.md
+++ b/src/site/content/en/handbook/markup-code/index.md
@@ -79,12 +79,13 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 ```
 
 ## Escaping special characters
-If you need to use double `{` braces in your code examples you will need to
+To use special characters (for example, double curly brackets) in your code samples, escape them like this:
 
 ```html
 {% raw %}{% raw %}{% verbatim %}{{ item.num | calculate }}{% endverbatim %}{% endraw %}{% endraw %}
 ```
 
+{% Aside %}
 The `{% raw %}{% verbatim %}{% endraw %}` tags will be visible when you stage
 locally but not in production.
 {% endAside %}
--- a/src/site/content/en/handbook/markup-code/index.md
+++ b/src/site/content/en/handbook/markup-code/index.md
@@ -78,7 +78,7 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 </label>
 ```
 
-## Escaping code
+## Escaping special characters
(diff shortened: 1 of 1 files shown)


src/site/content/en/handbook/markup-code/index.md
```diff
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 </label>
 ```
 
+## Escaping code
+If you need to use double `{` braces in your code examples you will need to
+escape them:
```
mfriesenhahn:
```suggestion
```

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
 </label>
 ```
 
+## Escaping code
```
mfriesenhahn:
```suggestion
## Escaping special characters
```

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
...
+If you need to use double `{` braces in your code examples you will need to
```
mfriesenhahn:
```suggestion
To keep the parser from interpreting special characters in your code samples (for example, `{{`), escape them using the `{% raw %}{% verbatim %}` shortcode combination:
```
Trying to generalize and make explicit that both shortcodes are needed. Not sure "parser" is the right word. Feel free to just fix and publish if it isn't.

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
...
+escape them:
+
+```html
+{% raw %}{% raw %}{% verbatim %}{{ item.num | calculate }}{% endverbatim %}{% endraw %}{% endraw %}
+```
+
+The `{% raw %}{% verbatim %}{% endraw %}` tags will be visible when you stage
```
mfriesenhahn:
```suggestion
{% Aside %}
The `{% raw %}{% verbatim %}{% endraw %}` shortcodes will be visible when you stage
```
Are we referring to these as tags or shortcodes? If tags, I need to fix elsewhere in the guide.

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -78,6 +78,16 @@ Show that lines have changed by using the pattern `language/lineNumAdded/lineNum
...
+locally but not in production.
```
mfriesenhahn:
```suggestion
locally but not in production.
{% endAside %}
```

src/site/content/en/handbook/markup-code/index.md
robdodson: I avoided using double curly braces in the example itself because we have to escape them :D

src/site/content/en/handbook/markup-code/index.md
robdodson: Not sure if it's too much information but...

Lower case things like `{% raw %}` are Nunjucks "blocks".
Uppercase things like `{% Aside %}` are eleventy "shortcodes".

src/site/content/en/handbook/markup-code/index.md
mfriesenhahn: Well, <facepalm> for not escaping it (sorry), but I do think it's clearer to just show the double curly braces than write out the phrase. But not a big deal either way.

src/site/content/en/handbook/markup-code/index.md
mfriesenhahn: Ah-ha. "Tags" it is :)

src/site/content/en/handbook/markup-code/index.md
robdodson: Totally agree. Unfortunately because we publish the handbook to devsite I don't think I'm able to show the double curlies without also showing the unescaped `{% verbatim %}` around them. I thought that might confuse folks so I went with "double `{`"

robdodson:
@mfriesenhahn do you think this is ready to merge? If so I think you can hit the button to land it. Btw I typically use the Squash merge feature on GitHub. You can click the down arrow next to the merge button to make sure it's set to use squash merge. This will clean up all of the interim commits.

mfriesenhahn:
> @mfriesenhahn do you think this is ready to merge? If so I think you can hit the button to land it. Btw I typically use the Squash merge feature on GitHub. You can click the down arrow next to the merge button to make sure it's set to use squash merge. This will clean up all of the interim commits.

The edits suggested in GH got messed up, so I had to push a commit to fix the aside. Which means you need to flip the CLA. 😞

robdodson:
CLA BOT CAN'T STOP ME!

image

robdodson:
I had to revert this change because it was breaking the build on devsite but I couldn't figure out why. Even though the verbatim tags are in the correct spot in the output it seems to be breaking for a different reason than before. Pulling this out for now so it doesn't block other things while we sort out a fix.

PR #1196, opened 2019-07-23
V2+dart sass

samthor:
Removes node-sass in favor of sass, which is native JS, rather than compiling the dependent libsass' C++ code. We use it as part of [rollup-plugin-postcss](https://github.com/egoist/rollup-plugin-postcss/issues/146), which supports either.

This allows a much wider range of Node versions to run the site, which makes getting started much easier, at the possible tradeoff of some runtime compile cost.

Also removes some other Gulp deps that didn't seem to be used.


robdodson:
rad thank you!

samthor:
Also: Travis seems unhappy with this PR—but the logs seem to show that the server started and it has just "timed out". So it seems like the check is bad rather than PR causing problems.

PR #1200, opened 2019-07-23
Changed `content-heading` to `content-encoding`

Steffan153:
`content-heading` didn't make sense in this context (it was talking about content encodings, not headers), don't know if it even exists. (in the illustration, it pointed to `content-encoding` too.)

* Changed `content-heading` to `content-encoding`

Thanks.


robdodson:
thank you!

Issue #1180, opened 2019-07-21
Firefox shows red outline for required input on load


alexandersandberg:
**Describe the bug**
Firefox adds a red outline to any input with the `required` attribute on load (see image 1). This leads to inconsistencies between browsers (see image 2).

**To Reproduce**
Visit [/measure](https://web.dev/measure) in Firefox.

**Expected behavior**
Since there's already a warning displayed through an event when the required input is empty, this default warning should probably be removed, or at least made consistent with other browsers.

**Screenshots**

![Screenshot of input warning in Firefox on load](https://user-images.githubusercontent.com/35671299/61596036-90bf2680-abfe-11e9-97e8-ceae7201e9df.png)
_Image 1. What the input looks like in Firefox, on page load._

![Graphic showing inconsistency with Firefox's styling of an input with the required attribute](https://user-images.githubusercontent.com/35671299/61595944-b5ff6500-abfd-11e9-875a-651ac87821c3.jpg)
_Image 2. After clicking 'Run audit' with empty input value, in different browsers._

robdodson:
Just submitted a fix. It will take a few days to roll out. Thanks for flagging this!

PR #1202, opened 2019-07-24
add web-url-chooser{,-container} to choose audit URL

samthor:
Adds `web-url-chooser` and `web-url-chooser-container` to choose URL to Audit. Pulls and saves to Firestore.

This diverges from the DevSite model a bit. Entering a URL has a reasonably complex state machine (e.g., once you "Switch URL", you cannot get your old URL back), and now it's tied to real-time storage (as opposed to in DevSite, which polled occasionally).

### DIFF
--- a/src/lib/components/UrlChooser/_styles.scss
+++ b/src/lib/components/UrlChooser/_styles.scss
@@ -8,24 +8,18 @@
 //
 // =============================================================================
 
-$border-radius: 6px;
-
 web-url-chooser {
-  $padding: 16px;
-  $cardsHeight: 219px;
-  $inputHeight: 56px;
+  $INPUT_HEIGHT: 56px;
 
   display: block;
-  border-radius: $border-radius;
+  border-radius: 6px;
 
-  &[switching] {
-    .lh-enterurl--switch {
-      display: block;
-    }
+  &:not([switching]) .lh-enterurl--switch {
+    display: none;
+  }
 
-    .lh-enterurl--selected {
-      display: none;
-    }
+  &[switching] .lh-enterurl--selected {
+    display: none;
   }
 
   .report_header_enterurl {
@@ -35,35 +29,37 @@ web-url-chooser {
   .lh-enterurl {
     position: relative;
     align-self: center;
+  }
 
-    .lh-enterurl__close {
-      position: absolute;
-      right: 0;
-      top: 0;
-      height: 100%;
-      padding: 0;
-      background: none;
-      box-shadow: none;
-      display: flex;
-      align-items: center;
+  /** The "X" button that appears to the right of a URL. */
+  .lh-enterurl .lh-enterurl__close {
+    position: absolute;
+    right: 0;
+    top: 0;
+    height: 100%;
+    padding: 0;
+    background: none;
+    box-shadow: none;
+    display: flex;
+    align-items: center;
 
-      &::after {
-        @include font-material-icon();
-        content: 'close';
-        color: $GREY_400;
-      }
+    &::after {
+      @include font-material-icon();
+      content: 'close';
+      color: $GREY_400;
+    }
 
-      &:focus,
-      &::-moz-focus-inner {
-        outline: 0;
-      }
+    &:focus,
+    &::-moz-focus-inner {
+      outline: 0;
+    }
 
-      &:focus::after {
-        color: $GREY_900;
-      }
+    &:focus::after {
+      color: $GREY_900;
     }
   }
 
+  /** The currently selected URL, not in editing mode. */
   .lh-enterurl--selected {
     font-weight: 500;
     padding-left: 24px;
@@ -77,12 +73,8 @@ web-url-chooser {
     }
   }
 
-  .lh-enterurl--switch {
-    display: none;
-  }
-
   .lh-input {
-    height: $inputHeight;
+    height: $INPUT_HEIGHT;
     width: 100%;
     transition: border 200ms ease-in-out;
     outline: none;
@@ -97,16 +89,21 @@ web-url-chooser {
     display: flex;
     flex-wrap: wrap;
     margin-top: 8px;
-
-    .w-button {
-      // List all vals so autoprefixer doesn't rewrite this to flex: 1 1 0%
-      // and make "switch url" text ellipsis.
-      flex: 1 1 auto;
-      height: $inputHeight;
-      justify-content: center;
-    }
   }
 
+  .lh-controls .w-button {
+    // List all vals so autoprefixer doesn't rewrite this to flex: 1 1 0%
+    // and make "switch url" text ellipsis.
+    flex: 1 1 auto;
+    height: $INPUT_HEIGHT;
+    justify-content: center;
+  }
+}
+
+web-url-chooser {
+  /// @deprecated
+  // TODO(robdodson): This should be replaced with the @media rules from breakpoints.scss, but the
+  // Measure page is special and currently does not have a matching value.
   @media screen and (min-width: $BREAKPOINT_VALUE_TABLET) {
     .report_header_enterurl {
       display: flex;
--- a/src/lib/components/UrlChooser/_styles.scss
+++ b/src/lib/components/UrlChooser/_styles.scss
@@ -1,0 +1,129 @@
+@import '../../../styles/settings/colors';
+@import '../../../styles/tools/mixins';
+
+// =============================================================================
+// URL CHOOSER
+//
+// Lets the user select a URL and trigger an audit.
+//
+// =============================================================================
+
+$border-radius: 6px;
+
+web-url-chooser {
+  $padding: 16px;
+  $cardsHeight: 219px;
+  $inputHeight: 56px;
+
+  display: block;
+  border-radius: $border-radius;
+
+  &[switching] {
+    .lh-enterurl--switch {
+      display: block;
+    }
+
+    .lh-enterurl--selected {
+      display: none;
+    }
+  }
+
+  .report_header_enterurl {
+    position: relative;
+  }
+
+  .lh-enterurl {
+    position: relative;
+    align-self: center;
+
+    .lh-enterurl__close {
+      position: absolute;
+      right: 0;
+      top: 0;
+      height: 100%;
+      padding: 0;
+      background: none;
+      box-shadow: none;
+      display: flex;
+      align-items: center;
+
+      &::after {
+        @include font-material-icon();
+        content: 'close';
+        color: $GREY_400;
+      }
+
+      &:focus,
+      &::-moz-focus-inner {
+        outline: 0;
+      }
+
+      &:focus::after {
+        color: $GREY_900;
+      }
+    }
+  }
+
+  .lh-enterurl--selected {
+    font-weight: 500;
+    padding-left: 24px;
+    // image is duplicated in lh-input.scss
+    background: url('data:image/svg+xml;utf8,<svg width="18px" height="18px" xmlns="http://www.w3.org/2000/svg"><path fill="%23212121" d="M7.992,0 C3.576,0 0,3.584 0,8 C0,12.416 3.576,16 7.992,16 C12.416,16 16,12.416 16,8 C16,3.584 12.416,0 7.992,0 Z M13.536,4.8 L11.176,4.8 C10.92,3.8 10.552,2.84 10.072,1.952 C11.544,2.456 12.768,3.48 13.536,4.8 Z M8,1.632 C8.664,2.592 9.184,3.656 9.528,4.8 L6.472,4.8 C6.816,3.656 7.336,2.592 8,1.632 Z M1.808,9.6 C1.68,9.088 1.6,8.552 1.6,8 C1.6,7.448 1.68,6.912 1.808,6.4 L4.512,6.4 C4.448,6.928 4.4,7.456 4.4,8 C4.4,8.544 4.448,9.072 4.512,9.6 L1.808,9.6 Z M2.464,11.2 L4.824,11.2 C5.08,12.2 5.448,13.16 5.928,14.048 C4.456,13.544 3.232,12.528 2.464,11.2 Z M4.824,4.8 L2.464,4.8 C3.232,3.472 4.456,2.456 5.928,1.952 C5.448,2.84 5.08,3.8 4.824,4.8 Z M8,14.368 C7.336,13.408 6.816,12.344 6.472,11.2 L9.528,11.2 C9.184,12.344 8.664,13.408 8,14.368 Z M9.872,9.6 L6.128,9.6 C6.056,9.072 6,8.544 6,8 C6,7.456 6.056,6.92 6.128,6.4 L9.872,6.4 C9.944,6.92 10,7.456 10,8 C10,8.544 9.944,9.072 9.872,9.6 Z M10.072,14.048 C10.552,13.16 10.92,12.2 11.176,11.2 L13.536,11.2 C12.768,12.52 11.544,13.544 10.072,14.048 Z M11.488,9.6 C11.552,9.072 11.6,8.544 11.6,8 C11.6,7.456 11.552,6.928 11.488,6.4 L14.192,6.4 C14.32,6.912 14.4,7.448 14.4,8 C14.4,8.552 14.32,9.088 14.192,9.6 L11.488,9.6 Z" id="icon/action/language_24px"></path></svg>') no-repeat 0 50%;
+
+    &::before {
+      // include zero-width space to force minimum height
+      content: '\200b';
+      user-select: none;
+    }
+  }
+
+  .lh-enterurl--switch {
+    display: none;
+  }
+
+  .lh-input {
+    height: $inputHeight;
+    width: 100%;
+    transition: border 200ms ease-in-out;
+    outline: none;
+    background-color: $WHITE;
+
+    &::-ms-clear {
+      display: none;
+    }
+  }
+
+  .lh-controls {
+    display: flex;
+    flex-wrap: wrap;
+    margin-top: 8px;
+
+    .w-button {
+      // List all vals so autoprefixer doesn't rewrite this to flex: 1 1 0%
+      // and make "switch url" text ellipsis.
+      flex: 1 1 auto;
+      height: $inputHeight;
+      justify-content: center;
+    }
+  }
+
+  @media screen and (min-width: $BREAKPOINT_VALUE_TABLET) {
+    .report_header_enterurl {
+      display: flex;
+    }
+
+    .lh-controls {
+      margin-top: 0;
+      flex-wrap: nowrap;
+    }
+
+    .lh-enterurl {
+      margin-right: 32px;
+      width: 100%;
+    }
+
+    .lh-input {
+      width: 100%;
+    }
+  }
+}
--- a/src/lib/components/UrlChooser/_styles.scss
+++ b/src/lib/components/UrlChooser/_styles.scss
@@ -22,7 +22,7 @@ web-url-chooser {
     display: none;
   }
 
-  .report_header_enterurl {
+  .lh-report-header-enterurl {
     position: relative;
   }
 
@@ -32,7 +32,7 @@ web-url-chooser {
   }
 
   /** The "X" button that appears to the right of a URL. */
-  .lh-enterurl .lh-enterurl__close {
+  .lh-enterurl__close {
     position: absolute;
     right: 0;
     top: 0;
@@ -42,6 +42,11 @@ web-url-chooser {
     box-shadow: none;
     display: flex;
     align-items: center;
+    justify-content: center;
+
+    &[disabled] {
+      display: none;
+    }
 
     &::after {
       @include font-material-icon();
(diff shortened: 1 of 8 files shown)


src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
+import {html} from "lit-element";
+import {store} from "../../store";
+import {BaseElement} from "../BaseElement";
+import {userRef, firebase} from "../../fb";
+
+/* eslint-disable require-jsdoc */
```
robdodson: Would you mind adding a fileoverview to help explain what this class does?

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+      if (ref) {
+        const p = ref.set(
+          {
+            userUrl: url,
+            userUrlUpdate: firebase.firestore.FieldValue.serverTimestamp(),
+          },
+          {merge: true},
+        );
+        p.catch((err) => {
+          // TODO: Firestore errors are problematic but we can still run with the new URL.
```
robdodson: Is there a TODO here or is this more of an informational note about how to handle the failure?

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+          console.warn("could not write URL", err);
+        });
+      }
+
+      this.switching = false;
+    }
+
+    // TODO: actually run Lighthouse. Currently just freezes the page.
```
robdodson: Should this element be running Lighthouse or should it just tell the outside world that it has a URL so a container component knows to then go run Lighthouse.

robdodson: Btw I should mention that for the ProfileSwitcher I used the React notion of Container components (Polymer also often refers to this as Mediator components). https://medium.com/@learnreact/container-components-c0e67432e005

So I was wondering if we'd want to do the same thing with the UrlChooser. It would be a rather dumb component (just moving through its various states depending on if the user has put in a URL or not) and otherwise it would communicate up to its Container when it was time to run Lighthouse.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+    if (document.activeElement !== input) {
+      // nb. will fail in Shadow DOM mode
+      window.requestAnimationFrame(() => {
+        input.focus();
+      });
+    }
+  }
+
+  onUrlKeyup(e) {
+    if (e.keyCode === /* ESC */ 27) {
```
robdodson: can we use `e.key == "Escape"` ?
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+      this.clearInput();
+    } else if (e.keyCode === /* ENTER */ 13) {
```
robdodson: same comment as above.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
+import {html} from "lit-element";
+import {store} from "../../store";
+import {BaseElement} from "../BaseElement";
+import {userRef, firebase} from "../../fb";
+
+/* eslint-disable require-jsdoc */
+class UrlChooser extends BaseElement {
+  static get properties() {
+    return {
+      href: {type: String},
```
robdodson: should we call this `this.url` instead? Seems like that's how we otherwise refer to it with local variables.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+      switching: {type: Boolean, reflect: true},
+      active: {type: Boolean},
+    };
+  }
+
+  constructor() {
+    super();
+    this.href = null; // when signed out or waiting for Firestore, this is null
+    this.switching = true;
```
robdodson: Can you add a similar comment like the one above to explain what `switching` is for

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+    this.active = false;
```
robdodson: Same comment as above

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+    // non-properties
```
robdodson: Are these private properties? If so maybe we underscore them?

src/lib/components/UrlChooser/index.js
robdodson: One last discussion point: I think a common practice for Container components is that they are the ones who talk to redux/unistore and handle application state. Their children just handle local state and inform the Container when they've changed. So one question would be if we decide to Containerize this thing, if we would want to move the calls to the store out of this widget and into its Container.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+      this.switching = false;
+    }
+
+    // TODO: actually run Lighthouse. Currently just freezes the page.
+    this.active = true;
+    console.warn("web-url-chooser in terminal state: should run Lighthouse");
+  }
+
+  switchUrl(ev) {
```
robdodson:
```suggestion
  switchUrl(e) {
```

I think elsewhere in the codebase I've been using `e` so maybe just nice to be consistent.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+      // if the user has signed out, clear the href and enter switching mode
+      this.href = null;
+      this.switching = true;
+    } else if (state.userUrl && !this.switching) {
+      // in all other cases, only update the URL if ther user isn"t switching
+      this.href = state.userUrl;
+    }
+  }
+
+  runAudit(ev) {
```
robdodson:
```suggestion
  runAudit(e) {
```

src/lib/components/UrlChooser/index.js
robdodson: On second thought I think I definitely want to pull the actual audit running logic out of this component. It feels like it's overloaded to both have it render the UI and also handle to logic for calling out to Lighthouse.

It also feels like when Lighthouse is running we should reflect that as a state in the store. There are other widgets that may want that info if we decide to break things up into smaller UI chunks.

I think we could put the logic that talks to Lighthouse either into a state action function, or into its own dedicated class. This widget would signal to the Container that it's time to rock and the container would either call the state action or call out to the class object.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+    this.href = url;
+    if (url !== this.urlInput.value) {
+      // Could also call this.update() here, this just short-circuits the change.
+      this.urlInput.value = url;
+    }
+
+    if (!this.urlInput.validity.valid) {
+      const detail = `Invalid URL. Please enter a full URL starting with https://.`;
+      const ce = new CustomEvent("web-error", {bubbles: true, detail});
```
robdodson: mind calling this `event` or `newEvent`?

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+  onUrlKeyup(e) {
+    if (e.keyCode === /* ESC */ 27) {
+      this.clearInput();
+    } else if (e.keyCode === /* ENTER */ 13) {
+      this.runLighthouseButton.click();
+    }
+  }
+
+  fixUpAndValidateUrl() {
```
robdodson: mind breaking this into two functions? One that formats the URL and one that verifies it?

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+  updated() {
+    this.urlInput = this.renderRoot.querySelector('input[type="url"]');
+    this.runLighthouseButton = this.renderRoot.querySelector("#run-lh-button");
+  }
+
+  onStateChanged() {
+    const state = store.getState();
+    const urlInputValue = this.urlInput ? this.urlInput.value : null;
+
+    if (this.active) {
```
robdodson: This seems like something that would be better tracked as state in the store (since other things may depend on it) versus the component itself. Although I do realize that the component might change its visual state when LH is running. Perhaps that info should be passed down to the component. See my comments below about Containers and state.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+        <div class="lh-enterurl lh-enterurl--selected">${this.href}</div>
+        <div class="lh-enterurl lh-enterurl--switch">
+          <input
+            type="url"
+            class="lh-input"
+            .value="${this.href}"
+            placeholder="Enter a web page URL"
+            pattern="https?://.*"
+            minlength="7"
+            required
```
robdodson: Remove required per https://github.com/GoogleChrome/web.dev/issues/1180
I think(?) required is only supposed to be added after a validation check has failed.

src/lib/components/_all.scss
```diff
@@ -1,3 +1,4 @@
 @import './ProfileSwitcher/styles';
 @import './ProgressBar/styles';
-@import './SparklineChart/styles';
\ No newline at end of file
+@import './SparklineChart/styles';
+@import './UrlChooser/styles';
```
robdodson: newline

robdodson: did the linter not grump about this? maybe I broke it at some point... I think it used to block on this.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+
+      // write URL to firestore and local
+      const url = this.urlInput.value;
+      store.setState({userUrl: url});
+      const ref = userRef();
+      if (ref) {
+        const p = ref.set(
+          {
+            userUrl: url,
+            userUrlUpdate: firebase.firestore.FieldValue.serverTimestamp(),
```
robdodson: can this be turned into a method on the `fb` object? I'm wondering if it would be better for that module to provide actions for external consumers as opposed to widgets calling into it directly.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,185 @@
...
+      if (!this.fixUpAndValidateUrl()) {
+        return;
+      }
+
+      // write URL to firestore and local
+      const url = this.urlInput.value;
+      store.setState({userUrl: url});
+      const ref = userRef();
+      if (ref) {
+        const p = ref.set(
```
robdodson: what does p stand for?

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,129 @@
...
+@import '../../../styles/tools/mixins';
+
+// =============================================================================
+// URL CHOOSER
+//
+// Lets the user select a URL and trigger an audit.
+//
+// =============================================================================
+
+$border-radius: 6px;
```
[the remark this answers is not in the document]
samthor: Done. I've actually removed `$BORDER_RADIUS` because it was only used one place.

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,129 @@
...
+    .lh-enterurl {
+      margin-right: 32px;
+      width: 100%;
+    }
+
+    .lh-input {
+      width: 100%;
+    }
+  }
+}
```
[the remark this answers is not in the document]
samthor: It was pretty grumpy about literally everything else.

src/lib/components/_all.scss
samthor: Fixed

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,129 @@
...
+    .w-button {
+      // List all vals so autoprefixer doesn't rewrite this to flex: 1 1 0%
+      // and make "switch url" text ellipsis.
+      flex: 1 1 auto;
+      height: $inputHeight;
+      justify-content: center;
+    }
+  }
+
+  @media screen and (min-width: $BREAKPOINT_VALUE_TABLET) {
```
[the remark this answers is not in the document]
samthor: Done

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,129 @@
...
+
+web-url-chooser {
+  $padding: 16px;
+  $cardsHeight: 219px;
+  $inputHeight: 56px;
+
+  display: block;
+  border-radius: $border-radius;
+
+  &[switching] {
```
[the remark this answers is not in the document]
samthor: So I actually disagree with this somewhat. I'm happy to be overruled.

But this file has about ~15-20 rules, many of which are long (because of `::before`, `::-moz-focus-inner`, other specials). I've _flattened_ it a bit, but I think nesting still gains us a bit.

Additionally, the only way to have scoped variables in Sass is to nest. So we define `$INPUT_HEIGHT` at the top inside `web-url-chooser`, then use it here only.

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done (I plan to show errors as part of `web-profile` or whatever later)

src/lib/components/UrlChooser/index.js
samthor: Done (well I've inlined the verifiying code)

src/lib/components/UrlChooser/index.js
samthor: I've broken them apart.

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: It now fires `audit` which requests that the controller do something, which could then talk to the state action or class object. For now though I've left that part as a TODO.

src/lib/components/UrlChooser/index.js
samthor: A bit of both. My point is we don't really care if this fails, because we're still going to run Lighthouse, but the user won't see the same URL on their next visit.

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
samthor: Promise.

src/lib/components/UrlChooser/index.js
samthor: Yes, but I've left it for now.

src/lib/components/UrlChooser/index.js
robdodson: Maybe the comment should say "Note:"? I guess since we don't have any plans to change this behavior

src/lib/components/UrlChooser/index.js
robdodson: Would you be ok with turning it into an exported function? My concern is that we may forget to do so.

src/lib/components/UrlChooserContainer/index.js
```diff
@@ -0,0 +1,78 @@
...
+    // if Lighthouse is active, ignore changes to the URL from Firestore
+    this.url = state.activeLighthouseUrl || state.userUrl;
+    this.active = state.activeLighthouseUrl !== null;
+  }
+
+  runAudit(e) {
+    const url = e.detail;
+
+    // write URL to firestore and local
+    // TODO(samthor): This should happen in a controller setting.
```
robdodson: I think the practice with Redux style apps is to create a function that performs the work and sets the state as a result. Then the various widgets call that function without having to do the state management themselves.

So store.js might export a `runAudit` function that sets a state flag to let the app know it's auditing, then does the fetch to lighthouse, then sets a state flag when it's done auditing and finally sets the state for the audit results.

src/lib/components/UrlChooserContainer/index.js
```diff
@@ -0,0 +1,78 @@
...
+        .active=${this.active}
+        @audit=${this.runAudit}
+      ></web-url-chooser>
+    `;
+  }
+
+  onStateChanged() {
+    const state = store.getState();
+
+    // if Lighthouse is active, ignore changes to the URL from Firestore
```
robdodson: is this checking if lighthouse is running or not? Maybe we should have explicit state flags for `isLighthouseAuditing` or something along those lines? Currently I'm having trouble telling at a glance what the relationship between `activeLighthouseUrl` and `active` is.

src/lib/components/UrlChooserContainer/index.js
```diff
@@ -0,0 +1,78 @@
...
+    this.active = state.activeLighthouseUrl !== null;
+  }
+
+  runAudit(e) {
+    const url = e.detail;
+
+    // write URL to firestore and local
+    // TODO(samthor): This should happen in a controller setting.
+    store.setState({
+      userUrl: url,
```
robdodson: Can you help me understand the difference between `userUrl` and `activeLighthouseUrl`?

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,126 @@
...
+
+  &:not([switching]) .lh-enterurl--switch {
+    display: none;
+  }
+
+  &[switching] .lh-enterurl--selected {
+    display: none;
+  }
+
+  .report_header_enterurl {
```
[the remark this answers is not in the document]
samthor: Changed to `.lh-report-header-enterurl`

src/lib/components/UrlChooser/_styles.scss
```diff
@@ -0,0 +1,126 @@
...
+  .lh-enterurl .lh-enterurl__close {
+    position: absolute;
+    right: 0;
+    top: 0;
+    height: 100%;
+    padding: 0;
+    background: none;
+    box-shadow: none;
+    display: flex;
+    align-items: center;
```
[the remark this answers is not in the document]
samthor: Yeah, that puts the X in the horizontal center. So I suppose that's better!

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,158 @@
...
+          >
+            Run Audit
+          </button>
+        </div>
+      </div>
+    `;
+  }
+
+  updated(changedProperties) {
+    this._urlInput = this.renderRoot.querySelector('input[type="url"]');
```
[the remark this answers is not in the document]
samthor: I always imagined that there was no guarantee of node stability. I guess if you have loops and dynamic stuff that is true, but we're just stamping it once.

I've moved it there.

src/lib/components/UrlChooser/index.js
samthor: Done

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,158 @@
...
+    } else if (!this.switching) {
+      // in all other cases, only update the URL if ther user isn't switching
+      this._urlInput.value = updateUrl;
+    }
+  }
+
+  requestAudit() {
+    let url = this.url;
+
+    if (this.switching) {
```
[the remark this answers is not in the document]
samthor: Done

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,158 @@
...
+        </div>
+      </div>
+    `;
+  }
+
+  updated(changedProperties) {
+    this._urlInput = this.renderRoot.querySelector('input[type="url"]');
+    this._runLighthouseButton = this.renderRoot.querySelector("#run-lh-button");
+
+    if (!changedProperties.has("url")) {
```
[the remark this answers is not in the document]
samthor: As discussed in the other PR, property change handlers are pure functions. I've made this a getter/setter for `url`.

src/lib/components/UrlChooser/index.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    const event = new CustomEvent("audit", {detail: url});
+    this.dispatchEvent(event);
+  }
+
+  switchUrl() {
+    this.switching = true;
+
+    const input = this._urlInput;
```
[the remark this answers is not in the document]
samthor: I've done something different. I've realised that the awkward `requestAnimationFrame` is actually because we haven't rendered yet.

So I've added a `updated()` handler which focuses on the `urlInput` if `switching` is now true, and made the lit-html template just set `switching = true` as an inline handler.

src/lib/components/UrlChooser/index.js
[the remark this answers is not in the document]
samthor: Actually I know I resolved this but it doesn't work.

We need to be able to set the value on `<input />` and we don't have it until after a first update. We don't want to have a property for that as it might blindly overwrite what the user is currently typing.

I can't think of a better solution than this?

src/lib/components/UrlChooserContainer/index.js
robdodson: ok

PR #1216, opened 2019-07-29
web-sparkline-chart

samthor:
Get sparkline working in v2.

This is a reasonably big change and makes quality-of-life improvements over v1.

* uses `ResizeObserver` (or fallback to `resize`) to resize the element
  * does not need `scoreHeight` and friends—reads size from element size
* better failure cases (supports single point, no points)
* does not show "Invalid Date" on poor data

Screen Shot 2019-07-29 at 13 31 27


src/lib/components/SparklineChart/_styles.scss
```diff
@@ -61,10 +70,21 @@ web-sparkline-chart {
...
+  .sl-cursor {
+    stroke-linecap: round;
+  }
+
   .path {
-    stroke-dasharray: 1000;
-    stroke-dashoffset: 1000;
-    animation: sparkline-dash-animation 300ms linear forwards;
+    stroke-linecap: round;
+    // TODO(robdodson): This doesn't animate even in the old site.
```
robdodson: I'm cool just removing this stuff. I think Eric added it originally and my knowledge of SVG animation is nil :P

src/lib/components/SparklineChart/index.js
```diff
@@ -571,64 +554,19 @@ class SparklineChart extends BaseElement {
...
-        const gradients = this.getElementsByClassName("gradient");
-        for (const gradient of gradients) {
-          gradient.classList.add("fadein");
-        }
-      });
-    }
+  firstUpdated() {
+    this.cursorElement_ = this.renderRoot.querySelector("#cursor");
+    this.scoreElement_ = this.renderRoot.querySelector("#score");
+    this.announcerElement_ = this.querySelector(".sr-announcer");
```
robdodson: maybe we should turn this into an id as well?

src/lib/components/SparklineChart/index.js
```diff
@@ -39,187 +41,121 @@ function clampTo2Decimals(val) {
...
   }
 
-  /**
-   * @return {boolean}
-   * @export
-   */
-  get fill() {
-    return this.hasAttribute("fill");
-  }
+  set demo(demo) {
```
robdodson: Do you want to keep this or remove it before we land?

src/lib/components/SparklineChart/_styles.scss
samthor: Removed

src/lib/components/SparklineChart/index.js
samthor: as discussed in chat—no!

src/lib/components/SparklineChart/index.js
samthor: I'm less worried since this isn't hundreds of kbs of data.

PR #1219, opened 2019-07-29
remove -webkit- references

argyleink:
see userland comment https://twitter.com/simevidas/status/1154960320487727105?s=20

having -webkit- in the code examples creates more questions than it helps answer

### DIFF
--- a/src/site/content/en/blog/backdrop-filter/index.md
+++ b/src/site/content/en/blog/backdrop-filter/index.md
@@ -82,14 +82,13 @@ You can combine filters for rich and clever effects, or use just one filter for 
 As with many features of the modern web, you'll want to know whether the user's browser supports `backdrop-filter` before using it. Do this with `@supports()`. For performance reasons, fall back to an image instead of a polyfill when `backdrop-image` isn't supported. The example below shows this.
 
 ```css
-@supports (-webkit-backdrop-filter: none) or (backdrop-filter: none) {
+@supports (backdrop-filter: none) {
 	.background {
 		backdrop-filter: blur(10px);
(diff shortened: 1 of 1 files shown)


mfriesenhahn:
Should we remove the reference to the `-webkit-` prefix from the copy in the second para as well?

argyleink:
i felt the reference in the 2nd paragraph was fine since it's contextually correct due to the "recent years" reference. it's also not code that someone may copy/paste 🤷‍♂ 

tldr; it's fine as a historical reference, but shouldnt be in our code blocks

mfriesenhahn:
> i felt the reference in the 2nd paragraph was fine since it's contextually correct due to the "recent years" reference. it's also not code that someone may copy/paste 🤷‍♂
> 
> tldr; it's fine as a historical reference, but shouldnt be in our code blocks

Cool, just wanted to double-check.

simevidas:
If possible, add a note about using Autoprefixer to generate the prefixes, since they are still needed for Safari. 

Ideally, this would become a global guideline for all articles on web.dev, but that’s a request for the person(s) in charge of the site 😅

argyleink:
i think the web.dev spirit would be to write a post on autoprefixer: the what, the why and the how. that way it can stand alone and be referenced. perhaps you want to write one!? this is an open blog you can submit to.

robdodson:
Might be worth adding a quick note in the meantime, something like:

```
{% Aside %}
Some older browsers require a vendor prefixed version of this property. You can use a tool like [Autoprefixer](https://css-tricks.com/autoprefixer/) to automatically add these prefixes to your CSS.
{% endAside %}
```

PR #1241, opened 2019-08-01
Add new cases for the Instruction component

### LINKED ISSUE #1133: Add new cases to Instruction widget
- The Instruction widget includes a `reload` case that contains an instruction to reload embedded sample apps, but the wording doesn't make sense for reloading a full page.
- There's no case for opening the DevTools **Command** menu, which is needed fairly often.

- Add a `devtools-command` case that reads: "Press `Control+Shift+P` (or `Command+Shift+P` on Mac) to open the **Command** menu."
- Change the name of the current `reload` case to `reload-app`.
- Add a new `reload-page` case that reads: "Reload the page."

mihajlija:
Fixes #1133

- Add new Glitch instruction
- Add separate instructions for reloading an app and a simple page
- Add DevTools Command menu instruction

### DIFF
--- a/src/site/_includes/components/Instruction.js
+++ b/src/site/_includes/components/Instruction.js
@@ -82,14 +82,24 @@ module.exports = (type, listStyle = 'ul') => {
       instruction = html`${bullet}Select the **Disable cache** checkbox.`;
       break;
 
-    case 'reload':
+    case 'reload-app':
       instruction = html`${bullet}Reload the app.`;
+      break;
+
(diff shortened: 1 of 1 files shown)


mihajlija:
@robdodson Thanks for the VS Code shortcut tip, now I just have to remember to use it consistently :)

robdodson:
yeah maybe I can figure out why it's not automated. With so many dotfiles in the project you would think one of them would do the right thing 😆

PR #1252, opened 2019-08-02
Update code samples on /handbook/web-dev-components

mfriesenhahn:
Add {% verbatim %} tags to code samples on /handbook/web-dev-components to make devsite happy.


robdodson:
oh I think I did this already :)

PR #1255, opened 2019-08-02
Wires up controller for v2

samthor:
Adds actions, CI code, and a `<web-lighthouse-scores-meta>` element (to show download links, errors).

Not perfect, just took the path of least resistance in order to get a review. The state/controller logic could do with special scrutiny. I'm trying to ensure that there's not multiple requests in-flight at once, and deal with the fact that the user's targeted URL can change while a Lighthouse run is happening.


src/lib/controller.js
```diff
@@ -0,0 +1,79 @@
+import {store} from "./store";
+import {updateUrl} from "./fb";
+import {runLighthouse, fetchReports} from "./ci";
+
+class Controller {
```
robdodson: Instead of creating a `Controller` class, what do you think of turning these into Redux/Unistore actions?

https://github.com/developit/unistore/blob/master/README.md#action

I think that would go something like:

```
export const requestRunLighthouse = store.action(async (state) => {
  .. do a bunch of async stuff ...
  return {
    userUrl: url,
    ...
  };
});
```

Maybe we'd create an actions directory and have one action per file?

src/lib/components/LighthouseScoresMeta/index.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * @fileoverview Shows meta information about a Lighthouse run, including an optional error.
+ */
+
+import {html} from "lit-element";
+import {BaseElement} from "../BaseElement";
+import {CI_HOST} from "../../ci";
```
robdodson: `LH_HOST`

src/lib/ci.js
```diff
@@ -0,0 +1,47 @@
+
+export const CI_HOST = "https://lighthouse-dot-webdotdevsite.appspot.com";
```
robdodson: `LH_HOST`

and maybe we rename this file to lighthouse-service.js or something...?

src/lib/ci.js
```diff
@@ -0,0 +1,47 @@
...
+
+export async function fetchMedians() {
```
robdodson: Should we comment this out for now with a TODO that it's broken?

src/lib/ci.js
```diff
@@ -0,0 +1,47 @@
...
+  const resp = await window.fetch(`${CI_HOST}/lh/medians?url=all`);
+  return await resp.json();
+}
+
+export async function runLighthouse(url, signedIn=false) {
+  const body = {
+    url,
```
robdodson: Can you comment these params? I'm not sure I grok what `replace` is doing?

src/lib/components/LighthouseScoresContainer/index.js
```diff
@@ -6,39 +8,79 @@
...
+
+    this.metaElement = null;
+    this.statsElement = null;
+    this.metricsElement = null;
+    this.auditsElement = null;
+
+    store.subscribe(this.onStateChanged.bind(this));
+    this.onStateChanged();
+
+    this.onCategoryChanged = (ev) => {
```
robdodson:
```suggestion
    this.onCategoryChanged = (e) => {
```

src/lib/components/LighthouseScoresContainer/index.js
```diff
@@ -6,39 +8,79 @@
...
+      this.auditsElement.filteringOn = ev.detail;
```
robdodson:
```suggestion
      this.auditsElement.filteringOn = e.detail;
```

src/lib/components/LighthouseScoresContainer/index.js
```diff
@@ -6,39 +8,79 @@
...
-    // LitElement, so wire things up once it's connected, and keep track of whether it's done.
-    if (this.firstUpdated) {
+  onStateChanged() {
+    const {
+      lighthouseError,
+      lighthouseResult,
+      activeLighthouseUrl,
+    } = store.getState();
+
+    if (!lighthouseResult) {
```
robdodson: should we also check that `lighthouseError` exists?

src/lib/components/LighthouseScoresContainer/index.js
```diff
@@ -6,39 +8,79 @@
...
+      // TODO: clear data?
```
robdodson: is this asking if we should clear out the user's urls in the store or in the input field? Seems like we'd want to leave everything in place and just show the error message so they can tweak what they've already typed in. But I'm probably misunderstanding :)

src/lib/components/LighthouseScoresMeta/_styles.scss
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /// @deprecated
+  @media screen and (max-width: $BREAKPOINT_VALUE_TABLET) {
+    .lh-report-meta__links {
+      display: flex;
+      justify-content: space-between;
+    }
+  }
+}
```
robdodson: newline

src/lib/ci.js
samthor: I've just added a TODO. There's nothing inherently wrong with the call, I think?

src/lib/ci.js
samthor: I'm not sure either—it was from the old codebase. I've left you a TODO :)

src/lib/components/LighthouseScoresContainer/index.js
samthor: Your "Seems like..." is what happens now. I'll update that so it's clear.

src/lib/components/LighthouseScoresContainer/index.js
samthor: Well, it's null if it doesn't, which won't display anything. So it's moot?

src/lib/controller.js
samthor: FWIW there's no actions-in-serial behavior here. I should have expected that from a @developit 350 byte special ;-)

samthor: Done, take a look.

src/lib/fb.js
```diff
@@ -61,26 +79,62 @@ export function userRef() {
...
-    {merge: true},
-  );
-  p.catch((err) => {
+    };
+
+    // TODO(robdodson): Users can pass (examples for the same site):
+    //   "https://google.com/", "http://google.com", "https://www.google.com/?foo"
+    // Is it worth simply using the bare hostname here? (would still have www. vs not)
+
+    const prevSeen = (data.userUrlSeen && data.userUrlSeen[url]) || null;
```
robdodson: In firebase, what do you think of renaming the properties to:

```
users: {
  [some_user_id]: {
    currentUrl: example.com,
    previousUrls: {
      foo.com: 2019-10-31,
      ...
    }
  }
}
```

I guess as I've been reading it's been a little hard for me to quickly know what userUrl and userUrlSeen were referring to. Open to other suggestions.

src/lib/fb.js
```diff
@@ -61,26 +79,62 @@ export function userRef() {
...
 
-export function updateUrl(url) {
+/**
+ * Update's the user's row in Firestore (if signed in) with an updated URL and optional audit time.
+ *
+ * @param {string} url to update the user's row with
+ * @param {!Date} auditedOn of the most recent Lighthouse run
+ * @return {!Date} the earliest audit seen for this URL
+ */
+export async function updateUrl(url, auditedOn = null) {
```
robdodson: I was trying to think of a more specific name for this function since "update" could mean a number of things.
Some options... `setCurrentUrl`, `replaceCurrentUrl`, `addOrReplaceCurrentUrl`?

src/lib/fb.js
```diff
@@ -40,10 +41,27 @@ firebase.auth().onAuthStateChanged((user) => {
...
+      const userUrlSeen = prevSeen ? prevSeen.toDate() : null;
+
+      console.info("snapshot: got userUser", userUrl, userUrlSeen);
       store.setState({
-        userUrl: data.userUrl || "",
+        userUrlSeen,
+        userUrl,
       });
+
+      // TODO(samthor): This will request reports as soon as the user is signed in, regardless of
```
robdodson: Could this be moved to router.js?

src/lib/controller.js
```diff
@@ -0,0 +1,85 @@
...
+  // off a Lighthouse run here. We store the active URL running through
+  // Lighthouse as their preferred URL could change because of another
+  // browser, and this prevents results being appropritioned to the wrong URL.
+  store.setState({
+    userUrl: url,
+    activeLighthouseUrl: url,
+    lighthouseError: null,
+  });
+
+  return serialLighthouseTask(async (state) => {
```
robdodson: Another possibility is compare the LH result with whatever the local user url is before displaying the results. If there's a mismatch just throw an error. That might let us remove some of the serialization code to simplify things a bit.

src/lib/fb.js
samthor: So we also use the "seen" dictionary to tell us when the current URL was first added. So "previousUrls" isn't quite accurate, but if you're happy with using it anyway, I am happy to make the change.

src/lib/fb.js
```diff
@@ -61,26 +79,62 @@ export function userRef() {
...
+    const update = {
       userUrl: url,
-      userUrlUpdate: firebase.firestore.FieldValue.serverTimestamp(),
-    },
-    {merge: true},
-  );
-  p.catch((err) => {
+    };
+
+    // TODO(robdodson): Users can pass (examples for the same site):
```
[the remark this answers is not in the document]
samthor: I think there's two goals here.

1. To have a common "high water mark" for the first time a user has scanned a URL. It's not specifically about returning the _same_ results, just more about—if I've scanned http://google.com a month ago, maybe I can also get access to https://google.com from a month ago.

2. There are some small ambiguities which we could resolve—google.com vs google.com/, for instance.

src/lib/fb.js
```diff
@@ -61,26 +79,62 @@ export function userRef() {
...
+
+    const prevSeen = (data.userUrlSeen && data.userUrlSeen[url]) || null;
+    if (prevSeen) {
+      // nb. There's already a valid timestamp here, so don't replace it with a future time,
+      // but grab it so we can inform a signed-in caller.
+      const cand = prevSeen.toDate();
+      if (cand.getTime() && cand.getTime() < auditedOn.getTime()) {
+        auditedOn = cand; // take earliest date
+      }
+      // TODO(robdodson): Do we care about migrating userdata from DevSite?
```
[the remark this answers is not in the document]
samthor: I feel like we need to do a migration step—something like:

* signing into Firebase from DevSite and uploading the results
* using localStorage to maintain URLs for users during a transition period
* or exposing a CORS-safe endpoint on DevSite for users to fetch their known URLs.

/shrug

src/lib/fb.js
samthor: `saveUserUrl`? It's about saving it to Firestore.

samthor:
I've updated with your requests (removed serial code). There's some duplication now to avoid clobbering other results.

I'm loathe to fix all the other TODOs, e.g. fetching data earlier than we should, in this PR.

src/lib/fb.js
robdodson: what if we call it `urlHistory` ?

robdodson: or you could simplify it to just `urls` ?

```
currentUrl: foo.com,
urls: {
  foo.com: 2019-10-31,
  bar.com: 2018-05-10,
}
```

src/lib/actions.js
```diff
@@ -0,0 +1,116 @@
+import {store} from "./store";
+import {saveUserUrl} from "./fb";
```
robdodson: Looks like there's a circular dependency where actions imports fb and fb imports actions.

robdodson: maybe it doesn't actually break anything tho...

samthor: Yeah, there's nothing inherently wrong with circular deps in module-land, but it can create ambiguities during loading or rollup.

I'll have a think.

src/lib/fb.js
samthor: I've renamed to `currentUrl` and `urls`.

src/lib/actions.js
samthor: I've resolved this by adding another var to the store telling us whether we _should_ request results.

PR #1258, opened 2019-08-02
Add support for specifying anchors in headers

### LINKED ISSUE #1257: Need support for custom anchors in header tags
WebFu supports specifying anchors in headers eg: `## hello {: #anchor }`. 

This is important for long headers, or in cases where the header may change but you want to keep things consistent. We also need this so that when we move content from WebFu to web.dev, we can maintain the same anchor tags.

Add support for https://www.npmjs.com/package/markdown-it-attrs, and use the same syntax as WebFu used `{: #something }`

Looked at customizing the slugify function, but it was too much custom code.

petele:
Fixes #1257 

Adds support for specifying custom anchors in headers using the same style as WebFu 

`## hello {: #anchor }`

This PR adds a new node dependency ([markdown-it-attrs](https://www.npmjs.com/package/markdown-it-attrs)), which is designed explicitly for this purpose. I customized the default use so that it matches the same syntax as WebFu (`{: #anchor }`).

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -87,7 +87,7 @@ module.exports = function(config) {
   const markdownItAttrsOpts = {
     leftDelimiter: '{:',
     rightDelimiter: '}',
-    allowedAttributes: [],
+    allowedAttributes: ['id', 'class', /^data\-.*$/],
   };
   config.setLibrary(
     'md',
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -19,6 +19,7 @@ const pluginSyntaxHighlight = require('@11ty/eleventy-plugin-syntaxhighlight');
 
 const markdownIt = require('markdown-it');
 const markdownItAnchor = require('markdown-it-anchor');
+const markdownItAttrs = require('markdown-it-attrs');
 const slugify = require('slugify');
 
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -83,9 +84,16 @@ module.exports = function(config) {
       });
     },
   };
+  const markdownItAttrsOpts = {
+    leftDelimiter: '{:',
+    rightDelimiter: '}',
+    allowedAttributes: [],
```
petele: @robdodson - I've left this blank, so folks can technically apply other attributes to elements, for example they could add a class, or `data-`, etc. If you want to lock it down to just `id`, we can easily do that, but the flexibility is nice. I didn't ever see it abused on WebFu, so I think it should be pretty safe.

robdodson: I updated the comment so it's scoped to `id`, `class`, and `data-*`. I figure it's nice to lock it down because if we start doing localization it might be easy for someone to sneak in an `onload` or something if we're not paying close attention.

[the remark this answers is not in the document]
petele: Updated PR per request

robdodson:
so cool, thank you!!!

PR #1259, opened 2019-08-03
Markdown lint rules

robdodson:
Adds an initial group of markdown linter rules. This includes everything in our existing rules doc except for:

- remark-lint-code (requires we fork remark-lint-code-eslint to make it compatible with eslint v5)
- remark-lint-no-empty-sections (broken)
- remark-lint-link-title-style (This is loaded but I didn't use the parens style. I left the default ").

There are 273 warnings with the rules that I added. @mfriesenhahn maybe you should take a look at the output to see if there's anything you disagree with and I can update the config in this PR. Then we can do a follow up PR to fix all the warnings.


mfriesenhahn:
Went through the warnings. Just a few issues/notes:
- It's interpreting YAML comments as headings, which means false positives on heading rules.
- The `no-repeat-punctuation` rule is flagging cases where an author has used three periods instead of the ellipsis character. That should indeed be flagged, but I'm wondering if we can modify the warning text--because I'm not sure everybody knows there _is_ an ellipsis character.
- The `no-shortcut-reference-link` rule is flagging cases where authors use square brackets to indicate a placeholder value. We either need a different convention for that or we need to add square brackets to our escape character guidance in the handbook.
- I think we need to remove the `no-duplicate-headings` rule. There are enough valid use cases for repeated headers (e.g., providing a "Problem" and "Solution" for multiple issues in a post) that I think it'll cause more harm than good.
- The handbook includes examples of errors (e.g., using curly apostrophes), so it'd be great if we can escape those.
- This is more a note for myself, but with this linting we can remove several manual checks from the checklist!

robdodson:
> It's interpreting YAML comments as headings, which means false positives on heading rules.
blargh...

> The no-repeat-punctuation rule is flagging cases where an author has used three periods instead of the ellipsis character. That should indeed be flagged, but I'm wondering if we can modify the warning text--because I'm not sure everybody knows there is an ellipsis character.

Unfortunately the message is [hard coded into the linter rule](https://github.com/laysent/remark-lint-plugins/blob/master/packages/remark-lint-no-repeat-punctuation/index.js#L17). We could fork it and just have our own version with a custom message.

> The no-shortcut-reference-link rule is flagging cases where authors use square brackets to indicate a placeholder value. We either need a different convention for that or we need to add square brackets to our escape character guidance in the handbook.

Sounds like in the short term we should disable this one.

> I think we need to remove the no-duplicate-headings rule. There are enough valid use cases for repeated headers (e.g., providing a "Problem" and "Solution" for multiple issues in a post) that I think it'll cause more harm than good.

It looks like there are two variations of this linter.
https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-no-duplicate-headings
https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-no-duplicate-headings-in-section

The second one ensures there isn't a duplicate heading in the same section, which may be the behavior that we want.

> The handbook includes examples of errors (e.g., using curly apostrophes), so it'd be great if we can escape those.

Can you link me to an example?

robdodson:
oh btw I think we can add comments in places to tell the linter to shut up.

```
<!--lint disable no-duplicate-headings-->
```

mfriesenhahn:
> > The no-shortcut-reference-link rule is flagging cases where authors use square brackets to indicate a placeholder value. We either need a different convention for that or we need to add square brackets to our escape character guidance in the handbook.
> 
> Sounds like in the short term we should disable this one.

Actually, I think this is mainly a problem in the handbook, so I can work around it. 
> 
> > I think we need to remove the no-duplicate-headings rule. There are enough valid use cases for repeated headers (e.g., providing a "Problem" and "Solution" for multiple issues in a post) that I think it'll cause more harm than good.
> 
> It looks like there are two variations of this linter.
> https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-no-duplicate-headings
> https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-no-duplicate-headings-in-section
> 
> The second one ensures there isn't a duplicate heading in the same section, which may be the behavior that we want.

Perfect.
> 
> > The handbook includes examples of errors (e.g., using curly apostrophes), so it'd be great if we can escape those.
> 
> Can you link me to an example?

Sure: The ["Don't" example for curly quotes](https://web.dev/handbook/grammar/#quotation-marks-and-apostrophes) uses curly quotes for a demonstration. (Or, at least, it should. Looks like I may have accidentally "fixed" it.)

robdodson:
- I think I fixed the `atx` header thing using remark-frontmatter.
- I forked the repeat-punctuation rule and added a custom message about using the ellipsis character.
- I disabled the no-shortcut-reference-link rule for now.
- Switched to only using remark-lint-no-duplicate-headings-in-section so we can hopefully still check for duplicate headings at the section level.
- Working on adding a new rule to identify unescaped `{{` and `}}` tags.

robdodson:
@mfriesenhahn lemme know if you think this is good to land.

wooorm:
Hi folks! 👋 Just wanted to drop into your DMs to suggest that remark can be combined with retext to do [natural language stuff](https://github.com/retextjs/retext/blob/master/doc/plugins.md#list-of-plugins). Such as the curly quotes checking. [Here’s how I do that in my preset](https://github.com/wooorm/remark-preset-wooorm/blob/master/index.js#L14-L19) (although I prefer the curlies).

robdodson:
Thanks @wooorm!

Is it possible to enable retext and use the smart quotes plugin in a `.remarkrc` file?

wooorm:
Yes it is, call in '.remarkrc.js' and write it as a commonjs module (code can be the same as the preset above)

robdodson:
oh wait I think I see. I can either create a plugin preset or convert `.remarkrc` to `.remarkrc.js`.

wooorm:
Yes!

robdodson:
@wooorm I think one issue I just uncovered with the smart quotes plugin is that we tend to use single quotes for arguments in our markdown short codes (https://github.com/GoogleChrome/web.dev/blob/a93b9b2aa1d65eff5767dac9f8cb671b2cd403f7/src/site/content/en/handbook/sandbox/asides/index.njk#L9) and it flags that because it thinks we're using quotation at the wrong level, I imagine it's trying to catch things like "Hello 'world'".

wooorm:
Ah, that’s hard. As it’s not standard markdown. So it’s seen as prose 🤔

This gets really hard because templates may be compiled to something interleaving with Markdown syntax, so it’s not possible to 100% support this well.

You could use `rehype`, `rehype-retext`, etc. instead, and run the natural language on the generated HTML.

Alternatively, a while back, mapbox had a similar problem with liquid tags. They came up with this solution: https://github.com/mapbox/retext-mapbox-standard/blob/master/lib/strip_liquid.js.
A more complete solution would parse the constructs as a new node type, and compile them back to markdown as well. But that gets a bit more involved, and it’ll be something like this: https://github.com/djm/remark-shortcodes/blob/master/index.js or https://github.com/Nevenall/remark-terms/blob/master/src/index.js (although your tags are simpler and you don’t need the `h*` support).

robdodson:
@wooorm no worries. I think I can stick to the custom rule I wrote to detect smart quotes. I wasn't sure if there was a simple config fix but it sounds like it might be easiest to go with the custom rule.

robdodson:
@mfriesenhahn Unfortunately the spell checker was a little too aggressive (we use a lot of technical words like sass, which it flags as misspellings). I removed the natural language processing stuff for now. Maybe we can land this initial set of rules and revisit after we feel like they've had a chance to settle.

mfriesenhahn:
Sounds good. I would like to explore the possibility of using a spell checker with white listing, but definitely agree this preliminary step will help a lot. (Soooo ready not to have to check quote curliness! 😄)

robdodson:
ok, i'll consider this PR done and kick off another one to fix the linting issues.

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -35,6 +35,23 @@ Include the [language name](https://prismjs.com/#supported-languages) after the
 ```
 ````
 
+For code blocks that do not need syntax highlighting (e.g. HTTP headers) set
```
mfriesenhahn:
```suggestion
For code blocks that don't need syntax highlighting (for example, HTTP headers), set
```

src/site/content/en/handbook/markup-code/index.md
```diff
@@ -35,6 +35,23 @@ Include the [language name](https://prismjs.com/#supported-languages) after the
...
+the language to `text`:
+
+````markdown
+```text
+Set-Cookie: promo_shown=1; Max-Age=2600000; Secure
+```
+````
+
+For terminal commands set the language to `bash` and omit leading `$` signs:
```
mfriesenhahn:
```suggestion
For terminal commands, set the language to `bash` and omit leading `$` signs:
```

jpmedley:
I can easily generate a near complete list of web interfaces and interface
members for a whitelist. I can't do enum members yet. Fortunately, web
enums seem to contain a high number of terms that appear in non-technical
English.
 816-678-7195

PR #1261, opened 2019-08-04
 /uses-rel-preconnect/ should include link to /preconnect-and-dns-pre…

Malvoz:
Add link from audit https://web.dev/uses-rel-preconnect to article https://web.dev/preconnect-and-dns-prefetch/


robdodson:
nice, thank you!

PR #1262, opened 2019-08-05
Experimental translation

agektmr:
This pull request is to see if l10n will work on our new infrastructure. Follow up on #568 .

- Creates `src/site/content/ja/accessible/control-focus-with-tabindex` directory
- Creates `src/site/content/ja/accessible/control-focus-with-tabindex/index.md` file with a translation of `src/site/content/en/accessible/control-focus-with-tabindex/index.md`

This fails to build. I'll describe the problem in a separate issue.


robdodson:
Going to close as Ewa is now working on this.

PR #1264, opened 2019-08-05
Add nvmrc

surma:
This repo doesn't build for me on Linux with Node 12, but it does with Node 10. 

My suggestion is to add an `.nvmrc` :)


robdodson:
Ah I think the failing build is because this branch uses node-sass. We've switched off of it in the v2 branch. But good call on adding the .nvmrc

PR #1277, opened 2019-08-06
Includes mention of auto lazy-loading with Lite Mode and updates bug link

housseindjirdeh:

- Includes an FAQ on future plans to expand this feature for Lite Mode
- Updates bug link

CC @addyosmani @bengreenstein

### DIFF
--- a/src/site/content/en/blog/native-lazy-loading/index.md
+++ b/src/site/content/en/blog/native-lazy-loading/index.md
@@ -151,6 +151,13 @@ A placeholder shows for lazy-loaded iframes that are still being fetched.
 
 ## FAQ
 
+### Are there plans to expand this feature?
+
+There are plans to change the default lazy-loading behavior of the browser to automatically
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/native-lazy-loading/index.md
```diff
@@ -151,6 +151,13 @@ A placeholder shows for lazy-loaded iframes that are still being fetched.
 
 ## FAQ
 
+### Are there plans to expand this feature?
+
+There are plans to change the default lazy-loading behavior of the browser to automatically
```
scott-little: nit: Would it be possible to mention here that Chrome would only target images and iframes that it determines to be good candidates for lazy loading? I'm worried that developers might be worried that all images and iframes will be lazily loaded indiscriminately for Lite Mode users

housseindjirdeh: That's a great point. Updated in #1280

PR #1281, opened 2019-08-07
Tweak example for feature detection and using lazysizes

trumbitta:
Using a custom data attribute instead of a CSS class, teaches young developers to avoid CSS classes for things they were never meant to be used for.  
If a young developer starts using classes as hooks for their JavaScript code, some day in the future some other young developer will *optimize* the HTML and remove that *unused* CSS class that is nowhere to be found in any stylesheet.

### DIFF
--- a/src/site/content/en/blog/native-lazy-loading/index.md
+++ b/src/site/content/en/blog/native-lazy-loading/index.md
@@ -218,21 +218,21 @@ library only when `loading` isn't supported. This works as follows:
 
 - Replace `<img src>` with `<img data-src>` to avoid an eager load in unsupported browsers. If the
   `loading` attribute is supported, swap `data-src` for `src`.
-- If `loading` is not supported, load a fallback (lazysizes) and initiate it. You can use the custom
-  `data-loading-fallback` attribute as a way to indicate to lazysizes which images to lazy-load.
+- If `loading` is not supported, load a fallback (lazysizes) and initiate it. As per lazysizes docs, you use the
+  `lazyload` class as a way to indicate to lazysizes which images to lazy-load.
 
 ```html
 <!-- Let's load this in-viewport image normally -->
 <img src="hero.jpg" alt="…">
 
 <!-- Let's lazy-load the rest of these images -->
-<img data-src="unicorn.jpg" alt="…" loading="lazy" data-loading-fallback="lazysizes">
-<img data-src="cats.jpg" alt="…" loading="lazy" data-loading-fallback="lazysizes">
-<img data-src="dogs.jpg" alt="…" loading="lazy" data-loading-fallback="lazysizes">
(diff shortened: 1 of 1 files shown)


trumbitta:
@googlebot I signed it!

housseindjirdeh:
Thanks a ton for this @trumbitta. Totally understand your concern of whether classes are a good approach to identify which images to load via JS.

The code snippet in the article leverages [lazysizes](https://github.com/aFarkas/lazysizes), a popular lazy loading lib, as a fallback. By default, it detects which images to lazy load by looking for `lazyload` classes which is why we included it in the markup (changing this class _is_ possible through its [API](https://github.com/aFarkas/lazysizes#js-api---options) but I don't know if there's a way to change it to use a data attr instead as a marker).

We could definitely explore the option of using a different lazy loading library ([verlok/lazyload](https://github.com/verlok/lazyload)) as an example in this article if everyone feels strongly about this, but defining which hook to use is ultimately up to which third-party library we decide to use :)

trumbitta:
@housseindjirdeh got it, thanks. I admit I never used lazysizes, and didn't care to check before submitting my PR.

Given your explanation, and now that I have read lazysizes's README, do you think the example could have both the class needed by lazysizes and the new data attribute?

housseindjirdeh:
@trumbitta appreciate it, and thanks a ton for updating it. Things look great now 👍 

> Do you think the example could have both the class needed by lazysizes and the new data attribute?

Although I'm inclined to say yes, I think it might add more confusion than what it's worth. This is especially because the new data attribute wouldn't be recognized by lazysizes (unless we do some wrangling with the API if I'm not mistaken).

src/site/content/en/blog/native-lazy-loading/index.md
```diff
@@ -218,8 +218,8 @@ library only when `loading` isn't supported. This works as follows:
 
 - Replace `<img src>` with `<img data-src>` to avoid an eager load in unsupported browsers. If the
   `loading` attribute is supported, swap `data-src` for `src`.
-- If `loading` is not supported, load a fallback (lazysizes) and initiate it. You can use
-  `class=lazyload` as a way to indicate to lazysizes which images to lazy-load.
+- If `loading` is not supported, load a fallback (lazysizes) and initiate it. As per lazysizes docs, you use the
```
robdodson:
```suggestion
- If `loading` is not supported, load a fallback (lazysizes) and initiate it. As per the lazysizes docs, use the
```

src/site/content/en/blog/native-lazy-loading/index.md
```diff
@@ -218,8 +218,8 @@ library only when `loading` isn't supported. This works as follows:
...
+  `lazyload` class as a way to indicate to lazysizes which images to lazy-load.
```
robdodson:
```suggestion
  `lazyload` class to indicate which images to lazy-load.
```

PR #1323, opened 2019-08-14
Release image CDN content

khempenius:


robdodson:
Do you mind updating the `date` field on all of these to today?

robdodson:
Wanted to mention that Katie and I discussed that the codelabs box at the bottom of the image CDNs post is linking to the thumbor blog post. This is different from other codelab boxes but we wanted to make sure thumbor got even footing.

I think we can do this temporarily and then switch over to using the "Suggested articles" widget to show the thumbor post once we have that shipped.

Issue #1199, opened 2019-07-23
Provide more hero and thumbnail options.


robdodson:
**Is your feature request related to a problem? Please describe.**
Blog post thumbnails reuse the hero image from the post itself. This creates problems when the hero image does not neatly scale down to the thumbnail dimensions.

**Describe the solution you'd like**
We should use the picture element to let authors provide different crops of these images.
We should also add the ability to have a separate thumbnail and hero image.e considered.

**Additional context**
https://github.com/GoogleChrome/web.dev/pull/1142#issuecomment-514237432

PR #1324, opened 2019-08-14
Let authors provide a thumbnail for posts. Addresses #1199.

robdodson:

- Adds new `thumbnail` and `thumbnail_alt` YAML properties.
- Renames `alt` to `hero_alt`.
- Updates post template with new properties.

Note, before landing this we should update all of the existing posts to use `hero_alt`.

### DIFF
--- a/src/site/_drafts/_template-post/index.md
+++ b/src/site/_drafts/_template-post/index.md
@@ -10,7 +10,11 @@ hero: hero.jpg
 # You can adjust the position of your hero image with this property.
 # Values: top | bottom | center (default)
 # hero_position: bottom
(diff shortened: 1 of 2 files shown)


mihajlija:
Could I have just this one post in the blog now, as an entry point to third-party JS topic? Visibility boost can't hurt...

image

robdodson:
I think I'm going to go with a slightly different approach in a new PR.

Authors will still be able to provide a different crop of their hero image for the thumbnail, but we'll reuse the same `alt` text for both images and we'll tell folks that the policy is that they need to be the same.

mihajlija:
Good call 👍

PR #1328, opened 2019-08-15
[hot-fix] Update preconnect recommendations

mihajlija:
There's an ongoing discussion about recommended limits for preconnect, but we're definitely not going to recommend a hard number. I've double-checked everything, this the last fix I need to get in (the sooner the better).

- Fix errors and remove hard numbers from the recommendation.


addyosmani:
@robdodson do we think we'll be able to land this change this week?... pushing my luck? :)

robdodson:
yep,I can push this out today.

robdodson:
This fix should be live now.

PR #1366, opened 2019-08-19
Add thumbnail to Third party JS post

mihajlija:

- Add a thumbnail to Third party JS post :)


mfriesenhahn:
Hey, is it possible to get a version of the thumbnail to use as the hero image as well? To help maintain continuity between thumbnail and post, I think we want to have a similar image in both places.

mihajlija:
Honestly I prefer to keep the post without the hero image, it pushes out the text (unnecessarily). The thumbnail is the first frame of the video, so I don't think we are really breaking the continuity?

mfriesenhahn:
Normally, I would totally agree with you, but even on my fairly large work monitor, the video is below the fold. I hear you about not wanting to push down the text though. I'm sure you've already considered this, but does this post _need_ a thumbnail? I ask because I'm not sure the still from the video conveys third-party JS that well without the page elements lighting up.

Anyway, I'm OK with it if you think it's best to leave this as it is--this case is admittedly on the edge. But in general I think we want to make sure the same image appears in both thumbnail and hero positions to provide a highly salient indicator that the link has taken you where you were expecting.

mihajlija:
Dropping the thumbnail then and suggest adding that guidance to the handbook. I was absolutely planning to use only the thumbnails going forward :D

mfriesenhahn:
Absolutely! On the docket :)

robdodson:
Should we close this PR?

PR #1378, opened 2019-08-19
Normalize "code splitting" punctuation

### LINKED ISSUE #1364: The term "code splitting" is hyphenated inconsistently throughout the site
We use both "code splitting" and "code-splitting" through web.dev articles, and (as far as I can tell) there's no pattern to when we hyphenate vs. when we don't.

- https://web.dev/reduce-javascript-payloads-with-code-splitting
- https://web.dev/code-splitting-suspense
- https://web.dev/route-level-code-splitting-in-angular

Looking at other sites on the internet, they do it too, so I'm not sure which one is correct.

- https://webpack.js.org/guides/code-splitting/
- https://rollupjs.org/guide/en/#code-splitting
- https://reactjs.org/docs/code-splitting.html

My preference would be for "code splitting" when used as a noun or a verb, and "code-splitting" (or "code-split") when used as an adjective, but either way, we should standardize and add our preferred form to the [word list](https://web.dev/handbook/word-list/).

mfriesenhahn:
Fixes #1364.

- Adds guidance about punctuation for _code splitting_ to handbook [Word list](https://web.dev/handbook/word-list).
- Corrects all variants of _code splitting_ in existing content.


src/site/content/en/fast/codelab-preload-critical-assets/index.md
```diff
@@ -193,8 +193,8 @@ when needed.
 ## Preloading and prefetching with webpack
 
 The
-[code splitting guide](/reduce-javascript-payloads-with-code-splitting)
-explores the use of dynamic imports to split a bundle into multiple chunks.
+[Reduce JavaScript payloads with code-splitting](/reduce-javascript-payloads-with-code-splitting)
```
philipwalton: This should be "code splitting", correct?

mfriesenhahn: Good catch

mfriesenhahn:
> Code splitting has come to take on a very specific meaning in web development (as a bundler feature), so much so that I'm not sure replacing it with "splitting" or "splitting your code" would carry the same meaning in every case.
> 
> While I think the changes in this PR are mostly fine (especially if you link to a guide on code splitting when using the term "splitting" generically), but I wonder if it's too strong to recommend never using "code-splitting" as a verb.

I hear that, though I still don't love the redundancy of saying things like "code-split your code." And, actually, many of the posts are already just using "split" as the verb form--I suspect for that reason.

Ultimately, though, I agree our language should be descriptive, not prescriptive. Is there a small group of folks who could weigh in on what usage is more conventional around the web?

mfriesenhahn:
@robdodson, can you review and merge when you get a sec?

PR #1383, opened 2019-08-19
Migrate web-codelab element.

### LINKED ISSUE #1374: Migrate web-codelab element.
We need to migrate the `web-codelab` element out of devsite and into our v2 branch.

robdodson:
Fixes #1374

- Adds `web-codelab` element.
- Updated `codelab.njk` to use web-codelab instead of local dev version.
-

### DIFF
--- a/src/lib/components/Codelab/_styles.scss
+++ b/src/lib/components/Codelab/_styles.scss
@@ -1,0 +1,23 @@
+@import '../../../styles/settings/colors';
+
+web-codelab {
+  display: flex;
+  height: 100%;
+}
+
+.web-codelab__headline {
+  margin-top: 0;
+  margin-bottom: 8px;
+}
+
+.web-codelab__instructions {
+  overflow-x: hidden;
+  overflow-y: scroll;
+  padding: 32px;
+  width: 600px;
+}
+
+.web-codelab__glitch {
+  flex: 1;
+  padding: 16px;
+}
--- a/src/lib/components/Codelab/_styles.scss
+++ b/src/lib/components/Codelab/_styles.scss
@@ -3,21 +3,22 @@
 web-codelab {
   display: flex;
   height: 100%;
+
+  .web-codelab__headline {
+    margin-top: 0;
+    margin-bottom: 8px;
+  }
+  
+  .web-codelab__instructions {
+    overflow-x: hidden;
+    overflow-y: scroll;
+    padding: 32px;
+    width: 600px;
+  }
+  
+  .web-codelab__glitch {
+    flex: 1;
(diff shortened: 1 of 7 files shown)


src/lib/components/Codelab/_styles.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../../styles/settings/colors';
+
+web-codelab {
+  display: flex;
+  height: 100%;
+}
+
+.web-codelab__headline {
```
samthor: I know you don't like nesting but it seems sensible to wrap this in at least the CE itself?

src/lib/components/Codelab/index.js
```diff
@@ -0,0 +1,69 @@
...
+    return container;
+  }
+
+  get src() {
+    if (!this.glitch) {
+      return;
+    }
+
+    const url = new URL(`${GLITCH_EMBED}/${this.glitch}`);
+    url.searchParams.append("path", this.path);
```
samthor: Not supported in IE11 without polyfill but we probably don't care.

src/lib/components/Codelab/index.js
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor() {
+    super();
+
+    this.glitch = "";
+    this.path = "index.html";
+  }
+
+  createRenderRoot() {
+    const container = document.createElement("div");
```
samthor: Why do this instead of just having `<div class="web-codelab__glitch">` returned as part of `render()`?

robdodson: The element has some existing Light DOM children so it goes:

```
<web-codelab>
  <div class="web-codelab__instructions">[this gets generated by 11ty]</div>
  .. [this is where web-codelab__glitch needs to render]
</web-codelab>
```

If I try to return `web-codelab__glitch` it will replace the `web-codelab__instructions` child. This works around that by appending a new child and using that as the render target.

src/lib/components/Codelab/_styles.scss
robdodson: hm yeah I just noticed the LH gauge does that. I don't think it gives any benefit since BEM already prevents collisions but it might be nice to be consistent with the other custom elements.

I was actually thinking that it might make sense to switch to [CSS Modules](https://github.com/css-modules/css-modules) at some point since they play nice with Sass and could give us essentially the same scoping as Shadow DOM without all the polyfilling.

src/lib/components/Codelab/index.js
samthor: That's fine but it might be worth just adding a comment here calling this out.

PR #1385, opened 2019-08-20
Server side logic for localization

agektmr:
This pull request adds localization to web.dev.
The server will serve the localized content from the lang directory under `/dist`.
User can specify the lang with `hl` query parameter or browser's `accept-language` header.
When changing the language preference, append `preferred_lang` to the query along with `hl`.


server.js
```diff
@@ -15,10 +15,68 @@
...
+app.use(cookieParser());
+app.use(localeHandler);
 app.use(express.static('dist'));
-app.use(express.static('dist/en'));
+
+app.get("*", function(req, res) {
+  const dir = req.path.split(path.sep)[1];
+  // Serve static files as is
+  if (staticFilesPath.includes(dir)) {
+    res.sendFile(`${__dirname}/dist${req.path}`);
```
robdodson:
```suggestion
    res.sendFile(path.join(__dirname, 'dist', req.path));
```

server.js
```diff
@@ -15,10 +15,68 @@
...
+    return;
+  }
+  // Serve the lang specific file from a localized directory
+  const _path = `${__dirname}/dist/${req.lang}${req.path}/index.html`;
```
robdodson:
```suggestion
  const localizedFilePath = path.join(__dirname, 'dist', req.lang, req.path, 'index.html');
```

server.js
```diff
@@ -15,10 +15,68 @@
...
+  if (fs.existsSync(_path)) {
```
robdodson:
```suggestion
  if (fs.existsSync(localizedFilePath)) {
```

server.js
```diff
@@ -15,10 +15,68 @@
...
+    res.sendFile(_path);
```
robdodson:
```suggestion
    res.sendFile(localizedFilePath);
```

server.js
```diff
@@ -15,10 +15,68 @@
...
+  } else {
+    res.sendFile(`${__dirname}/dist/${defaultLocale}${req.path}/index.html`);
```
robdodson: I'd suggest creating another variable after `localizedFilePath`. This variable should be named `fallbackFilePath` and contain this path.

agektmr:
I've made changes that address feedback + other issues.

server.js
```diff
@@ -15,10 +15,67 @@
...
+
 const app = express();
 
+const staticFilesPath = [
+  'images', 'test', 'app.css', 'app.js'
+];
+
+// Detect all locale directories
+const supportedLocales = [];
+const dirs = fs.readdirSync(`${__dirname}/dist`);
```
robdodson:
```suggestion
const dirs = fs.readdirSync(path.join(__dirname, 'dist'));
```

agektmr:
@robdodson I just noticed the cookie for storing preferred language in DevSite (v1?) is `django_language`. Also, how it changes preferred language is to POST a request to `/i18n/setlang`.
I wonder how much should we care about backward compatibility with DevSite.

server.js
```diff
@@ -14,12 +14,68 @@
...
-app.use(express.static("dist/ja"));
+const staticFilesPath = [
+  'images', 'test', 'app.css', 'app.js'
+];
+
+// Detect all locale directories
+const supportedLocales = [];
+const dirs = fs.readdirSync(path.join(__dirname, 'dist'));
+for (const dir of dirs) {
+  if (staticFilesPath.includes(dir)) continue;
```
samthor: nit: please always `{}` and multiline to avoid problems in future

server.js
```diff
@@ -14,12 +14,68 @@
...
+  supportedLocales.push(dir);
+}
+const defaultLocale = 'en';
+
+app.use(locale(supportedLocales, defaultLocale));
+app.use(cookieParser());
+app.use(express.static('dist'));
+
+app.get("*", function(req, res) {
```
samthor: nit: use ' throughout

server.js
```diff
@@ -14,12 +14,68 @@
...
+  }
+  // Update preferred language (with sanity check)
+  if (req.query.preferred_lang !== undefined &&
+      supportedLocales.includes(lang)) {
+    res.cookie('preferred_lang', lang,
+      {maxAge: 10 * 365 * 24 * 60 * 60 * 1000, httpOnly: true});
+  }
+
+  // Serve the lang specific file from a localized directory
+  const localizedFilePath = path.join(__dirname, 'dist', lang, req.path, 'index.html');
```
samthor: This is a security hole. I could set my `preferred_lang` cookie to be "../../../escape/the/root".

Can we check that the lang is in "[-_a-zA-Z]" (i.e., '-', '_' or English upper/lower case)?

server.js
```diff
@@ -14,12 +14,68 @@
...
-app.use(express.static("dist"));
-// app.use(express.static("dist/en"));
-app.use(express.static("dist/ja"));
+const staticFilesPath = [
+  'images', 'test', 'app.css', 'app.js'
+];
+
+// Detect all locale directories
+const supportedLocales = [];
+const dirs = fs.readdirSync(path.join(__dirname, 'dist'));
```
samthor: Do we actually need this? Could we allow the user to set any language (... checking that it's valid, as per my security comment below) and then try the file / fallback to 'en'?

server.js
samthor: GitHub ate my underscore and treated it as italics. It should be, any underscore, dash, or English upper/lower case.

robdodson:
@devnook Eiji did some work on translations a while back. Putting it on your radar in case you want to use some of this in your current work. Going to close the pull request for now.

PR #1418, opened 2019-08-21
update axe links to axe 3.3

paulirish:
to avoid seeing this header when clicking through

Screen Shot 2019-08-21 at 1 17 18 PM


mfriesenhahn:
I was just talking to @robdodson about this. I kinda think we shouldn't be linking to the full list of axe rules anyway and instead just link to the rule(s) relevant to a given audit. Thoughts?

paulirish:
yup. i agree with removing the links to the full listing.

PR #1424, opened 2019-08-21
v0 search code

samthor:
Adds an Algolia-based search index. Definitely a WIP.

* adds `testindex.js` which compares the git revisions of articles, and reconciles by only doing required inserts/deletes
  * stores state in `indexed-rev.json`, which would need to be committed as part of a deploy step - is this practical?
* adds a very basic `web-search` component
  * this *includes* our API key, but it's the Search only-key, so this seems intended

Open questions:

* Should we store an index version number (so we can reindex all things if the _way_ we index changes)- or should we just reindex _everything_ on deploy?
* Not currently doing full-text search. I don't know if Algolia is happy with Markdown or HTML.
* Tag articles with i18n?


samthor:
* After exploring our current bundle size, I don't think it's worth building our own search code—their lightweight bundle is 40k, which is way not worth worrying about right now.

* The approach you referenced is probably better (in that Eleventy itself shouldn't be invoking Algolia) but isn't fundamentally different to what we're doing—just use a magic collection to output something. Actually creating a file which outputs it like magic might be more idiomatic but we could just write it inline where the index is generated too. ¯‍\‍_‍(‍ツ‍)‍_‍/‍¯

robdodson:
@samthor were you thinking of picking up this PR or did you want to start over?

samthor:
yes, I'll redo this. I think we should be using the Algolia JS for now, our bundle size isn't anything to write home about.

PR #1430, opened 2019-08-22
Update compare widget

mfriesenhahn:

- Use the `<figure>` and `<figcaption>` elements in the compare widget to improve semantics.
- Restyle compare widget to make it easier to see what content is being compared.
- Make compare widget work in the columns widget.
- Make the columns widget breakpoint match the phone media query breakpoint.
- Update all existing content pages that use the compare widget.

Notes:

- The Markdown parser isn't parsing fenced code blocks correctly when they appear in HTML elements. (For example, it's not applying syntax highlighting.) Until that's fixed, nesting code blocks in the compare widget won't work.
- I made the figure caption a parameter for the compare short code. While that leads to some ugliness in the MD, it allows the short code to be used in all cases, which should hopefully make things easier for contributors.


mfriesenhahn:
Oh, and I forgot to put the caption above the example.

mfriesenhahn:
@robdodson, I added a `CompareCaption` shortcode to match the approach we agreed to for the `Details` shortcode.

src/styles/components/_compare.scss
```diff
@@ -10,7 +10,6 @@
 // =============================================================================
 
 .w-compare {
-  font-size: .875em;
   line-height: 1.5em;
   margin: -16px 0 32px;
```
robdodson: I think you can remove the negative top margin. It seems to look better without it.

src/styles/components/_columns.scss
```diff
@@ -14,3 +14,43 @@
   // sass-lint:enable-all
   grid-template-columns: repeat(auto-fit, minmax(30ch, 1fr));
 }
+
+.w-columns .w-compare {
+  margin: 0;
+  /// adds the breakpoint styles
```
robdodson: I'd suggest removing the notes about adding breakpoint styles.

src/styles/components/_compare.scss
mfriesenhahn: I left it in because I think it improves the connection between paired compare widgets by reducing the vertical spacing between them, but I can remove it.

src/styles/components/_columns.scss
mfriesenhahn: OK. Was just following the convention in other style sheets.

mfriesenhahn:
@kaycebasques, can you review this PR when you have a moment?

robdodson:
looks like there's a few merge conflicts. Might be because I just merged the Details component PR.

PR #1446, opened 2019-08-28
Switch v2 to node-sass

robdodson:

- Update npm scripts to use node-sass for a 3x improvement in sass compilation times.


samthor:
We should drop this since #1447 is in now.

PR #1465, opened 2019-08-30
Responsive side nav

### LINKED ISSUE #1371: Rebuild responsive header and side-nav.
As part of the devsite migration, we need to create our own navigation header and responsive side-nav.

I have a side-nav that I stole from Paul Lewis which animates well and uses the inert polyfill. I think we could quickly use this.

robdodson:
Fixes #1371

- Adds a Header component.
- Adds a SideNav component.
- Adds wicg-inert polyfill.
- Adds :focus-visible polyfill.
- Defines actions for opening and closing the SideNav, as well as inerting other parts of the page.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -18,6 +18,7 @@ module.exports = [
       format: "iife",
       entryFileNames: outputPattern,
       chunkFileNames: outputPattern,
+      sourcemap: true,
     },
     watch: {
       clearScreen: false,
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -116,3 +116,11 @@ export const requestFetchReports = store.action((state, url, startDate) => {
     return update;
   });
 });
+
+export const openSideNav = store.action(() => {
+  return {isSideNavExpanded: true};
+});
+
+export const closeSideNav = store.action(() => {
+  return {isSideNavExpanded: false};
+});
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -119,6 +119,7 @@ export const requestFetchReports = store.action((state, url, startDate) => {
 });
 
 export const openSideNav = store.action(() => {
+  document.body.classList.add("web-side-nav--expanded");
   const main = document.querySelector("main");
   const header = document.querySelector("web-header");
   main.inert = true;
@@ -127,6 +128,7 @@ export const openSideNav = store.action(() => {
 });
 
 export const closeSideNav = store.action(() => {
+  document.body.classList.remove("web-side-nav--expanded");
   const main = document.querySelector("main");
   const header = document.querySelector("web-header");
   main.inert = false;
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -118,7 +118,7 @@ export const requestFetchReports = store.action((state, url, startDate) => {
   });
 });
 
-export const openSideNav = store.action(() => {
+export const expandSideNav = store.action(() => {
   document.body.classList.add("web-side-nav--expanded");
   const main = document.querySelector("main");
   const header = document.querySelector("web-header");
@@ -127,7 +127,7 @@ export const openSideNav = store.action(() => {
   return {isSideNavExpanded: true};
 });
 
-export const closeSideNav = store.action(() => {
+export const collapseSideNav = store.action(() => {
   document.body.classList.remove("web-side-nav--expanded");
   const main = document.querySelector("main");
   const header = document.querySelector("web-header");
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -1,6 +1,7 @@
 import {store} from "./store";
 import {saveUserUrl} from "./fb";
 import {runLighthouse, fetchReports} from "./lighthouse-service";
+import "wicg-inert";
 
 export const requestRunLighthouse = store.action((state, url) => {
   const p = (async () => {
@@ -118,9 +119,17 @@ export const requestFetchReports = store.action((state, url, startDate) => {
 });
 
 export const openSideNav = store.action(() => {
+  const main = document.querySelector("main");
+  const header = document.querySelector("web-header");
+  main.inert = true;
+  header.inert = true;
   return {isSideNavExpanded: true};
 });
 
 export const closeSideNav = store.action(() => {
+  const main = document.querySelector("main");
+  const header = document.querySelector("web-header");
+  main.inert = false;
+  header.inert = false;
   return {isSideNavExpanded: false};
 });
--- a/src/lib/components/Header/_styles.scss
+++ b/src/lib/components/Header/_styles.scss
@@ -19,27 +19,23 @@
 @import '../../../styles/tools/breakpoints';
 @import '../../../styles/tools/mixins';
 
-web-header:not(:defined) {
-  padding: 0 24px 0 197px;
-}
-
-web-header:not(:defined),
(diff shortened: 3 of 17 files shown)


src/lib/components/Header/_styles.scss
```diff
@@ -0,0 +1,132 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import '../../../styles/settings/colors';
+@import '../../../styles/settings/global';
+@import '../../../styles/tools/breakpoints';
+@import '../../../styles/tools/mixins';
+
+web-header:not(:defined) {
```
samthor: does this work in a polyfilled universe?

src/lib/components/Header/index.js
```diff
@@ -0,0 +1,67 @@
...
+        </div>
+      </div>
+
+      <web-profile-switcher-container></web-profile-switcher-container>
+    `;
+  }
+
+  firstUpdated() {
+    this.hamburgerBtn = this.querySelector(".web-header__hamburger-btn");
+    this.hamburgerBtn.addEventListener("click", openSideNav);
```
samthor: Shouldn't this be a `@click=...` part of the rendered HTML?

src/lib/components/SideNav/index.js
```diff
@@ -0,0 +1,199 @@
...
+
+  firstUpdated() {
+    this.sideNavContainerEl = this.querySelector(".web-side-nav__container");
+    this.addEventListeners();
+    store.subscribe(this.onStateChanged);
+    this.onStateChanged();
+  }
+
+  addEventListeners() {
+    this.addEventListener("click", this.onCloseSideNav);
```
samthor: For everything not on `this`, same as my other comment—shouldn't this be done inside lit?

We could even add a top-level element that's the same size that handles the other on... handlers.

src/lib/components/SideNav/index.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  set expanded(val) {
+    if (this.expanded_ === val) {
+      return;
+    }
+
+    const oldVal = this.expanded_;
+    this.animatable = true;
+    this.addEventListener("transitionend", this.onTransitionEnd);
```
samthor: `{once: true}`?

samthor: And then it can just be inline handler?

src/lib/components/Header/_styles.scss
robdodson: probably not. I also think that any FOUC with this header is bad so I'm going to move all of its guts to the light dom

src/lib/components/Header/index.js
robdodson: keeping this as is because I moved all the children back to light dom.

src/lib/components/SideNav/index.js
robdodson: I fixed this for the handler referencing sideNavContainerEl. But I kept the addListeners block for the stuff pointing at `this`. I explored creating a wrapper element, but [lit's syntax for passing options to a declarative handler](https://github.com/Polymer/lit-html/pull/531) seems to make the code more confusing than it's worth.

Basically I was ending up creating objects that contained the touch code, and the functions within those objects have to be bound back to `this` so it's just kinda gross.

src/lib/components/SideNav/index.js
robdodson: Added the `once: true` object and removed the `removeEventListener` in the handler. But I didn't inline this listener. See above for my reasons re: lit's options object.

robdodson:
@samthor did some updates based on your feedback. Would you mind taking a look at the two commits?

Along with the stuff I commented on I also:
- Made the header fixed position to match what devsite does.
- Added margin styles to `<main>` to account for fixed position header.
- Tweaked z-index for a few elements. I'm planning to create a separate PR to introduce an actual z-index scale so we don't end up with a bunch of arbitrary values. But didn't want to complicate this PR.
- Added an `overflow: hidden` class to body during the `openSideNav()`/`closeSideNav()` actions.

rollup.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = [
       format: "iife",
       entryFileNames: outputPattern,
       chunkFileNames: outputPattern,
+      sourcemap: true,
```
samthor: this seems unrelated but sure 👍

src/lib/actions.js
```diff
@@ -1,6 +1,7 @@
 import {store} from "./store";
 import {saveUserUrl} from "./fb";
 import {runLighthouse, fetchReports} from "./lighthouse-service";
+import "wicg-inert";
```
samthor: I'm not against inert but an alternative approach is that focusing outside the nav bar implicitly closes it (it's what we do on Santa). Just my 2c and doesn't require the polyfill.

src/lib/components/Header/index.js
samthor: sure. I wonder if it's worth calling out `BaseElement` subclasses which don't have a `render()` method.

i.e., there's no value to `BaseElement` in this case?

src/lib/components/SideNav/index.js
```diff
@@ -0,0 +1,195 @@
...
+  set expanded(val) {
+    if (this.expanded_ === val) {
+      return;
+    }
+
+    const oldVal = this.expanded_;
+    this.animatable = true;
+    this.addEventListener("transitionend", this.onTransitionEnd, {once: true});
+    this.expanded_ = val;
+    if (this.expanded_) {
```
samthor: why not `this.inert = !this.expanded_;` ?

src/lib/components/SideNav/index.js
```diff
@@ -0,0 +1,195 @@
...
+    this.requestUpdate("expanded", oldVal);
+  }
+
+  get expanded() {
+    return this.expanded_;
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    store.unsubscribe(this.onStateChanged);
```
samthor: The way I read the docs for `firstUpdated`, it seems to be one-time, not after every addition to the page.

https://lit-element.polymer-project.org/guide/lifecycle#firstupdated

Maybe move the subscribe call to `connectedCallback`?

src/lib/components/Header/index.js
robdodson: yeah maybe I should just switch to extending LitElement.

src/lib/actions.js
robdodson: I don't think that would prevent a screen reader from breaking outside of the side nav tho. Are you also marking the other areas as aria-hidden?

src/lib/components/Header/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+import {expandSideNav} from "../../actions";
+
+class Header extends HTMLElement {
+  connectedCallback() {
+    this.hamburgerBtn = this.querySelector(".web-header__hamburger-btn");
+    this.hamburgerBtn.addEventListener("click", expandSideNav);
+  }
+
+  manageFocus() {
```
samthor: Maybe comment where this is called from.

Alternatively, is there any precedent to overloading `.focus()` itself?

robdodson: ermmmmmm I feel a little sketchy about overloading focus... I'll add a comment instead :P

PR #1471, opened 2019-08-31
Add Image and Figure custom tags.

robdodson:
Adding @samthor as a reviewer since it'll be Monday his time tomorrow :)

- Add `Image` and `Figure` custom tags so we can migrate to using an image CDN with srcset and webp fanciness.

To keep things from getting super complicated, I'd like to land this PR first. Then follow up with a second PR to add srcset/webp goodness. And then follow up with a final PR that converts all existing images over to the new tags.

Example usage in Markdown files:

```
{%
  Image
  src="./label-a11y-lint.png",
  alt="hello",
  maxWidth="300px"
%}
```

```
{%
  Figure
  src="./label-a11y-lint.png",
  alt="hello",
  type="fullbleed screenshot",
  maxWidth="300px"
%}
Figcaption text goes here! Can be **markdown**.
{% endFigure %}
```

### DIFF
--- a/src/site/_data/site.js
+++ b/src/site/_data/site.js
@@ -25,5 +25,8 @@ module.exports = {
   banner: `Register for this year's \`#ChromeDevSummit\` happening on Nov. 11-12
   in San Francisco to learn about the latest features and tools coming to the
   Web. [Request an invite here](https://developer.chrome.com/devsummit/).`,
+  // Note that the imageCdn value is only used when we do a production build
+  // of the site. Otherwise all image paths are local. This means you can
+  // develop locally without having to mess with the CDN at all.
   imageCdn: 'https://webdev.imgix.net',
 };
--- a/src/site/_includes/components/tags/Image.js
+++ b/src/site/_includes/components/tags/Image.js
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Adds two new shortcodes to Eleventy: Image and Figure.
+ * These shortcodes will convert local image paths to image CDN paths depending
+ * on the build environment. This allows authors to dev with local images but
+ * switch to using image CDN images with srcset and other fanciness during
+ * production.
+ *
+ * We use the custom tags API instead of Eleventy's universal shortcodes here
+ * because we need access to the page context object (specifically the URL for
+ * the page that's being rendered). Unfortunately Eleventy's universal
+ * shortcodes don't have access to this information, so custom tags it is!
+ *
+ * Example usage:
+ *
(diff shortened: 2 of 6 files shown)


src/site/_includes/components/tags/Image.js
```diff
@@ -0,0 +1,211 @@
...
+  if (process.env.ELEVENTY_ENV === 'prod') {
+    imagePath = path.join(stripLanguage(ctx.page.url), src);
+    imagePath = new URL(imagePath, imageCdn).href;
+  }
+  return imagePath;
+}
+
+/**
+ * Render an image element as an HTML string.
+ * @param {!Object} args An args object containing src and optional alt and
```
samthor: nit: you can write this like

```js
/**
 * @param {{src: string, alt: stirng, maxWidth: number}} args An args ...
*/
```

src/site/_data/site.js
```diff
@@ -21,4 +21,5 @@ module.exports = {
   url: 'https://web.dev',
   repo: 'https://github.com/GoogleChrome/web.dev',
   subscribe: 'https://web.dev/subscribe',
+  imageCdn: 'https://webdev.imgix.net',
```
samthor: Is it worth adding a comment to this? It's something in our main config which references a third-party site. We'll be adding more of these, but it seems like lock-in for someone wanting to run our code themselves.

src/site/_includes/components/tags/Image.js
```diff
@@ -0,0 +1,211 @@
...
+ * @param {!Object} ctx An eleventy context object.
+ * @return {string}
+ */
+function renderImage({src, alt, maxWidth}, ctx) {
+  const imagePath = getImagePath(src, ctx);
+  let style;
+  if (maxWidth) {
+    style = `max-width: ${maxWidth};`;
+  }
+  // Using oneLine seems to make Nunjucks' SafeString function happy.
```
samthor: is it handling escaping of `${imagePath}` and etc? If I provide an image with "'s is that fine?

src/site/_includes/components/tags/Image.js
```diff
@@ -0,0 +1,211 @@
...
+};
+
+// -----------------------------------------------------------------------------
+// FIGURE
+// -----------------------------------------------------------------------------
+
+/**
+ * Render a figcaption element as an HTML string.
+ * @param {?string} caption A markdown string of caption text.
+ * @param {Array<string>} figCaptionClasses An Array of classes to apply to the
```
samthor: `!Array<string>` is non-nullable, which is what you want: `null` would be invalid here since you call a method on the array

src/site/_includes/components/tags/Image.js
```diff
@@ -0,0 +1,211 @@
...
+  // figcaption element.
+  return oneLine`
+    <figcaption class="${figCaptionClasses.join(' ')}">${md.renderInline(caption.trim())}</figcaption>
+  `;
+}
+
+/**
+ * Render a figure element as an HTML string.
+ * @param {!string} image An HTML string for an image element.
+ * @param {?Object} args An args object containing an optional type field.
```
samthor: `{{type: string}}`

src/site/_data/site.js
robdodson: yeah I can add a comment. This should only kick in if the person is doing a production build.

src/site/_includes/components/tags/Image.js
robdodson: The `oneLine` method just trims all the whitespace and newlines so the text appears as a single line.

The `SafeString` method doesn't seem to be well documented but it looks like if there is anything fishy with the HTML it just escapes the whole thing, turning brackets into `&lt;` and `&gt;`. I found that if my HTML string has newlines in it, it seems to consider that suspect and escapes the whole thing.

PR #1484, opened 2019-09-06
Make focus color contrast conformant

mfriesenhahn:
- Change focus ring color to `$WEB_PRIMARY_COLOR` so it's contrast conformant on white.
- Comment out `$WEB_SECONDARY_COLOR and flag it as non-conformant.


src/styles/settings/_colors.scss
```diff
@@ -222,13 +222,13 @@ $YELLOW_800: #f9a825;
 // googleguidance@ and they said we're allowed to make reasonable tweaks to
 // the palette.
 $WEB_PRIMARY_COLOR: #3740ff;
-$WEB_SECONDARY_COLOR: #3fc4ff;
+// $WEB_SECONDARY_COLOR: #3fc4ff; // Not contrast conformant on white
```
robdodson: I think we should delete this if it's not used anywhere else.

mfriesenhahn: OK, deleted. Wasn't sure if you wanted to leave it as a placeholder for a replacement secondary color down the road.

PR #1498, opened 2019-09-12
Create basic GitHub Actions workflow for indexing

samthor:
I have no idea what I'm doing. 🤷


samthor:
I think from that point of view this PR is mostly irrelevant since it just runs those commands.

So there's no reason to actually add GH actions at all. Or any sort of CI deploy.

PR #1533, opened 2019-09-17
Update path link code styling

mfriesenhahn:

- Reduce assertiveness of code font styling in path links on collection landing pages so they're not so intrusive.
- Match code font color in path links so it matches rest of link.


jpmedley:
(I've been wondering why we have borders around inline code at all. Makes total sense with standalone examples, whether multi or single line. For inline stuff, I wonder what it adds.)

PR #1541, opened 2019-09-19
Add w-footer.

### LINKED ISSUE #1372: Rebuild footer.
As part of the devsite migration, we need to recreate the footer in a way that doesn't rely on their CSS.

Tidy up the footer markup and CSS. Ideally it should look the same.

robdodson:
Fixes #1372.

- Adds `.w-footer` which is a simplified version of the devsite footer.
- Adds sourcemaps to `sass.sh`


src/site/_includes/partials/footer.njk
```diff
@@ -0,0 +1,128 @@
+<footer class="w-footer">
+  <nav class="w-footer__linkboxes">
+    <ul class="w-footer__linkboxes-list">
+      <li class="w-footer__linkbox">
+        <h3 class="w-footer__linkbox-heading">Contribute</h3>
+        <ul class="w-footer__linkbox-list">
+          <li class="w-footer__linkbox-item">
+            <a href="https://github.com/GoogleChrome/web.dev/issues/new?assignees=&labels=bug&template=bug_report.md&title="
+              class="w-footer__linkbox-link
+            gc-analytics-event" data-category="Site-Wide Custom Events" data-label="Footer Link (index 1)">
```
samthor: gc-analytics-event seems ignored, we just use data-category and data-label to trigger the event.

Maybe this is an omission in DevSite v2.

src/site/_includes/partials/footer.njk
```diff
@@ -0,0 +1,128 @@
+<footer class="w-footer">
+  <nav class="w-footer__linkboxes">
+    <ul class="w-footer__linkboxes-list">
+      <li class="w-footer__linkbox">
+        <h3 class="w-footer__linkbox-heading">Contribute</h3>
+        <ul class="w-footer__linkbox-list">
+          <li class="w-footer__linkbox-item">
+            <a href="https://github.com/GoogleChrome/web.dev/issues/new?assignees=&labels=bug&template=bug_report.md&title="
```
samthor: `target="_blank" rel="noopener"` on all links?

src/site/_includes/partials/footer.njk
```diff
@@ -0,0 +1,128 @@
...
+      </ul>
+    </nav>
+    <nav class="w-footer__utility-nav">
+      <ul class="w-footer__utility-list">
+        <li class="w-footer__utility-item
+            ">
+          <a class="w-footer__utility-link
+            gc-analytics-event" href="https://policies.google.com/" data-category="Site-Wide Custom Events"
+            data-label="Footer Terms and Privacy link">
+            Terms & Privacy
```
samthor: &amp;amp;

src/site/_includes/partials/footer.njk
robdodson: I think the devsite policy is not to use target=_blank.

> We recommend not using target="_blank" to force such links to open a new tab; developers know how to open new tabs if they want to, so we should respect their decisions and leave the choice to them. For more thoughts on use of target="_blank", read [Paul Ingraham's blog](http://www.paulingraham.com/target-blank-is-evil.html).

The linked post makes a decent argument so I'm inclined to keep them if that's ok.

PR #1547, opened 2019-09-20
include the webcomponents polyfill

### LINKED ISSUE #1376: Add custom elements polyfill.
When we migrate off of devsite we should consider if we want to include a custom elements polyfill. Currently IE and Edge make up something like <1% of our traffic but it may still be worth doing in the short term, at least until Edgium has gained more traction?

Bundle the custom elements polyfill into our site.

samthor:
Adds the Web Components loader to our source tree. Resolves #1376.

Note that our JS actually isn't being loaded now either (since it has a `.hexcode.js`) suffix. That's a separate issue.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -68,8 +68,7 @@ gulp.task('copy-content-assets', () => {
 gulp.task('copy-node_modules-assets', () => {
   return gulp
     .src([
-      `./node_modules/@webcomponents/webcomponentsjs/**/*.js`,
-      `!./node_modules/@webcomponents/webcomponentsjs/src/**`,
+      `./node_modules/@webcomponents/webcomponentsjs/bundles/*.js`,
     ])
     .pipe(gulp.dest('./dist/lib/webcomponents/'));
 });
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -65,9 +65,19 @@ gulp.task('copy-content-assets', () => {
     .pipe(gulp.dest('./dist/en'));
 });
 
+gulp.task('copy-node_modules-assets', () => {
+  return gulp
+    .src([
+      `./node_modules/@webcomponents/webcomponentsjs/**/*.js`,
+      `!./node_modules/@webcomponents/webcomponentsjs/src/**`,
+    ])
+    .pipe(gulp.dest('./dist/lib/webcomponents/'));
+});
+
 gulp.task('build', gulp.parallel(
   'copy-global-assets',
   'copy-content-assets',
+  'copy-node_modules-assets',
 ));
 
 gulp.task('watch', () => {
--- a/src/lib/bootstrap.js
+++ b/src/lib/bootstrap.js
@@ -5,6 +5,7 @@
  * app.js (or TODO: entrypoints).
  */
 
+import config from "./bootstrap-config";
 import "@webcomponents/webcomponentsjs/webcomponents-loader.js";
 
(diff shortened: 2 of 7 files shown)


samthor:
It occurs to me the other way to solve this is to bundle the -loader.js code and tell it where to find the various possible bundles (so we'd still need the gulp step).

It would be fewer network requests, but 🤷

gulpfile.js
```diff
@@ -65,9 +65,19 @@ gulp.task('copy-content-assets', () => {
     .pipe(gulp.dest('./dist/en'));
 });
 
+gulp.task('copy-node_modules-assets', () => {
```
robdodson: I was hoping to get rid of gulp at some point but I guess it's still proving useful as a file copy thinger... Though I'm wondering if we could drop it for an npm script or something...

src/site/_includes/layout.njk
```diff
@@ -18,6 +18,8 @@
       href="//fonts.googleapis.com/css?family=Google+Sans:400,500|Roboto:400,400italic,500,500italic|Roboto+Condensed:400,700|Roboto+Mono:400,500|Material+Icons"
     />
     <link rel="stylesheet" href="/app.css" />
+    {# Module code is loaded deferred (the "asynchronous" case in the polyfill docs) but we only load other module code anyway. #}
+    <script type="module" src="/lib/webcomponents/webcomponents-loader.js"></script>
```
robdodson: The docs seem to indicate that the polyfills themselves will be loaded async. Is webcomponents-loader just a boostrap that then script-appends the polyfills? If so, we might have to use their `waitFor` API, which feels like a bummer but maybe we can just use it to boostrap app.js or something.

samthor: Ugh, yeah. I think I failed to brain yesterday, of course we now have to `waitFor` this to be ready.

I'll have a think and update this soon.

gulpfile.js
samthor: I think folks overthink build systems. I've dropped Gulp from a bunch of my projects over the last couple of years in favour of literally just simple Node scripts which copy things into place. (You don't just need to try to do it in bash :) ).

src/site/_includes/layout.njk
samthor: I feel dirty (?) but I've inlined the loader now.

robdodson: I guess another option would be to create a bootstrap.js that dynamically imports app.js. I think I might prefer this approach as we might do more in bootstrap.js as we're planning to break up app.js and be a bit smarter about what we load and when.

samthor: I've done this now. I like it. I can also preload `app.js` in a `<link>` tag, which I think makes me happy.

samthor:
this is still broken because I'm not passing the root where the bundles exist properly. To be fixed...

src/lib/bootstrap.js
```diff
@@ -0,0 +1,21 @@
+/**
```
robdodson: you mentioned you have a similar file in santa that contains the version number. I was wondering if we should include something like that as well? Might help with future debugging in prod?

samthor: I've added that now. Take a look—I use Rollup to generate a virtual config file we can import.

robdodson:
seems ok to me. Mind fixing the conflict?

Issue #1539, opened 2019-09-18
Pull Firebase code from CDN


samthor:
We're bundling Firebase, but should instead be pulling down only specific parts of it from the CDN to benefit from global caching. It's just Auth and Firestore for now.

samthor:
So adding/removing `firebase/firestore` from our codebase shows that it's adding about ~1mb to our bundle (~500kb when minified), yet the downloaded size from CDN is about ~350kb (which we presume is minified).

🤔

PR #1548, opened 2019-09-20
add custom Firebase loader

samthor:
Full disclosure, I'm not convinced this is the right idea (only because of maintenance cost) but it _does_ one-up the standard way of loading Firebase.

This adds a a loader that loads the parts of Firebase that we need. Notably, while it always loads core (app/auth), it means we don't load Firestore until we need it (a user is signed in). That's a ~360kb saving.

We're also _already_ preconnect'ing to www.gstatic.com.


package.json
```diff
@@ -28,10 +28,10 @@
   "license": "Apache-2.0",
   "dependencies": {
     "express": "^4.16.4",
-    "firebase": "^6.2.4",
     "focus-visible": "^5.0.2",
     "lit-element": "^2.2.0",
     "navaid": "^1.0.5",
+    "terser": "^4.3.1",
```
robdodson: is this being used?

samthor: Nope, I just installed it to compare minified sizes (we should use it later).

robdodson:
@davideast 👋

We're exploring ways to break up and lazy load firebase auth + firestore to bring our bundle size down. Sam explored the idea of loading various parts via CDN in this PR but I also wanted to see if you have any advice here.

src/lib/fb.js
```diff
@@ -141,6 +192,7 @@ export async function saveUserUrl(url, auditedOn = null) {
 
 // Sign in the user
 export async function signIn() {
+  await firebaseReady;
```
robdodson: I think this is a bit of a drawback for me. Over time it feels like we might need to sprinkle these ready checks or forgetting to do so might lead to weird bugs.

samthor:
Also, see #1539 for comments on size.

src/lib/fb.js
samthor: We could always load app/auth (even as a module import side-effect from the actual CDN so it's happening in the same 'place' as the lazy-loading) but optionally load Firestore via the lazy-loading approach.

That way at least we'd reduce the complexity.

robdodson: yeah that might be good. I'm also curious to hear if David has any thoughts. I'm hoping there's a magical, well trod path that he can point us at 🤞

samthor:
So, I've updated this PR a bit.

I thought I could maybe load the Firebase deps as a static import inside an ES6 module. This didn't work for two reasons:

* we currently emit an 'iife' from Rollup (but this might change if we use Rollup bundling of ES modules), and this does not support `import`
* gstatic doesn't serve Firebase with CORS headers, so it cannot be loaded that way.

I've gone back to including the `<script>` tags for Firesbase directly. With `defer`, it's still async-ish, and runs before our module code (without blocking the page). But, I've left Firestore as a dynamic import, although placed the URL in a similar place to others so we update them at once.

I could revisit this when we have a bootstrap script working properly.

robdodson:
Sounds good. Let's forge ahead and maybe revist when we get into the code splitting stuff. Can you fixup the package.json conflict?

package.json
samthor: FYI I removed this as part of the merge.

PR #1549, opened 2019-09-20
add compile-css script which runs autoprefixer

### LINKED ISSUE #1377: Add autoprefixr to our CSS build.
We currently rely on Devsite's use of autoprefixr. When we migrate, we'll need to add this back ourselves.

Add autoprefixr to the CSS build. Likely targeting last two versions of each browser. Maybe IE11 is as high as we'd go for IEs. Or omit IE entirely and just favor Edge?

samthor:
Fixes #1377.

Does _not_ set browser targets. Apparently we can do that centrally [in package.json](https://github.com/browserslist/browserslist#browserslist-).


compile-css.js
```diff
@@ -0,0 +1,69 @@
...
+  omitSourceMapUrl: true, // since we just read it from the result object
+};
+if (isProd) {
+  compiledOptions.outputStyle = 'compressed';
+}
+log('Compiling', process.argv[2]);
+const compiledResult = sassEngine.renderSync(compiledOptions);
+
+// #2: Run postcss for autoprefixer.
+// TODO(samthor): This could also be disabled not in prod.
```
robdodson: Would you mind setting it so this only runs during a prod build. Seems like it adds slowness to a dev build where we won't benefit from it.

compile-css.js
```diff
@@ -0,0 +1,69 @@
+/*
```
robdodson: maybe we should copy the devsummit site and put build related tasks into a `lib/` directory in the root.

As a follow up, I want to rename the various `lib/` directories we have in `src/` because I think they're a bit confusing at the moment.

compile-css.js
samthor: Done

compile-css.js
samthor: Sure, but this PR doesn't change the status quo now. I probably agree but I wonder how big it will get.

robdodson:
Before landing this should we set a browser target? I guess I'm ok with last 2 versions unless there are any objections?

samthor:
I've added the browser target.

package.json
```diff
@@ -84,5 +86,8 @@
   },
   "optionalDependencies": {
     "node-sass": "^4.12.0"
-  }
+  },
+  "browserslist": [
```
robdodson: btw I've been trending toward providing .rc files for a lot of our tools but I don't have a specific reason for doing this. Do you have any reasons why you prefer putting stuff into package.json versus .rc? I guess one reason is that it's less files to keep track of?

samthor: I don't honestly have a strong opinion. Probably fewer files is ideal.

package.json
```diff
@@ -84,5 +86,8 @@
...
+    "last 2 versions"
```
robdodson: I think we've said that we don't support IE in the past, so I'd add `not ie`

package.json
robdodson: ok. we can add this to package.json and think about if we want to clean up the other .rc files in a future PR

package.json
robdodson: some data, IE is 0.42% of our traffic over the last week and I honestly feel like a certain portion of that has to be developers just checking to see if the site works in IE :P

samthor: Done

PR #1552, opened 2019-09-22
V2 snackbar

### LINKED ISSUE #1386: Recreate our cookies banner.
As part of the DevSite migration we'll need to recreate the cookie banner notification.

A cookies banner!

robdodson:
Fixes #1386

- Adds a Snackbar element with built-in cookies banner.
- Adds state to store, and actions for checking and accepting cookies.

This shows the cookie banner to all users, not just those coming from EU countries, which I'm ok with. GDPR all the things!


robdodson:
btw I think since the `actions.js` file is starting to get crowded, I might break it up into smaller files in a future PR.

src/lib/components/Snackbar/index.js
```diff
@@ -0,0 +1,100 @@
...
+      ms = CLOSING_ANIMATION_TIME;
+    }
+
+    this.animatable = true;
+    setTimeout(() => (this.animatable = false), ms);
+  }
+
+  onAccept() {
+    this.open = false;
+    // acceptAction is a property set by an ancestor or an action.
```
samthor: I feel like idiomatically we should be dispatching an `Event` here.

src/lib/components/Snackbar/index.js
```diff
@@ -0,0 +1,100 @@
...
+    if (this.acceptAction) {
+      this.acceptAction();
+      this.acceptAction = null;
+    }
+  }
+
+  get cookiesTemplate() {
+    return html`
+      <div class="web-snackbar__label" role="status" aria-live="polite">
```
samthor: TIL "polite".

src/lib/components/Snackbar/index.js
```diff
@@ -0,0 +1,100 @@
...
+        We serve cookies on this site to analyse traffic, remember your
+        preferences, and optimise your experience.
```
samthor: I hate to correct your spelling but "optimize" and "analyze"?

src/lib/components/Snackbar/_styles.scss
```diff
@@ -0,0 +1,120 @@
...
+web-snackbar[animatable] {
+  visibility: visible;
+
+  .web-snackbar__surface {
+    transition: opacity 75ms cubic-bezier(.4, 0, 1, 1),
+                transform 75ms cubic-bezier(0, 0, .2, 1);
+  }
+}
+
+web-snackbar[stacked] {
```
samthor: Why is this an attribute? Do you plan to have different snackbars?

src/lib/components/Snackbar/index.js
robdodson: Yeah, I think technically it's redundant since I'm using `role=status` and not `role=alert`.

status is implied aria-live=polite
alert is implied aria-live=assertive.

Good reminder that I need to test this with a screen reader before landing it :)

src/lib/components/Snackbar/index.js
robdodson: These are the British spellings that I stole from DevSite :D
But I'll switch them because we're going to show it to everyone anyway.

src/lib/components/Snackbar/_styles.scss
robdodson: I was copying the Material Components version. If you remove it, it should all layout horizontally. I was thinking we might use that for the "We have a new service worker, hit refresh!" notifications, if we choose to do that pattern.

image

src/lib/components/Snackbar/index.js
robdodson: I was using the UK spelling but I'll switch to the correct variation.

samthor:
FYI I'm waiting until you build the -container or tell me you've decided not to :)

src/lib/components/Snackbar/index.js
robdodson: Slightly changed this so we pass in an action. Adding unique event listeners was making LitElement recreate the elements which broke the CSS transitions.

src/lib/components/Snackbar/index.js
robdodson: I made it so the snackbar is first in the DOM. This is nice because when a user lands on the site they don't immediately have their focus stolen. But it's essentially the first thing they reach with a screen reader.

robdodson:
Container is in. PTAL :)

src/lib/actions.js
```diff
@@ -135,3 +135,25 @@ export const collapseSideNav = store.action(() => {
...
+
+  return {showingSnackbar: true, snackbarType: "cookies"};
+});
+
+export const setUserAcceptsCookies = store.action(() => {
+  localStorage.setItem("web-accepts-cookies", 1);
+  return {
+    userAcceptsCookies: true,
+    showingSnackbar: false,
+    // Note we don't set the snackbarType to null because that would cause the
```
samthor: I think we can solve this in the element too (if we see "type" => null, just hold it for the length of the animation). This is also fine, especially as you've left a comment. 👍

src/lib/components/Snackbar/index.js
```diff
@@ -0,0 +1,92 @@
...
+    let ms;
+    if (Boolean(val)) {
+      this.setAttribute("open", "");
+      ms = OPENING_ANIMATION_TIME;
+    } else {
+      this.removeAttribute("open");
+      ms = CLOSING_ANIMATION_TIME;
+    }
+
+    this.animatable = true;
```
samthor: There's a tiny race condition here (we clear `animatable` for the first anim, not the last) if you toggle this lots but it's probably fine.

robdodson: yeah my hunch is this won't be a high frequency toggle. One thing we may look into in the future is queuing snackbars (I think devsite does this). But for now we only have 1.

Issue #1505, opened 2019-09-16
Add route base code-splitting


robdodson:
**Describe the solution you'd like**
To keep the site nice and snappy, we should only load JavaScript that the user needs for the current route. This means a lot of the stuff for /measure could be deferred (though preloaded since we know that route is popular).

samthor:
My 2c on this is just to create perhaps three different entry points:

* signed in normal page
* logged out normal page
* Measure page (since it has the bulk of our code requirements)

These entry points do very little aside import further JS. We give the three of them to Rollup and let it *magically* come up with a graph of code, which we just ship as `type="module"`, importing missing content as users change states between the three possibilities above.

samthor:
Some thoughts: we could load our CSS per-module (in JS) and have them blit it out on use. This way we just use Rollup and our CSS becomes giant JS variables.

I'll do a proper writeup today, just some thoughts.

samthor:
We now need to add injecting CSS in JS and potentially any other "splits" we want (right now it's measure vs everything else). This will involve telling our build script how to render Sass so we can inline CSS.

robdodson:
sgtm!

PR #1581, opened 2019-09-30
switch to dynamic Rollup build system

samthor:
Partial solution to #1505 (does not attack CSS yet).

Adds Rollup-based code splitting. Changes to an imperative build process so we can use the output of bundling "pages" in bootstrap.

One Q I'd like to resolve: what a "URL" is could potentially be inconsistent. We e.g. load "index.html" in the router, and use URL as everything _after_ the initial slash. But pages can be served under e.g. "measure/" **and** "measure" (note missing trailing slash).

Is the allowed URL schema literally "/[a-z0-9]+"? Do we have any nested pages? Should pages canonically be served with a trailing slash?


robdodson:
> Is the allowed URL schema literally "/[a-z0-9]+"?
Pretty much.

> Do we have any nested pages?
Yes but only in the [handbook](https://web.dev/handbook) so far.
Example: https://web.dev/handbook/quick-start/

> Should pages canonically be served with a trailing slash?
Good question. I think your other PR addressed this and decided to canonically remove the trailing slash.

package.json
```diff
@@ -7,7 +7,7 @@
     "lint:scss": "sass-lint -c .sasslintrc -v -q",
     "build:gulp": "npx gulp build",
     "build:eleventy": "eleventy",
-    "build:rollup": "rollup -c rollup.config.js",
+    "build:rollup": "node build.js",
     "build:sass": "node ./compile-css.js src/styles/all.scss dist/app.css",
     "watch:gulp": "npx gulp build && gulp watch",
     "watch:eleventy": "eleventy --watch",
```
robdodson: Mind adding the `--quiet` flag to eleventy?

package.json
```diff
@@ -7,7 +7,7 @@
     "lint:scss": "sass-lint -c .sasslintrc -v -q",
     "build:gulp": "npx gulp build",
     "build:eleventy": "eleventy",
```
robdodson: Mind adding the `--quiet` flag to eleventy?

robdodson:
(don't punch me) 👊

I'm a little anxious about the amount of custom code in the PR. I think it may be easy for you to understand as the author, but I worry that myself and other folks on the team might have a hard time holding it in our heads.

I'm not saying we should throw this out, but I was wondering if you'd be willing to try writing a separate PR that:

- Puts the router itself into bootstrap.js so it's immediately available.
- Uses dynamic import instead of the custom `config` object.
- Uses the rollup plugin you authored to output the hash names to a .json file which eleventy can require.
- Updates the eleventy page templates to pull in that json and add their own `modulepreload` tags. It seems like almost every page will just `modulepreload` the `measure.js` and `measure.js` will just `modulepreload` the `default.js`.

I'd like to compare the two options if possible, to see which one is easier to grok. I might be ok leaving a little perf on the table if it means that the code is easier to reason about and more folks can contribute. Happy to also hop on a GVC to discuss if you want.

samthor:
I don't mind but I really think it trades complexity in one place for
another. Intermediate JSON files are also very hard to rationalize about,
and in this case the output from one task is literally the input to a
related one.




On Tue, 1 Oct 2019, 07:18 Rob Dodson, <notifications@github.com> wrote:

> (don't punch me) 👊
>
> I'm a little anxious about the amount of custom code in the PR. I think it
> may be easy for you to understand as the author, but I worry that myself
> and other folks on the team might have a hard time holding it in our heads.
>
> I'm not saying we should throw this out, but I was wondering if you'd be
> willing to try writing a separate PR that:
>
>    - Puts the router itself into bootstrap.js so it's immediately
>    available.
>    - Uses dynamic import instead of the custom config object.
>    - Uses the rollup plugin you authored to output the hash names to a
>    .json file which eleventy can require.
>    - Updates the eleventy page templates to pull in that json and add
>    their own modulepreload tags. It seems like almost every page will
>    just modulepreload the measure.js and measure.js will just
>    modulepreload the default.js.
>
>
This sounds wrong. Why would a non-measure page preload measure.js?

The point of preload is to pull in dependencies that the main entrypoint
needs, but which are shared with other entrypoints, so you don't incurr
excess request chains. These dependencies are the chunks that Rollup
generates.

And yes, there's only two for now. But imagine only loading e.g. the
Codelab component for a page which needs it.

My other concern, which we talked about, is injecting modulepreload in
_actual_ HTML causes us to touch every file every build if JS changes,
because the hashes will be new. This causes challenges for SW caching.

PR #1582, opened 2019-09-30
Adds loading bar to v2

samthor:
Adds loading bar for slow loads. No particular issue.

Slightly diverges (simplified) from DevSite styles.

### DIFF
--- a/src/lib/router.js
+++ b/src/lib/router.js
@@ -1,4 +1,5 @@
 import navaid from "navaid";
+import {store} from "./store";
 
 const router = navaid();
 let isFirstRun = true;
@@ -33,7 +34,7 @@ async function swapContent(url) {
     return;
   }
 
-  document.body.classList.add("loading");
+  store.setState({duringPageLoad: true});
 
   const main = document.querySelector("main");
   // Grab the new page content
@@ -46,7 +47,7 @@ async function swapContent(url) {
     window.location.href = window.location.href;
     throw e;
   } finally {
-    document.body.classList.remove("loading");
+    store.setState({duringPageLoad: false});
   }
   // Remove the current #content element
   main.querySelector("#content").remove();
--- a/src/lib/router.js
+++ b/src/lib/router.js
@@ -34,7 +34,7 @@ async function swapContent(url) {
     return;
   }
 
-  store.setState({duringPageLoad: true});
+  store.setState({isPageLoading: true});
 
   const main = document.querySelector("main");
   // Grab the new page content
@@ -47,7 +47,7 @@ async function swapContent(url) {
     window.location.href = window.location.href;
     throw e;
   } finally {
-    store.setState({duringPageLoad: false});
+    store.setState({isPageLoading: false});
   }
   // Remove the current #content element
   main.querySelector("#content").remove();
--- a/src/lib/router.js
+++ b/src/lib/router.js
@@ -33,6 +33,8 @@ async function swapContent(url) {
     return;
   }
 
+  document.body.classList.add("loading");
+
   const main = document.querySelector("main");
   // Grab the new page content
   let page;
@@ -43,6 +45,8 @@ async function swapContent(url) {
     // TODO(robdodson): In future, failure pages might be HTML themselves
     window.location.href = window.location.href;
     throw e;
+  } finally {
+    document.body.classList.remove("loading");
   }
   // Remove the current #content element
   main.querySelector("#content").remove();
--- a/src/lib/store.js
+++ b/src/lib/store.js
@@ -33,6 +33,9 @@ const initialState = {
   isSideNavExpanded: false,
   isSearchExpanded: false,
 
+  // Whether to show the progressbar and mark the main content as busy, during a load.
+  duringPageLoad: false,
+
   // When a user lands on the page, check if they have accepted our
   // cookie policy.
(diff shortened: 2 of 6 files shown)


src/styles/components/_progress.scss
```diff
@@ -0,0 +1,45 @@
...
+    content: '';
+    left: 0;
+    min-width: 320px;
+    position: absolute;
+    top: 0;
+    width: 50vw;
+  }
+}
+
+body:not([loading]) .w-loading-progress {
```
robdodson: looks like you're using an attribute here, but your code in `router` sets a class on `body`.

src/styles/components/_progress.scss
```diff
@@ -0,0 +1,45 @@
...
+  position: fixed;
+  right: 0;
+  top: 0;
+  transform: scaleY(1);
+  transform-origin: 50% 0%;
+  transition: transform $TRANSITION_SPEED ease;
+  transition-delay: 1s;  // show loading hint only after waiting this long
+  z-index: 100000;
+
+  // Shows a white "flurry" on top of the loading indicator.
```
robdodson: I think there may not be enough contrast between the foreground and background in the animation. I was actually staring at it for a sec and came back to look at the CSS because I remembered the "flurry" comment but didn't think I was seeing it.

src/site/_includes/layout.njk
```diff
@@ -38,6 +38,7 @@
       snackbar will be the first stop for the screen reader / focus.
     #}
     <web-snackbar-container></web-snackbar-container>
+    <div class="w-loading-progress"></div>
```
robdodson: Would using `<progress>` make life difficult? Was just thinking it might be nice semantically. But if it's fiddly, don't worry about it.

If we're going with a div, I'd suggest adding `role=progressbar aria-valuemin=0 aria-valuemax=100`. By not providing an `aria-valuenow` attribute, it conveys that it's indeterminate. [Explainer](https://www.w3.org/TR/wai-aria-1.1/#progressbar).

src/lib/router.js
```diff
@@ -33,6 +33,8 @@ async function swapContent(url) {
     return;
   }
 
+  document.body.classList.add("loading");
```
robdodson: Maybe a dumb question because I haven't thought through the load order but...

Would it be possible to turn these into states in the store (IS_PROGRESSBAR_SHOWING) and put this work into an action and have the router call the action to toggle the state? It might also be good to put `aria-busy=true` on the `<main>` element while we're doing this and set it back to "false" when we're done.

samthor: Done

src/site/_includes/layout.njk
samthor: I tried `<progress>` but it was difficult to style it correctly (because of `-webkit-appearance` and friends). I did the second approach.

src/styles/components/_progress.scss
samthor: I've made the flurry go to full-white in the middle. I think it's more obvious now. You need to really delay the load speed to see this ever show up.

src/styles/components/_progress.scss
samthor: fixed by using `&[hidden]` instead.

src/lib/store.js
```diff
@@ -33,6 +33,9 @@ const initialState = {
   isSideNavExpanded: false,
   isSearchExpanded: false,
 
+  // Whether to show the progressbar and mark the main content as busy, during a load.
+  duringPageLoad: false,
```
robdodson: Elsewhere I've been using an `is` prefix for boolean state stuff like `isSearchExpanded`.
wdyt about renaming to `isPageLoading` ?

src/styles/components/_progress.scss
```diff
@@ -0,0 +1,47 @@
...
+  left: 0;
+  pointer-events: none;
+  position: fixed;
+  right: 0;
+  top: 0;
+  transform: scaleY(1);
+  transform-origin: 50% 0%;
+  transition: transform $TRANSITION_SPEED ease;
+  transition-delay: 1s;  // show loading hint only after waiting this long
+  z-index: 100000;
```
robdodson: Can you check the z-index for the side nav drawer? I was curious if this would put it above that.

src/styles/components/_progress.scss
```diff
@@ -0,0 +1,47 @@
...
+  to   { visibility: visible; }
+}
+
+@keyframes loading-progress-before {
+  from { transform: translate(-50%); }
+  to   { transform: translate(100vw) translate(-50%); }
+}
+
+.w-loading-progress {
+  background: #3fc4ff;
```
robdodson:
```suggestion
  background: $WEB_SECONDARY_COLOR;
```

You'll need to import the _colors.scss for this.

robdodson:
Minor comments. Feel free to merge after you update.

src/styles/components/_progress.scss
samthor: Done

src/lib/store.js
samthor: Done

src/styles/components/_progress.scss
samthor: Yes. I've put it below. We should clean this up with Sass vars at some point.

PR #1584, opened 2019-09-30
add incredibly basic SW

samthor:
(Note that this rebases off #1581 and that GitHub might not play nice with that)

(Also, this is wrong for #1583).

Adds an incredibly basic Workbox-based SW as part of the build system.

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -1,0 +1,48 @@
+importScripts(
+  "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js",
+);
+
+// TODO(samthor): rollup a list of assets to precache on install.
+// TODO(samthor): match Google Fonts, Firebase and friends.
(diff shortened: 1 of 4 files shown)


src/lib/sw.js
```diff
@@ -0,0 +1,48 @@
+importScripts(
+  "https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js",
+);
+
+// TODO(samthor): rollup a list of assets to precache on install.
+// TODO(samthor): match Google Fonts, Firebase and friends.
```
robdodson: I think I have code for that which I stole from the workbox docs:
https://github.com/GoogleChrome/web.dev/blob/workbox/src/lib/service-worker.js#L26

robdodson:
Looks good to me so far. Happy to take a second look when we get #1581 sorted.

Issue #1370, opened 2019-08-19
Rebuild service worker.


robdodson:
**Is your feature request related to a problem? Please describe.**
As part of the devsite migration, we'll need to author our own service worker and PWA manifest. We'll need to make sure we transition smoothly off of the old service worker.

**Describe the solution you'd like**
A new service worker that dynamically caches pages as the user views them. Likely built with [Workbox](https://developers.google.com/web/tools/workbox/).

jeffposnick:
Just wanted to throw some random thoughts out there after being looped in on the discussion from https://github.com/GoogleChrome/web.dev/pull/1719#discussion_r337572735

First,

https://github.com/GoogleChrome/web.dev/blob/93694df6ec4874a85975127ce4a1856066e76bf5/src/lib/sw.js#L9

is not a good idea. (See https://github.com/GoogleChrome/workbox/issues/2259.) It will precache the contents of `/offline.html` once for each user, when the service worker is first installed, and then never update it, even if you change the contents of `/offline.html` and rebuild your site. The recommended approach is to use one of the Workbox build tools to generate the precache manifest (which will add a revision parameter containing a hash of the current `/offline.html` contents automatically), or alternatively, to use something like [this recipe](https://developers.google.com/web/tools/workbox/guides/advanced-recipes#warm_the_runtime_cache) to manually add `/offline.html` to the cache, which will end up refreshing things each time an updated service worker is installed.

I'm also concerned about an approach that caches full HTML documents and serves them via a stale-while-revalidate strategy,

https://github.com/GoogleChrome/web.dev/blob/93694df6ec4874a85975127ce4a1856066e76bf5/src/lib/sw.js#L20-L23

You're going to end up caching a lot of duplicated HTML (basically, all of the common headers/footers on a page), and more importantly, whenever you do update those headers/footers and redeploy your site, users are going to effectively see an arbitrarily old version of HTML for each URL that they revisit.

There's background on why this is a problem at
- https://jeffy.info/2017/01/24/offline-first-for-your-templated-site-part-2.html
- https://developers.google.com/web/updates/2018/05/beyond-spa#common_pitfalls

Working around this issue requires a more nuanced response strategy than runtime caching of full HTML.

CC: @philipwalton

jeffposnick:
Just a couple of notes based on our further conversation:

[...]

PR #1585, opened 2019-09-30
add manifest

samthor:
Adds manifest. For #1370.

I'd like to lazy-load PWACompat but we can try that with the Rollup build stuff using a dynamic `import()` when that's in.


src/misc/manifest.webmanifest
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "web.dev",
+  "short_name": "web.dev",
+  "start_url": "/",
+  "display": "standalone",
```
robdodson: btw I realized recently that the display: standalone version of the existing pwa makes it hard to share links. We have a share button but it is really just a tweet button. We might want to think about adding additional options for sharing in the future.

src/misc/manifest.webmanifest
```diff
@@ -0,0 +1,56 @@
+{
```
robdodson: maybe we should put this in `src/pwa` or something?

samthor: I don't know what else would go in here though. I think it would only be this file?

For misc, I was thinking of other files like "robots.txt", "opensearch.xml", ... anything misc and top-level.

robdodson: I was curious what devsite does and I think they put it in a pwa directory but the only other thing in their is the icons.

I guess misc is fine ¯\_(ツ)_/¯

PR #1596, opened 2019-10-02
V2+build all rollup

samthor:
Here's an alternative to #1581.

I do like using Rollup's built-in `import()` splitting support. It's easy^.

^ except we need a polyfill for Edge

But fundamentally we can't include the generated chunk names in the JS for `modulepreload`. Options are:
* don't do `modulepreload` at all
* modify every HTML page, risking cache invalidations for **every** article the second we touch a byte of JS
* ship another JS file that we `import()` or include, which could be `modulepreload`-ed (same number of request chains as #1581) but this is a whole different type of complexity.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     mocha: true,
     browser: true,
   },
+  parser: "babel-eslint",
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: "module",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -4,13 +4,6 @@ import commonJs from "rollup-plugin-commonjs";
 const isProd = process.env.ELEVENTY_ENV === "prod";
 const webdevConfigName = "webdev_config";
 
-let outputPattern;
-if (isProd) {
-  outputPattern = "[name].[hash].js";
-} else {
-  outputPattern = "[name].js";
-}
-
 const config = {
   prod: isProd,
   webcomponentsPath: isProd
@@ -39,18 +32,20 @@ const configPlugin = {
 
 module.exports = [
   {
-    input: "src/lib/app.js",
+    input: "src/lib/bootstrap.js",
     output: {
       dir: "dist",
-      format: "iife",
-      entryFileNames: outputPattern,
-      chunkFileNames: outputPattern,
+      format: "esm",
+      chunkFileNames: "_[hash].js",
+      entryFileNames: "[name].js", // this will only be "bootstrap.js"
       sourcemap: true,
+      dynamicImportFunction: "polyfillImport",
     },
     watch: {
       clearScreen: false,
     },
     plugins: [
+      configPlugin,
       resolve(),
       commonJs({
         include: "node_modules/**",
@@ -62,33 +57,11 @@ module.exports = [
     output: {
       dir: "dist/test",
       format: "iife",
-      entryFileNames: outputPattern,
-      chunkFileNames: outputPattern,
     },
     watch: {
       clearScreen: false,
     },
     plugins: [
-      resolve(),
-      commonJs({
-        include: "node_modules/**",
-      }),
-    ],
-  },
-  {
-    input: "src/lib/bootstrap.js",
-    output: {
-      dir: "dist",
-      format: "iife",
-      entryFileNames: outputPattern,
-      chunkFileNames: outputPattern,
-      sourcemap: true,
-    },
-    watch: {
-      clearScreen: false,
-    },
-    plugins: [
-      configPlugin,
       resolve(),
       commonJs({
         include: "node_modules/**",
--- a/src/lib/bootstrap.js
+++ b/src/lib/bootstrap.js
@@ -7,18 +7,18 @@
 
 import config from "./bootstrap-config";
 import "@webcomponents/webcomponentsjs/webcomponents-loader.js";
+import {router, entrypointLoaded} from "./router";
+import entrypointForRoute from "./entrypoint-for-route";
 
 console.info("web.dev", config.version);
 
 WebComponents.waitFor(async () => {
-  return new Promise((resolve, reject) => {
-    // nb. import() is fairly well supported (although not as much as raw modules), but we just
-    // don't need it
-    const s = document.createElement("script");
-    s.type = "module";
-    s.onerror = reject;
-    s.onload = () => resolve();
-    s.src = "/app.js";
-    document.head.appendChild(s);
-  });
+  // ... only load the first entrypoint if the user hasn't changed URLs in the meantime
+  if (!entrypointLoaded) {
+    await entrypointForRoute(window.location.pathname.substr(1));
+  }
 });
+
+// Run as long-lived router w/ history & "<a>" bindings
+// Also immediately calls `run()` handler for current location
+router.listen();
--- a/src/lib/polyfill/import.js
+++ b/src/lib/polyfill/import.js
@@ -1,0 +1,35 @@
+/**
+ * Simple polyfill for the `import()` function.
+ *
+ * Limitations:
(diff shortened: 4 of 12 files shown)


.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     mocha: true,
     browser: true,
   },
+  parser: "babel-eslint",
```
robdodson: is this needed for dynamic import?

src/lib/router.js
```diff
@@ -52,7 +61,7 @@ async function swapContent(url) {
 
 router
   .on("/", async () => {
-    return swapContent("index.html");
+    return swapContent("");
```
robdodson: When we dispatch the pageView event in the swapContent function, I think the url for the page will be blank. Seems like we should add a condition there.

src/lib/bootstrap.js
```diff
@@ -7,18 +7,18 @@
...
-    // don't need it
-    const s = document.createElement("script");
-    s.type = "module";
-    s.onerror = reject;
-    s.onload = () => resolve();
-    s.src = "/app.js";
-    document.head.appendChild(s);
-  });
+  // ... only load the first entrypoint if the user hasn't changed URLs in the meantime
+  if (!entrypointLoaded) {
```
robdodson: Did you add this here because the router returns if `isFirstRun` is true?
I was thinking that this work should live entirely in the router if possible.

src/lib/router.js
```diff
@@ -44,6 +50,9 @@ async function swapContent(url) {
     window.location.href = window.location.href;
     throw e;
   }
+
+  // Wait for code to be ready
+  await entrypointPromise;
```
robdodson: Should this go into a try/catch?

src/lib/polyfill/import.js
```diff
@@ -0,0 +1,35 @@
+/**
```
robdodson: Mostly curious—I was wondering why you chose not to use the [dynamic-import-polyfill](https://github.com/GoogleChromeLabs/dynamic-import-polyfill)?

src/lib/router.js
```diff
@@ -33,6 +35,10 @@ async function swapContent(url) {
     return;
   }
 
+  // Load the entrypoint simultaneously with content.
+  const entrypointPromise = entrypointForRoute(url);
```
robdodson: Maybe we should split `swapContent` so it contains two functions, one that handles fetching the js and one that handles fetching the html. Doing both things in the same function makes it a little confusing.

robdodson:
I think I'd like to rope in someone from the Speed team to give us some advice. We want to use a static site generator, code split, modulepreload, and cache our pages. But that combination of items seems incompatible without some really weird hoop jumping.

robdodson:
I think we may need to write a short doc laying out our solutions and get comments from folks. I guess this can also be a nice perf artifact :)

.eslintrc.js
samthor: No, it's needed because the current `eslint` doesn't think `import()` is allowed—it says it's a parse error. It expects only top-level import:

```js
import 'foo';
```

src/lib/router.js
samthor: This just matches the other URLs, like "measure", "learn", which don't have a prefix "/".

What do you want to do? Something special?

src/lib/router.js
samthor: The main thing I want to get across here is that the loads can happen in parallel. I can refactor it so it's two separate functions if you like, but we end up just awaiting on both results regardless.

rollup.config.js
```diff
@@ -39,18 +32,20 @@ const configPlugin = {
...
   {
-    input: "src/lib/app.js",
+    input: "src/lib/bootstrap.js",
     output: {
       dir: "dist",
-      format: "iife",
-      entryFileNames: outputPattern,
-      chunkFileNames: outputPattern,
+      format: "esm",
+      chunkFileNames: "_[hash].js",
```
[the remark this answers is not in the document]
samthor: It could be. In that case it generates two types of files: "measure.[hash].js" as well as lots of "chunk.[hash].js". It doesn't really matter in practice—are we going to care what our dist JS bundles are named?

samthor:
Re: meaningful routing.

Routing doesn't need Web Components or other site code to be ready—it's self-contained. And the "complexity" is a single conditional statement with a bunch of comments.

src/lib/bootstrap.js
samthor: I answered this down below but I'll put it again here (oops).

Routing doesn't need Web Components or other site code to be ready—it's self-contained. And the "complexity" is a single conditional statement with a bunch of comments.

src/lib/polyfill/import.js
samthor: No polyfill is actually "correct", because they cannot resolve relative paths. They also do more work, because they care about resolving the results of an import, whereas we explicitly _do not_ need that.

It's up to you, but in general I think a few well-commented lines here are clearer than a magic black box that is basically doing the same thing.

robdodson:
> Routing doesn't need Web Components or other site code to be ready—it's self-contained. And the "complexity" is a single conditional statement with a bunch of comments.

OK, I think I'm missing something, bare with me :)

The router contains code to load the entrypoint files, but it seems like the conditional does the same thing. The entrypoint files depend on the web components polyfills because they contain code that calls customElements.define(). I was wondering if the conditional could be replaced by moving `router.listen()` into that position:

```
WebComponents.waitFor(async () => {
  router.listen(); // looks for entrypoints for the current route
});
```

I think one change that would need to be made is currently the router contains an `isFirstRun` flag which tells it not to call `swapContent()` ([link](https://github.com/GoogleChrome/web.dev/blob/1bf2dbf3116f6e7651c9bf42de013bd63b540c62/src/lib/router.js#L31)). If we split `swapContent()` into separate functions, one for loading html and one for loading the entrypoint js, then we could have the function that loads entrypoint js execute on first run.

.eslintrc.js
robdodson: Sorry if I wasn't clear, that's what I meant. The linter isn't recognizing dynamic import.

rollup.config.js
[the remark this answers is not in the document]
robdodson: I found it helpful just now because I was curious what went into each bundle.

src/lib/router.js
robdodson: Ah I think it's a moot point because `ga("send", "pageview")` doesn't actually pass along the url (I guess it's using the location).

I thought the next route, `/*`, is what matches pages like /measure and /learn. In those cases it seems fine since `'measure'` and `'learn'` are passed as the `url`. So when the handler calls:

```
document.dispatchEvent(new CustomEvent("pageview", {detail: url})); 
```

it works as expected.

In the example I highlighted, since `url` is an empty string, I thought we were telling google analytics to track the url `""` instead of `"/"`. My mistake 😅

src/lib/polyfill/import.js
robdodson: yep that's cool with me 👍

src/lib/router.js
robdodson: yep, understood. I was thinking it might be good as two functions in case we continue to add logic to it. Being able to separate at a glance which parts are related to loading the js and which parts are related to loading the html might be helpful.

robdodson:
It seems like if we want to maximize performance then #1581 may be the right option. I guess the dynamic import model is nice for a SPA that always sends down index.html but then just fetches data for subsequent pages and templates them out. In our case, as you pointed out, we would need to invalidate all of our html pages when we need to update a hash and that seems...ungood.

I'm going to see if I can chat with Jeff or Phil tomorrow just to see if they have any ideas. The easiness of dynamic import is a bummer to lose, but it just might not make sense in a funky SSG scenario.

I am a bit curious how something like Gatsby handles this situation...

PR #1611, opened 2019-10-07
Add styling for external links

mfriesenhahn:

- Add styling for links with a [rel=noopener] attribute.
- Add guidance to handbook to help ensure rel=noopener is used for external links.

### DIFF
--- a/src/site/content/en/handbook/web-dev-components/index.md
+++ b/src/site/content/en/handbook/web-dev-components/index.md
@@ -27,6 +27,7 @@ guidance about how to use them effectively.
 1. [Glitches](#glitches)
 1. [Images](#images)
 1. [Instructions](#instructions)
+1. [Links (external)](#links-(external))
 1. [Lists](#lists)
 1. [Stats](#stats)
 1. [Tables](#tables)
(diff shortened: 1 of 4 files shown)


src/site/content/en/handbook/web-dev-components/index.md
```diff
@@ -722,6 +723,18 @@ for unordered lists.
 - Lorem ipsum dolor sit amet, consectetur adipiscing elit. Proin dictum a massa
   sit amet ullamcorper.
 
+## Links (external)
+
+```html
+Lorem ipsum dolor sit amet,
+<a href="example.com" rel="noopener">consectetur adipiscing elit</a>.
```
kaycebasques: This example isn't using `target="_blank"`

robdodson:
@mfriesenhahn do you happen to know of a good url where i can see this in action?

robdodson:
oh wait there's an example in the handbook. ignore me

kaycebasques:
image

src/site/content/en/handbook/web-dev-components/index.md
robdodson: I think css works if it's using `rel=noopener` or `target=_blank`.

robdodson:
I think Kayce's approach sounds like it could work. We could probably also enforce it via some linting rules.

mfriesenhahn:
I thought there was some discussion about adding some logic to the Markdown parser to (a) flag any links with `web.dev` in them and (b) add `rel="noopener"` to external links? I remember Pete didn't want to have to remember to use markup for external links, which, you know, fair. So, if we do that, this PR needs reworked. 

Other than that, happy to do Kayce's approach for the CSS.

mfriesenhahn:
Well, actually, I guess if we handle external linking in the parser, the CSS matters less since we can just set it to whatever the parser outputs for external links.

If not, I'd kinda prefer to _add_ Kayce's approach to the `rel=noopener` rule to make it less likely we'll have false negatives.

robdodson:
Tagging @MichaelSolati who understands the markdown parser stuff better than I do.

We could do this as a two phase approach. Phase 1 would be to land the PR as is, since we have recently landed PRs to ensure external links contained rel=noopener.

Phase 2 would be to update the parser to look for external links and spit out the noopener attribute on them.

mfriesenhahn:
Works for me. We'll just need to remember to update the handbook when that phase 2 PR lands.

Lemme add Kayce's rule to the CSS!

mfriesenhahn:
Added the rule

robdodson:
Heads up, I think I'm going to need to revert this commit. It seems to have introduced some weirdness. My bad for not testing better.

image

image

PR #1634, opened 2019-10-09
Update blog/css-props-and-vals/index.md

WORMSS:
Looked like a small copy/paste error.

```diff
-  transition: --stop-color: 1s;
+  transition: --stop-color 1s;
```

### DIFF
--- a/src/site/content/en/blog/css-props-and-vals/index.md
+++ b/src/site/content/en/blog/css-props-and-vals/index.md
@@ -170,7 +170,7 @@ following CSS can be written with the same registered property:
 button {
   --stop-color: red;
   background: linear-gradient(var(--stop-color), black);
-  transition: --stop-color: 1s;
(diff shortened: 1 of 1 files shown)


WORMSS:
@googlebot I signed it!

robdodson:
@Snugug should it be `transition: var(--stop-color) 1s;` ?

WORMSS:
If that is the case, the external example also needs to be updated. My fix made it match the example. But the example didn't work for me but I was assuming my browser couldn't handle the feature yet.

Snugug:
@robdodson nope, @WORMSS is correct. It should be `transition: --stop-color 1s;`

robdodson:
> @robdodson nope, @WORMSS is correct. It should be transition: --stop-color 1s;

ok cool, thanks for taking a look!

PR #1647, opened 2019-10-10
Remove aside from ot include.

robdodson:
It looks like we created this include but it isn't used anywhere yet. We should probably do a separate PR to add it to existing articles :)

**Changes proposed in this pull request:**

-  Remove the `Aside` from the origin-trial include.

If folks would like to use this include within an Aside they can still do so:

```
{% Aside %}
{% include 'content/origin-trials.njk' %}
{% endAside %}
```

cc @petele @jpmedley


petele:
Can we land this ASAP? It removes the aside wrapper, but it also fixes the broken link to OT Guide for Web Devs.

robdodson:
yep, just landed

PR #1679, opened 2019-10-16
migrate to supporting /measure/ (trailing slash) pages

samthor:
As per your requests, Rob :)

We still have to work around Navaid's awkwardness. To be fair, if you remove all links to "/foo" (and "/foo/index.html", although I don't they actually exist anywhere) inside our codebase, some of this isn't needed—but I feel like folks might accidentally add them back in.

### DIFF
--- a/server.js
+++ b/server.js
@@ -17,37 +17,9 @@
 const express = require('express');
 const app = express();
 
-// Match e.g., "/hello", "/hello/foo", but not "/blah/foo.svg". Does not match "/": this is checked
-// in the conditional below on its own.
-const contentPageMatch = /^(\/\w+)+$/;
-
-// Disallow requests to "index.html" in any subdirectory (send a 302) along with
-// requests to a subdirectory that end with "/". This means that e.g., the
-// measure page, technically at "https://web.dev/measure/index.html", will
-// only be served under "https://web.dev/measure".
-// This resolves ambiguity for Service Workers and the Navaid router (which
-// internally trims trailing slashes for JS-based routing).
-const contentHandler = (req, res, next) => {
-  if (req.url === '/') {
-    req.url = '/index.html';
-  } else if (req.url.endsWith('/index.html') || req.url.endsWith('/')) {
-    const index = req.url.lastIndexOf('/');
-    const redir = req.url.substr(0, index);
-
-    // 301 means that Chrome and friends don't put the incorrect URL in the browser's history.
-    res.writeHead(301, {Location: redir});
-    return res.end();
-  } else if (contentPageMatch.exec(req.url)) {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -17,37 +17,9 @@
...
-  }
-
-  return next();
-};
-
 app.use(
-  express.static('dist', {index: false}),
-  contentHandler,
-  express.static('dist/en', {index: false, redirect: false}),
+  express.static('dist', {index: false, redirect: false}),
```
samthor: pre-emptive comment: this folder should _never_ contain actual HTML content, so I figure disabling defaults here is fine

robdodson: Would you mind adding a comment to the code that explains that.
What's the benefit to disabling the defaults?

src/lib/router.js
```diff
@@ -52,10 +52,23 @@ async function swapContent(url) {
...
+      const stripped = params.wild.slice(0, -"index.html".length);
+      return swapContent(`/${stripped}`);
+    } else if (window.location.pathname.endsWith("/")) {
+      // Navaid strips a trailing "/" on its own, so ensure it is added again before loading.
+      return swapContent(`/${params.wild}/`);
+    }
+
+    // Otherwise, this is a request for e.g. "/measure". By calling history.replaceState() to
+    // re-add the trailing slash, Navaid will re-run this code and trigger the latter conditional
+    // above. (If we also call swapContent(), two requests will fire.)
```
robdodson: > (If we also call swapContent(), two requests will fire.)

I wasn't sure what this comment was getting at. Is it saying if we call swap content here, and then update the path it'll trigger the conditional and call swapContent again? Might omit this part since it's implied. I think the comment  may have made it seem more complex than it is.

src/lib/router.js
```diff
@@ -52,10 +52,23 @@ async function swapContent(url) {
...
+    // This doesn't modify the back stack, so further navigation works fine.
+    window.history.replaceState(null, null, window.location.pathname + "/");
```
robdodson: For the future...

This makes me think that we're working around navaid enough that it might be worth us either using a different router (page.js maybe?) or just roll our own. Not something we need to do now though.

server.js
samthor: I think your implicit comment is correct; I've reverted to the default. I think there are a bunch of edge cases here that I'm too worried about.

src/lib/router.js
samthor: I've made the comment much simpler.

src/lib/router.js
samthor: Yes, I agree with this sentiment.

PR #1681, opened 2019-10-16
Tweaks style of search box

samthor:
Just my opinion but I think we can make this look nicer.

* we were using the default `<input>` font—migrate to use the page's font (matches DevSite)
* replace the blue focus outline with an _underline_

Looks a bit like this:

Screen Shot 2019-10-16 at 17 39 10

Using an underline rather than a surrounding box for focus makes the appear animation less jarring—if it's a whole box, you don't see the right side of the focus until the animation is done. The underline makes the input box seem more like it's growing. (DevSite grows the `<input>` itself but we don't, for whatever reason).

### DIFF
--- a/src/lib/components/Search/_styles.scss
+++ b/src/lib/components/Search/_styles.scss
@@ -101,6 +101,16 @@ web-search {
     border: none;
     padding: 8px 8px 8px 40px;
     width: 100%;
+    font: inherit;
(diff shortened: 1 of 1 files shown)


robdodson:
ah yeah, when I removed the devsite CSS in #1667 it caused these changes. Suggestions LGTM!

PR #1656, opened 2019-10-11
Collection for PWAs

petele:
Fixes #1440 

Changes proposed in this pull request:
- Creates the collection for PWAs

PR #1699, opened 2019-10-17
Let paths be marked as drafts.

robdodson:
Related to #1656

- Quick fix to let paths be marked as drafts. To mark a learning path as a draft, add `draft: true` to the learning path's `path` property in its 11tydata.js file.

Example:

```
module.exports = {
  // Tags are inherited by all posts.
  tags: ['pathItem', 'accessible'],
  path: {
    draft: true,
    ...
  },
};
```

### DIFF
--- a/src/site/_data/postHost.js
+++ b/src/site/_data/postHost.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-const data = require('../content/en/learn/learn.11tydata.js');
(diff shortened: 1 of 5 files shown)


robdodson:
Since this touches postHost, which is used by the /measure page to connect docs to LH audits, we need to test that we didn't break anything there before we push to production.

robdodson:
@petele fyi

Issue #1601, opened 2019-10-02
Router should scroll to the top when a page changes.


robdodson:
When navaid updates a page, we should scroll to the top of the page.

PR #1702, opened 2019-10-17
nav tweaks

### LINKED ISSUE #1600: Router should change page title and manage focus.
**Describe the bug**
When loading a new page into the shell, the router should update the title for the site and should move focus to the first heading in the new page to signal to screen readers that the content is loaded.

samthor:
Fixes #1600 and #1601.

Unrelated but I'm inclined to remove Navaid since we're working around more of it. For instance, I'd love to retain scroll position in history state, so that the 'back' button feels more natural (going "back" will just reset scroll to zero). But that's [not possible](https://github.com/lukeed/navaid/issues/6).

### DIFF
--- a/src/lib/router.js
+++ b/src/lib/router.js
@@ -58,8 +58,13 @@ async function swapContent(url) {
 
   // Grab the new page content
   let page;
+  let content;
   try {
     page = await getPage(url);
+    content = page.querySelector("#content");
+    if (content === null) {
+      throw new Error(`no #content found: ${url}`);
+    }
     await entrypointPromise;
   } catch (e) {
     // If something fails, just make a browser URL change
@@ -76,9 +81,13 @@ async function swapContent(url) {
   // Remove the current #content element
   main.querySelector("#content").remove();
   // Swap in the new #content element
-  main.appendChild(page.querySelector("#content"));
+  main.appendChild(content);
   // Update the page title
   document.title = page.title;
+  // Focus on the first title (or fallback to content itself)
+  (content.querySelector("h1, h2, h3, h4, h6, h6") || content).focus();
+  // Scroll to top
+  document.scrollingElement.scrollTop = 0;
 }
 
 router
--- a/src/lib/router.js
+++ b/src/lib/router.js
@@ -35,7 +35,8 @@ async function getPage(url) {
 }
 
 /**
- * Force the user's cursor to the target element.
+ * Force the user's cursor to the target element, making it focusable if needed.
+ * After the user blurs from the target, it will restore to its initial state.
  *
  * @param {?Element} el
  */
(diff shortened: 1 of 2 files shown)


src/lib/router.js
```diff
@@ -71,9 +76,13 @@ async function swapContent(url) {
   // Remove the current #content element
   main.querySelector("#content").remove();
   // Swap in the new #content element
-  main.appendChild(page.querySelector("#content"));
+  main.appendChild(content);
   // Update the page title
   document.title = page.title;
+  // Focus on the first title (or fallback to content itself)
+  (content.querySelector("h1, h2, h3, h4, h6, h6") || content).focus();
```
robdodson: Wanted to mention that this won't do anything unless the headings or content have tabindex=-1. We could add the tabindex in our various page templates.

robdodson: Also noticed there are two h6s on line 83

samthor: oops re: h6

I've added a `forceFocus` function which sets a tabindex imperatively to move the cursor. WDYT of that approach from an a11y point of view?

src/lib/router.js
```diff
@@ -33,6 +33,23 @@ async function getPage(url) {
...
+ */
+function forceFocus(el) {
+  if (!el) {
+    // do nothing
+  } else if (el.hasAttribute("tabindex")) {
+    el.focus();
+  } else {
+    el.tabIndex = -1;
+    el.focus();
+    el.deleteAttribute("tabindex");
```
robdodson:
```suggestion
    el.removeAttribute("tabindex");
```

robdodson: I just tested this and removing the attribute causes focus to blur and go back to the body.

If you want to suppress the focus ring, I think instead of removing the attribute we could add a class, something like `.w-force-focus` which sets `outline: 0 !important;`

We could then either leave the class in place, or add a `once` listener to the element's blur event to remove the class.

samthor: So you're not wrong but _keyboard focus_ now starts from this location. I've used this trick in a few places (`deleteAttribute` being wrong notwithstanding 🤣) and I always thought it was good enough.

Isn't it weird to have a focused thing with no highlight? Can I set some aria-property to imply that focus is here instead, at least until I focus on something new?

robdodson: > So you're not wrong but keyboard focus now starts from this location. I've used this trick in a few places (deleteAttribute being wrong notwithstanding 🤣) and I always thought it was good enough.

Yeah it may move the focus navigation start point, but the screen reader cursor doesn't go to the header. I think it actually goes to the nearest landmark which is main in this case. So if you turn on a screen reader and click a link, instead of reading the header text (which is nice) it says "main". IOW, you've improved the experience for folks using keyboard and switch devices, but not for folks using screen readers.

> Isn't it weird to have a focused thing with no highlight?

Managing focus like you're doing in this case is basically the one exception where you're allowed to put "focus" on something and hide the ring. The ring is meant to indicate that the user can interact with something. In this case we're putting focus on a non-interactive header to both move the focus navigation start point and to have a screen reader tell the user that new content has loaded.

> Can I set some aria-property to imply that focus is here instead, at least until I focus on something new?

I don't think so. ARIA technically doesn't affect focus behavior at all. It's just meant to add semantic info to the tree. I think you might be thinking of things like [aria-owns](https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-owns) where you can kind of say "hey even though this element isn't my child, pretend like it is."

samthor: I've incorporated your suggested changes now!

PR #1709, opened 2019-10-21
Header style tweaks for v2

samthor:
@robdodson I think some of these changes might have just been broken in your DevSite code removal.

* Restores the blue underline indicating which top-level page the user is on (JS + CSS changes)
  * This also makes `web-search` have a white background so it covers this while open
* Removes the default grey background in responsive mode for "Sign In" and the search open/close toggle
* Make the "SIGN IN" button have `cursor: pointer` (matchs DevSite, where it's technically a link)

### DIFF
--- a/src/lib/components/Header/_styles.scss
+++ b/src/lib/components/Header/_styles.scss
@@ -104,6 +104,10 @@ web-header {
     margin: 0;
     padding: 0 24px;
     position: relative;
+
+    &[active] {
+      box-shadow: 0 -2px 0 $WEB_PRIMARY_COLOR inset;
+    }
   }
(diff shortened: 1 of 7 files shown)


samthor:
It occurs to me another approach to the events I'm listening for in `web-header` is to add URL to global state. WDYT?

robdodson:
> It occurs to me another approach to the events I'm listening for in web-header is to add URL to global state. WDYT?

Yeah I was wondering if we should put the active page into the `store` in case other things need it in the future. Do you think the router could set this state?

samthor:
Done the router state change.

PR #1710, opened 2019-10-21
fix searching with no results

samthor:
Fixes a few issues.

* Hitting "Enter" with no search results caused a JS error
* Adds a message indicating that no results were found with a link to Googs
* Simplifies the `focusout` handler by just invoking `.click()` programatically

### DIFF
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -43,6 +43,7 @@ class Search extends BaseElement {
     this.showHits = false;
     this.cursor = -1;
     this.query = "";
+    this.timeout = 0;
 
     // On smaller screens we don't do an animation so it's ok for us to fire off
     // actions immediately. On larger screens we need to wait for the searchbox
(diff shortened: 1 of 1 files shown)


src/lib/components/Search/index.js
```diff
@@ -107,10 +108,28 @@ class Search extends BaseElement {
...
+    if (!this.showHits) {
       return "";
     }
 
+    if (!this.hits.length) {
+      if (!this.query) {
+        return "";
+      }
+
+      const query = "web.dev " + this.query.trim();
```
robdodson: Did you want to scope search results just to web.dev? If so I think you could do:

```
const query = "site:web.dev " + this.query.trim();
```

samthor: I intentionally didn't. I think Algolia is pretty good and Google finds our site most of the time anyway, but this just broadens the possible answers for users.

I've added a comment.

samthor: And an Analytics event.

PR #1714, opened 2019-10-21
make Measure signin button work

samthor:
Adds `web-signin-button`, which displays the giant sign in button for the Measure page. This previously wasn't hooked up.

Also adds `BaseStateElement`. If you like it, I'll migrate all the state-subscribing elements to listen to it.

### DIFF
--- a/src/lib/components/BaseStateElement/index.js
+++ b/src/lib/components/BaseStateElement/index.js
@@ -1,0 +1,30 @@
+import {store} from "../../store";
+import {BaseElement} from "../BaseElement";
+
+/**
+ * Base element which subscribes to global state.
+ *
+ * @extends {BaseElement}
+ */
+export class BaseStateElement extends BaseElement {
+  constructor() {
+    super();
+    this.onStateChanged = this.onStateChanged.bind(this);
+  }
+
+  connectedCallback() {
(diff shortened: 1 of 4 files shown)


samthor:
Also, as an alternative, the signin button could _not_ listen to state, and instead be hidden or shown by the global CSS override that we set in app.js. But its click handler needs to point to Firebase code, which is why it must be an element.

robdodson:
>Also, as an alternative, the signin button could not listen to state, and instead be hidden or shown by the global CSS override that we set in app.js. But its click handler needs to point to Firebase code, which is why it must be an element.

I think I prefer doing this through the store so +1 for the approach in the PR.
I think there's just one little span that otherwise relies on that `lh-signedin` class. Kinda wish we could remove it but I guess it's fine for now.

One thing to be mindful of is that this introduces some FOUC that wasn't there before. It's not a big deal, and maybe we could use the approach devsite uses of adding a little opacity fade-up when these controls upgrade?

samthor:
I agree about opacity fade-up as well as what we discussed over chat, which is storing what we _think_ the signed-in state is, inside `localStorage`. There might still be a flash, but for fully-cached clients, it should be almost imperceptible.

PR #1717, opened 2019-10-21
_redirect.yaml support for v2

samthor:
Adds a handler that works with our existing `_redirect.yaml` file format.

Otherwise links like e.g., "/subscribe" don't work.

Removes the other disused yaml files from DevSite.


server.js
```diff
@@ -14,13 +14,32 @@
...
+// In development, Eleventy isn't guaranteed to have run, so read the actual
+// source file.
+const redirectsPath = isProd
+    ? 'dist/en/_redirects.yaml'
+    : 'src/site/content/en/_redirects.yaml';
+
+// Don't block loading the server if the redirect handler couldn't build.
+try {
+  const redirectHandler = buildRedirectHandler(redirectsPath);
+  handlers.push(redirectHandler);
```
samthor: nb. This could be `unshift` instead, if we think that this should win over actual content.

robdodson: I think push is the right behavior... If I recall the way devsite works is, if the file exists, it will use it. Otherwise it will redirect. This seems to match that behavior.

robdodson:
This is great, thank you for catching this one!

PR #1719, opened 2019-10-21
V2+sw stuff

samthor:
Adds a 404 and offline page. Serves the offline page via SW only.

Notes:

* this is a bit hard to test because the SW only responds to "/foo/", yet most links are still "/foo". (Express is doing the redirect)—see TODOs in sw
* we serve the 404 html for things like missing PNG assets
* we should later annotate the offline page so that the frontend code can re-request when it gets an 'online' event.

### DIFF
--- a/server.js
+++ b/server.js
@@ -19,24 +19,34 @@ const isProd = Boolean(process.env.GAE_APPLICATION);
 const express = require('express');
 const buildRedirectHandler = require('./redirect-handler.js');
 
+const redirectHandler = (() => {
+  // In development, Eleventy isn't guaranteed to have run, so read the actual
+  // source file.
+  const redirectsPath = isProd
+      ? 'dist/en/_redirects.yaml'
+      : 'src/site/content/en/_redirects.yaml';
+
+  // Don't block loading the server if the redirect handler couldn't build.
+  try {
+    return buildRedirectHandler(redirectsPath);
+  } catch (e) {
+    console.warn(e);
+    return (req, res, next) => next();
+  }
+})();
+
+// 404 handlers aren't special, they just run last.
+const notFoundHandler = (req, res, next) => {
+  const options = {root: 'dist/en'};
+  res.sendFile('404/index.html', options, (err) => err && next(err));
+};
+
 const handlers = [
   express.static('dist'),
   express.static('dist/en'),
+  redirectHandler,
+  notFoundHandler,
 ];
-
-// In development, Eleventy isn't guaranteed to have run, so read the actual
-// source file.
-const redirectsPath = isProd
-    ? 'dist/en/_redirects.yaml'
-    : 'src/site/content/en/_redirects.yaml';
-
-// Don't block loading the server if the redirect handler couldn't build.
-try {
-  const redirectHandler = buildRedirectHandler(redirectsPath);
-  handlers.push(redirectHandler);
-} catch (e) {
-  console.warn(e);
-}
 
 const app = express();
 app.use(...handlers);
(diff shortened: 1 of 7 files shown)


robdodson:
looks like travis is grumping about some eslint/prettier things

server.js
```diff
@@ -17,9 +17,16 @@
 const express = require('express');
 const app = express();
 
+// 404 handlers aren't special, they just run last.
+const notFoundHandler = (req, res, next) => {
+  const options = {root: 'dist/en'};
+  res.sendFile('404/index.html', options, (err) => err && next(err));
```
robdodson: I guess a TODO we should add is to write up a legit error handler that logs the errors...somewhere :P
I guess app engine logs it for us but maybe we should do something nicer?

src/site/content/en/404.njk
```diff
@@ -0,0 +1,15 @@
...
+title: 404
+description: |
+  Page Not Found
+---
+
+<div class="w-layout-container--large">
+  <header class="w-page-header">
+    <h1 class="w-page-header__headline w-code">{{ title }}</h1>
+    <p class="w-page-header__copy">
+      Sorry, we couldn't find that page.
```
robdodson: Should we include a link back to the main page?

src/lib/sw.js
```diff
@@ -19,3 +23,11 @@ workbox.routing.registerRoute(
   new RegExp("/([\\w-]+/)*$"),
   new workbox.strategies.StaleWhileRevalidate(),
 );
+
+workbox.routing.setCatchHandler(({event}) => {
```
robdodson: Do we hit this only when the user is offline or does it trigger if any of the routes throws an error? Maybe it's fine, but I was curious if we should double check `navigator.onLine` (even though it's terrible...) and if the user is online but a route just exploded, maybe we give them an 'oops! something went wrong' page? Maybe that'll never happen in practice tho ¯\_(ツ)_/¯

src/styles/components/_page-header.scss
```diff
@@ -30,6 +30,10 @@
 
 .w-page-header__headline {
   margin: 0 auto 11px;
+
+  &.w-code {
```
robdodson: I think we updated code inside of headers on master to look nicer. I'm not 100% sure if you'll still need this if we rebase against master.

src/lib/sw.js
samthor: This was the advice in the official docs: https://developers.google.com/web/tools/workbox/guides/advanced-recipes#provide_a_fallback_response_to_a_route

Remember that 404's aren't errors. So I think this is safe. Honestly I am probably just going to ask @jeffposnick to look at the site before we launch it :) 👋

src/styles/components/_page-header.scss
samthor: I'll let you deal :)

DevSite makes the 404 page appear in code font. I like it as a stylistic thing.

src/lib/sw.js
jeffposnick: Right, `catchHandler` will fire whenever `workbox-routing` encounters an exception when trying to generate a response. This would most likely happen due to being offline without an appropriate cache entry being available to fall back to.

I'm super-happy to review the full service worker, and in particular, I can help you move to Workbox v5.

robdodson: > Right, catchHandler will fire whenever workbox-routing encounters an exception when trying to generate a response. This would most likely happen due to being offline without an appropriate cache entry being available to fall back to.

👍

> I'm super-happy to review the full service worker, and in particular, I can help you move to Workbox v5.

Thank you Jeff! I think we're getting pretty close so we'll give you a shout when we hit a stopping point.

PR #1724, opened 2019-10-22
Prettify header and user image

samthor:
This tries to fix causes of FOUC.

* We store whether we think the user was previously signed in inside `localStorage` which propagates to the global state, so it's available before Firebase tells us anything
* Fades in the user's profile picture only after it loads (and shows a placeholder silhouette until then)
* Fixes a bug with `<web-search>` (it would "click" on the thing you focused on _outside_ itself)
* Expands the allowed content on the Measure page so that FOUC is less annoying on large screens (the top line of text now doesn't wrap)

### DIFF
--- a/src/lib/components/ProfileSwitcher/_styles.scss
+++ b/src/lib/components/ProfileSwitcher/_styles.scss
@@ -57,12 +57,16 @@ web-profile-switcher {
 }
 
 .w-profile-toggle {
-  border: 4px solid transparent;
+  background: #F1F3F4;
+  border: 4px solid white;
   border-radius: 50%;
   box-shadow: none;
+  box-sizing: content-box;
   display: flex;
   height: auto;
   justify-content: center;
+  min-height: 32px;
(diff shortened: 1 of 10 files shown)


src/lib/components/ProfileSwitcher/_styles.scss
```diff
@@ -57,12 +57,16 @@ web-profile-switcher {
 }
 
 .w-profile-toggle {
-  border: 4px solid transparent;
+  background: #F1F3F4;
```
robdodson:
```suggestion
  background: $GREY_100;
```

src/lib/components/ProfileSwitcher/_styles.scss
```diff
@@ -57,12 +57,16 @@ web-profile-switcher {
...
+  border: 4px solid white;
```
robdodson:
```suggestion
  border: 4px solid $WHITE;
```

PR #1725, opened 2019-10-22
V2+replace navaid

samthor:
Swap navaid out for custom router.

I actually like this a bunch and it works around issues in Navaid and friends, at the cost of a purely-imperative loader function—which is fine for us since every URL loads the same way.

(If it helps I might publish it separately and we can depend on it 🤣)


src/lib/utils/router.js
```diff
@@ -0,0 +1,154 @@
...
+    // nb. this avoids collision with top menubars etc
+    target.scrollIntoView({block: "center"});
+  } else {
+    document.documentElement.scrollTop = +target || 0;
+  }
+}
+
+/**
+ * The caller wants to change the active URL. Let them, without triggering any
+ * few loads.
```
robdodson:
```suggestion
 * new loads.
```

PR #1729, opened 2019-10-22
Add CDS promo module.

robdodson:

- Add a promotional CDS module to the home page. Also has a commented out video player that we'll swap in during the event.


src/site/content/en/images/modules/cds19-lockup-stack.svg
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0" encoding="utf-8"?>
```
samthor: Run this through svgomg?

Issue #1375, opened 2019-08-19
Migrate LH data.


robdodson:
**Is your feature request related to a problem? Please describe.**
As part of the devsite migration, we're switching over to storing data in Firestore. We need to make sure that when we switch over, users don't lose their historical data.

**Describe the solution you'd like**
It seems like the easiest solution will be to let anyone look at the historical data for a URL. I think this may also solve the problem of people "losing" their URL data when they switch URLs.

Next steps are to remove the timestamp requirement in lighthouse-keeper.

robdodson:
Did some digging and here's what I've come up with so far.

In `lighthouse-service.js` we have two functions:
- [runLighthouse](https://github.com/GoogleChrome/web.dev/blob/93694df6ec4874a85975127ce4a1856066e76bf5/src/lib/lighthouse-service.js#L23) which gets called when the user audits a URL for the first time. If the user is not signed in, we call this function.
- [fetchReports](https://github.com/GoogleChrome/web.dev/blob/93694df6ec4874a85975127ce4a1856066e76bf5/src/lib/lighthouse-service.js#L54) which gets called when we're trying to gather historic data for a url. If the user is signed in, we call this function.

In lighthouse-keeper, when we hit the route for fetchReports, [it will check if there is a timestamp](https://github.com/GoogleChrome/lighthouse-keeper/blob/e275bcd3bcabacb9de90e37e59c1b7ff53bcdedf/server.mjs#L250-L279). If you comment out the `sinceDate` stuff then it will return the last 10 reports (which is its max limit).

In the `UrlChooserContainer` if you call `requestFetchReports` instead of `requestRunLighthouse` then it will give you the historic data for any url that's in the system, even if you're signed out. However it doesn't go through proper state changes (the field doesn't disable and it doesn't show an animating progress bar).

One issue I found is if you were previously signed in and had a URL associated with your account (`foo.com`), then you sign out, audit a different url (`bar.com`), then sign in again, it will replace `bar.com` in the text field with `foo.com` and fetch the results for `foo.com`.

samthor:
I think `requestFetchReports` is intended to be more of, you're opening the site anew, and want to quickly get historic data. So we don't disable the field since maybe you're coming today to audit some _different_ site.

The foo.com vs bar.com—we could prioritize any local non-signed in URL.

samthor:
I think we're done now—historical data will show up for any URL, but users' specific _preference_ won't, right?

robdodson:
correct.

PR #1735, opened 2019-10-23
retain previous URL on signin

samthor:
Related to #1375.

* Retains the active URL if you'd run Lighthouse on it before signing in. This does *not* write that to your profile, however (that wouldn't be _too_ hard, but perhaps a touch racey).

* Clears any previous Lighthouse result on signout. This won't stop an in-flight run (as that's indicated via `activeLighthouseUrl`).

The changes to `fb.js` are a lot of comments, but I think it's definitely worth it.

### DIFF
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -1,7 +1,21 @@
 import {store} from "./store";
 import {saveUserUrl} from "./fb";
 import {runLighthouse, fetchReports} from "./lighthouse-service";
-import "wicg-inert";
+
+export const clearSignedInState = store.action(() => {
+  const {isSignedIn} = store.getState();
+  if (isSignedIn) {
+    return {
+      userUrlSeen: null,
+      userUrl: null,
+      checkingSignedInState: false,
+      isSignedIn: false,
+      user: null,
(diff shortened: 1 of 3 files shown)


robdodson:
Thanks for putting this together!

> Retains the active URL if you'd run Lighthouse on it before signing in. This does not write that to your profile, however (that wouldn't be too hard, but perhaps a touch racey).

I think we do need to write this to their profile because the experience is a tad confusing without it. Near the sign in button we state `Sign in to track your progress over time.` but if you sign in and then refresh the page, it clears out the URL. I think the alternative would be that the user would have to know to re-audit the URL they just audited after they sign in, in order to store it.

samthor:
I think asking the users to "audit again" is odd, so I probably agree, we should do something else.

1) Save to profile automatically on signin
2) add "Save" hyperlink to the right of the URL

WDYT?

robdodson:
> Save to profile automatically on signin

I think we should just go with this. It feels like the expected flow.

samthor:
Cool, this is done, PTAL. Just some tweaks to `fb.js`.

PR #1741, opened 2019-10-24
Add imgix to hero and post cards

robdodson:

- Use imgix to srcset hero images and post card images when we're in a prod environment.
- `format=auto` will serve webp images if the browser supports it.
- `fit=max` means imgix won't try to scale up an image if the source is smaller than the requested dimensions (IOW, it won't stretch a 1600px image to be 3200px).

### DIFF
--- a/src/site/_includes/components/Hero.js
+++ b/src/site/_includes/components/Hero.js
@@ -15,16 +15,24 @@
  */
 
 const {html} = require("common-tags");
-const path = require("path");
 const stripLanguage = require("../../_filters/strip-language");
+const getImagePath = require("../../_utils/get-image-path");
+const getSrcsetRange = require("../../_utils/get-srcset-range");
 
 /* eslint-disable max-len */
 module.exports = ({page, hero, alt, heroPosition, heroFit = "cover"}) => {
+  const imagePath = getImagePath(hero, stripLanguage(page.url));
+  const srcsetRange = getSrcsetRange();
+
+  // prettier-ignore
   return html`
     <img
-      class="w-hero w-hero--${heroFit}
-      ${heroPosition ? `w-hero--${heroPosition}` : ""}"
-      src="${stripLanguage(path.join(page.url, hero))}"
+      class="w-hero w-hero--${heroFit} ${heroPosition ? `w-hero--${heroPosition}` : ""}"
+      sizes="100vw"
+      srcset="${srcsetRange.map((width) => html`
+        ${imagePath}?auto=format&fit=max&w=${width} ${width}w,
+      `)}"
+      src="${imagePath}"
       alt="${alt}"
     />
   `;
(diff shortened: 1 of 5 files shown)


src/site/_includes/components/Hero.js
```diff
@@ -15,16 +15,24 @@
...
 const {html} = require("common-tags");
-const path = require("path");
 const stripLanguage = require("../../_filters/strip-language");
+const getImagePath = require("../../_utils/get-image-path");
+const getSrcsetRange = require("../../_utils/get-srcset-range");
 
 /* eslint-disable max-len */
 module.exports = ({page, hero, alt, heroPosition, heroFit = "cover"}) => {
+  const imagePath = getImagePath(hero, stripLanguage(page.url));
+  const srcsetRange = getSrcsetRange();
```
samthor: Is this saying, use all srcset sizes because the hero image should be page-size?

If the source image (that's in our codebase) is smaller than a srcset value, it's going to be included for all sizes, even where imgix will return literally the same image?
(I wonder if our build step could look at the max width and limit srcset to that)

robdodson: > Is this saying, use all srcset sizes because the hero image should be page-size?

yep

> If the source image (that's in our codebase) is smaller than a srcset value, it's going to be included for all sizes, even where imgix will return literally the same image?

I'm not sure I understand the question. Just confirming that we're on the same page:

For hero images we used to ask everyone to upload a 1600px width hero, but we've recently updated the handbook to ask for a 3200px version. If I request a 3200px image from imgix, but the source is only 1600px, then it will send back the 1600px version (it won't force it to scale up). If I request an 800px image, imgix will scale the image down to 800px.

> (I wonder if our build step could look at the max width and limit srcset to that)

Yeah there are probably more clever ways to do this. Or more precise ways to do it (i.e. just hand code the srcset sizes you want depending on the component).

imgix also has a library, [imgix.js](https://www.imgix.com/imgix-js) that tries to figure this out for you at runtime but I wasn't a huge fan of including it as a client-side dep. It'd be interesting to poke around at the source to see if we could turn it into an eleventy plugin.

samthor: My point is if my screen starts at 3200px wide, I'll fetch the 3200px definition.

If I then resize my screen (say rotate a tablet), I might fetch the 1600px definition. This could be _the same file_.

Could this Eleventy component run something to work out the maximum size of the image it's including (I don't quite appreciate when it runs in the build pipeline) and not include srcset definitions for > the image size?

robdodson: > Could this Eleventy component run something to work out the maximum size of the image it's including (I don't quite appreciate when it runs in the build pipeline) and not include srcset definitions for > the image size?

Possibly. My main concern would be having to read in the image data for every image during a build and having that slow things down.

PR #1748, opened 2019-10-26
Add shortcode for page meta tags and hide 404/offline pages from Search

### LINKED ISSUE #1713: Add meta og tags for all pages.
We should test share our landing page URLs and blog posts and verify that they generate the correct thumbnails and descriptions.

We'll likely need to add meta og tags to all of these pages. For blog posts it would be nice if we could use the hero image in the meta og tag. We have imgix set up for web.dev already so we could use an imgix url to make sure the thumbnail is sized correctly.

robdodson:
Fixes #1713

- Add a new `Meta` shortcode to handle all the fiddly logic associated with meta tags and fallbacks.
- Hide /404/ and /offline/ from search indexing.
- Add missing description to blog 😬

### DIFF
--- a/src/site/_includes/components/Meta.js
+++ b/src/site/_includes/components/Meta.js
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


src/site/_includes/components/Meta.js
```diff
@@ -0,0 +1,116 @@
...
+
+  function renderFacebookMeta() {
+    const meta = getMetaByPlatform("facebook");
+    // nb. This will mark pages like /404/ and /offline/ as "article" but that's
+    // probably fine as those shouldn't show up in search results anyway.
+    const type = pageUrl === "/" ? "website" : "article";
+    // Filter out tags that we don't want to show up in Search.
+    // These tags are only used internally to determine which layout a page
+    // should use.
+    let tags = pageData.tags || [];
```
samthor: I think this might be clearer if you start off with

```js
let tags = (type === "article" ? pageData.tags : null) || [];
```

(or similar). The block below can just be the mapping and I can see more clearly that tags are not allowed for the top-level.

PR #1751, opened 2019-10-28
V2+cache more

samthor:
Most of #1370. 

* Caches misc images and assets on install
* The build script must run after other scripts (in production) in order to pick up asset hashes
* Stores pages in Workbox with ".../index.html" trailer to pick up Workbox's default support for that (as "/" will load "/index.html", but **not** the other way around)
* Caches images (e.g., author profiles) at runtime
  * .... doesn't cache hero images (since they're assets under individual articles)—although I thought this moves to imgix in #1741

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -1,3 +1,15 @@
+const normalizeIndexCacheKeyPlugin = {
+  cacheKeyWillBeUsed({request, mode}) {
+    // Take advantage of Workbox's built-in handling of .../index.html routes and ensure that its
+    // cache keys always include it. (e.g., requests for foo/ will load foo/index.html: but
+    // foo/index.html will not load foo/).
+    if (request.url.endsWith("/")) {
+      return request.url + "index.html";
+    }
+    return request;
+  },
+};
+
 import manifest from "cache-manifest";
 
 importScripts(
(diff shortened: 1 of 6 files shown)


src/lib/sw.js
```diff
@@ -1,31 +1,86 @@
...
+  event.respondWith(p);
+});
+
+workbox.routing.setCatchHandler(async ({event}) => {
   if (event.request.destination === "document") {
     // TODO(samthor): Annotate this page so the client knows it's being displayed because the user
     // is offline, and force it to rerequest when navigator.onLine is true.
-    return caches.match("/offline/");
+    // We have to ignoreSearch because Workbox adds __WB_REVISION__ to all its cached contents.
+    return caches.match("/offline/index.html", {ignoreSearch: true});
```
jeffposnick: You could use `workbox.precaching.getCacheKeyForURL("/offline/index.html")` as an alternative to setting `ignoreSearch`. (That has the advantage of always returning the version of `index.html` corresponding to the currently active service worker, which admittedly isn't likely to be a big deal for something like an offline page.)

samthor: I was going to tell you to write this in the docs then I realised that it's literally right there. 🤦‍♂ Thanks.

src/lib/sw.js
```diff
@@ -1,31 +1,88 @@
...
+  new RegExp("/images/.*"),
   new workbox.strategies.StaleWhileRevalidate(),
 );
 
-workbox.routing.setCatchHandler(({event}) => {
+/**
+ * Untrailed requests are network-only, but with a fallback to redirecting to the same page with a
+ * trailing slash.
+ */
+self.addEventListener("fetch", (event) => {
```
robdodson: Does this mean if we leave a `/` off of a link it will try to fetch from the network first?

jeffposnick: @samthor This part is a bit buried in the docs, but would https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.precaching.html#~urlManipulation help?

It's designed to let you use the built-in `workbox-precaching` `fetch` handler while providing a custom list of alternative cache keys to check.

Issue #1295, opened 2019-08-08
content: migrate "Experimenting with Periodic Background Sync"


jpmedley:
**Author**: @petele 
**Reviewer**: @jpmedley 

**Target publish date:** 2019-10-16
- [ ] Check this box if this is a hard deadline.

- [ ] Check this box if you'd like your reviewer to fix grammar and punctuation issues for you.

**Process**

Check out the [web.dev writing process](https://web.dev/handbook/quick-start/).

When you're ready, make a copy of [this template](https://docs.google.com/document/d/1lgaNIEnXZf-RB8_p9RK22QEgpXJqnu77pLWVWVy4nuw/edit)
and fill in the proposal and outline sections. When you're finished, drop a link
to the doc in the 1st step below and check the box. A member of the web.dev team
will assign this ticket to themselves and work as your reviewer to help you
complete the remaining steps.

- [x] 1. I've done my proposal https://docs.google.com/document/d/1CNrSJxLcoUTpUojLPiiJYvR1SgGC_YVjpfBfhrVn-gM/edit#heading=h.s6q03wgfh878
- [x] 2. Proposal approved — _reviewer checks this_
- [ ] 3. I've added a draft to my proposal doc
- [ ] 4. Draft approved — _reviewer checks this_
- [x] 5. I've submitted a pull request #1835
- [ ] 6. Pull request approved — _reviewer checks this_

jpmedley:
The WebFu version [has been merged](https://github.com/google/WebFundamentals/pull/7993). We will need to rewrite before moving to web.dev.

jpmedley:
There is a version of this on [/Updates](https://developers.google.com/web/updates/2019/08/periodic-background-sync) that will need considerable rework before moving to web.dev.

jeffposnick:
Hey @jpmedley & co. (@petele)—I see that there's a deadline of Oct. 9 associated with this, and just wanted to check in to see what's specifically needed from me.

I know that Joe would like to see a rewrite of what was initially published on WebFu.

If I'm going to perform the rewrite, I'll need some additional guidance as to what needs to be cut. Or was Joe planning on taking a first stab at the rewrite?

jpmedley:
@jeffposnick I've been traveling. Sorry about the delay. 

Some of the dates on these things are more aspirational that not.

[...]

PR #1752, opened 2019-10-28
Migrate Periodic Background Sync article

tomayac:
Fixes https://github.com/GoogleChrome/web.dev/issues/1295.

- Migrate the Web Fundamentals article https://developers.google.com/web/updates/2019/08/periodic-background-sync to web.dev.

CC: @petele @jpmedley @jeffposnick @carolrivero


jeffposnick:
In as far as this looks to be a direct copy of the text currently on d.g.c, I'm 👍

tomayac:
This is indeed simply the current text repackaged in web.dev markup. @jpmedley might want to make some editorial changes?!

kaycebasques:
Did @jpmedley review the original article that was published on /web?

kaycebasques:
In the related issue, @jpmedley said that he wanted to rewrite this article. Is that still the plan?

https://github.com/GoogleChrome/web.dev/issues/1295#issuecomment-538485651

tomayac:
If planned, the review (or rewrite) can happen in this pull request then.

kaycebasques:
I think we're going to use #1835 instead. We can always re-open this if I'm mistaken. See https://github.com/GoogleChrome/web.dev/issues/1295#issuecomment-550091158 for context.

PR #1756, opened 2019-10-28
Add sitemap.xml and fix errors.

robdodson:

- Adds a sitemap.xml
- Fixes pages without dates
- Excludes pages which shouldn't be in the sitemap


robdodson:
A note about the `htmlDateString` filter that I added...

Even though we put the dates in our YAML as `2019-10-31`, when eleventy sees a date in YAML it converts it to a date object. So the filter just converts it back to `2019-10-31`.

PR #1758, opened 2019-10-29
Add custom dimensions to analytics.

robdodson:

- Add custom dimensions to analytics. Our property already has dimensions specified by devsite but we probably won't use those immediately or possibly ever:

image

- Revs the tracking version to 3 (it's currently at '2' for the live version of web.dev).

### DIFF
--- a/src/site/_data/site.js
+++ b/src/site/_data/site.js
@@ -34,6 +34,9 @@ module.exports = {
   analytics: {
     ids: {
       prod: "UA-126406676-2",
+      // TODO (robdodson): These properties exist in GA but we don't use them.
+      // Adding a note to inject these into pages when we create a fancier
+      // staging environment.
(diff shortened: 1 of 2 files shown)


src/site/_data/site.js
```diff
@@ -31,4 +31,16 @@ module.exports = {
   // of the site. Otherwise all image paths are local. This means you can
   // develop locally without having to mess with the CDN at all.
   imageCdn: "https://webdev.imgix.net",
+  analytics: {
+    ids: {
+      prod: "UA-126406676-2",
+      staging: "UA-126406676-3",
```
samthor: We don't use non-prod at all right now? That's intentional?

robdodson: Hm good point. I guess I can toggle which analytics to use based on the ELEVENTY_ENV

PR #1760, opened 2019-10-29
V2+edge misc fixes

samthor:
This fixes some issues for older browsers.

* change `path` => `root` for the Web Components loader (oops)
* set `overflow-x: hidden` because Edge was allowing scrolling for some reason ¯‍\‍_‍(‍ツ‍)‍_‍/‍¯
* use `<body class="unresolved">` to work around lack of support for `:defined`—this isn't actually perfect because we lazy-load element code
  * ... so also remove `unresolved` on `<web-side-nav>` on its own until after its first render (prevents screenshot below)

Screen Shot 2019-10-29 at 13 45 58

I can't really test this without deploying, so this is done on my Mac first.

### DIFF
--- a/src/lib/components/SideNav/_styles.scss
+++ b/src/lib/components/SideNav/_styles.scss
@@ -24,6 +24,7 @@ body.web-side-nav--expanded {
   overflow: hidden;
 }
 
+web-side-nav:not(:defined),
 body.unresolved web-side-nav {
   display: none;
 }
@@ -41,6 +42,7 @@ web-side-nav {
 
 // Prevent FOUC by hiding children until after firstUpdated(), as children are
 // reparented within .web-side-nav__container before use
+// This isn't ":not(:defined)" because firstUpdated occurs after a frame.
 web-side-nav.unresolved * {
   display: none !important;
 }
--- a/src/lib/components/SideNav/_styles.scss
+++ b/src/lib/components/SideNav/_styles.scss
@@ -24,7 +24,8 @@ body.web-side-nav--expanded {
   overflow: hidden;
 }
 
-web-side-nav:not(:defined) {
+// Ideally this would be ":undefined", but that's not supported in old browsers.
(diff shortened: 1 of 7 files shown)


src/lib/components/SideNav/_styles.scss
```diff
@@ -24,7 +24,7 @@ body.web-side-nav--expanded {
   overflow: hidden;
 }
 
-web-side-nav:not(:defined) {
+body.unresolved web-side-nav {
```
robdodson: Should we do something like:

```
web-side-nav:not(:defined),
body.unresolved web-side-nav {
```

I imagine when Edgium ships we'll want to remove the web components loader stuff but we'll still want the `:not(:defined)` styles.

src/lib/components/SideNav/_styles.scss
```diff
@@ -39,6 +39,12 @@ web-side-nav {
   z-index: 300;
 }
 
+// Prevent FOUC by hiding children until after firstUpdated(), as children are
+// reparented within .web-side-nav__container before use
+web-side-nav.unresolved * {
```
robdodson: Could we add `:not(:defined)` here as well?

robdodson:
Thanks for adding these fixes!

Left a couple comments regarding `:not(:defined)` but feel free to merge whenever.

src/lib/components/SideNav/_styles.scss
samthor: No, because there's a change after a frame (due to reparenting) and so the flash still occurs. I think maybe Chrome is showing me a frame before lit renders.

samthor: I've added a comment to this effect.

src/lib/components/SideNav/_styles.scss
samthor: Ack

PR #1761, opened 2019-10-29
Use compression on dev server.

robdodson:
Related to #1757

- Add compression to dev server. This will benefit Lighthouse CI since it uses our dev server to audit the site.

### DIFF
--- a/server.js
+++ b/server.js
@@ -16,6 +16,7 @@
 
 const isProd = Boolean(process.env.GAE_APPLICATION);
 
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -51,6 +52,12 @@ const handlers = [
 ];
 
 const app = express();
+// For dev we'll do our own compression. This ensures things like Lighthouse CI
```
samthor: Oh good. I would hate to do gzip in JS.

PR #1762, opened 2019-10-29
More unsupported browser fixes

samthor:
The `WebComponents.root` var wasn't being passed with a leading slash. But I realised we can restructure our configs to be a bit simpler too.

This also adds a notice informing users that Measure might be unsupported on old browsers.


samthor:
There's some other JS fixes for Edge.

src/site/content/en/measure/index.njk
```diff
@@ -22,6 +22,9 @@ to remove them as part of the v1 migration. #}
         </span>
       </p>
       <web-signin-button></web-signin-button>
+      <p class="w-page-header__copy w-unsupported-notice">
+        Measure isn't available on your browser. Try a supported browser.
```
robdodson: I can already feel the grumpy tweets 😁

Maybe tweak the language to say. "Sorry, the measure tool does not work with legacy browsers. Please try again using the latest Firefox, Safari, or Chrome."

src/styles/components/_page-header.scss
```diff
@@ -41,3 +41,17 @@
...
+body:not(.unresolved) .w-unsupported-notice {
+  display: none !important;
+}
+
+@keyframes unsupported-appear {
+  from { visibility: hidden; }
+  to   { visibility: hidden; }
+}
+
+.w-unsupported-notice {
```
robdodson: Is it worth fading this out or should it just be persistent?

samthor: I think it's fine. It only shows up after 5 seconds anyway and we don't expect many users to see it.

src/site/content/en/measure/index.njk
samthor: Sounds good.

PR #1789, opened 2019-10-30
Clean up layout template. Add favicon. Remove outdated hash file.

robdodson:

- Add missing favicon.
- Remove entrypoint.hashmanifest.json. This is an old file I previously experimented with but am no longer using.


samthor:
Do we also need a `favicon.ico` (or did we used to have one)?

robdodson:
> Do we also need a favicon.ico (or did we used to have one)?

I don't think so. This is a copy of what is on the existing site.

PR #1800, opened 2019-10-31
Fixes 1772 - removes extra styles in LighthouseScores

devnook:
Fixes 1772 - removes extra styles in LighthouseScores


robdodson:
Thank you @devnook !!!

PR #1805, opened 2019-10-31
Removing bounce stat from defer non-critical CSS article.

demianrenzulli:
Removing reference to [this article](https://www.thinkwithgoogle.com/intl/en-154/insights-inspiration/research-data/need-mobile-speed-how-mobile-latency-impacts-publisher-revenue/), which contains an incorrect stat.

cc // @robdodson


kaycebasques:
Thanks @demianrenzulli, I already pushed a commit to fix this

PR #1812, opened 2019-10-31
travis: adjust git depth

paulirish:
the default is 50.
depth of only 1 breaks a ancestor hash lookup we do in lhci

this shouldn't matter much for CI time either. (especially now that lhci added ~3min to the total times 😉)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
 sudo: false
 dist: trusty
 git:
-  depth: 1
   submodules: false
 cache:
   directories:


robdodson:
Just curious, do you need it to go to 50 or would a depth of like 10 suffice? Mostly wondering if it's worth trimming it to just what lhci needs and not more

paulirish:
50 is best because its reasonable for someone to have >10 commits in their PR's branch.

also, the difference in clone time between these is very small. :)

image

robdodson:
ok sounds good.

PR #1814, opened 2019-11-01
send pageview for new pages

### LINKED ISSUE #1795: Analytics don't fire for initial page view.
**Describe the bug**
When a user first lands on the site, analytics don't fire a pageview.

**To Reproduce**
1. Go to any page on web.dev

**Expected behavior**
There should be a pageview ping to GA.

samthor:
Possibly fixes #1795.

Simplifies the GA behavior.

* Moves the Analytics code above importing our code (I think this would never be a problem because modules are deferred, but it just seems less ambiguous about whether `ga` is available)
* Sets the new page as we change URL and explicitly call `ga("set", "page", ...)`, which I think is needed (cc @philipwalton )

### DIFF
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -47,10 +47,3 @@ document.addEventListener("click", (e) => {
 
   trackEvent(data);
 });
-
-/**
- * Configure tracking events for page changes. Emitted in `router.js`.
- */
-document.addEventListener("pageview", () => {
-  ga("send", "pageview");
-});
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -90,7 +90,6 @@ function forceFocus(el) {
  * @return {!Promise<void>}
  */
 export async function swapContent(url, isFirstRun) {
-  document.dispatchEvent(new CustomEvent("pageview", {detail: url}));
   const entrypointPromise = loadEntrypoint(url);
 
   // If we disagree with the URL we're loaded at, then replace it inline
@@ -129,6 +128,9 @@ export async function swapContent(url, isFirstRun) {
       currentUrl: url,
     });
   }
+
+  ga("set", "page", window.location.pathname);
+  ga("send", "pageview");
 
   // Remove the current #content element
   main.querySelector("#content").remove();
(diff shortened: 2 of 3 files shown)


src/lib/analytics.js
```diff
@@ -47,10 +47,3 @@ document.addEventListener("click", (e) => {
 
   trackEvent(data);
 });
-
-/**
- * Configure tracking events for page changes. Emitted in `router.js`.
- */
-document.addEventListener("pageview", () => {
-  ga("send", "pageview");
```
philipwalton: Doesn't removing this mean you won't get pageviews on SPA views?

src/lib/loader.js
```diff
@@ -130,6 +129,9 @@ export async function swapContent(url, isFirstRun) {
     });
   }
 
+  ga("set", "page", window.location.pathname);
```
philipwalton: > Sets the new page as we change URL and explicitly call ga("set", "page", ...), which I think is needed (cc @philipwalton )

Not necessarily needed for the first pageview, but doing this helps get rid of unwanted query params in your analytics (which is generally what you want).

However, you're going to want to *also* do this when you call `pushState()` in your SPA router. This guide should have more details: https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications.

src/lib/analytics.js
samthor: No, if you look at the PR, we used to fire a custom event which triggers this. This is now running in "loader".

src/lib/loader.js
samthor: This is where we load new content (i.e., related to calling `pushState`). `swapContent` is the callback into web.dev loading code from our router.

The question I had for you was more about calling `ga("set", "page", ...)`, which we _were not_ doing. I presume just triggering a pageview doesn't implicitly update the page, if GA sees a new location. Is that correct?

samthor:
Phil, appreciate the comment but I think you need to reread what this PR does :)

We send our initial hit inline in our HTML (see the changed .njk file - it's just adjacent). And I'm explicitly adding a pageview event and page set where we do our SPA content swap.

src/lib/loader.js
philipwalton: Gah, sorry. For some reason I had it in my head that this was the template code (for the initial page load). Now I see that it's not.

> The question I had for you was more about calling ga("set", "page", ...), which we were not doing. I presume just triggering a pageview doesn't implicitly update the page, if GA sees a new location. Is that correct?

Yes, that's correct. You need to set the new `page` value, which you're doing. I took a second look at the staged PR and it looks correct to me.

You still might want to call `ga("set", "page", window.location.pathname);` in the template code before the first pageview to remove any query params from the URL that gets sent, but that's up to you.

philipwalton:
Ignore me, carry on! :)

PR #1822, opened 2019-11-01
Update codeowners.

robdodson:
Get more granular with codeowner permissions. Lock down specific file types (.js, .njk, etc).


robdodson:
@kaycebasques @samthor fyi

PR #1826, opened 2019-11-03
Cookies without a `SameSite` default fix

agaripian:
Cookies without a `SameSite` attribute will be treated as `SameSite=Lax`. According to the docs https://www.chromium.org/updates/same-site and https://tools.ietf.org/html/draft-west-cookie-incrementalism-00 the default without `SameSite` attribute will act at `SameSite=Lax` NOT `SameSite=None`


agaripian:
@googlebot I signed it!

rowan-m:
Wow, yes - good catch. `SameSite=Lax` is the default.

PR #1831, opened 2019-11-04
Fix code in headers.

robdodson:
@mfriesenhahn does this look good to you? I think the recent change may have bumped the font size down for code in headers. Example: https://web.dev/duplicate-id/

### DIFF
--- a/src/styles/generic/_code.scss
+++ b/src/styles/generic/_code.scss
@@ -76,8 +76,10 @@ h3 code,
 h4 code,
 h5 code,
 h6 code {
+  background: transparent;
   border: 0;
   color: inherit;
+  font-size: inherit;
   margin: 0;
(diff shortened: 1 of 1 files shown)


mfriesenhahn:
Woops! Completely forgot to check headers. `font-size` looks good. Background looks funky to me though. I think the differences between monospace and sans-serif fonts at titling sizes are sufficient without it?

robdodson:
> Background looks funky to me though. I think the differences between monospace and sans-serif fonts at titling sizes are sufficient without it?

This PR should set the background to be transparent. Without that change the code section has a light grey colored background. Maybe i'm misunderstanding what you're saying?

mfriesenhahn:
Never mind! Moving too quickly.

PR #1842, opened 2019-11-06
Correct margins in buttons on mobile. Fixes #890

### LINKED ISSUE #890: Adjacent buttons should wrap properly on mobile.
https://web-dev-staging.appspot.com/sandbox/#buttons

image

devnook:
Fixes #890 

Let's use top&right margins instead of left, so that the margin is not obtrusive on small screen sizes. Also we should allow wrapping so that all text in button can be visible on small screen, wdyt?


robdodson:
>Also we should allow wrapping so that all text in button can be visible on small screen, wdyt?

I think we're supposed to avoid wrapping according to the material spec
image

Did you see examples where the text was so long it needed to wrap?

src/styles/generic/_buttons.scss
```diff
@@ -7,9 +7,6 @@
 // =============================================================================
 
 button {
+  margin: 16px 16px 0 0;
```
robdodson: I think this may add unnecessary top margin to all of our buttons, the measure page is a good example:

image

For what it's worth, most folks don't use buttons in their article. So I'm wondering if we should just ask authors to put their buttons into a container that adds the necessary margin at smaller screen sizes? @argyleink any suggestions?

devnook: I think the only sane way to tackle this is to let the parent container decide margins of the buttons inside.

src/styles/generic/_buttons.scss
```diff
@@ -7,9 +7,6 @@
 // =============================================================================
 
 button {
+  margin: 0 16px 0 0;
```
robdodson: I think we may want to remove this, otherwise buttons which are meant to be centered on the screen get shifted to the left.

Here's a preview with the PR enabled, notice the measure button:
image

Here's the current button on prod:
image

robdodson:
I'm ok with the idea of the parent container needing to define the button spacing for wrapping buttons.

PR #1846, opened 2019-11-07
disallow www.web.dev, redirect to naked domain

### LINKED ISSUE #1833: Redirect www.web.dev => web.dev
This would be a redirect done in "server.js".

Importantly, we can't just do a blind redirect. We should also serve "/sw.js" as a blank SW that replaces the previous one and forces clients to reload. If that file doesn't exist then clients who have "www.web.dev" installed will have a bad time.

samthor:
Fixes #1833. Serves a dummy SW that nukes any clients.

### DIFF
--- a/server.js
+++ b/server.js
@@ -44,7 +44,19 @@ const notFoundHandler = (req, res, next) => {
     .sendFile("404/index.html", options, (err) => err && next(err));
 };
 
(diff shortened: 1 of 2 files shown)


robdodson:
Does navigating to www.web.dev automatically request the sw.js even if we redirect to https://web.dev? IOW, can you explain a bit more why you added the nuke-sw part?

PR #1858, opened 2019-11-08
Update page description when we load new content.

robdodson:

- A quick fix to update page descriptions juuuuuuuust in case googlebot is running JS on our site and navigating it as if it's a SPA. Ultimately we should replace this with merging all of the meta tags from the new document, but I wanted to throw this in real quick as a bandaid.

### DIFF
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -140,10 +140,8 @@ export async function swapContent(url, isFirstRun) {
   document.title = page.title;
   // Update the page description
   const description = page.querySelector("meta[name=description]");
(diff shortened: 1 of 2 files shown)


src/lib/loader.js
```diff
@@ -138,6 +138,12 @@ export async function swapContent(url, isFirstRun) {
 
   // Update the page title
   document.title = page.title;
+  // Update the page description
+  const description = page.querySelector("meta[name=description]");
+  if (description && description.content) {
```
samthor: const updatedContent = description ? description.content : '';

PR #1860, opened 2019-11-08
don't eat hash

### LINKED ISSUE #1859: CMD/CTRL clicking a link to open it in a new tab drops the hash in the URL
**Describe the bug**
CMD/CTRL clicking a link to open it in a new tab drops the hash in the URL

**To Reproduce**
1. Go to any page with a link that contains a hash at the end of the URL
2. CMD/CTRL+CLICK the link to open it in a new tab
3. Notice that the URL in the new tab doesn't include the hash portion of the original link

**Expected behavior**
Hash URLs should be preserved

samthor:
Fixes #1859.

We were actually just eating it on _load_, nothing to do with nav specifically.

### DIFF
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -84,18 +84,19 @@ function forceFocus(el) {
 }
 
 /**
- * Swap the current page for a new one.
- * @param {string} url url of the page to swap.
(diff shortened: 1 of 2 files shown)


philipwalton:
Thanks! I just checked the staged version and everything works as expected.

PR #1863, opened 2019-11-09
Update share article urls and swap in new hero.

robdodson:

- Update web share article urls to be `web-share` and `web-share-target`
- Swaps in a new hero for `web-share`

@jpmedley @petele fyi

If y'all want to source different hero images I'm happy to swap em around but wanted to get something up that was decent so we could get the initial launch out the door.


jpmedley:
I encourage you all to go back to using the same image for both. A few months back, Rob mused out loud (if I'm remembering this correctly) that we might get rid of the hero image because they didn't communicate anything. I used the hero image to communicate that two articles are related to each other.

robdodson:
A better option might be to reach out to Harleen to see if we could get a  custom illustration done that conveys these two posts are related. When we used the same image, and the two cards were next to each other, it looked (to me at least) as if someone had made a mistake and included the same image for two posts. As an outside observer, it didn't come across as intentional.

PR #1856, opened 2019-11-08
Add all metrics docs

philipwalton:
Fixes #1569, fixes: #1570, fixes: #1571, fixes: #1572, fixes: #1573, fixes: #1574, fixes: #1855

This PR branches off of the work done in #1854 and adds all the remaining metrics docs and redirects.

It also fixes the issue in #1855 by ignoring URLs in `<a>` elements that end in `jpg`, `png`, `gif`, and `svg`. @samthor let me know if you'd prefer a different solution to handle this issue, but I need *some* solution as my new docs are linking to SVG and PNG files.

PR #1864, opened 2019-11-09
Revert TTI doc changes for consistency

philipwalton:
#1856 was merged when some changes for the TTI doc was in flux, so this PR reverts that change to be consistent with the other docs. We can revisit the change for all the docs in a subsequent PR.

### DIFF
--- a/src/site/content/en/metrics/tti/index.md
+++ b/src/site/content/en/metrics/tti/index.md
@@ -13,13 +13,15 @@ tags:
   - metrics
 ---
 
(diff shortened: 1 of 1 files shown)


kaycebasques:
Sorry, didn't realize work was still in-flight

philipwalton:
No worries! Michael had suggested making the intro sections not an <aside> (just inlining them), and we were experimenting with what that might look like. We can revisit in a later PR if you agree.

PR #1873, opened 2019-11-10
Web share article: Correct link to web share target article

tomayac:
The link should be https://web.dev/web-share-target/

CC: @petele @jpmedley


ryym:
Hi, 
It looks like the link from the [`web-share-target` article][web-share-target] is broken too.

https://github.com/GoogleChrome/web.dev/blob/2276b285b8960de6264c56c4463f7838ebcb467a/src/site/content/en/blog/web-share-target/index.md#L31-L34

The link should be <https://web.dev/web-share/>, isn't it?


[web-share-target]: https://web.dev/web-share-target/

tomayac:
Thanks, @ryym, fixed the other way round, too.

PR #1878, opened 2019-11-10
Adopt responsive images w/ [srcset] and image CDN

paulirish:
I noticed this while preparing my CDS talk. ;)

------------

Before:

```html
<img 
  src="evergreen-googlebot.png" 
  alt="Googlebot holding the Chrome logo" style="max-width: 400px;">
```

After:

```html
<img sizes="100vw" srcset="
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=240   240w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=480   480w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=768   768w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=1045 1045w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=1434 1434w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=1730 1730w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=1959 1959w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=2195 2195w,
    https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=2880 2880w,
    https://webdev.imgix.net/javascript-and-google-evergreen-googlebot.png?auto=format&amp;fit=max&amp;w=3200 3200w"
  src="https://webdev.imgix.net/javascript-and-google-search-io-2019/evergreen-googlebot.png"
  alt="Googlebot holding the Chrome logo"/>
```

PR #1880, opened 2019-11-11
Add lite-youtube-embed for ~422X faster-loading youtube embeds

paulirish:
Before and after:
image

Note: "422.1X" is not accurate at all.

(Builds on top of #1878)


paulirish:
sorry for the PR noise (again)

LHCI doesnt post github status on PR coming from forks, hence the doubling up im doing.

this PR redone in 1882

PR #1887, opened 2019-11-12
misc: update lhci to 0.3.x

patrickhulce:

- Moves the script to the safer, non-breaking change `0.3.x` tag
- Cleans up the script to take advantage of `startServerCommand`

~~WIP: looks like LHCI was already failing with the commiter timestamp change, so wait for that to be fixed before merging~~ should be fixed in 0.3.1 :)

### DIFF
--- a/tools/lhci/.lighthouserc
+++ b/tools/lhci/.lighthouserc
@@ -10,6 +10,7 @@
       ]
     },
     "upload": {
+      "token": "d96145f6-c3d2-4b22-9e53-6913cdd2df03",
       "serverBaseUrl": "https://lhci-canary.herokuapp.com/"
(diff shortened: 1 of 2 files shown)


robdodson:
@patrickhulce I added one comment regarding the token. Is this otherwise ready to merge?

patrickhulce:
Yep! I don't see your comment though, something I need to take action on?

robdodson:
I was asking if we should move the token to an environment variable so it's
not checked in to source

patrickhulce:
Oh, eh not really. It's no longer a secret when you want to allow external contributions to be uploaded, so I don't see a strong reason to.

PR #1893, opened 2019-11-12
Change "What is a good TTI" to "What is a good TBT"

lencioni:
This seems like a more appropriate heading for this section.

### DIFF
--- a/src/site/content/en/metrics/tbt/index.md
+++ b/src/site/content/en/metrics/tbt/index.md
@@ -123,7 +123,7 @@ TBT](/lighthouse-total-blocking-time) for usage details.
   measure [First Input Delay (FID)](/fid/).
 {% endAside %}
 
-## What is a good TTI?
+## What is a good TBT?
 
 For details on how your page's TBT affects your Lighthouse performance score,
 see [How Lighthouse determines your TBT


kaycebasques:
Thanks for the fix @lencioni. Let me know if you're cool with signing the Contributor License Agreement otherwise I can put the fix in in a separate PR.

lencioni:
Yep, I can do that tomorrow

lencioni:
@googlebot I signed it!

robdodson:
Wanted to call out that this will break existing deep links to this heading. That may be fine, but I wanted to mention it  in  case it comes as a surprise

jpmedley:
Speaking of that, are there instructions somewhere for configuring a redirect?

robdodson:
I don't think so. They essentially follow the same rules as devsite. You'd add your redirects to [our _redirects.yaml file](https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/_redirects.yaml).

philipwalton:
I don't believe hash URLs work in redirects, so if we wanted to ensure existing hash URLs don't break, we'd probably have to add a wrapper <div id="what-is-a-good-tti"> to the heading.

IMO, I think it's fine to just merge as-is. Those links will still go to the right page, and developers can probably still find the content they're looking for. (We also don't know for sure that such links exist in the wild, so we might not be breaking anyone).

robdodson:
sgtm. For future reference, we do support `{: some-hash-name}` syntax if you want to manually give an id to a heading. That way you could manually use the old name to not break links. But I agree, in this case it's probably not a big deal.

PR #1899, opened 2019-11-16
Auto-size table if device is small

andreas0607:
The table will overflow and have hidden content if the device width is smaller than 512px. 
See here for an example: [https://web.dev/periodic-background-sync/#current-status](https://web.dev/periodic-background-sync/#current-status)

Couldn't find any issue related to this. I just noticed it on my phone earlier and it annoyed me. 

- Change style of table to have `min-width:auto` on sm-breakpoint
- 
-

### DIFF
--- a/src/styles/generic/_tables.scss
+++ b/src/styles/generic/_tables.scss
@@ -1,5 +1,5 @@
 @import '../settings/type';
-
+@import '../tools/breakpoints';
 // =============================================================================
 // TABLES OVERVIEW
 //
@@ -21,6 +21,9 @@ table {
   margin: 0;
   min-width: 512px;
   width: 100%;
+  @include bp(sm) {
+    min-width:auto;
+  }
 }
 
 tr {
--- a/src/styles/pages/_post.scss
+++ b/src/styles/pages/_post.scss
@@ -20,6 +20,9 @@
 
   @include bp(sm) {
     margin-bottom: 160px;
+    table{
+      min-width:auto;
+    }
   }
 }
 


andreas0607:
@googlebot I signed it!

robdodson:
I believe the tables should horizontally scroll on smaller displays. We could think about adding a drop shadow on the right  hand side  to  indicate that it scrolls?

andreas0607:
Hey! Thanks for your reply, I realized that this branch didn't really fix the issue, thats why I closed it. 

I still am neither able to scroll horizontally on tables on my phone, or in devtools displaying any mobile device. I believe it's because the post-wrapper uses `overflow:visible` in a compination with body having `overflow-x:hidden` and table with `min-width: 512px`.  The min-width overrides the `width:100%` causing the table to overflow. I suspect that it might be the case for all blog-posts with inline-html in the markdownfiles. 
I made a new pull-request where this is fixed: [https://github.com/GoogleChrome/web.dev/pull/1916]( https://github.com/GoogleChrome/web.dev/pull/1916)

robdodson:
oh we have a helper class that these tables should be contained within. I think the author omitted that.

robdodson:
@jpmedley I think most of the capabilities posts use a markdown table wrapped in a table-wrapper div ([example](https://github.com/GoogleChrome/web.dev/blame/2930e971a639b16b81c285f9f665f4eba1cbcb78/src/site/content/en/blog/badging-api/index.md#L66-L76))

Would you mind updating the sms-receiver and periodic-background-sync posts to use the same format? The main thing is adding that wrapper div to make sure they overflow correctly. I've opened #1917 to see if we can create a plugin to do this automatically in the future.

PR #1905, opened 2019-11-18
Update handbook components post

mfriesenhahn:

- Remove `{% verbatim %}` tags. (No longer needed now that site has been migrated from dev site.)
- Add missing code samples for all components.
- Call out all variants of the Instruction component.
- Call out the two variants of the Banner component.
- Add explanations for components that shouldn't need to be used by post authors.
- Add note about `codelabs` YAML field to `/handbook/markup-post-codelab`.


robdodson:
I think you could add that blockquote now supports `>`. Although, I think they need to use the html version if they want to include a `cite`.

robdodson:
LGTM after adding the blockquote item.

robdodson:
oh wait, ignore my previous comments. I remember [the change](https://github.com/GoogleChrome/web.dev/commit/2c5d5ca6da886b24966fc7e0ce7315b3b169a4aa) was to not require folks to use the w-blockquote class. But they'll still need to use html to include the cite. Definitely seems like another thing we could turn into a shortcode at some point.

mfriesenhahn:
> oh wait, ignore my previous comments. I remember [the change](https://github.com/GoogleChrome/web.dev/commit/2c5d5ca6da886b24966fc7e0ce7315b3b169a4aa) was to not require folks to use the w-blockquote class. But they'll still need to use html to include the cite. Definitely seems like another thing we could turn into a shortcode at some point.

I can do that if you like. Before I start, do y'all agree that we should always include the source? If so, I can make that a required argument. @robdodson, @kaycebasques

robdodson:
thinking about this a bit more.. I think the citation should be optional.

mfriesenhahn:
> thinking about this a bit more.. I think the citation should be optional.

Mind sharing your thinking? I started to make it optional but then thought: how else do we know whom the quote is coming from? And you might say, "Well, people can just provide an introductory sentence or something." But I'd argue that for such cases people should really use an inline quote rather than a block quote, which based on the web.dev styling is really functioning as a pull quote. And pull quotes are supposed to be more or less freestanding.

robdodson:
I guess the two examples that come to mind are if folks are pull quoting something said earlier in the article to give it additional emphasis, or if they're using the blockquote for a common saying, "Measure twice, cut once", something like that.

I don't have a strong opinion though

PR #1910, opened 2019-11-18
Don't use code character in titles and headings.

jpmedley:


robdodson:
Hey @jpmedley, I was wondering if you removed these because they were showing up in the URLs and the meta page title?

jpmedley:
This is purely typographical.

jpmedley:
The backticks themselves are showing up in the search results.

image

robdodson:
> The backticks themselves are showing up in the search results.

Yeah, I was wondering if we should strip them from indexing but still leave them when we render the html so it's clear we're talking about code?

kaycebasques:
@robdodson the least-blocked path seems to be:

* Remove the backticks until they're stripped from indexing
* Allow people to add them back once that feature is complete

In that case I will merge this PR

@jpmedley can you also add a line to https://web.dev/handbook/content-checklist/#mechanics mentioning that the title should not include backticks because they show up in search results

robdodson:
ok sgtm

jpmedley:
Where's the source?
 816-678-7195

kaycebasques:
@jpmedley the handbook source code is `src/site/content/en/handbook`

PR #1918, opened 2019-11-19
feat: add underline to links of codelab, learn and blog pages, fixes #1683

### LINKED ISSUE #1683: Add underlines to links in articles.
For accessibility we shouldn't be relying on color alone to convey that something is interactive. The original design for links in our content posts did not call for them to have underlines, but I think it might be good for us to add this affordance.

image

MichaelSolati:
Fixes #1683 

Adds underline to links inside of `.w-post-content` and `.web-codelab__instructions` (except for author links). Also cleaned up some of the styling for `a` (and similar stuff) using SCSS parent selectors.


src/styles/generic/_links.scss
```diff
@@ -16,22 +16,28 @@ a,
...
-:link:active,
-:visited:active {
-  outline: 1px solid $FOCUS_COLOR;
-  text-decoration: underline;
+.w-post-content,
+.web-codelab__instructions {
+  a,
+  :link,
+  :visited {
+    &:not(.w-author__link) {
```
robdodson: I bet we can remove this and allow these links to be underlined as well.

src/styles/generic/_links.scss
```diff
@@ -16,22 +16,28 @@ a,
...
   text-decoration: none;
-}
 
-a:hover,
-:link:hover,
-:visited:hover {
-  text-decoration: underline;
+  &:focus,
+  &:active,
+  &:hover {
```
robdodson: We may want to make it so hovering and pressing a link removes its underline so there's some signal to the user that it's working.

PR #1920, opened 2019-11-20
feat: add support for `subtopics` inside of site content `topics`, fixes #1137

### LINKED ISSUE #1137: Consider adding subheadings to collections
It might be nice to add subheadings to big topic areas. For example, in the frameworks sections (/react, etc) we have sections for "Fast". It might be nice to subdivide these into "Loading performance" and "Runtime performance".

image

Moving to internal tracking tool

MichaelSolati:
Fixes #1137.

Inside of a collection topics can support subtopics (embedded collections?)

Here's what it would look like _if_ we had subtopics in the Angular collection:

Screen Shot 2019-11-20 at 2 08 24 PM

No where currently uses subtopics, but effectively topic objects in the topics array of a post (like [`angular.11tydata.js`](https://github.com/GoogleChrome/web.dev/blob/4ec23419bbdac156ea5fb6410821db2306304bd4/src/site/content/en/angular/angular.11tydata.js#L15)) now support a `subtopics` field, which is just an array of objects of the below data structure:

```TypeScript
type SubTopic = {
     title: string;
     pathItems: string[];
}
```

### DIFF
--- a/src/site/_includes/components/ArticleNavigation.js
+++ b/src/site/_includes/components/ArticleNavigation.js
@@ -28,15 +28,13 @@ const md = require("markdown-it")();
  * @return {Array} An array of pathItem slugs.
  */
 function getPathItemsFromTopics(topics) {
-  const temp = topics.reduce((reduced, topic) => {
+  return topics.reduce((reduced, topic) => {
     const subPathItems = (topic.subTopics || []).reduce((accumulator, subTopic) => {
       return ([...accumulator, ...(subTopic.pathItems)]);
     }, []);
     topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
     return reduced.concat(topic.pathItems);
   }, []);
-
-  return temp;
 }
 
 /**
--- a/src/site/_includes/components/ArticleNavigation.js
+++ b/src/site/_includes/components/ArticleNavigation.js
@@ -28,10 +28,15 @@ const md = require("markdown-it")();
  * @return {Array} An array of pathItem slugs.
  */
 function getPathItemsFromTopics(topics) {
-  return topics.reduce(
-    (pathItems, topic) => pathItems.concat(topic.pathItems),
-    [],
-  );
+  const temp = topics.reduce((reduced, topic) => {
+    const subPathItems = (topic.subTopics || []).reduce((accumulator, subTopic) => {
+      return ([...accumulator, ...(subTopic.pathItems)]);
+    }, []);
+    topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
+    return reduced.concat(topic.pathItems);
+  }, []);
+
+  return temp;
 }
 
 /**
--- a/src/site/_includes/components/ArticleNavigation.js
+++ b/src/site/_includes/components/ArticleNavigation.js
@@ -29,8 +29,8 @@ const md = require("markdown-it")();
  */
 function getPathItemsFromTopics(topics) {
   return topics.reduce((reduced, topic) => {
-    const subPathItems = (topic.subTopics || []).reduce((accumulator, subTopic) => {
-      return ([...accumulator, ...(subTopic.pathItems)]);
+    const subPathItems = (topic.subtopics || []).reduce((accumulator, subtopic) => {
+      return ([...accumulator, ...(subtopic.pathItems)]);
     }, []);
     topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
     return reduced.concat(topic.pathItems);
--- a/src/site/_includes/components/ArticleNavigation.js
+++ b/src/site/_includes/components/ArticleNavigation.js
@@ -29,10 +29,13 @@ const md = require("markdown-it")();
  */
 function getPathItemsFromTopics(topics) {
   return topics.reduce((reduced, topic) => {
-    const subPathItems = (topic.subtopics || []).reduce((accumulator, subtopic) => {
-      return ([...accumulator, ...(subtopic.pathItems)]);
-    }, []);
-    topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
+    const subPathItems = (topic.subtopics || []).reduce(
+      (accumulator, subtopic) => {
+        return [...accumulator, ...subtopic.pathItems];
+      },
+      [],
+    );
+    topic.pathItems = [...(topic.pathItems || []), ...subPathItems];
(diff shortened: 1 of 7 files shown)


robdodson:
cc @petele, @jpmedley, @kaycebasques that we're working on this. I think the PWA collection needs this feature.

src/site/_includes/components/ArticleNavigation.js
```diff
@@ -28,10 +28,15 @@ const md = require("markdown-it")();
  * @return {Array} An array of pathItem slugs.
  */
 function getPathItemsFromTopics(topics) {
-  return topics.reduce(
-    (pathItems, topic) => pathItems.concat(topic.pathItems),
-    [],
-  );
+  const temp = topics.reduce((reduced, topic) => {
```
robdodson: maybe this var should be named `pathItems`?

robdodson:
In terms of design I think we're imagining it would look something like this:

image

src/site/_includes/components/PathCard.js
```diff
@@ -29,6 +29,16 @@ function getPostCount(learningPath) {
   // in path.njk. Ideally we should do this in the learningPath .11ty.js files
   // but eleventy hasn't parsed all of the collections when those files get
   // initialized so we can't look up posts by slug.
+
+  // Merge subTopic pathItems
+  learningPath.topics = (learningPath.topics).map((topic) => {
+    const subPathItems = (topic.subTopics || []).reduce((accumulator, subTopic) => {
+      return ([...accumulator, ...(subTopic.pathItems)]);
+    }, []);
+    topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
```
robdodson: My understanding is this will put subpath items before path items. I wonder if we want to reverse it so pathItems happen first. I think that would allow folks to include intro posts, similar to this:

## Topic: Fast
- An introduction to performance

### Subtopic: Measuring speed
- Something about measuring

### Subtopic: Optimizing speed
- Something about optimizing

src/site/_includes/components/ArticleNavigation.js
```diff
@@ -28,10 +28,15 @@ const md = require("markdown-it")();
...
+    const subPathItems = (topic.subTopics || []).reduce((accumulator, subTopic) => {
+      return ([...accumulator, ...(subTopic.pathItems)]);
+    }, []);
+    topic.pathItems = [...subPathItems, ...(topic.pathItems || [])];
```
robdodson: See comment below about putting path items ahead of subpath items.

src/site/_includes/components/ArticleNavigation.js
MichaelSolati: This just shouldn't have been there 😅

MichaelSolati:
Resolved the feedback, here's what it would look like _if_ we had subtopics in the Angular collection:

Screen Shot 2019-11-20 at 2 08 24 PM

No where currently uses subtopics, but effectively topic objects in the topics array of a post (like [`angular.11tydata.js`](https://github.com/GoogleChrome/web.dev/blob/4ec23419bbdac156ea5fb6410821db2306304bd4/src/site/content/en/angular/angular.11tydata.js#L15)) now support a `subtopics` field, which is just an array of objects of the below data structure:

```TypeScript
type SubTopic = {
     title: string;
     pathItems: string[];
}
```

src/site/_includes/partials/topic.njk
```diff
@@ -1,21 +1,21 @@
...
-    {% else %}
-      <li class="w-path-listitem">
-        <a
-          class="w-path-link"
-          href="{{ item.url | stripLanguage }}"
-        >
-          {{ item.data.title | md | safe }}
-        </a>
-      </li>
+<div class="w-topic">
```
robdodson: What do you think about doing something like this:

```
<div class="w-topic">
  <h2 id="{{ topic.title | slug }}" class="no-link w-headline--three w-topic-header">
    {{ topic.title }}
    <a class="w-headline-link" href="#{{ topic.title | slug }}" aria-hidden="true">#</a>
  </h2>
  <ul class="w-path-list">
    {% if topic.pathItems %}
      {% set section = topic %}
      {% include 'partials/item.njk' %}
    {% endif %}
  </ul>
  {% if topic.subtopics %}
    {% for section in topic.subtopics %}
      <h3 id="{{ section.title | slug }}" class="no-link w-headline--four w-subtopic-header">
        {{ section.title }}
        <a class="w-headline-link" href="#{{ section.title | slug }}" aria-hidden="true">#</a>
      </h3>
      <ul class="w-path-list">
        {% include 'partials/item.njk' %}
      </ul>
    {% endfor %}
  {% endif %}
</div>
```

I think this'll avoid the need for negative margin on the subtopic heading.

src/styles/components/_topic.scss
```diff
@@ -0,0 +1,58 @@
...
+// Topics are the sections of a collection and can be broken into subtopics.
+//
+// =============================================================================
+
+$TOPIC_SPACING_SIZE_LARGE: 48px;
+$TOPIC_SPACING_SIZE_MEDIUM: 40px;
+$TOPIC_SPACING_SIZE_SMALL: 32px;
+
+.w-topics {
+  .w-topic {
```
robdodson: I think we could BEM these up a bit so they don't rely on descendant selectors. As much as possible we try to avoid descendant selectors to avoid specificity weirdness.

Maybe something like:

```
.w-topic {...}
.w-topic__header {...}
.w-topic__subtopic-header {...}
```

src/styles/components/_topic.scss
```diff
@@ -0,0 +1,58 @@
...
+//
+// Topics are the sections of a collection and can be broken into subtopics.
+//
+// =============================================================================
+
+$TOPIC_SPACING_SIZE_LARGE: 48px;
+$TOPIC_SPACING_SIZE_MEDIUM: 40px;
+$TOPIC_SPACING_SIZE_SMALL: 32px;
+
+.w-topics {
```
robdodson: Is this class needed?

src/site/_includes/path.njk
```diff
@@ -60,10 +60,10 @@ layout: layout
   </section> #}
 
   <section class="w-layout-container">
-    <div class="w-numbered-headers">
+    <div class="w-topics">
```
robdodson: You might be able to get rid of this div if it's not targeted by any css

robdodson:
LGTM! Nice work 🙌

PR #1938, opened 2019-11-22
feat(CopyCode): add custom element to allow code blocks to be copied, fixes #1871

### LINKED ISSUE #1871: Add button for copying code

Blocks of code like this should be copy-able with a single button click:

image

Add a **Copy to clipboard** button to the top-right of each code block.

None.

**Additional context**

See DevSite for example:

image

https://cloud.google.com/nodejs/getting-started/hello-world

Would also be helpful to fire an analytics event every time that the code is copied. This would enable us to track how many users actually copy the code when doing a codelab, for example.

MichaelSolati:
Fixes #1871

Add custom element that wraps fenced code blocks and gives them a button to copy the code. I would love @kaycebasques thoughts on how it looks though.

### DIFF
--- a/src/lib/components/CopyCode/index.js
+++ b/src/lib/components/CopyCode/index.js
@@ -11,47 +11,39 @@ import {BaseElement} from "../BaseElement";
  * @final
  */
 class CopyCode extends BaseElement {
-  static get properties() {
-    return {
-      code: {type: String},
-    };
-  }
-
   constructor() {
     super();
+    this.copyCode = this.copyCode.bind(this);
   }
 
   connectedCallback() {
     super.connectedCallback();
-    this.setAttribute("tabindex", "0");
 
-    if (this.firstChild) {
-      const copyButton = document.createElement("input");
-      copyButton.setAttribute("type", "button");
-      copyButton.setAttribute("tabindex", "0");
-      copyButton.className = "w-copy-code-button";
+    this.copyButton = document.createElement("button");
+    this.copyButton.className = "web-copy-code__button";
+    this.copyButton.addEventListener("click", this.copyCode);
 
-      this.firstChild.prepend(copyButton);
+    const copyContainer = document.createElement("div");
+    copyContainer.className = "web-copy-code__container";
+    copyContainer.append(this.copyButton);
 
-      this.addEventListener("click", (e) => {
-        if (e.target === copyButton) {
-          e.preventDefault();
-          this.copyCode();
-        }
-      });
-    }
+    this.prepend(copyContainer);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    this.copyButton.removeEventListener("click", this.copyCode);
   }
 
   copyCode() {
-    const textarea = document.createElement("textarea");
-    textarea.value = decodeURIComponent(this.code);
-    textarea.setAttribute("readonly", "");
-    textarea.style.cssText = "position: fixed; bottom: -80px; height: 80px;";
-    document.body.appendChild(textarea);
-    textarea.select();
+    window.getSelection().removeAllRanges();
+    const range = document.createRange();
+    range.selectNode(this.querySelector("code"));
+    window.getSelection().addRange(range);
     document.execCommand("copy");
-    document.body.removeChild(textarea);
+    window.getSelection().removeAllRanges();
   }
 }
 
-customElements.define("copy-code", CopyCode);
+customElements.define("web-copy-code", CopyCode);
--- a/src/lib/components/CopyCode/index.js
+++ b/src/lib/components/CopyCode/index.js
@@ -13,7 +13,7 @@ import {BaseElement} from "../BaseElement";
 class CopyCode extends BaseElement {
   constructor() {
     super();
-    this.copyCode = this.copyCode.bind(this);
+    this.onCopy = this.onCopy.bind(this);
   }
 
   connectedCallback() {
@@ -21,7 +21,7 @@ class CopyCode extends BaseElement {
 
     this.copyButton = document.createElement("button");
     this.copyButton.className = "web-copy-code__button";
-    this.copyButton.addEventListener("click", this.copyCode);
+    this.copyButton.addEventListener("click", this.onCopy);
 
     const copyContainer = document.createElement("div");
     copyContainer.className = "web-copy-code__container";
@@ -33,10 +33,10 @@ class CopyCode extends BaseElement {
   disconnectedCallback() {
     super.disconnectedCallback();
 
-    this.copyButton.removeEventListener("click", this.copyCode);
+    this.copyButton.removeEventListener("click", this.onCopy);
   }
 
-  copyCode() {
+  onCopy() {
     window.getSelection().removeAllRanges();
     const range = document.createRange();
     range.selectNode(this.querySelector("code"));
--- a/src/lib/components/CopyCode/index.js
+++ b/src/lib/components/CopyCode/index.js
@@ -1,0 +1,57 @@
+/**
+ * @fileoverview Element that renders copyable code.
+ */
+
+import {BaseElement} from "../BaseElement";
+
+/**
+ * Renders code block that can easily be copied.
+ *
+ * @extends {BaseElement}
+ * @final
+ */
+class CopyCode extends BaseElement {
+  static get properties() {
+    return {
+      code: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute("tabindex", "0");
+
+    if (this.firstChild) {
+      const copyButton = document.createElement("input");
+      copyButton.setAttribute("type", "button");
+      copyButton.setAttribute("tabindex", "0");
+      copyButton.className = "w-copy-code-button";
+
+      this.firstChild.prepend(copyButton);
+
+      this.addEventListener("click", (e) => {
+        if (e.target === copyButton) {
+          e.preventDefault();
+          this.copyCode();
+        }
+      });
+    }
+  }
+
+  copyCode() {
+    const textarea = document.createElement("textarea");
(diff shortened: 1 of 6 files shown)


src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+    };
+  }
+
+  constructor() {
+    super();
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute("tabindex", "0");
```
robdodson: I think you can get rid of this because focus will go to the `copyButton`.

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (this.firstChild) {
+      const copyButton = document.createElement("input");
```
robdodson: I think you can replace this with `"button"`, then you don't need to set the type.

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+      copyButton.setAttribute("type", "button");
+      copyButton.setAttribute("tabindex", "0");
```
robdodson: input and button are implicitly focusable so you can remove setting the tabindex

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+    textarea.setAttribute("readonly", "");
+    textarea.style.cssText = "position: fixed; bottom: -80px; height: 80px;";
+    document.body.appendChild(textarea);
+    textarea.select();
+    document.execCommand("copy");
+    document.body.removeChild(textarea);
+  }
+}
+
+customElements.define("copy-code", CopyCode);
```
robdodson:
```suggestion
customElements.define("web-copy-code", CopyCode);
```

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.setAttribute("tabindex", "0");
+
+    if (this.firstChild) {
+      const copyButton = document.createElement("input");
+      copyButton.setAttribute("type", "button");
+      copyButton.setAttribute("tabindex", "0");
+      copyButton.className = "w-copy-code-button";
```
robdodson: nit: `web-copy-code__button`.

src/lib/components/CopyCode/styles.scss
```diff
@@ -0,0 +1,38 @@
+@import "../../../styles/settings/colors";
+@import '../../../styles/settings/global';
+
+.w-copy-code-button {
```
robdodson: Apologies for being a bit inconsistent with our naming.

For custom elements, we always prefix with `web-*`.
For other CSS components we just use the `w-*` namespace.

I realize that's hecka weird but we should be consistent until we update everything. I think we can name this component `web-copy-code` for now, and when I get back the week after next we can do a bit of house cleaning and give everything the same prefix.

src/lib/components/CopyCode/styles.scss
```diff
@@ -0,0 +1,38 @@
...
+  background-image: url("../../../images/icons/file_copy.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+  border: none;
+  border-radius: $GLOBAL_ROUNDED;
+  float: right;
+  height: 48px;
+  padding: 0;
+  width: 48px;
+  z-index: 100;
```
robdodson: Is this needed?

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+      this.firstChild.prepend(copyButton);
+
+      this.addEventListener("click", (e) => {
```
robdodson: Could you do something like:

```
this.copyButton.addEventListener("click", this.onCopy)
```

I think stylistically that may be consistent with some of our other custom elements. We've tended to declare handlers with the `on*` prefix versus doing them inline.

robdodson: Note that you'll probably want to do:

```
this.onCopy = this.onCopy.bind(this);
```

in this element's constructor to properly scope that handler.

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+        if (e.target === copyButton) {
+          e.preventDefault();
+          this.copyCode();
+        }
+      });
+    }
+  }
+
+  copyCode() {
```
robdodson: This might be a good opportunity to use the newer [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) and fallback to `execCommand` if the browser doesn't support it.

```
if ('clipboard' in navigator) {
  navigator.clipboard.writeText(this.innerText)
} else {
  // fallback
}
```

I think the MDN docs say the Clipboard API requires permissions but in my testing in Chrome and Firefox it didn't prompt for permission. It is worth noting that Safari doesn't support the Clipboard API so we'd need to use a fallback for it. My hope is that eventually they'll ship support and we can remove the fallback.

Btw, I think we might be able to get away with using `innerText` instead of relying on the attribute.

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+    const textarea = document.createElement("textarea");
```
robdodson: Another way to go about this might be to select the text of the firstChild itself. Taking a look at [this doc](https://developers.google.com/web/updates/2015/04/cut-and-copy-commands) I tried:

```
window.getSelection().removeAllRanges();
const range = document.createRange();
range.selectNode(this.firstChild);
window.getSelection().addRange(range);
document.execCommand("copy");
window.getSelection().removeAllRanges();
```

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,57 @@
...
+    return {
+      code: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
+  }
+
+  connectedCallback() {
```
robdodson: Suggest also adding a `disconnectedCallback` to remove any event listeners and clean up after the element.

src/lib/components/CopyCode/styles.scss
```diff
@@ -0,0 +1,38 @@
+@import "../../../styles/settings/colors";
+@import '../../../styles/settings/global';
+
+.w-copy-code-button {
+  background: none;
+  background-image: url("../../../images/icons/file_copy.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+  border: none;
+  border-radius: $GLOBAL_ROUNDED;
```
robdodson: Also add `cursor: pointer`

src/lib/components/CopyCode/styles.scss
```diff
@@ -0,0 +1,38 @@
...
+  float: right;
+  height: 48px;
+  padding: 0;
+  width: 48px;
+  z-index: 100;
+  opacity: 0;
+
+  &:hover,
+  &:focus {
+    background-color: rgba($WHITE, 0.5);
```
robdodson: I think the white on the grey background is a little too low contrast. Maybe we can try one of the other colors to see if it provides a better visual. Maybe $WEB_SECONDARY_COLOR?

src/lib/components/CopyCode/styles.scss
MichaelSolati: Are we referring to the z-index? Probably not, I think I was tinkering with how I would position the button and it lingered.

src/lib/components/CopyCode/index.js
MichaelSolati: I was under the impression that event listeners belonging to a custom element, or elements that belong to a custom element are [removed on the elements destruction](https://stackoverflow.com/a/35108388)? I'm assuming this would have only been applicable if the element was using the shadow dom?

src/lib/components/CopyCode/styles.scss
MichaelSolati: I'm not sure if I love the color, so I lowered the opacity. I also applied a transition which smoothes out the animation.

src/lib/components/CopyCode/index.js
MichaelSolati: I'm digging this solution.

src/lib/components/CopyCode/index.js
```diff
@@ -0,0 +1,49 @@
...
+    this.prepend(copyContainer);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    this.copyButton.removeEventListener("click", this.copyCode);
+  }
+
+  copyCode() {
```
robdodson: tiny style nit, would you mind renaming this function to `onCopyCode` or maybe just `onCopy`? I think elsewhere we've used the `on*` prefix to denote event handlers.

MichaelSolati: I have no qualms renaming it, but if you don't mind me asking, is there a reason to make the change? I completely believe that event handlers (or emitted events) should be prefixed by `on`. However if it's an event internal to the component and not bubbled up, does it still count?

robdodson: oh it's just a code style thing. I tend to always prefix handlers with `on*` so if someone is scanning they can quickly grok that it's a handler and not a utility method ([example](https://github.com/GoogleChrome/web.dev/blob/master/src/lib/components/SideNav/index.js#L192)). Just personal preference. But since it's been done a lot already in the existing code, I was thinking it would be good to keep the pattern for consistency sake.

src/lib/components/CopyCode/index.js
robdodson: I think you're correct that the handler will be automatically deleted. Previously Sam and I were encouraging one another to add `disconnectedCallback` to make sure we disconnected elements from the `store`. I think it was muscle memory that encouraged me to leave my previous comment. In this instance it's probably fine to remove the `disconnectedCallback` if you'd like to shave some bytes.

PR #1956, opened 2019-12-04
fix(w-card): set fixed height of `.w-card` when on breakpoints smaller than smart phones, fixes #1653

### LINKED ISSUE #1653: Posts without a thumbnail are collapsed a bit on mobile.
**Describe the bug**
See screenshot:
image

**Expected behavior**
The post should have a bit more min-height so it's size matches its peers. We do this on desktop, so it must be an oversight on mobile.

MichaelSolati:
Fixes #1653.

In order to keep the cards in the blog view the same height, regardless of if in mobile view (single card per row) or in desktop view (multiple cards per row) I've given the cards a set height (which is the height of a thumbnail card when the bp is less than md). When md or larger it'll default to auto where cards without a thumbnail will match the height of the other cards (with a thumbnail).

### DIFF
--- a/src/styles/components/_card.scss
+++ b/src/styles/components/_card.scss
@@ -13,13 +13,13 @@
 .w-card {
   border-radius: 8px;
   display: block;
-  height: 462px;
+  min-height: 462px;
   overflow: hidden;
   position: relative;
   transition: box-shadow $TRANSITION_SPEED;
 
   @include bp(md) {
-    height: auto;
+    min-height: auto;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/styles/components/_card.scss
```diff
@@ -13,9 +13,14 @@
 .w-card {
   border-radius: 8px;
   display: block;
+  height: 462px;
```
robdodson: looks good. one small suggestion would be to use `min-height` instead here so things don't get cut off if they have a lot of text:

image

PR #1966, opened 2019-12-05
feat(SnackbarContainer): only allow PWA install prompt if cookies accepted, fixes #1771

### LINKED ISSUE #1771: We should defer the PWA install banner until after the cookie banner.
When a new user visits the site, we show the cookie banner and the app install banner at the same time. Because they overlap, the add to homescreen banner actually covers the ok/more info cookies buttons.

Defer the add to home screen banner until after the user has accepted cookies.

MichaelSolati:
Fixes #1771

On connection of the component add an event listener for before the PWA install prompt appears. Stop it and save it until the user has accepted cookies.

### DIFF
--- a/src/lib/components/SnackbarContainer/index.js
+++ b/src/lib/components/SnackbarContainer/index.js
@@ -34,6 +34,7 @@ class SnackbarContainer extends BaseElement {
 
   constructor() {
     super();
+    this.onBeforeInstallPrompt = this.onBeforeInstallPrompt.bind(this);
     this.onStateChanged = this.onStateChanged.bind(this);
   }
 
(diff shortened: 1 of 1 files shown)


src/lib/components/SnackbarContainer/index.js
```diff
@@ -34,24 +34,51 @@ class SnackbarContainer extends BaseElement {
...
+    this.onBeforeInstallPrompt = this.onBeforeInstallPrompt.bind(this);
     this.onStateChanged = this.onStateChanged.bind(this);
   }
 
   connectedCallback() {
     super.connectedCallback();
     store.subscribe(this.onStateChanged);
     this.onStateChanged();
+
+    if (!this.acceptedCookies) {
```
robdodson: I was wondering if it's possible that the user may have already accepted cookies, but hasn't been hit with the install prompt yet. I don't know if it's still the case, but Chrome used to have a heuristic that waited until the user had visited a site more than once before showing them the install prompt. If that's still the case then they may have accepted cookies on the first visit but not seen the prompt yet and this listener would never set up.

src/lib/components/SnackbarContainer/index.js
```diff
@@ -34,24 +34,51 @@ class SnackbarContainer extends BaseElement {
...
     store.unsubscribe(this.onStateChanged);
+    window.removeEventListener(
+      "beforeinstallprompt",
+      this.onBeforeInstallPrompt,
+    );
+  }
+
+  onBeforeInstallPrompt(e) {
+    e.preventDefault();
+    this.installPrompt = e;
```
robdodson: Is it worth checking to see if the user has already accepted cookies here? If they have, we could let the event proceed and if they haven't we could capture the event.

robdodson:
@samthor would you mind taking a look as well?

src/lib/components/SnackbarContainer/index.js
MichaelSolati: Well if the user has accepted cookies then we don't need to wait for them to accept cookies and then present it. We only want to prevent the install prompt if they haven't accepted cookies yet. Otherwise let chrome run it's normal business.

src/lib/components/SnackbarContainer/index.js
MichaelSolati: This will never run if the user has accepted cookies. It's attached as a listener if line 46 is false. I just didn't want to attach the listener at all if they've accepted cookies.

src/lib/components/SnackbarContainer/index.js
robdodson: ah yes, good point!

src/lib/components/SnackbarContainer/index.js
robdodson: got it, sounds good

robdodson:
LGTM. I'll wait for Sam to give it a second look before landing.

PR #1969, opened 2019-12-06
Update Word list

jpmedley:
Note that in code 'referer' is misspelled, a fact that was not caught before  being widely implemented in web servers and browsers. https://tools.ietf.org/html/rfc7231#section-5.5.2


kaycebasques:
Thanks @jpmedley

PR #1970, opened 2019-12-06
fix(table): tables written in markdown are automatically wrapped by `<div class="w-table-wrapper">`, fixes #1917

### LINKED ISSUE #1917: Create a markdown-it plugin to wrap tables
Tables should be contained within a `<div class="w-table-wrapper">` class so they overflow correctly on mobile.

A [markdown-it](https://github.com/markdown-it/markdown-it) plugin that wraps tables in the aforementioned div.

MichaelSolati:
Fixes #1917

Add some rendered rules in order to wrap table elements with a `<div class="w-table-wrapper">`. This does not work with an HTML table, as from what I understand HTML doesn't go through the custom element renderers (I think it's just passed through?).

So possibly we could implement a rule for HTML and modify it, but I think that would be ugly... Maybe something like jsdom to pull out tables and add the div? (Something like this...)

```JavaScript
const jsdom = require("jsdom");

const htmlRule = (html) => {
   const dom = new JSDOM(html);

   dom.window.document.querySelectorAll("table").forEach((table) => {
    const parent = table.parentNode;
    if (!(parent instanceof HTMLDivElement && parent.classList.contains("w-table-wrapper"))) {
      const newParent = dom.window.document.createElement("div");
      newParent.classList.add("w-table-wrapper");
      parent.insertBefore(newParent, table);
      newParent.appendChild(table);
     }
  });
}
```

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -102,13 +102,19 @@ module.exports = function(config) {
     .use(markdownItAnchor, markdownItAnchorOptions)
     .use(markdownItAttrs, markdownItAttrsOpts);
 
-  // custom renderer for fences
+  // custom renderer rules
   const fence = mdLib.renderer.rules.fence;
-  mdLib.renderer.rules.fence = function (tokens, idx, options, env, slf) {
-    const fenced = fence(tokens, idx, options, env, slf)
-    const token = tokens[idx];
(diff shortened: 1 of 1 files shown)


robdodson:
The HTML rule is interesting but maybe we can hold off on it for now. I agree it's a bit much.

PR #1971, opened 2019-12-08
Blog (lazy-loading): Remove the `auto` value, per spec changes

Malvoz:
The `auto` value has been removed from the draft spec, see https://github.com/whatwg/html/pull/3752#issuecomment-555328635 and https://github.com/whatwg/html/pull/3752#issuecomment-555531017 for the motivation.

This PR removes the `auto` value. Would you rather include a note that "there was initially a third value; the `auto` value, which was the default value"~ or is it acceptable to just delete it?

/cc @housseindjirdeh, @robdodson

### DIFF
--- a/src/site/content/en/blog/native-lazy-loading/index.md
+++ b/src/site/content/en/blog/native-lazy-loading/index.md
@@ -70,10 +70,8 @@ and iframes that can be reached by scrolling:
 
 Here are the supported values for the `loading` attribute:
 
-- `auto`: Default lazy-loading behavior of the browser, which is the same as not
-  including the attribute.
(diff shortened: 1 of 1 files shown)


domfarolino:
Thanks - just for clarity, while the spec PR no longer has the "auto" attribute, Chrome currently ships "auto" support ("auto" == "eager" in non-Lite-mode Chrome) and does not have any plans to un-ship it.

In the future, once Document Policy is further along, we're planning on re-introducing "auto" to the spec, and having a Document Policy control the behavior of "auto".

Malvoz:
Thanks @domfarolino. I think it's best to leave this as is. Closing.

PR #1983, opened 2019-12-10
feat(post): add table of contents to posts, fixes #1931

### LINKED ISSUE #1931: Add table of contents to articles.
We'll need some design help on this. I think having each article lead off with a TOC may push a bunch of stuff down the page, so we might consider using a button to open the TOC?

There are eleventy plugins for generating TOCs (https://www.npmjs.com/package/eleventy-plugin-toc) so we probably just need to chuck one of these into our project and style it.

MichaelSolati:
Fixes #1931


samthor:
Could this be in a closed `<details>` element? Clearly this is useful to have, but articles like [this one](https://deploy-preview-1983--web-dev-staging.netlify.com/prefers-color-scheme/) end up having a huge amount of cruft before you get to the article. Especially on mobile.

robdodson:
I pinged Kayce in the bug to ask if he could provide some examples of TOCs that he likes. I think it may also be worth reaching out to our designer Jiwoong after Kayce shares examples so he can style something for us.

kaycebasques:
Michael and I talked about this in person a while back but I'll just note it here for the record:

DevSite's approach of having the ToC in a third column on the right is pretty standard for documentation sites:

image

On narrow viewports DevSite shows the first few ToC items but then hides the rest and that approach seems fine to me:

image

I'm not too invested in the specifics of where the ToC goes just so long as that it's immediately findable when I first visit a page

robdodson:
My 2c is that the layout of devsite and similar docs sites like [microsoft's edge docs](https://docs.microsoft.com/en-us/intune/apps/apps-edge-macos) are a bit information heavy and cluttered feeling and reducing information density was one of the goals I wanted for our content pages. FWIW css-tricks, smashing mag, alligator.io, Medium, etc. don't do TOCs —  not saying we should block on that but I did want to highlight that design consideration.

I'd be curious to see if we could explore putting the TOC inside of some piece of UI, maybe like the 'show docs outline' that appears in google docs?
image

robdodson:
@jimoong if you have spare cycles at some point could we work with you on a TOC treatment?

jimoong:
Sure! Will take a look at this.

jimoong:
As Rob mentioned, I'd also prefer keeping the visual style of article more close to blog article style then tech doc because that's the direction we took when originally designing the blog system. So I took the Rob's icon idea to generate some mocks as an example. Details might be off since this is a quick sketch, so please forgive. PTAL and let me know what you think.

1. ToC collapsed and icon on top right corner of the article.
ToC collapsed

2. ToC expanded as a RHS panel. Trying to keep content area clean.
ToC expanded

3. User selecting a section
ToC selected

Translated to mobile version
ToC-mobile

robdodson:
I like the look of these.

@jimoong earlier in the thread Sam noted that [this article](https://deploy-preview-1983--web-dev-staging.netlify.com/prefers-color-scheme/) has a really long table of contents. How do you think we should handle something like that on mobile? Maybe we limit the height of the table of contents and make it scroll?

MichaelSolati:
> I like the look of these.
> 
> @jimoong earlier in the thread Sam noted that [this article](https://deploy-preview-1983--web-dev-staging.netlify.com/prefers-color-scheme/) has a really long table of contents. How do you think we should handle something like that on mobile? Maybe we limit the height of the table of contents and make it scroll?

I kinda like the idea of keeping the side menu even on mobile.

samthor:
It's not an ultimate solution but I also wonder whether we could only show the first-level of headings (or a sensible minimum count) on small screens. Similarly I wonder whether even on desktop we should only show say, 2 levels.

Thomas' article is particularly detailed though—most aren't like that :)

kaycebasques:
I'm open to any UI just as long as it's immediately discoverable when I first load a page.

jpmedley:
> It's not an ultimate solution but I also wonder whether we could only show the first-level of headings (or a sensible minimum count) on small screens. Similarly I wonder whether even on desktop we should only show say, 2 levels.

I don't know how much of a solution this is. With 1200 words as a goal, articles with multiple levels of structure should be outliers, only occurring when the subject itself is complex. Even with more complex topics, I look for ways to limit TOC complexity. I like TOCs to clearly communicate something about the structure of an article. (Why waste an opportunity to communicate.)

jpmedley:
Also truncating the TOC, as opposed to closing all the nesting, lets readers get a sense of the whole TOC with a single click instead of, potentially, multiple clicks.

kaycebasques:
@jimoong's [proposal](https://github.com/GoogleChrome/web.dev/pull/1983#issuecomment-589228536) LGTM.

robdodson:
Some initial thoughts.

- I think we need to link h3s as well as h2s.
- I don't think the implementation matches Jiwoong's design. See:
https://user-images.githubusercontent.com/1054804/74964092-a3355480-53c7-11ea-94b5-5e34eb7336a3.png

It looks like h2's are larger and have a bit more space around them. Also looks like he might be using Google Sans instead of the default Roboto. Note that only the active item is blue and there's a blue bar next to it to indicate active state.

- On mobile we probably need to make it so the button to open the ToC is not sticky and scrolls away when you scroll past the top of the article. Otherwise it takes up too much space. This also includes when you click on a link.
- The label/checkbox CSS trick is not keyboard accessible (you can't focus the label using tab key). I'd suggest using a `<button>`. You may need to turn this into a real component.

src/site/_includes/layout.njk
```diff
@@ -165,6 +165,7 @@ show_banner: true
     </web-side-nav>
     <main>
       <div id="content">
+        <input class="hidden" id="w-toc__open-state" type="checkbox">
```
robdodson, line 168: Hm I think this'll be added to every page. Maybe we don't need this if we decide to use a real component that's only displayed on posts.

src/site/_includes/partials/post.njk
```diff
@@ -10,6 +10,19 @@
     } %}
   {% endif %}
 
+  <div class="w-toc__button-wrapper">
+    <label for="w-toc__open-state" class="w-toc__button w-button w-button--secondary w-button--icon" data-icon="list_alt" aria-label="Toggle table of contents"></label>
+  </div>
+  <div class="w-toc">
+    <div class="w-toc__label">
+      <span>In this article</span>
```
robdodson, line 18: May as well make this a div since you want a block layout.

src/site/_includes/partials/post.njk
```diff
@@ -10,6 +10,19 @@
...
+      <span class="w-toc__label-spacer"></span>
```
robdodson, line 19: hm this seems a bit odd? why do you need this?

jpmedley:
> * I think we need to link h3s as well as h2s.

That depends very much on the article. Will there be a way to control this?

MichaelSolati:
@robdodson I think the design should be much more inline with @jimoong's design (mind you it's hard to be exact). I know that extracting the headers to generate the TOC is something I still need to work on, but if you could PTAL at the general appearance and behavior that would be awesome.

@jpmedley Unfortunately I don't think I can let authors choose which headings they want. The library that I forked and am using for this right now seems to be somewhat broken (in the sense that i couldn't pass in options at the template.) I forked it so I can set which headers I want it to use when the plugin is added to eleventy, not when it's invoked on a page.

robdodson:
Experientially it's seeming pretty good with one exception... I think on mobile we may want the button to be statically positioned at the top of the article. Otherwise it's just too cluttered:

image

robdodson:
In other words, on mobile we wouldn't do a sticky positioned element. It would scroll away as the user scrolled down the page.

src/styles/components/_toc.scss
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+}
+
+web-table-of-contents {
+  background-color: $WHITE;
+  border-bottom: 1px solid $BORDER_COLOR;
+  border-top: 1px solid $BORDER_COLOR;
+  display: none;
+  padding: 24px;
```
robdodson, line 48: I think putting padding on the element in this way will prevent you from being able to do the little blue rectangle to the left of the scrollSpy'd anchor.

image

src/site/_includes/partials/post.njk
```diff
@@ -10,6 +10,10 @@
     } %}
   {% endif %}
 
+  <web-table-of-contents title="{{ title | md | safe }}" title-id="{{ title | md | safe | slugify }}">
```
robdodson, line 13: it looks like you pipe title through a number of filters both here and in h1.

to dry it up a bit you could use `set` maybe and set it as a variable?

src/styles/components/_toc.scss
```diff
@@ -0,0 +1,162 @@
...
+.w-toc__active--border {
+  border-left: 3px solid $WEB_PRIMARY_COLOR;
+  margin-left: -24px;
+  padding-left: 21px;
+}
+
+div#content.w-toc-open {
+  > * {
+    @include bp(md) {
+      width: calc(100% - 268px);
```
robdodson, line 30: I think you may want to put the ToC width into a variable. Because it's so important you might want to put it in settings/_dimensions.scss. That's where the variable for the topnav is.

robdodson: I also feel a bit weird about this rule. It clearly works, but it's also very broad.
I was wondering if this could be done with flexbox by giving the toc a flex-basis to 268px when it's open and setting its flex-shrink to 0. Might need to fiddle around with it a bit, but this might be safer. For instance, we might put an absolutely positioned child in `#content` and we wouldn't want its width to get broken when the ToC is open.

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,154 @@
...
+      'w-button--icon',
+    );
+    tocButton.setAttribute('data-icon', 'list_alt');
+    tocButton.addEventListener('click', this.open);
+
+    tocWrapper.append(tocButton);
+    this.after(tocWrapper);
+  }
+
+  scrollSpy() {
```
robdodson, line 86: I noticed on this page https://deploy-preview-1983--web-dev-staging.netlify.app/prefers-color-scheme/
when you scroll near "Why dark mode" the active link disappears. I'm guessing because there's an h4 or something?

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,154 @@
...
+    this.open = this.open.bind(this);
+    this.scrollSpy = this.scrollSpy.bind(this);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    this.tocInnerDiv = document.createElement('div');
+    this.tocInnerDiv.classList.add('w-toc__content');
+    this.tocInnerDiv.innerHTML = this.innerText;
```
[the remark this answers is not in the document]
MichaelSolati: The post is `/use-lazysizes-to-lazyload-images`, I hadn't been able to figure out why though...

Screen Shot 2020-06-11 at 3 54 44 PM

Possibly because there is an `` in a title? => https://web.dev/use-lazysizes-to-lazyload-images/#update-lessimggreater-andor-lesspicturegreater-tags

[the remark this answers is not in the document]
MichaelSolati: Hold on, may have found a fix...

In `/src/site/_transforms/responsive-images/index.js` I can check if there is an `originalSrc` and if there isn't skip that img element.

Screen Shot 2020-06-11 at 4 01 57 PM

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,154 @@
...
+    this.addOpenButton();
+
+    this.scrollSpy();
+    document.addEventListener('touchmove', this.scrollSpy);
+    document.addEventListener('scroll', this.scrollSpy);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.close();
```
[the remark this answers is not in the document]
MichaelSolati: This removes `.w-toc-open` from the `div#content`

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,162 @@
...
+ * Element that renders table of contents.
+ * @extends {BaseElement}
+ * @final
+ */
+class TableOfContents extends BaseElement {
+  static get properties() {
+    return {
+      title: {type: String},
+      titleId: {type: String, attribute: 'title-id'},
+      opened: {type: Boolean, converter: stringToBoolean, reflect: true},
```
robdodson, line 31: You shouldn't need a converter, lit will do it for you.

Here's an example:
https://glitch.com/edit/#!/adjoining-erratic-cake?path=index.html%3A7%3A25

A boolean attribute is true if it's present on the html and false if it is absent.
e.g.

```html
<my-foo opened> // opened is true
<my-foo opened=""> // opened is true
<my-foo> // opened is false
```

robdodson: So you should be able to do something like:

```js
updated(changedProperties) {
  if (changedProperties.has('opened')) {
    const opened = changedProperties.get('opened');
    if (opened) {
      this.scrollSpy();
      document.addEventListener('touchmove', this.scrollSpy, {passive: true});
      document.addEventListener('scroll', this.scrollSpy, {passive: true});
    } else {
      document.removeEventListener('touchmove', this.scrollSpy);
      document.removeEventListener('scroll', this.scrollSpy);
    }
  }
}
```

MichaelSolati: So the converter works for cases when `<my-foo opened="false">`.

As for the `updated` method I still think we should have the `open` and `close` methods. Especially the `close` one. The document event listened can exist after the element is removed, so I call the `close` when the element is removed.

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+        </h2>
+        ${html([this.tocHTML])}
+      </div>
+    `;
+  }
+
+  updated(changedProperties) {
+    if (
+      changedProperties.has('opened') &&
+      changedProperties.get('opened') !== this.opened
```
MichaelSolati, line 77: So the diff check between the previous state and the new state is necessary for when you open/close a the toc then go to another page. If I just checked `changedProperties.get('opened')` (the previous value) it would falsely trigger the `#content` to have to `.w-toc-open` but would not update the property of the element. It was weird, so this seems to work well (I'm not 100% why it behaves this way)

package.json
```diff
@@ -73,6 +73,7 @@
     "chokidar-cli": "^1.2.2",
     "common-tags": "^1.8.0",
     "dotenv": "^8.2.0",
+    "eleventy-plugin-toc": "github:MichaelSolati/eleventy-plugin-toc#feat/options",
```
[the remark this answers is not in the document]
MichaelSolati: So a few things my fork did was 1) allow for us to apply options for the plugin when we configure it with 11ty. The inline options just would not work. Also I added options so we can use an `ul` instead of an `ol` as well as allow for a flat list, rather than embedded `ul`s.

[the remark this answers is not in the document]
MichaelSolati: I guess we need to discuss what we should do with the library, if we want to continue with it, etc... Chat on Monday?

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+  constructor() {
+    super();
+    this.scrollSpy = this.scrollSpy.bind(this);
+    this.openedToFalse = this.openedToFalse.bind(this);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.tocHTML = this.innerHTML;
+    this.divContent =
```
robdodson, line 44: Can you pass #content in as a property or attribute?

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+      document.querySelector('#content') || document.createElement('div');
+    this.headings = this.divContent.querySelectorAll('h1[id], h2[id], h3[id]');
+
+    this.observer = new IntersectionObserver(this.scrollSpy);
+    this.headings.forEach((heading) => this.observer.observe(heading));
```
robdodson, line 49: I'd suggest looking into if you can use the rootMargin option of the observer to create a smaller "window" only at the top of the page that you're spying. Otherwise you're spying every header on the page and that means as soon as the article opens it's already the heading at the bottom.

image

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+      </div>
+    `;
+  }
+
+  updated(changedProperties) {
+    if (
+      changedProperties.has('opened') &&
+      changedProperties.get('opened') !== this.opened
+    ) {
+      if (this.opened) {
```
robdodson, line 79: Cross component side-effects should live with the action in actions.js. That way the element is less tightly coupled. Take a look at how the isModalOpen action works.

src/styles/components/_toc.scss
```diff
@@ -0,0 +1,161 @@
...
+      color: $WEB_PRIMARY_COLOR;
+    }
+  }
+}
+
+@mixin TOC_PADDING {
+  padding: 0 24px;
+}
+
+.w-toc__active--border {
```
robdodson, line 22: I think you're reaching for a BEM modifier when that's not the correct syntax to use. It looks like you're hitting the "grandchild selector" issue https://www.smashingmagazine.com/2016/06/battling-bem-extended-edition-common-problems-and-how-to-avoid-them/

If the border is just an element inside of the toc block then it's fine to do:

```
.w-toc__border
```

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+    this.divContent.classList.remove('w-toc-open');
+    this.observer.disconnect();
+  }
+
+  onStateChanged({tocOpened}) {
+    this.opened = tocOpened;
+  }
+
+  highlightActiveLink() {
+    const activeLink = 'w-toc__active--link';
```
robdodson, line 99: See the article I linked below about grandchild selectors. I think you're using BEM modifier syntax when you don't want to be.

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+    closeToC();
+  }
+
+  scrollSpy(headings) {
+    const links = [...this.querySelectorAll('a')];
+    for (const heading of headings) {
+      const href = `#${heading.target.getAttribute('id')}`;
+      const link = links.find((l) => l.getAttribute('href') === href);
+      if (heading.intersectionRatio > 0) {
+        link.classList.add('w-toc__visible--link');
```
robdodson, line 133: Would this be your firstVisibleLink or are the headings not guaranteed to come back in the correct order?

src/lib/components/TableOfContents/index.js
```diff
@@ -0,0 +1,144 @@
...
+        this.previouslyActiveHeading = heading.target.getAttribute('id');
+      } else {
+        link.classList.remove('w-toc__visible--link');
+      }
+
+      this.highlightActiveLink();
```
robdodson, line 139: Since you've already queried the links I wonder if you could move the logic from highlightActiveLink into this function? I'm also wondering if it can be optimized/simplified somehow but I guess that depends on whether or not the `headings` are in order.

src/lib/actions.js
```diff
@@ -217,3 +217,15 @@ export const setLanguage = store.action((state, preferredLanguage) => {
     userPreferredLanguage: preferredLanguage,
   };
 });
+
+export const closeToC = store.action(() => {
+  return {
+    tocOpened: false,
```
robdodson, line 223: change to isTocOpened. That'll match other booleans like isModalOpen, isSideNavExapnded, etc

src/lib/actions.js
```diff
@@ -217,3 +217,15 @@ export const setLanguage = store.action((state, preferredLanguage) => {
...
+  };
+});
+
+export const openToC = store.action(() => {
+  return {
+    tocOpened: true,
```
robdodson, line 229: isTocOpened

package.json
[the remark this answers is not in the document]
robdodson: Can you copy this into a directory in the _filters dir?

src/lib/components/TableOfContentsButton/index.js
```diff
@@ -0,0 +1,58 @@
...
+import {html} from 'lit-element';
+import {BaseStateElement} from '../BaseStateElement';
+import {openToC} from '../../actions';
+
+/**
+ * Element that renders table of contents open button.
+ * @extends {BaseStateElement}
+ * @final
+ */
+class TableOfContentsButton extends BaseStateElement {
```
robdodson, line 26: I don't think you need to extend BaseStateElement because this element will always call openToC—it doesn't care whether the ToC is already open or not.

So you could rewrite it as:

```js
class TableOfContentsButton extends BaseElement {
  constructor() {
    super();
  }

  render() {
    return html`
      <button
        class="w-toc__button--open w-button w-button--secondary w-button--icon"
        data-icon="list_alt"
        aria-label="Open Table of Contents"
        @click="${openToC}"
      ></button>
    `;
  }
```

src/lib/components/TableOfContents/index.js
robdodson: Suggest renaming to something like `articleContent`, as i'm not sure including `div` signifies much.

Also might want to log a warning if it's empty as I don't think that's a state we want to be in.

src/lib/components/TableOfContentsButton/index.js
MichaelSolati: It does care when the ToC is open though, in `src/styles/components/_toc.scss` it's set to `display: none;` when the `opened` attribute exists.

src/lib/components/TableOfContents/index.js
MichaelSolati: So this is reacting to a change made by:

```
onStateChanged({isTocOpened}) {
    this.opened = isTocOpened;
  }
```

which is part of the TableOfContents class. Pretty much I'm having the element subscribe to the state, but also as an input of the element's attribute.

src/lib/components/TableOfContentsButton/index.js
```diff
@@ -0,0 +1,58 @@
...
+    super();
+    this.openedToTrue = this.openedToTrue.bind(this);
+  }
+
+  render() {
+    return html`
+      <button
+        class="w-toc__button--open w-button w-button--secondary w-button--icon"
+        data-icon="list_alt"
+        aria-close="Open Table of Contents"
```
samthor, line 43: I think `aria-close` is a typo.

MichaelSolati:
> So I'm aware this has been bikeshedded to the nth degree but:
> 
> * why is this two elements? If it's just one, it doesn't need to be part of global state
> * `content-id` is needless config that we don't use. Can this instead just look for its `.closest('main')` and use that?

So passing in the ID was feedback, though I don't think I implemented it correctly. However the two elements and global state was also [feedback based on this](https://github.com/GoogleChrome/web.dev/pull/1983#discussion_r436167374).

Anyway, I fixed the aria attribute and opted to use `closest('main')` (cause it looks neater). Would you mind looking at it again @samthor?

robdodson:
> So passing in the ID was feedback, though I don't think I implemented it correctly. However the two elements and global state was also feedback based on this.

Yeah I found it a bit odd that the ToC itself would create a button, then go exploring outside of itself, up the ancestor tree to insert that button.

Having the open/closed state of the ToC be part of global state matches what we do elsewhere with hiding/showing modals and would give other elements the opportunity to react if they need to change appearance based on that state.

src/lib/components/TableOfContents/index.js
robdodson: I think `onStateChanged` is called by `connectedCallback` and will always fire before `updated` which means `this.opened` will always be false, even if the element has the attribute set.

I realized that BaseStateElement should probably connect to the store in `firstUpdated`, not `connectedCallback`. I'm not sure how many elements it would affect to change that now, but I'm doing a quick test.

PR #1988, opened 2019-12-11
Updates to "Multi origin PWA" article.

### LINKED ISSUE #1973: Various problems with multi-origin-pwas page
For page: /site/content/en/blog/multi-origin-pwas/index.md

1. Installation section recommends as a workaround that the `start_url` redirects to the main origin. It doesn't mention the fact that this will cause the installed app to show a permanent URL bar inside the app window, which might be a deal-breaker. I personally don't think it's worth giving this advice at all since it results in an obviously bad experience, but if we're going to include it, we should at least mention this as a downside. (Note that this same downside is mentioned in the first paragraph of Standalone mode section, but it isn't quite clear from reading it that it's a consequence of doing the redirect hack.)
2. In Standalone mode, it says "For example, the latest Chrome versions open a Chrome custom tab, when a user moves out of the scope in fullscreen mode." This should say "in standalone mode".
3. I'm not sure we should recommend "Device/Platform" as a reason to have a separate origin under "The Good". Having a separate "mobile." origin is questionable practice when it might be best to simply make the site adapt based on the screen size, rather than having an entirely separate origin. Having different origins may cause many of the same issues mentioned in the doc. Obviously many legacy sites are built this way, but that doesn't mean we should recommend it for new sites.
4. "Permissions are also scoped to origins. This means that if a user granted a given permission to the origin `https://maps.google.com`, it won't carry over to other origins, like `https://www.google.com`." This isn't the best example, since `https://maps.google.com` is a redirect to `https://google.com/maps` (so this particular site never experienced this problem in the first place). Since all other examples in the page use `example.com`, why not just say "This means that if a user granted a given permission to the origin `https://section.example.com`, it won't carry over to other origins, like `https://www.example.com`."

demianrenzulli:
Fixes #1973.

Changes proposed in this pull request: recommendations made by @mgiuca and others.

cc // @robdodson @petele


src/site/content/en/blog/multi-origin-pwas/index.md
```diff
@@ -96,20 +94,13 @@ Since there's no way to share permissions across origins, the only solution here
...
-
-- **Showing the prompt once, across different subdomains:** It's also possible to let each subdomain show the installation prompt. (This can make sense if some subdomains are more frequently visited). In this case, it's important to avoid showing the prompt to a user multiple times. To that purpose, a cookie could be used, to track if the prompt has already been shown, and check for its existence when the `beforeinstallprompt` event is triggered. If the cookie is present, it means that the user already received the prompt somewhere else, so `preventDefault()` can be called, to avoid showing it again.
-
+To mitigate this problem you can make sure that the prompt is shown only on a single domain. If the site has many subdomains that pass the installation criteria, the `beforeinstallprompt` event could be used and `preventDefault()` called (as explained [here](https://developers.google.com/web/fundamentals/app-install-banners/#listen_for_beforeinstallprompt)), to prevent the prompt from appearing in unintended parts of the site (like subdomains), while continue to show it in other parts (e.g. the home page).
```
kaycebasques: @demianrenzulli I see a few ways this paragraph can be improved. I'm going to give you guidelines but leave it up as an exercise for you to try.

1. Rephrase the link text following these directions: https://developers.google.com/style/link-text
1. Rephrase this section to use the [active voice](https://developers.google.com/style/voice): `the beforeinstallprompt event could be used and preventDefault() called`
1. Convert this section into a numbered list. I'll give you a hint on how it should start:

```
To mitigate this problem, make sure that the prompt is shown only on a single domain:

1. ...
```

src/site/content/en/blog/multi-origin-pwas/index.md
```diff
@@ -96,20 +94,13 @@ Since there's no way to share permissions across origins, the only solution here
...
 
 ### Standalone Mode
 
-While navigating in a standalone window, the browser will behave differently when the user moves outside of the scope set by the PWA's manifest. The behavior depends on each browser version and vendor. For example, the latest Chrome versions open a [Chrome custom tab](https://developer.chrome.com/multidevice/android/customtabs), when a user moves out of the scope in fullscreen mode.
+While navigating in a standalone window, the browser will behave differently when the user moves outside of the scope set by the PWA's manifest. The behavior depends on each browser version and vendor. For example, the latest Chrome versions open a [Chrome custom tab](https://developer.chrome.com/multidevice/android/customtabs), when a user moves out of the scope in standalone mode.
```
kaycebasques:
```suggestion
While navigating in a standalone window, the browser will behave differently when the user moves outside of the scope set by the PWA's manifest. The behavior depends on each browser version and vendor. For example, the latest Chrome versions open a [Chrome Custom Tab](https://developer.chrome.com/multidevice/android/customtabs), when a user moves out of the scope in standalone mode.
```

`Chrome Custom Tab` should be capitalized

FYI the link text you provided `Chrome Custom Tab` is an example of good link text because it matches the title of the document

src/site/content/en/blog/multi-origin-pwas/index.md
```diff
@@ -121,6 +112,10 @@ In most cases, there's no solution for this, but a workaround can be applied for
 The previous technique can help mitigating the potential UI change in a small part of the site, where the user can perform an action in a subdomain and return to the main origin (like in a login flow), but won't be an efficient technique to implement for entire paths, including many pages hosted in subdomains (like entire site sections).
 {% endAside %}
 
+{% Aside %}
+In the context of [Trusted Web Actitivies](https://developers.google.com/web/updates/2019/02/using-twa), there's a recommended way of avoiding this issue, by validating all origins using Digital Asset Links. You can find more information [here](https://github.com/GoogleChrome/android-browser-helper/tree/master/demos/twa-multi-domain/).
```
kaycebasques: Please rephrase the second link's text by following these guidelines: https://developers.google.com/style/link-text

demianrenzulli:
Thanks a lot for the feedback @kaycebasques! Just added modifications based on your comments. Please, let me know if it's ok!

src/site/content/en/blog/multi-origin-pwas/index.md
kaycebasques: Looks great! Good job @demianrenzulli.

src/site/content/en/blog/multi-origin-pwas/index.md
kaycebasques: Perfect. Good job again @demianrenzulli.

PR #1991, opened 2019-12-12
Add devtools-headsup include

mfriesenhahn:
Most codelabs that rely on Chrome DevTools don't explicitly state that Chrome is a dependency. If users try to follow the DevTools instructions in another browser, they won't work. To address that, this PR:
- Adds a devtools-headsup include with a link to download Chrome.
- Adds the new include to the start of all codelabs that rely on Chrome DevTools.
- Adds the new include to the codelab template in `/src/site/drafts` with a comment about when to keep it.


mfriesenhahn:
@kaycebasques, mind giving this a look when you have a sec? 😃

kaycebasques:
I think I was expecting Michael to merge this for some reason. Or maybe I just submitted the approval and got distracted before I merged. Anywho seems like I can merge now.

PR #1998, opened 2019-12-12
Fix multiple broken URLs that result in 404s

tcope25:
There are multiple URLs on web.dev that 404 or no longer found. This fixes a few of them.


robdodson:
thank you!

PR #1999, opened 2019-12-12
feat(blog): add pagination to blog landing page, fixes #719

### LINKED ISSUE #719: Add pagination to blog landing page.
I've asked Jiwoong for a design for this but we should also figure out how we want to do it with eleventy.

MichaelSolati:
Fixes #719

So this is currently designed to look like how a google search does pagination. Having looked at the designs, it seems like we'd show a few pages, then an ellipses, then the last page... Like so:

```
1  2  3  4  ...  9
```

Which I don't love. My implementation looks like this when there are less than 10 pages:

Screen Shot 2019-12-12 at 2 47 55 PM

Screen Shot 2019-12-12 at 2 48 03 PM

Screen Shot 2019-12-12 at 2 48 09 PM

And lets say there are more than 10:

Screen Shot 2019-12-12 at 2 50 29 PM

Screen Shot 2019-12-12 at 2 50 20 PM

Screen Shot 2019-12-12 at 2 50 46 PM


robdodson:
Adding @devnook as an additional reviewer.

samthor:
Is anyone actually going through 10 pages? Mostly I wonder if we should really just keep this simple and push people to use search 🤔

MichaelSolati:
> Is anyone actually going through 10 pages? Mostly I wonder if we should really just keep this simple and push people to use search 🤔

I'm not sure if anyone ever will, but I wanted to code in the condition now. Also, search would require dynamic page generation (or client side modifying the page), which we don't do yet. I do agree search would be good, but this seems to be the solution with our current tooling.

robdodson:
@samthor can you explain a bit more what kind of search you're imagining?

samthor:
This is mostly a drive-by review, so feel free to ignore. I just have
opinions™ that adding ten links on the homepage implies that we actually
want you to iterate through them, when blog content is about being
up-to-date, and that we already have a successful search component.

I would vote for just having "1 2 3 >>" and then writing the links all out
when you're on page 4, 5, 6, ... etc.

src/site/content/en/blog/index.njk
```diff
@@ -24,9 +28,33 @@ date: 2018-11-05
...
+        <a href="{{ pagination.previousPageHref }}">Previous</a>
+      </div>
+    {% endif %}
+
+    {%- for pageEntry in pagination.hrefs | paginate(page.url)  %}
+      <div class="w-blog__pagination-link">
+        {% if page.url == pageEntry.href %}
+          <span aria-current="page">{{ pageEntry.index }}</span>
+        {% else %}
+          <a href="{{ pageEntry.href }}">{{ pageEntry.index }}</a>
```
robdodson: Suggest adding `aria-label="Page {{pageEntry.index}}"`

src/site/content/en/blog/index.njk
```diff
@@ -24,9 +28,33 @@ date: 2018-11-05
...
+    {% if pagination.previousPageHref %}
+      <div class="w-blog__pagination-link">
+        <a href="{{ pagination.previousPageHref }}">Previous</a>
+      </div>
+    {% endif %}
+
+    {%- for pageEntry in pagination.hrefs | paginate(page.url)  %}
+      <div class="w-blog__pagination-link">
+        {% if page.url == pageEntry.href %}
+          <span aria-current="page">{{ pageEntry.index }}</span>
```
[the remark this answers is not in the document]
robdodson: Researching this a bit...

I think you can leave the current page as an anchor, but style it to look different and give it the `aria-current` state. If it's not an anchor, aria-current won't work.

src/site/_filters/paginate.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Take array of paginated hrefs and select the pages to display.
+ * @param {string[]} pages An array of hrefs of all the pages.
+ * @param {string} current Href of current page.
+ * @return {object[]} An array of up to 10 items to display, including href and index.
+ */
+module.exports = function paginate(pages, current) {
```
robdodson: I'm not sure I understand what this filter is for. Can you explain a bit more why we need it?

robdodson:
I'll throw in my 2 cents.

Personally, I like seeing `1 2 3 … 27` because it helps me quickly grok how much content there is. If we do `Previous 1 2 3 4 5 Next` then it seems like there's maybe only 5 pages worth of stuff.

I think limiting it to five items before we ellipsis feels like the right amount.

Maybe we'd add the previous / next buttons if they're on anything greater than the first page, for example: `Previous 2 3 4 5 6 … 27 Next`

Happy to continue discussing this.

devnook:
I agree with Rob - for me, seeing how many pages are there is important. I'd also prefer a symbol (>, >>) rather than words (next, prev) as it is one less language-dependent item on the page, wdyt?

src/site/_filters/paginate.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Take array of paginated hrefs and select the pages to display.
+ * @param {string[]} pages An array of hrefs of all the pages.
```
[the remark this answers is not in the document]
devnook: +1 to switch to Typescript

src/site/_filters/paginate.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {string} current Href of current page.
+ * @return {object[]} An array of up to 10 items to display, including href and index.
+ */
+module.exports = function paginate(pages, current) {
+  const currentIndex = pages.findIndex((page) => page === current);
```
devnook: Since page is a string, maybe consider indexOf instead of findIndex for less code? There's no need for a callback.

src/site/_filters/paginate.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (currentIndex < 0) return [];
+
+  let curated = pages.slice(0, 10);
+  const start = currentIndex > 5 ? currentIndex - 5 : 0;
+  const end = currentIndex + 5;
+
+  if (currentIndex > 5) curated = pages.slice(start, end);
```
devnook: It might be more readable, if you have this condition checked only once, eg:

let start = 0;
let end = 10;
if (currentIndex > 5) {
  start = currentIndex - 5;
  end = currentIndex + 5;
}
const curated = pages.slice(start, end);

You can also consider a more descriptive name than "curated", e,g, pagesToShow.

src/site/_filters/paginate.js
MichaelSolati: So it exists so that we do not display all the pages at the same time, just 10 pages at max 
(and it picks the 10 pages to show depending on the index of the page theyre currently on)

MichaelSolati:
Ok, so I've attempted to take all the feedback into account when making these updates. The templating and logic on the pagination filter are probably more complex, but here's the end result:

On the first page you'll see up to 8 pages, the 8th/last being the last page in number form. You'll also see a `>` for next page and `>>` for the last page:
Screen Shot 2019-12-16 at 7 52 08 PM

On the second page everything pretty much stays the same except there is now a `<<` for the first page. It made sense to me since a previous page button would take you to the first page, so why not just use a first page button:
Screen Shot 2019-12-16 at 7 52 18 PM

Now on the third page you'll have a `<` for the previous page, as well as the first, next, and last pages (this pretty much stays the same until we get towards the end):
Screen Shot 2019-12-16 at 7 52 27 PM

Now, I don't really want to keep explaining what's happening here, but I apply a similar logic as I did above to the next and last buttons as we get to the end (removing the ellipses when it is no longer needed):
Screen Shot 2019-12-16 at 7 53 11 PM
Screen Shot 2019-12-16 at 7 53 26 PM
Screen Shot 2019-12-16 at 7 53 36 PM
Screen Shot 2019-12-16 at 7 53 48 PM

src/site/_includes/components/Meta.js
```diff
@@ -20,7 +20,9 @@ const stripLanguage = require("../../_filters/strip-language");
 const {html} = require("common-tags");
 
 module.exports = (locale, page, collections) => {
```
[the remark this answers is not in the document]
MichaelSolati: Noted, could/should discuss to see if you'd like me to fix it.

src/site/_filters/paginate.js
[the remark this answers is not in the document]
MichaelSolati: I only saw this now! One more commit 😅

robdodson:
To tidy up the design a tad...

Let's remove the `<` and `>` buttons. I don't think they offer a ton of value.
Let's change the `<<` and `>>` buttons to use `&raquo;` and `&laquo;`.

src/styles/pages/_pagination.scss
```diff
@@ -0,0 +1,46 @@
+@import '../settings/colors';
+
+$W_PAGINATION_LINK_SPACING: 6px;
+
+@mixin w-pagination__links($COLOR, $ACTIVE: false) {
+  display: flex;
+  font-size: 20px;
+  justify-content: center;
+  min-width: 20px;
+  padding: 0px $W_PAGINATION_LINK_SPACING;
```
robdodson: I think we may need to increase this padding (perhaps adding some to the top and bottom as well) such that the tap target size is 40 x 40. I think right now it's around 23 x 32.

src/styles/pages/_pagination.scss
```diff
@@ -0,0 +1,46 @@
+@import '../settings/colors';
+
+$W_PAGINATION_LINK_SPACING: 6px;
+
+@mixin w-pagination__links($COLOR, $ACTIVE: false) {
+  display: flex;
+  font-size: 20px;
```
robdodson: I was wondering about switching this to `1rem` to make it a bit smaller and help it scale with the rest of the type. wdyt?

src/styles/pages/_pagination.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.w-pagination {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+  width: 100%;
+}
+
+.w-pagination > .w-pagination__active {
```
robdodson: I think the BEM style is that modifiers should have two dashes instead of underscores. So it would be:

`.w-pagination__link .w-pagination--active`

I realize that kind of blows up the mixin though. IMO it may be good to separate the modifier from the link styles in case we decide to add more modifiers in the future. It's also good to avoid unnecessary selector specificity if we can help it. So removing the `>` would be good too.

src/styles/pages/_pagination.scss
```diff
@@ -0,0 +1,46 @@
...
+$W_PAGINATION_LINK_SPACING: 6px;
+
+@mixin w-pagination__links($COLOR, $ACTIVE: false) {
+  display: flex;
+  font-size: 20px;
+  justify-content: center;
+  min-width: 20px;
+  padding: 0px $W_PAGINATION_LINK_SPACING;
+
+  a {
```
robdodson: We try to avoid styling generic selectors because it can lead to specificity battles later. I was thinking you could probably get rid of the outer `div` and make this anchor `.w-pagination__link`. The outer div doesn't seem to be adding much.

src/styles/pages/_pagination.scss
```diff
@@ -0,0 +1,46 @@
...
+  flex-wrap: wrap;
+  justify-content: center;
+  width: 100%;
+}
+
+.w-pagination > .w-pagination__active {
+  @include w-pagination__links($BLACK, true);
+}
+
+.w-pagination > .w-pagination__link {
```
robdodson: I think you can remove the child selector `>` to reduce specificity.

src/site/content/en/blog/index.njk
```diff
@@ -24,9 +28,61 @@ date: 2018-11-05
...
+    {%- for pageEntry in pagination.hrefs | paginate(page.url)  %}
+      <div
+        {% if page.url == pageEntry.href %}
+          class="w-pagination__active"
+        {% else %}
+          class="w-pagination__link"
+        {% endif %}
+      >
+        {% if pageEntry.showEllipses %}
+          <span class="w-pagination__spacer">...</span>
```
robdodson: Use ellipsis character `…` instead of three dots. It'll announce better for screen readers.

robdodson:
Hope it's ok, I started nerding out on the CSS for the arrows and did a variation that I'd like to get your thoughts on?

https://github.com/GoogleChrome/web.dev/tree/feat/blog-variation

image

Here's what I tweaked:

- Removed the `.w-paginatino__container` and made the anchors bigger to make the tap targets larger.
- Removed descendant selectors in the CSS. I think BEM style dictates not using descendants whenever possible to avoid specificity battles.
- Added material icon font for the next/prev arrows.
- Removed a bit of bottom padding on the grid so the space between the articles, the pagination, and the footer was even.

robdodson:
Awesooooooooome! 

giphy (3)

PR #1992, opened 2019-12-12
Prepare getInstalledRelatedApps() article for shipping.

jpmedley:
Fixes #1909 

Note: If this is written content for the site please include the issue number from your [content proposal](https://github.com/GoogleChrome/web.dev/issues?q=is%3Aissue+is%3Aopen+label%3A%22content+proposal%22).

Changes proposed in this pull request:

- Updates the article for leaving an OT in Chrome 80.

**This feature may not ship in Chrome 80. DO NOT MERGE WITHOUT WRITTEN CONFIRMATION FROM ME OR @petele.**

PR #2008, opened 2019-12-16
Correct a technical error.

jpmedley:
Changes the description of one of the suggested use cases. It is not currently possible for the native app to switch to the web app, but it is possible for web to switch to native.


jpmedley:
I cherry-picked this change to the pending Chrome 80 update(#1992)

PR #2022, opened 2019-12-18
Add CDS playlist to front page of site and deeplink id.

robdodson:
This was a request from @harleenkbatra08.

We'd like to run an ad campaign to direct folks to the embed on the frontpage so I've added an id as well.

Also had to switch away from the YouTube component because it doesn't support playlist embedding. This PR needs to go out today 😅

### DIFF
--- a/src/site/content/en/index.njk
+++ b/src/site/content/en/index.njk
@@ -110,13 +110,21 @@ date: 2018-11-05
     </a>
   </section> #}
 
-  <section class="w-cds-promo">
+  <section class="w-cds-promo" id="cds">
     <div class="w-cds-promo__grid">
       <div>
-        <img
-          src="/images/modules/cds19-lockup-stack.svg"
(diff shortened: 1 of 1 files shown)


src/site/content/en/index.njk
```diff
@@ -110,13 +110,21 @@ date: 2018-11-05
...
+  <section class="w-cds-promo" id="cds">
     <div class="w-cds-promo__grid">
       <div>
-        <img
-          src="/images/modules/cds19-lockup-stack.svg"
-          alt="chrome dev summit 2019"
-        >
+        <div class="w-cds-promo__video">
+          <div class="w-youtube">
+            <iframe
```
MichaelSolati: Simple question, do we not want to use the custom element @paulirish put out to keep the ifram from loading? I know we're lazy loading it, but wouldn't it still have a perf hit on lci?

PR #2027, opened 2019-12-19
Make deep links scroll properly.

### LINKED ISSUE #1976: Post headers hidden when navigated to via link
**Describe the bug**
When the user follows a link to a post section header, the header is hidden behind the top navigation bar. Without the header, it's less immediately clear that the link has taken the user to the intended location in the post. (This issue will be more pressing once ToCs are implemented.)

For example, here's the scroll position for `/how-to-measure-speed/#lab-data-vs-field-data`:
image

**To Reproduce**
1. Go to any post with `<h2>` elements.
2. Click on the link for an `<h2>` element.
4. See error.

**Expected behavior**
When the user follows a link to a post section header, the header should be visible. For example, here's what `/how-to-measure-speed/#lab-data-vs-field-data` should look like:
image

robdodson:
Fixes #1976

- Prevent top navigation from overlapping with headers when we deep link.

### DIFF
--- a/src/styles/generic/_shared.scss
+++ b/src/styles/generic/_shared.scss
@@ -7,6 +7,8 @@
 //
 // =============================================================================
 
+$WEB_HEADER_HEIGHT: 65px;
+
 // Set box-sizing globally to handle padding and border widths
 *,
 *::before,
@@ -68,7 +70,8 @@ var {
 *:not([href])[id]::before {
   content: ' ';
   display: block;
-  height: 65px;
-  margin-top: -65px;
+  height: $WEB_HEADER_HEIGHT;
+  margin-top: -#{$WEB_HEADER_HEIGHT};
+  pointer-events: none;
   visibility: hidden;
 }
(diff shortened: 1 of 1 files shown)


src/styles/generic/_shared.scss
```diff
@@ -62,3 +62,13 @@ var {
   -webkit-hyphens: none;
   hyphens: none;
 }
+
+// Ensures deep linking to headers has enough space to account for the
+// top navigation so they don't overlap.
+*:not([href])[id]::before {
```
samthor: Gross. I love it.

samthor: Several thoughts:

1) this should at least be `pointer-events: none` so it can't accidentally prevent clicks on something above it
2) can we use a sass variable for `65px` e.g. `$header-size`
3) could the pseudo-element just be a simple `position: absolute` element which is `top: -65px`? I suppose that then requires the heading to be `position: relative`, so, ... hm.
4) we could avoid all this by having the header disappear on scroll, but that's a big design change.

robdodson: I've added the first two suggestions. I wasn't sure about making headers disappear or making them all relative...

robdodson:
Oh it looks like u don't need to interpolate, just adding a `-` will do the thing. Updated!

PR #2034, opened 2019-12-20
Add icon to external links (replaces #1611)

mfriesenhahn:

- Add `launch` icon to external links using a `w-link--external` class.
- Add guidance to handbook to help ensure correct markup is used for external links until Markdown parsing is implemented.
- Fix spacing between links in the author lockup.

Notes:
- In #1611, @kaycebasques proposed using a selector that matches all absolute URLs. That proved very challenging to scope. (For example, I couldn't find a straightforward way to not match the links in the author lockup without referring to their class, which would create an inter-component dependency.)
- So, instead, I went with using a class, which is what /web does. The goal would be to have the Markdown parser add that class and `rel=noopener` to all absolute links.
- We should also add a rule to remark that disallows `//web.dev` to enforce relative internal links.
- Rather than a pseudo-element using the Material icon font, I went with an SVG background image for the link element. Here's why:
  - When the pseudo is set to `display: inline`, it inherits the anchor's `text-decoration`, which produces a thick, ugly underline beneath the icon.
  - Using `display: inline-block` fixes that problem but allows line wrapping between the last word of the link text and the icon.
  - Using `display:inline` and `border-bottom` instead of `text-decoration:underline` solves the wrapping and underline problems, but the border collides with code font styling.
- The downside of the background image approach is that we'll have to load different SVGs if we ever implement theming (e.g., with `prefers-color-scheme`). So, if anybody has a better idea, please share!

cc: @robdodson, @MichaelSolati

### DIFF
--- a/src/images/icons/launch.svg
+++ b/src/images/icons/launch.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M0 0h24v24H0z" fill="none"/><path fill="#3740ff" d="M19 19H5V5h7V3H5c-1.11 0-2 .9-2 2v14c0 1.1.89 2 2 2h14c1.1 0 2-.9 2-2v-7h-2v7zM14 3v2h3.59l-9.83 9.83 1.41 1.41L19 6.41V10h2V3h-7z"/></svg>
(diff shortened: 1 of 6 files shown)


mfriesenhahn:
You can see examples:
- In the [handbook](https://deploy-preview-2034--web-dev-staging.netlify.com/handbook/web-dev-components/#links-(external)) (in body copy)
- In [the `backdrop-filter` post](https://deploy-preview-2034--web-dev-staging.netlify.com/backdrop-filter) (in figcaptions)

kaycebasques:
@robdodson do we support the nice markdown syntax that was available on DevSite for adding classes to links?

    [DevTools](https://devtools.chrome.com){: .w-link--external }

robdodson:
@kaycebasques yes I believe that should work.

samthor:
I know this is explicitly in reaction to #1611 but I really think this is a huge burden on authors.

I actually think we should do this either in JS (just on the article content, so there's no oddities with CE's) or at the Markdown level.

Just my 2c.

mfriesenhahn:
Agree with @samthor: my original hope was that authors wouldn't have to think about whether a link is external and just use standard MD syntax.

robdodson:
I'm fine with that if anyone wants to take a shot at it?

PR #2052, opened 2020-01-06
Add self-assessments batch 1

mfriesenhahn:
- Add first four self-assessment sets in:
  - `control-focus-with-tabindex`
  - `use-semantic-html`
  - `labels-and-text-alternatives`
  - `same-origin-policy`
- Add horizontal scrolling when tabs overflow viewport. (Matches behavior of Material tabs, except scrollbar is visible to improve discoverability and usability for mouse users.)
- Add Markdown parsing to AssessmentHint `<summary>` element.
- Tidy editorial issues in `same-origin-policy` post.
- Tidy editorial issues in `codelab-explore-network-panel` post. (Was going to add a set to this post but decided against it after tidying.)
- Fix instruction list formatting in `defer-non-critical-css`. (Same as above.)

### DIFF
--- a/src/site/content/en/accessible/control-focus-with-tabindex/index.md
+++ b/src/site/content/en/accessible/control-focus-with-tabindex/index.md
@@ -142,6 +142,114 @@ You can learn more about them in our guide on
 [screen reader basics](/semantics-and-screen-readers).
 {% endAside %}
 
+{% AssessmentCallout 'Use the drop-down below each code sample to check your understanding of tab order.' %}
+{% Tabs 'Samples for knowledge self check' %}
+{% Tab 'sample' %}
+
+This HTML renders a modal dialog:
+```html
+<div role="dialog" aria-labelledby="dialog-header">
+  <button aria-label="Close"></button>
+  <h2 id="dialog-header">
+    Do you want to allow notifications from this website?
+  </h2>
+  <button>No</button>
+  <button>Yes</button>
+</div>
+```
+
+{% AssessmentHint 'Which elements from the sample are included in the tab order?' %}
+Only the `<button>` elements are included in the tab order
+because they're the only native HTML form elements.
+
+To insert other elements into the tab order, you would add a `tabindex` attribute.
+{% endAssessmentHint %}
+
+{% endTab %}
+{% Tab 'sample' %}
+
+```html
+<section tabindex="-1">
+  <h2>Cat facts</h2>
+  <ul>
+    <li>A group of cats is called a <a href="https://m-w.com/dictionary/clowder">clowder</a>.</li>
+    <li>Most cats are <a href="https://www.catfacts.org/catnip.html"> unaffected by catnip</a>.</li>
+  </ul>
+</section>
+```
+
+{% AssessmentHint 'Which elements from the sample are included in the tab order?' %}
+Only the `<a>` elements are included in the tab order.
+
+The `<section>` element is not in the tab order
+because it has a negative `tabindex` value.
+(It can, however, be focused using the `focus()` method.)
+{% endAssessmentHint %}
+
+{% endTab %}
+{% Tab 'sample' %}
+
+```html
+<form action="./cat-info.json">
+  <p>Where did you get your cat?</p>
+  <label>
+    <input type="radio" name="source" value="shelter">
+    An animal shelter
+  </label>
+  <label>
+    <input type="radio" name="source" value="petshop">
+    A pet shop
+  </label>
+  <label>
+    <input type="radio" name="source" value="breeder">
+    A breeder
+  </label>
+  <p>Describe your cat:</p>
+  <input type="text">
+  <button type="submit">Submit</button>
+</form>
+```
+
+{% AssessmentHint 'How many tab stops does the sample include?' %}
+The sample includes **three** tab stops:
+- One for the radio group
+- One for the text input
+- One for the **Submit** button
+
+Radio groups have a
+[roving tabindex](#create-accessible-components-with-"roving-tabindex")
+by default, so the entire group has only one tab stop.
+{% endAssessmentHint %}
+
+{% endTab %}
+{% Tab 'sample' %}
+
+This HTML renders a popup menu followed by a search input:
+
+```html
+<div role="menu" tabindex="0">
+  <a role="menuitem" href="/learn/" tabindex="-1">Learn</a>
+  <a role="menuitem" href="/measure/" tabindex="-1">Measure</a>
+  <a role="menuitem" href="/blog/" tabindex="-1">Blog</a>
+  <a role="menuitem" href="/about/" tabindex="-1">About</a>
+</div>
+<input tabindex="1" type="text" role="search" aria-label="Search" placeholder="Search">
+```
+
+{% AssessmentHint 'Which element in the sample comes first in the tab order?' %}
+The **Search** text input comes first in the tab order.
+Because it has a `tabindex` greater than 1, it jumps to the front of the tab order.
+
+(This behavior is likely to cause confusion
+  if the menu is positioned on the page before the search input.
+  This is an example of why `tabindex>0` is considered an anti-pattern.)
+
+{% endAssessmentHint %}
+
+{% endTab %}
+{% endTabs %}
+{% endAssessmentCallout %}
+
 ## Keyboard access recipes
 
 If you're unsure what level of keyboard support your custom components might
--- a/src/site/content/en/accessible/control-focus-with-tabindex/index.md
+++ b/src/site/content/en/accessible/control-focus-with-tabindex/index.md
@@ -169,27 +169,6 @@ To insert other elements into the tab order, you would add a `tabindex` attribut
 {% Tab 'sample' %}
 
 ```html
-<section tabindex="-1">
-  <h2>Cat facts</h2>
-  <ul>
-    <li>A group of cats is called a <a href="https://m-w.com/dictionary/clowder">clowder</a>.</li>
-    <li>Most cats are <a href="https://www.catfacts.org/catnip.html"> unaffected by catnip</a>.</li>
-  </ul>
-</section>
-```
-
-{% AssessmentHint 'Which elements from the sample are included in the tab order?' %}
-Only the `<a>` elements are included in the tab order.
-
-The `<section>` element is not in the tab order
-because it has a negative `tabindex` value.
-(It can, however, be focused using the `focus()` method.)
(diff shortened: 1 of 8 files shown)


robdodson:
LGTM

src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -142,6 +142,113 @@ You can learn more about them in our guide on
 [screen reader basics](/semantics-and-screen-readers).
```
kaycebasques: on a side note @robdodson I haven't read this guide and it was so dang good

src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -142,6 +142,113 @@ You can learn more about them in our guide on
...
+  <h2>Cat facts</h2>
+  <ul>
+    <li>A group of cats is called a <a href="https://m-w.com/dictionary/clowder">clowder</a>.</li>
+    <li>Most cats are <a href="https://www.catfacts.org/catnip.html"> unaffected by catnip</a>.</li>
+  </ul>
+</section>
+```
+
+{% AssessmentHint 'Which elements from the sample are included in the tab order?' %}
+Only the `<a>` elements are included in the tab order.
```
kaycebasques: I thought the answer would be "none of them" because the tabindex of the `section` was `-1` and the `li` tags were children of the `section`. I did quick scan of the guide and I don't believe this idea is covered.

Self-assessments already proving their ability to really enhance knowledge, v exciting!

Which brings up an interesting topic. How do we capture the various incorrect answers? Ideally I think our solution text should try to explain each possible answer, and why it was wrong. But how do we solicit those incorrect answers from our users?

cc @robdodson @mfriesenhahn @devnook

src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -142,6 +142,113 @@ You can learn more about them in our guide on
...
+  <div role="radio" aria-checked="false" tabindex="0">Persian</div>
+  <div role="radio" aria-checked="false" tabindex="-1">Bengal</div>
+  <div role="radio" aria-checked="false" tabindex="-1">Maine Coon</div>
+</div>
+```
+
+To keep things simpler, ignore the
+[`aria-*` and `role` attributes](/semantics-and-screen-readers) for now.
+
+{% AssessmentHint 'For a roving `tabindex`, what logic should run when a user presses the `Right` arrow key while a radio element is focused?' %}
```
kaycebasques: This rephrasing would have made the question easier to grok for me:

> Suppose you have implemented roving tabindex logic on the `role="radio" elements and you're currently focused on the first one. What should happen after you press the Right Arrow Key?

src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -142,6 +142,113 @@ You can learn more about them in our guide on
...
+- Change the `tabindex` values for all radio elements in the group to -1.
+- If there's a radio element after the one that's focused,
+  set its `tabindex` value to 0.
+- If there's no radio element after the one that's focused,
+  set the `tabindex` value of the first radio element in the group to 0.
+- Focus the radio element that now has a `tabindex` of 0.
+
+That's a lot—and it doesn't even include ARIA attributes!
+This is an example of why it's easier to use native elements
+with built-in keyboard behavior whenever you can.
```
kaycebasques: Dang, pretty effective way to demonstrate why custom tabindex logic is a hassle. Because as I was reading through the list I was like "this is complicated"

src/site/content/en/accessible/labels-and-text-alternatives/index.md
```diff
@@ -334,3 +334,105 @@ promotional offers?" like in the VoiceOver example below:
...
+{% AssessmentCallout 'Use the drop-down below each code sample to check whether the HTML element in the sample has an accessible name.' %}
+{% Tabs 'Samples for knowledge self check' %}
+{% Tab 'sample' %}
+
+```html
+<img src="temp.svg" alt="temp">
+```
+
+{% AssessmentHint 'Does this image have an accessible name?' %}
+**Yes**. Give an image an accessible name by adding an `alt` attribute.
```
kaycebasques: First impression: It's technicall correct but doesn't seem very descriptive?

src/site/content/en/accessible/labels-and-text-alternatives/index.md
```diff
@@ -334,3 +334,105 @@ promotional offers?" like in the VoiceOver example below:
...
+what text will be saved when the form is submitted,
+but assistive technologies don't read it.
+To provide an accessible name,
+add an `id` attribute with the value `subscribe`
+to associate the checkbox with the label.
+{% endAssessmentHint %}
+
+{% endTab %}
+{% endTabs %}
+{% endAssessmentCallout %}
```
kaycebasques: Got a couple of these wrong and learned some stuff!

I'll note that I didn't read the page's content and instead jumped straight to the self-assessment. Which honestly doesn't seem bad IMO. It leads me to an interesting idea. We could lead off articles with "how much do you know?" self-assessments, and the answers would have links that jump users to the explanations for the things they got wrong... pretty efficient way to learn... you skip over the stuff you already know, and focus on the stuff you get wrong

cc @mfriesenhahn @robdodson @devnook

src/site/content/en/accessible/use-semantic-html/index.md
```diff
@@ -109,6 +109,96 @@ The reason for this is that buttons and links are announced differently by
...
+Those pages may be separate resources in a static site,
+or they may be views in a
+[single-page application](https://developers.google.com/web/fundamentals/architecture/app-shell).
+Either way, since the user is navigating somewhere else, links are semantic.
+{% endAssessmentHint %}
+
+{% endTab %}
+{% endTabs %}
+{% endAssessmentCallout %}
+
```
kaycebasques: Love it

src/site/content/en/secure/same-origin-policy/index.md
```diff
@@ -54,49 +56,137 @@ cross-origin resource is blocked.
...
+var context = document.getElementById('bargraph').getContext('2d');
+var img = new Image();
+img.onload = function() {
+  context.drawImage(img, 0, 0);
+};
+img.src = 'https://example.com/graph-axes.svg';
+```
+
+{% AssessmentHint 'Can this image be drawn on the canvas?' %}
+**It depends!** 😅 The image is on a different origin.
```
kaycebasques: lol great use of an emojus

src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -142,6 +142,113 @@ You can learn more about them in our guide on
...
+</section>
+```
+
+{% AssessmentHint 'Which elements from the sample are included in the tab order?' %}
+Only the `<a>` elements are included in the tab order.
+
+The `<section>` element is not in the tab order
+because it has a negative `tabindex` value.
+(It can, however, be focused using the `focus()` method.)
+{% endAssessmentHint %}
```
kaycebasques: One thing that could be nice as we iterate is a thumbs-up / thumbs-down next to each answer to get feedback from users on whether it made sense

src/site/content/en/accessible/control-focus-with-tabindex/index.md
mfriesenhahn: 1st thing: This is actually an important issue. These are comprehension questions, so we shouldn't ask about concepts that aren't explicit in the parent article. @robdodson, what would you think about adding a para about this issue and maybe `inert`? Either that or I need to change the item.

2nd thing: Once we add multiple-choice formats, you're right, we'll need to write rationales for all options. I don't think we need to _solicit_ all possible responses from users though. The way one usually writes distractors (i.e., wrong answers) is to provide common misconceptions or other cognitive errors. If we think a correct response to a single item isn't likely to capture all the nuances of a particular issue, we should generally write another item. (I've tried to do that in these sets--for example, by asking about both same-domain and cross-domain links in the links-vs-buttons set.) To see how other people are handling rationales, check out [Khan Academy](https://www.khanacademy.org/) and/or [IXL](https://www.ixl.com/).

src/site/content/en/accessible/labels-and-text-alternatives/index.md
mfriesenhahn: What do you think would be more helpful?

src/site/content/en/accessible/labels-and-text-alternatives/index.md
mfriesenhahn: So, one note: once we start cranking out more sets, I do think reviewers should read the parent post first. If we ask questions about concepts not covered in the post, they're gonna seem unfair and are likely to lead to negative reactions. ("How was I supposed to know that? You didn't tell me!")

That said, what you're talking about is typically called pre-assessment or formative assessment. I'm a little gun shy about communicating to users that they don't need to read an article if they answer the questions correctly. To have high confidence that a user has achieved mastery of a domain of any complexity, you usually need them to answer a fair number of questions that you're pretty confident are all accurately measuring that domain. All of that is a non-trivial amount of work in the short term, and I don't want to make promises to our users that may not be true. However, in the long term, I think it'd be awesome to experiment with a larger item bank that can be organized into quizzes presented as peers to articles [the way Khan Academy does](https://www.khanacademy.org/humanities/grammar/parts-of-speech-the-noun).

src/site/content/en/accessible/control-focus-with-tabindex/index.md
mfriesenhahn: I'd love to have a mechanism for per-item feedback! You usually need some context to understand how a user is interpreting an item differently than you had intended, so maybe a link to a feedback form might be useful?

Sorry to cite Khan again, but they've figured out a lot of this stuff. Here's what they do:
image

And here's the form after you click the "Report a problem" link:
image

src/site/content/en/accessible/control-focus-with-tabindex/index.md
mfriesenhahn: Talked to @robdodson about issue #1 and will add a para in a separate PR.

src/site/content/en/accessible/control-focus-with-tabindex/index.md
mfriesenhahn: Addressing in a separate PR.

PR #2055, opened 2020-01-07
Register analytics event for elements with gc-analytics-event class

devnook:
As mentioned in the comment (https://github.com/GoogleChrome/web.dev/pull/1595#pullrequestreview-320230019), we should be bale to track events for other elements than links.

I'm using existing gc-analytics-event class to do that, while keeping tracking of all a#href clicks too. If we agree on this approach, I'll add the class to other trackable custom elements too. 

- 
- 
-

### DIFF
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -31,16 +31,18 @@ function trackEvent({category, action, label, value}) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


robdodson:
looks good to me, thank you @devnook!

PR #2056, opened 2020-01-07
Make sure ids are unique in custom elements

devnook:
Make sure ids are unique when Tabs custom element is used multiple times on a page. 

I added a salt generator utility to BaseElement to generate unique ids - let me know if there is a better place for it!


robdodson:
LGTM, tagging in both Michael's for them to also give it a look.

PR #2059, opened 2020-01-08
Bug/fix broken images

robdodson:
I noticed some images looked a little off on mobile so I made some changes. Maybe we should  enact these as image policies?

- If an author wants a specific width for an image, they should use the `width` attribute and ideally set it in a pixel value instead of a percentage. They *should not* use `style="max-width: ..."` because images are set to be max-width 100% by default and by overriding that value with something like max-width: 400px it causes the images to break out of the container.

### DIFF
--- a/src/site/_drafts/_template-post/index.md
+++ b/src/site/_drafts/_template-post/index.md
@@ -97,7 +97,7 @@ ac blandit ullamcorper. Proin et semper nibh, sit amet imperdiet velit. Morbi at
 quam sem.
 
 <figure class="w-figure">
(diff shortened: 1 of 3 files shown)


mihajlija:
Thanks for fixing this @robdodson! 

I checked the handbook and right now it actually says you _should_ use `style="max-width: ..."`, here's the section that needs updating: https://web.dev/handbook/markup-media/#styling-body-images

robdodson:
Thank you @mihajlija! I've also updated the handbook and template post to include this guidance.

PR #2060, opened 2020-01-08
Fix sticky hover and focus styles

mfriesenhahn:
iOS makes `:hover` styles stick while an element is focused, which was causing issues with the chevron animations in the Details and AssessmentHint components.

- Add a `hover()` mixin that can be used when components need to show hover states only on devices that support `:hover`.
- Apply the `hover()` mixin to the AssessmentHint, Details, Tabs, and Tooltip components. 
- Hide tab highlight color on AssessmentHint, CopyCode, Button, Details, and Tabs components so it doesn't interfere with our styling.
- Hide focus outline for AssessmentHint, CopyCode, Details, and Tabs components to match behavior of rest of site.
- Show tooltips on long press on mobile devices to better align with Material specs.
- Fixes to the CopyCode component:
  - Resequence active style in stylesheet so it actually appears.
  - Make focus outline round rather than square to match button shape.

**Note:** I didn't apply `hover()` to the following components, so hover is still sticky for them. With the possible exceptions of buttons and cards (for which you can easily activate hover state while swiping through a collection page), these elements navigate the user elsewhere, so sticky hover isn't a huge problem. I can add `hover()` if folks think it'd be helpful though.
- Buttons
- Action buttons
- Article nav buttons
- Cards
- Chips
- Cross-links
- Links

### DIFF
--- a/src/styles/components/_details.scss
+++ b/src/styles/components/_details.scss
@@ -22,6 +22,7 @@ $BORDER_COLOR: $GREY_300;
 }
 
 .w-details__summary {
+  -webkit-tap-highlight-color: transparent;
   cursor: pointer;
   flex-grow: 1;
   font: inherit;
@@ -39,13 +40,19 @@ $BORDER_COLOR: $GREY_300;
     outline: none;
   }
 
-  &:hover .w-details__header::after,
-  &:focus .w-details__header::after {
-    background: $LIGHT_BLUE_50;
+  @include hover() {
+    &:hover .w-details__header::after,
+    &:focus .w-details__header::after {
+      background: $LIGHT_BLUE_50;
+    }
   }
 
   &:focus .w-details__header::after {
     box-shadow: 0 0 0 1px $FOCUS_COLOR;
+  }
+
+  .js-focus-visible &:focus:not(.focus-visible) .w-details__header::after { // sass-lint:disable-line class-name-format
+    box-shadow: none;
   }
 
   &:active .w-details__header::after {
@@ -134,6 +141,9 @@ h2.w-details__header:only-child {
   display: none;
 }
 
-.w-details__summary:hover .w-details__preview::after {
-  color: $LINK_COLOR;
+@include hover() {
+  .w-details__summary:hover .w-details__preview::after,
+  .w-details__summary:focus .w-details__preview::after {
+    color: $LINK_COLOR;
+  }
 }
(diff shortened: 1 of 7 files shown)


src/styles/components/_details.scss
```diff
@@ -39,15 +40,21 @@ $BORDER_COLOR: $GREY_300;
...
+    &:focus .w-details__header::after {
+      background: $LIGHT_BLUE_50;
+    }
   }
 
   &:focus .w-details__header::after {
     box-shadow: 0 0 0 1px $FOCUS_COLOR;
   }
 
+  .js-focus-visible &:focus:not(.focus-visible) .w-details__header::after { // sass-lint:disable-line class-name-format
```
robdodson: Do you want to add the `.js-focus-visible` thing to other places where focus-visible is used in this PR?

mfriesenhahn: It wasn't working when used with web components (Tabs and CopyCode). Wasn't sure why. Help appreciated!

mfriesenhahn:
Hey, @robdodson and @MichaelSolati, the chevron animation on the Details and AssessmentHint components still isn't working on iOS. It doesn't play at all in Chrome and gets stuck about halfway through in Safari. Do y'all know why? I don't want to go down a rabbit hole chasing this one if you've already bumped into this.

robdodson:
> It doesn't play at all in Chrome and gets stuck about halfway through in Safari.

Are you referring to Chrome running on iOS? Essentially that's still Safari (it uses the webkit rendering engine). I haven't run into this particular issue before but let me check on my iOS device to reproduce it first. Might want to see if someone like Adam Argyle has any thoughts if you can point him at a running example.

robdodson:
I noticed the animation does play if you're long pressing the arrow, but it's still very fiddly.

mfriesenhahn:
> Are you referring to Chrome running on iOS? Essentially that's still Safari (it uses the webkit rendering engine). 

Right, though the behavior was slightly different.

> I haven't run into this particular issue before but let me check on my iOS device to reproduce it first. Might want to see if someone like Adam Argyle has any thoughts if you can point him at a running example.

I'll link him to the staged version and see if he has suggestions.

mfriesenhahn:
After some investigation I just killed the animation. It doesn't seem worth fussing over.

src/styles/components/_details.scss
robdodson: Did you try:

```
.js-focus-visible .web-code-copy ...
```

?
```

mfriesenhahn: Fixed. I apparently can't read nested rules... <sigh>

PR #2067, opened 2020-01-09
Address issues in 1st self-assessment sets

mfriesenhahn:

- Fixes collisions between `<summary>` text and the chevron hover state in Details and AssessmentHint widgets.
- Addresses feedback from @kaycebasques on PR #2052:
  - Added a paragraph to `control-focus-with-tabindex` to make explicit that adding a `tabindex` value to an element doesn't affect its children.
  - Reframed item 4 in the `control-focus-with-tabindex` set to hopefully (!) clarify the task.

### DIFF
--- a/src/styles/components/_details.scss
+++ b/src/styles/components/_details.scss
@@ -4,6 +4,7 @@
 @import '../tools/mixins';
 
 $CHEVRON_SIZE: 24px;
+$CHEVRON_MARGIN: 12px;
 $BORDER_COLOR: $GREY_300;
 
 .w-details {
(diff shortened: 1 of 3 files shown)


src/styles/components/_details.scss
```diff
@@ -58,7 +58,7 @@ $BORDER_COLOR: $GREY_300;
 .w-details__header {
   color: $LIGHT_BLUE_600;
   margin: 0;
-  padding: .75rem #{$CHEVRON_SIZE + 12px} .75rem 0;
+  padding: .75rem #{$CHEVRON_SIZE + 2px + 12px} .75rem 0;
```
robdodson: Initially I was thinking of changing this to `+ 14px` but I figured you were explicit about the values to hint to the reader that they relate to other parts of the code, is that correct?

mfriesenhahn: Yeah, it's padding + font-size for the chevron pseudo-element. (Padding changes based on media query to make this fit better on feature phones.) I'll comment it.

PR #2074, opened 2020-01-10
Switch to NetworkFirst sw strategy.

### LINKED ISSUE #2069: Stale-While-Revalidate isn't the right strategy for content pages
I just fell for a stale article and wasted time updating it until I realized that I had a stale version of it in my cache.

I *strongly* suggest instead of the current [`new workbox.strategies.StaleWhileRevalidate()`](https://github.com/GoogleChrome/web.dev/blob/a5102ed4e19d14ebafe0976fbcdb7a709a7a2cd9/src/lib/sw.js#L119-L124) we use [`new workbox.strategies.NetworkFirst()`](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#network_first_network_falling_back_to_cache).

robdodson:
Fixes #2069

- Changes most sw fetches to be NetworkFirst
- Leaves Google fonts as StaleWhileRevalidate

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -118,7 +118,7 @@ const untrailedContentPageRe = new RegExp("^(/[\\w-]+)+$");
  */
 workbox.routing.registerRoute(
   contentPageRe,
-  new workbox.strategies.StaleWhileRevalidate({
+  new workbox.strategies.NetworkFirst({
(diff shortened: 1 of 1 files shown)


robdodson:
Note, I didn't rev `const serviceWorkerArchitecture = "v2";` because I wasn't sure if this constituted enough of a change to force clients to claim. I figured it could happen organically.

PR #2077, opened 2020-01-13
Changing subscribe fabs as per design

### LINKED ISSUE #1191: Make FABs take up less space on mobile. Eng work for #953

devnook:
Fixes #1191 

- Move subscribe FAB under the article on mobile
- Still need to add animation to the share button (separate pr)

### DIFF
--- a/src/site/_includes/partials/post.njk
+++ b/src/site/_includes/partials/post.njk
@@ -81,6 +81,16 @@
       <a href="#">Suggest translation</a> #}
     </div>
 
+    <a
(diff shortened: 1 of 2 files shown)


samthor:
* I don't know what the design suggests but can we remove the underline from the button? it looks more like a link rather than a CTA.

* is it worth calling out why it's in the .njk file and not in our component?

* Unrelated, but the `src/site/_includes/components/Actions.js` file seems to now be disused in favor of a web component. Want to remove that file?

PR #2084, opened 2020-01-15
Synchronize tags in fugu posts.

jpmedley:
Makes the tags for fugu/capabilities consistent across the blog.

### DIFF
--- a/src/site/content/en/blog/maskable-icon/index.md
+++ b/src/site/content/en/blog/maskable-icon/index.md
@@ -13,6 +13,7 @@ alt: Icons contained inside white circles compared to icons covering its entire 
 tags:
   - post
   - capabilities
+  - fugu
   - progressive-web-apps
 ---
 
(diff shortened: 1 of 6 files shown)


robdodson:
I recall a while back hearing that we didn't want to publicly use the term `fugu` (although I realize it's used in the chromium bug tracker). Is the policy still that we don't want to rely on that term? Would it be possible to put all of these under the `capabilities` tag?

jpmedley:
I don't know about policy, but I think that ship has sailed. Not only did we give away Fugu stickers at both IO and CDS, Pete has previously published multiple articles with both a fugu and a capabilities tag. This PR just makes the tagging consistent across all Fugu articles. Multiple members of the Fugu subgroup have even [tweeted the name](https://twitter.com/search?q=chrome%20fugu&src=typed_query).

kaycebasques:
Thanks for providing the paper trail, @jpmedley

PR #2095, opened 2020-01-18
Disable pointer events on action button children

robdodson:
I noticed the share button wasn't working if you clicked on the span text inside of it. I think it's because it uses `e.target.href` to open the little share window.

- Disable pointer events on action button children.


robdodson:
Btw I chose to go with `.w-actions__fab *` instead of `.w-actions__fab span` or giving the span a unique class like `.w-actions__fab-text` because I wanted to future proof things in case we ever add more stuff to one of these buttons. But if that's gross then I'm happy to give the span a class name and just refer to it that way.

PR #2101, opened 2020-01-24
Clarify CrUX bucketing for CLS

philipwalton:
This PR clarifies the difference between how CrUX and the Layout Instability API report CLS. The ambiguity was brought to my attention via this tweet: https://twitter.com/philwalton/status/1220798835754061824.

### DIFF
--- a/src/site/content/en/metrics/cls/index.md
+++ b/src/site/content/en/metrics/cls/index.md
@@ -264,8 +264,8 @@ To calculate the cumulative layout shift score for your pages, declare a
 variable that stores the current cumulative layout shift score, and then
 increment it any time a new, unexpected layout shift is detected.
 
-For consistency with how the [Chrome User Experience
-Report](https://developers.google.com/web/tools/chrome-user-experience-report)
+For consistency with how the [Chrome User Experience Report
+(CrUX)](https://developers.google.com/web/tools/chrome-user-experience-report)
 measures CLS, stop observing scores when the page's [lifecycle
 state](https://developers.google.com/web/updates/2018/07/page-lifecycle-api)
 changes to hidden:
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/cls/index.md
```diff
@@ -307,8 +307,14 @@ Ideally, pages would not have any layout shifts, so a CLS of `0` is the goal.
 Realistically, in some cases it's impossible to avoid 100% of layout shifts, so
 minimizing what you can is best.
 
-The Chrome Use Experience Report categorizes any page with a CLS less than 5 to
-be considered "small" (i.e. good).
+CrUX categorizes any page with a CLS less than `0.05` to be considered "small"
```
igrigorik: What do we actually mean by this? We don't assign labels in CrUX.. =/

philipwalton: I guess when I said CrUX I should have said the CrUX Dashboard, which does assign labels. My understanding was it was going to use the <5 bucket for "small", though now I can't find where I originally saw that. @rviscomi any ideas? If we haven't assigned a label yet, I can remove this (though I think the note is still helpful to explain that CrUX buckets are percentages).

igrigorik: Ah. If memory serves, <5 is because that bucket covers the 0-5 range and ideally CLS is 0 — that's definitely a good place to be. That said, given that this is not our official guidance for CLS thresholds, I would probably suggest we drop this part, but keep the rest.

philipwalton: Sure. I removed the section for now. We can re-add it once we've finalized our recommendations.

adamrights:
Began adding to WSJ pages and my dashboard was showing everything as fast! Dug into CRuX and realized it must have been 0.05. Of course there's a PR already. Glad to see will be updated soon, I would assume it's throwing some people off.

philipwalton:
> Has the recommendation been removed because these categories are still in flux (even though represented on CRuX materialized since May)?

Yes, we're in the process of finalizing our recommendations, so it made sense to remove any unofficial recommendations until we had something official.

philipwalton:
@kaycebasques could we get this PR merged? Looks like this is causing some confusion for a few people, so it'd be great to get the update published.

robdodson:
@kaycebasques if it's ok I'm happy to approve and merge this. Seems like a small change.

PR #2128, opened 2020-02-07
Center misaligned image

robdodson:

- Centers an image that was left aligned.


robdodson:
Going to merge as it's a small change. This does make me think that perhaps we should just default all images to being center aligned...

PR #2129, opened 2020-02-07
Add handler for serving language redirections based on preferred locale

devnook:
Add a handler that redirects the request based on requested locale, according to priority:
 1. Locale directly expressed in the url, e.g. /en/some-post
 2. Locale in accept-language HTTP header.
 3. Locale in "preferred_locale" cookie.
 If requested locale's content does not exist, falls back to default ('en').

Note for testing:
This PR makes more sense when combined with the PR #2123 as you need some "localised" content, e.g. /pl/ directory to test the handler.


locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+const DEFAULT_LOCALE = "en";
+
+const isSupported = (locale) => SUPPORTED_LOCALES.indexOf(locale) > -1;
+
+/**
+ * A handler that redirects the request based on requested locale,
+ * according to priority:
+ * 1. Locale directly expressed in the url, e.g. /en/some-post
+ * 2. Locale in accept-language HTTP header.
+ * 3. Locale in "preferrec_locale" cookie.
```
samthor: This should be "preferred_lang"

locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+ * If requested locale's content does not exist, falls back to default.
+ * @param {Object} req Request object.
+ * @param {Object} res Response object.
+ * @param {Function} next Middleware next handler.
+ * @return {!Function}
+ */
+module.exports = (req, res, next) => {
+  // Check if url is navigational.
+  if (req.url.endsWith("/")) {
```
samthor: style nit: this block always returns, and then we have a tiny alternative case below (`return next();`). Can we flip this around?

```js
  if (!req.url.endsWith("/")) {
    return next();
  }
  // less indented code
```

locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+    if (lang === DEFAULT_LOCALE) {
+      // If this is alread default language, continue.
+      return next();
+    }
+
+    const localizedFilePath = path.join(
+      __dirname,
+      "dist", // Must serve from dist directory even in dev mode.
+      lang,
```
samthor: `lang` is untrusted. I could set my `preferred_lang` cookie to "../../escape/our/root/" and Node will join it.

robdodson: are you thinking it should be sanitized with a `isSupported()` check?

locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+  if (req.url.endsWith("/")) {
+    // Check if language is specified in the url.
+    const pathSegment = req.path.split("/")[1];
+    const isLangInPath = isSupported(pathSegment);
+    let lang;
+    let filePath;
+
+    if (isLangInPath) {
+      lang = pathSegment;
+      const pathParts = req.path.split("/");
```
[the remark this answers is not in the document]
devnook: done

locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const fs = require("fs");
+const localeCode = require("iso-639-1");
+const path = require("path");
+
+const isProd = Boolean(process.env.GAE_APPLICATION);
+const contentDir = isProd ? "dist" : "src/site/content";
```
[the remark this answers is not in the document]
devnook: At this point of build process /dist might not be populated yet.

locale-handler.js
```diff
@@ -0,0 +1,83 @@
...
+    const localizedFilePath = path.join(
+      __dirname,
+      "dist", // Must serve from dist directory even in dev mode.
+      lang,
+      filePath,
+      "index.html",
+    );
+
+    if (fs.existsSync(localizedFilePath)) {
+      return isLangInPath ? next() : res.redirect(`/${lang}${filePath}`);
```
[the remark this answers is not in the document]
devnook: done

locale-handler.js
devnook: done

locale-handler.js
devnook: done

locale-handler.js
devnook: I whitelist allowed values in isSupported. Do you think there should be some additional check here?

locale-handler.js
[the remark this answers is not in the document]
robdodson: I'm not sure I understand why? I don't think we pull from `src/` for any other part of the server do we? I thought we usually ran `build` before starting the server.

locale-handler.js
robdodson: I see. It looks like line 57 sanitizes this value I think.

locale-handler.js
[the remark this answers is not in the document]
devnook: I based this assumption on this comment in server.js: https://github.com/GoogleChrome/web.dev/blob/master/server.js#L24  I think there is no guarantee that build runs before the server starts (e.g. if you run npm run start in stead of npm run dev)

Issue #1808, opened 2019-10-31
Add support for tags.


kaycebasques:
> In a book index, which shares underlying theory with tags, one of these would have 'see the other` instead of a page number. Synonyms are a must for indexes and tags because it's a way to account for a reader having a different term in their head than what's used in the article.

--- @jpmedley

https://github.com/GoogleChrome/web.dev/pull/1726/files#r340843948

robdodson:
Can you explain a bit more what you'd like to see here?

kaycebasques:
* When browsing tags, users might search for different synonyms. Some might search for `vr` whereas others might search for `virtual-reality`, so it'd be good to support both. This is what Joe was getting at.
* Suppose author A only tags their post with `vr`, whereas author B only tags their post with `virtual-reality`. Without a tag synonym infrastructure, when a user browses the posts tagged with `vr`, they won't actually see our full list of VR content.
* We therefore need some sort of infrastructure that will enable us to specify that a certain tag is synonymous with other tags. So when a user visits the `vr` tag they'll also see the posts that are tagged with `virtual-reality`.

First structure that comes to mind:

```
tags: {
  vr: {
    synonyms: [
      'virtual-reality'
    ]
}
```

jpmedley:
VR is probably the worst example of this. This is one of the few paper conventions that translates directly to the web, but never has been.

kaycebasques:
@jpmedley am I describing the situation correctly? Please elaborate if not.

jpmedley:
You've captured it.

robdodson:
> When browsing tags, users might search for different synonyms. Some might search for vr whereas others might search for virtual-reality, so it'd be good to support both. This is what Joe was getting at.

At the moment we don't have designs for tag search, though [we do have designs for tags UI](https://github.com/GoogleChrome/web.dev/issues/1188#issuecomment-548195437). That UI assumes a fixed set of tags.

Currently we use [Algolia](https://www.algolia.com/) to power the search bar at the top of the site. Aloglia is super powerful and we're just using a tiny fraction of what it provides. It might be possible to offload the synonym stuff to it somehow.

> Suppose author A only tags their post with vr, whereas author B only tags their post with virtual-reality. Without a tag synonym infrastructure, when a user browses the posts tagged with vr, they won't actually see our full list of VR content.

[...]

PR #2139, opened 2020-02-07
Add a feature card class to 1st post on homepage.

### LINKED ISSUE #2028: Homepage wraps top three articles in a weird way on tablet/small desktop.
image

Our three articles are fixed size cards so they wrap on medium sized displays. A better approach might be to make the top card span both columns on smaller screens.

robdodson:
Fixes #2028 

On the homepage, when you resize to a smaller display, the recent posts cards wrap in a way that looks broken. This resolves the issue by changing the layout for the most recent card so it's in more of a featured position.

This is a short term fix, but really I'd like us to redesign the homepage at some point 👨‍🎨

Since a lot of work has been done on #1808, I think we should wait until that PR lands before merging this one. I can rebase against master after #1808 lands and clean things up in this PR. This PR is read for review though, so PTAL!

**Before**
image

**After**
image

- Changes the `recentPosts` collection so it ensures all posts have a thumbnail
- Adds a `featured` argument to the `PostCard` shortcode so it can apply a featured modifier class.
- Adds a featured grid layout.
- Adds a featured variation to the PostCard styles.
- Changes the `breakpoints.scss` so it's easier to use the actual breakpoint values in cases where you might want to do a max-width breakpoint.

### DIFF
--- a/src/styles/components/_post-card.scss
+++ b/src/styles/components/_post-card.scss
@@ -12,6 +12,7 @@
 
 .w-post-card {
   @include w-body-text();
+  height: 100%;
   padding: 24px;
   position: relative;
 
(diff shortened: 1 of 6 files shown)


src/styles/components/_post-card.scss
```diff
@@ -151,3 +152,36 @@
...
+
+    .w-post-card__blurb {
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      padding-left: 64px;
+    }
+
+    .w-post-card__desc::before {
+      display: none;
```
samthor: it took me a while to work out what this was (the blue underline that indicates it's a link at small sizes). I don't really understand why we need to overwrite it again but I'm happy with it. Maybe a comment?

robdodson: good call. I'll see if I can remove it.

robdodson: turns out i can't! but added a comment

PR #2145, opened 2020-02-11
Add Percy snapshot testing.

robdodson:
Adds support for http://percy.io/ snapshot testing. Currently testing specific landing pages. After landing this PR I'd like to do a follow up to let us test our UI library (web components, shortcodes, etc) but that'll take more work.

- Filter post collections based on test env.
- Add a new GitHub workflow to run Percy tests.
- Set cookie banner to accepted if we're in a test or dev env.

@mfriesenhahn @kaycebasques fyi


robdodson:
> I presume this just warns us of the screenshot change delta :)

yep, it's quite fancy. I sent you an invite to view the project. It takes shots at a couple different screen sizes and grumps if there are changes. For now i'm just testing Chrome at two sizes. We get 5000 shots a month for free.

src/lib/store.js
```diff
@@ -42,7 +42,9 @@ const initialState = {
 
   // When a user lands on the page, check if they have accepted our
   // cookie policy.
-  userAcceptsCookies: false,
+  // We disable this check in dev and test environments so the cookie banner
+  // doesn't interfere with testing.
+  userAcceptsCookies: config.prod ? false : true,
```
samthor: super big nit but I think this might be more readable as

```js
   userAcceptsCookies: !config.prod,
```

samthor: and update the comment to have fewer negatives, so e.g. "We automatically accept cookies anywhere but prod so that ..."

robdodson:
@MichaelSolati PTAL when you get a chance

PR #2151, opened 2020-02-12
add Firestore backup workflow

samthor:
We could target e.g. /backups/YYYY-MM-DD, but it would be a more complex workflow (have to set an env var with `date "+%Y-%m-%d"`, then pass it to the gcloud command, ...).

This will work for now.

### DIFF
--- a/.github/workflows/snapshot-firestore.yml
+++ b/.github/workflows/snapshot-firestore.yml
@@ -1,0 +1,21 @@
+name: Snapshot firestore
+on:
+  schedule:
+    # At 19:00 UTC / 7:00 PST on every day-of-week from Monday through Thursday.
+    - cron: '0 19 * * 1-4'
+
+jobs:
+  ci:
+    name: Run gcloud to snapshot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Snapshot
+        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: '270.0.0'
+          service_account_email: ${{ secrets.GCLOUD_EMAIL }}
+          service_account_key: ${{ secrets.GCLOUD_KEY }}
+
+      # nb. gcloud uses a new timestamp'ed dir only when pointing at a top-level bucket. Thus, we
+      # can't naïvely target a subfolder (it will fail with "Path already exists").
+      - run: gcloud firestore export gs://web-dev-production-1.appspot.com
--- a/.github/workflows/snapshot-firestore.yml
+++ b/.github/workflows/snapshot-firestore.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Snapshot
-        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.1
         with:
           version: '270.0.0'
           service_account_email: ${{ secrets.GCLOUD_EMAIL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot-firestore.yml
```diff
@@ -0,0 +1,21 @@
...
+    # At 19:00 UTC / 7:00 PST on every day-of-week from Monday through Thursday.
+    - cron: '0 19 * * 1-4'
+
+jobs:
+  ci:
+    name: Run gcloud to snapshot
+    runs-on: ubuntu-latest
+    steps:
+      - name: Snapshot
+        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
robdodson: Would you mind updating this and the continuous-integration workflow to use the 0.1.1 tag
https://github.com/GoogleCloudPlatform/github-actions/releases

I think that's what we're supposed to do after looking at the OSPO policy today.

.github/workflows/snapshot-firestore.yml
```diff
@@ -0,0 +1,21 @@
...
+        with:
+          version: '270.0.0'
+          service_account_email: ${{ secrets.GCLOUD_EMAIL }}
+          service_account_key: ${{ secrets.GCLOUD_KEY }}
+
+      # nb. gcloud uses a new timestamp'ed dir only when pointing at a top-level bucket. Thus, we
+      # can't naïvely target a subfolder (it will fail with "Path already exists").
+      - run: gcloud firestore export gs://web-dev-production-1.appspot.com
```
robdodson: So this will generate new subdirectories like:

web-dev-production-1.appspot.com/2020-02-12T05:11:17_63121/

Not opposed, just want to make sure I understand.

.github/workflows/snapshot-firestore.yml
```diff
@@ -0,0 +1,21 @@
+name: Snapshot firestore
+on:
+  schedule:
+    # At 19:00 UTC / 7:00 PST on every day-of-week from Monday through Thursday.
+    - cron: '0 19 * * 1-4'
```
robdodson: Btw I think there's a way to set a GCS bucket rule where it deletes stuff after a certain amount of time. Do you want to add a rule to clean out this bucket after like 15 days so we don't have to remember to do it manually?

.github/workflows/snapshot-firestore.yml
robdodson: Also, should we make a new top level bucket for this to go into? firestore-backup.web-dev-production-1.appspot.com maybe?

.github/workflows/snapshot-firestore.yml
samthor: Done

.github/workflows/snapshot-firestore.yml
samthor: I've created `web-dev-production-firestore-archive`, owned by our production project.

.github/workflows/snapshot-firestore.yml
samthor: I've set it to 90 days. Partially because the bucket itself is Nearline, which penalizes you if you delete before 30 days.

I've also changed this to just do a weekly backup on Mondays.

samthor:
I wonder if I should document these buckets somewhere.

robdodson:
> I wonder if I should document these buckets somewhere.

Probably a good idea. I spent a few days discovering that app engine creates (and auto cleans) the `staging.*` bucket and the `us.artifacts.*` bucket. It stores build images in the artifacts bucket but they're safe to delete. It'd be good to document the rest. You can do it on our internal team page?

PR #2153, opened 2020-02-12
feat(authors): add pages for each authors posts

MichaelSolati:
Works in much the same way as the recent tags pr #1808 but it provides pagination based on author.


MichaelSolati:
@samthor @robdodson PTAL

src/site/_includes/components/Author.js
```diff
@@ -43,10 +43,10 @@ module.exports = ({
     <div class="w-author">
       <img
```
robdodson: What do you think about also wrapping the author image in an anchor? I'd suggest adding tabindex=-1 to it so we don't create too many tab stops for folks navigating via keyboard or switch device.

src/site/_includes/components/AuthorInfo.js
```diff
@@ -73,7 +73,9 @@ module.exports = ({post, author, showSocialMedia = false}) => {
       class="w-author__info"
       style="display: flex; flex-direction: column; justify-content: center;"
     >
-      <cite class="w-author__name">${fullName}</cite>
+      <cite class="w-author__name">
+        <a href="/authors/${id}">${fullName}</a>
```
robdodson: Linking the author's name on the post card and the page itself changes the appearance. I think we should write some some CSS to preserve the previous look, but add a :hover state that doubles as the :focus state by combining the selectors:

```
:hover,
:focus {
  // probably let the items turn blue
}

**Comment by robdodson** [comment]
Looks cool and is a very nice feature! Awesome work on this :D

File: src/site/_includes/components/AuthorInfo.js (hunk 2 above)
Comment (MichaelSolati) [reply src/site/_includes/components/AuthorInfo.js]: So this affects a posts page, I can change the card, but how should I address a page?

File: src/site/_includes/components/Author.js (hunk 1 above)
Comment (MichaelSolati) [reply src/site/_includes/components/Author.js]: Should this be added to cards as well? The second author link may be hard to click, but I figured I'd ask.

Comment (robdodson) [reply src/site/_includes/components/Author.js]: yep i think it should. I agree the 2nd author one will be a little hard to click, but they can click on the name as well

File: src/site/_includes/components/AuthorInfo.js (hunk 2 above)
Comment (robdodson) [reply src/site/_includes/components/AuthorInfo.js]: Maybe on the card we don't have a hover state at all (since we have one big one for the card itself). But we would have a :focus state which underlines the text (similar to how the focus state for the card heading works).

On the post page I would have it work similar to the breadcrumbs where it's grey but gets an underline on hover. And a blue outline + grey underline on focus.

image

File: src/site/_collections/paginated-posts-by-tag.js [hunk 3]
```diff
@@ -25,7 +25,7 @@ const postDescending = require("./post-descending");
  * This is because we can not paginate something already paginated... Pagination is effectively a loop, and we can't have an embedded loop O^2.
  *
  * @param {any} collection Eleventy collection object
- * @return {Array<{ title: string, href: string, description: string, tag: string, posts: Array<object>, index: number, pages: number }>} An array where each element is a paged tag with some meta data and n posts for the page.
+ * @return {Array<{ title: string, href: string, description: string, posts: Array<object>, index: number, pages: number }>} An array where each element is a paged tag with some meta data and n posts for the page.
  */
 module.exports = (collection) => {
```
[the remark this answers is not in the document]
MichaelSolati: Good point, will change

MichaelSolati:
@robdodson made the changes you recommended, PTAL

src/site/_collections/paginated-posts-by-author.js
```diff
@@ -0,0 +1,54 @@
...
+  posts.forEach((post) => {
+    const authors = post.data.authors || [];
+    authors.forEach((author) => {
+      const postsByAuthor = mapValue(authorsMap, author);
+      postsByAuthor.push(post);
+      authorsMap.set(author, postsByAuthor);
+    });
+  });
+
+  let authors = [];
```
samthor: const

src/site/_utils/map-value.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Returns value from map, or default if no value found.
+ *
+ * @param {Map<string, any>} map Map to get value from.
+ * @param {string} key Key to get value from map.
+ * @param {any} def Default value to return.
+ * @return {any} Value found, or default if no value returned.
+ */
+module.exports = (map, key, def = []) => {
```
samthor: FWIW this is a good idea but Python for example has [.setdefault](https://docs.python.org/2/library/stdtypes.html#dict.setdefault) which is basically more convenient.

```js
setDefault(map, foo, []).push(123);
```

src/site/_includes/components/Author.js
```diff
@@ -41,12 +41,14 @@ module.exports = ({
   const fullName = `${author.name.given} ${author.name.family}`;
   return html`
     <div class="w-author">
-      <img
-        class="w-author__image ${small && `w-author__image--small`}"
-        src="/images/authors/${avatar}.jpg"
-        alt="${fullName}"
-      />
-      ${AuthorInfo({post, author, showSocialMedia})}
+      <a href="/authors/${id}" tabindex="-1">
```
samthor: Is there a reason people can't tab to us? :(

robdodson: ah that was my mistake. I think I was trying to cut down on tab targets within the card since the image, headline, and body text all link to the same resource.

but yeah, we should remove the tabindex here because authors link to a completely unique url and that needs to be keyboard accessible.

src/site/_collections/paginated-posts-by-author.js
MichaelSolati: `authors` gets reassigned on line 50.

PR #2154, opened 2020-02-12
fix(Meta): use `path.title` and `path.description` in `getMetaByPlatform`, fixes #2002

### LINKED ISSUE #2002: Collection pages have empty meta data
**Describe the bug**
In the dist directory, take a look at at the fast/index.html. Notice most of the meta data is blank. This is true for all collections pages. We probably need to update either the layout.njk or Meta.js to get data for these pages and populate it correctly. The data itself should live in the collection's 11tydata.js file.

MichaelSolati:
Fixes #2002.


MichaelSolati:
@robdodson PTAL

robdodson:
Nice! Thank you for working on this 🙌

PR #2157, opened 2020-02-12
feat(SignPost): add SignPost for pages in collections, fixes #1193

### LINKED ISSUE #1193: Add sign posts so users knows when a post is part of a collection. Eng work related to #973

MichaelSolati:
Fixes #1193

Adds links to collections a page belongs to and removes the breadcrumbs (except on blog posts, those keep bread crumbs).

[This is best seen here](https://deploy-preview-2157--web-dev-staging.netlify.com/is-on-https/)


robdodson:
I noticed on desktop the signpost has a bit too much space between it and the heading. Might also be good to confirm it's 88px below the hero image.

image

Also looks like it has some left margin that isn't needed. Should be left justified with the heading.
image

Also need to test the layout on mobile. Here's what it should look like:
image

But I think it's losing its padding:
image

src/site/_includes/partials/post.njk
```diff
@@ -14,9 +14,16 @@
   {# Make sure Actions are first in the tab order. #}
   <web-actions actions="share|subscribe" authors="{{ authors | expandContributors | join('|') }}"></web-actions>
 
-  <div class="w-layout-container--narrow w-post-breadcrumbs">
-    {% Breadcrumbs path %}
-  </div>
+  {% if path.slug === 'blog' %}
```
robdodson: May want to wrap all this stuff in a `w-layout-container`

src/site/_collections/post-to-collections.js
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module.exports = (collections) => {
+  const postToCollections = {};
+  const paths = collections
```
robdodson: Suggest renaming this to `pathItems`. At a glance I thought this was referring to the 11tydata files for each path but then realized it was referring to every post with the path-item tag.

src/site/_collections/post-to-collections.js
```diff
@@ -0,0 +1,46 @@
...
+    .getFilteredByTag("pathItem")
+    .map((p) => p.data.path);
```
robdodson: I think the array that this returns is larger than it needs to be. I think we have <20 paths, but this returns an array of a few hundred items which are just the same path object duplicated in many instances.

src/site/_collections/post-to-collections.js
```diff
@@ -0,0 +1,46 @@
...
+
+  paths.forEach((path) => {
+    const details = {
+      title: path.title,
+      slug: path.slug,
+      href: `/${path.slug}/`,
+    };
+
+    path.topics.forEach((topic) => {
+      topic.pathItems.forEach((pathItem) => {
```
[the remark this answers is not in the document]
robdodson: Also suggest using the variable name `postSlug` instead of `pathItem`. I think then you'll end up with:

```
postToCollections[postSlug][path.slug]
```

which might be easier to grok at a glance.

src/site/_collections/post-to-collections.js
```diff
@@ -0,0 +1,46 @@
...
+        if (!postToCollections[pathItem]) {
+          postToCollections[pathItem] = {};
+        }
+
+        postToCollections[pathItem][path.slug] = details;
```
robdodson: Would it be possible to remove the need for the code on line 41 if this instead created an array and pushed the details to it?

src/site/_collections/post-to-collections.js
```diff
@@ -0,0 +1,46 @@
...
+      });
+    });
+  });
+
+  Object.keys(postToCollections).forEach((pathItem) => {
```
samthor: This looks like it's.. literally placing the values back on the same object? If it's not, it definitely needs a comment.

src/site/_collections/post-to-collections.js
MichaelSolati: I'm not sure what's the best way to address that though, I checked and it looks like i get around 260 items...

MichaelSolati:
@samthor I'd say all the changes have been made. Not sure how to respond to @robdodson 's last issue. Would you mind giving it another look through?

samthor:
Minor nit: I'm not a designer but I want the text and links to be... less distracting. It runs into the title.

I'm not suggesting I'm a good designer, but what about this:
Screen Shot 2020-03-03 at 08 39 55

Aside from that LGTM. I'd be happy for this to go in and then we revisit.

src/site/_collections/post-to-collections.js
robdodson: Would something like this work?

```
const paths = new Map();
collections.getFilteredByTag("pathItem").forEach((p) => {
  const path = p.data.path;
  paths.set(path, path);
});

paths.forEach((path) => ...
```

robdodson:
To Sam's last point.

What do y'all think about shifting it below the published date and giving it a similar text size and color.

image

MichaelSolati:
@robdodson made the changes, can I get a LGTM?

src/site/_includes/components/SignPosts.js
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const {html} = require("common-tags");
+const stripLanguage = require("../../_filters/strip-language");
+
+module.exports = ({postToCollections, url}) => {
+  // When @devnook lands her localization work we might need to update this file.
+  const strippedUrl = stripLanguage(url).replace(/\//g, "");
```
samthor: Is this stripping all slashes from a page?
What about e.g. https://web.dev/handbook/content-checklist/ ?

MichaelSolati: It's only used so that for the keys of an object from `posts-to-collections.js` which doesn't include handbook urls. Maybe this is something to be addressed later, if at all (I'm not sure)?

PR #2160, opened 2020-02-13
Clarify that NFC is only available on Chrome for Android

kaycebasques:

- Changes title from "Interact with NFC devices on the web" to "Interact with NDV devices on Chrome for Android"
- Removes draft status so that the page is visible on the blog again

### DIFF
--- a/src/site/content/en/blog/nfc/index.md
+++ b/src/site/content/en/blog/nfc/index.md
@@ -1,6 +1,6 @@
 ---
-title: Interact with NFC devices on the web
-subhead: Reading and writing to NFC tags is now possible on the mobile web.
+title: Interact with NFC devices on Chrome for Android
+subhead: Reading and writing to NFC tags is now possible
 authors:
   - beaufortfrancois
 date: 2020-02-12
@@ -8,12 +8,11 @@ hero: hero.jpg
 thumbnail: thumbnail.jpg
 alt: A photo of NFC tags
 description: |
-  Reading and writing to NFC tags is now possible on the mobile web.
+  Reading and writing to NFC tags is now possible on Chrome for Android.
 tags:
   - post # post is a required tag for the article to show up in the blog.
   - capabilities
   - fugu
-draft: true
 ---
 
 {% Aside %}


kaycebasques:
web.dev does not use the phrase "on the web" or "on the mobile web" unless there is cross-browser support for the feature.

I don't like to make changes without author buy-in so here's the full rationale for why this change is necessary:

No other browser has supported the feature, 1 other browser is actively opposed to it:

image

We recently got feedback from the Web GDEs expressing frustrating that we were using the phrase "on the web" for features only available on Chrome for Android.

web.dev's 2020 content mission is to help the **open web** thrive. I haven't socialized the content vision across the wider Web DevRel team yet but I do have buy-in from Kinlan, Dion, and the Area Leads (including Pete) and I've made it clear to these stakeholders that web.dev needs to be careful about how it phrases new features in order to make sure that it's truly representing the open web.

kaycebasques:
@robdodson Percy barked about this PR. The CI passed and the staged site looks fine and I don't have access to the Percy dashboard so I'm assuming this is just a false alarm.

beaufortfrancois:
@kaycebasques I believe the feedback we received from Web GDEs was about desktop vs mobile, not about browser support. I'd prefer, if possible, to have a title that looks like "Interact with NFC devices on the mobile web" as I'm quite sure some blink-based browsers will ship Web NFC if the origin trial is successful.

For info, https://deploy-preview-2160--web-dev-staging.netlify.com/nfc/#use clearly states that Web NFC is only for Chrome for Android for now as it's an origin trial post.

Having a title that doesn't change over time would be great for SEO as well.

kaycebasques:
Pretty sure the gist of the Web GDE frustration is that we're using "mobile web" and "web" too liberally.

> I'd prefer, if possible, to have a title that looks like "Interact with NFC devices on the mobile web"

Is it supported on mobile Safari or Firefox? I don't even see a caniuse table for this API.

> Having a title that doesn't change over time would be great for SEO as well.

Yes, agreed, it's unfortunate that we can't use "mobile web" right now, but the reality is that the API doesn't have cross-mobile-browser support (if my understanding of Safari and Firefox support is correct) so it's not appropriate to use that phrase right now. Building rapport with our audience is more important than SEO in this case.

One possible option is to return the "Interact with NFC devices on the mobile web" but put a disclaimer at the very top of the article (literally before any other content) that says that the API is currently only supported on Chrome for Android.

PR #2167, opened 2020-02-13
Hide yt iframe during screenshot tests.

robdodson:

- Hides the yt iframe during screenshot testing cuz it's flaky and keeps shifting the channel logo around a few pixels which breaks the tests. Returning `""` instead of a undefined return because eleventy will literally write "undefined" if we do that and it just looks ugly :P

### DIFF
--- a/src/site/_includes/components/YouTube.js
+++ b/src/site/_includes/components/YouTube.js
@@ -29,6 +29,13 @@ module.exports = (id, startTime) => {
     throw new Error("Cannot create YouTube component if id is undefined");
   }
 
+  // Don't load YT iframe in our test environment. This specifically affects
+  // screenshot testing where the iframe can be slow or flaky to load and fail
+  // because YouTube is always fiddling with their UI.
+  if (process.env.ELEVENTY_ENV === "test") {
+    return "";
+  }
+
   let src = `https://www.youtube.com/embed/${id}`;
   if (startTime) {
     src += `?start=${startTime}`;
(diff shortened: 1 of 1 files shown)


src/site/_includes/components/YouTube.js
```diff
@@ -29,6 +29,13 @@ module.exports = (id, startTime) => {
     throw new Error("Cannot create YouTube component if id is undefined");
   }
 
+  // Don't load YT iframe in our test environment. This specifically affects
+  // screenshot testing where the iframe can be slow or flaky to load and fail
+  // because YouTube is always fiddling with their UI.
+  if (process.env.ELEVENTY_ENV === "test") {
```
samthor: `.w-youtube` is still going to create a box of a certain w/h (with the `padding` ratio hack). Is it worth including the physical box and giving it a background color so it looks like a red/black square?

robdodson: I don't think it will because since we're returning early the shortcode never writes out `<div class="w-youtube">`. It just writes out a blank text node.

samthor: My point is that we _could_ write out that div, just not include the iframe, so the layout is more akin to the real one if anyone is actually looking at the visual diffs.

I could perhaps accidentally add 100s of YouTube videos and I would not be alerted to that fact. If it's a giant red box of the right size, at least I'd know.

robdodson:
Added a placeholder to occupy the space. Should look like this in Percy now:

image

robdodson:
I tweaked it one more time to add a bit of text so it'll appear like this:
image

Just in case future people are like "why is my iframe a solid box?"

samthor:
That's quite aesthetically pleasing. Good color choice.

PR #2180, opened 2020-02-14
Only run Percy if .js, .njk, or .scss files change.

### LINKED ISSUE #2166: Only have percy run if a .njk, .js, or .css file is changed.
It doesn't make sense to run percy every time someone writes a new post because we're not snapshotting those pages anyway. So we should probably ignore `.md` files, or at least any `.md` files that we don't care about.

robdodson:
Fixes #2166

- Adds a bash script to the Percy workflow which checks if `.js`, `.njk`, or `.scss` files were touched in the PR and only runs Percy if they were.


robdodson:
shucks, looks like `BRANCH=$(git rev-parse --abbrev-ref HEAD)` doesn't work as expected in an action.

PR #2181, opened 2020-02-14
Only run Percy on certain file types.

### LINKED ISSUE #2166: Only have percy run if a .njk, .js, or .css file is changed.

robdodson:
Fixes #2166
Better version of #2180

- Use GitHub Actions' [request paths syntax](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths) to limit when Percy runs. Should only run if .njk, .js, or .scss files are changed.

### DIFF
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -7,10 +7,17 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '**.js'
+      - '**.njk'
+      - '**.scss'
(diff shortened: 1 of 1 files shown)


robdodson:
I tested adding a .scss or .js change and it triggered Percy, so I think this is working as intended.

PR #2187, opened 2020-02-19
Revert #2123

robdodson:
My mistake for not catching this sooner.

The changes in #2123 moved everything under /handbook/... to be under /...

Unfortunately that means we're now squatting on important URLs like /quick-start/, /voice/, etc.

I'd like to revert this change and see if we can update the work in #2123 to keep handbook under its own path.


robdodson:
Superseded by #2188

PR #2189, opened 2020-02-19
Add GH Action to lint yaml and verify updated field.

### LINKED ISSUE #1959: Throw linting error when `updated` field is not updated

1. Someone updates content
1. They forget to update the `updated` field in the front matter

1. The automated build/linter system throws an error instructing them to update the field

Assigning to @robdodson for triage

robdodson:
Fixes #1959 

- Adds a new GitHub Actions workflow that runs against pull requests which contain markdown files in `src/site/content/**`.
- Adds a GitHub Action that lints yaml front matter. Currently checks for `date` field, and verifies the `updated` field has been added to modified files and that it's current.
- See the README in the `.github/actions/lint-front-matter` directory which explains how to add your own custom rules.

cc @petele @kaycebasques @mfriesenhahn @jpmedley

### DIFF
--- a/.github/actions/lint-front-matter/linters/all-files.js
+++ b/.github/actions/lint-front-matter/linters/all-files.js
@@ -1,5 +1,5 @@
-const getYamlFrontMatter = require('../utils/get-yaml-front-matter.js.js');
-const checkHasProperty = require('../rules/has-property.js/index.js.js');
+const getYamlFrontMatter = require('../utils/get-yaml-front-matter');
+const checkHasProperty = require('../rules/has-property.js/index');
 
 /**
  * @param {Array<string>} files Files that should be linted.
--- a/.github/actions/lint-front-matter/linters/all-files.js
+++ b/.github/actions/lint-front-matter/linters/all-files.js
@@ -1,0 +1,15 @@
+const getYamlFrontMatter = require('../utils/get-yaml-front-matter.js.js');
+const checkHasProperty = require('../rules/has-property.js/index.js.js');
+
+/**
+ * @param {Array<string>} files Files that should be linted.
+ * @return {Array<string>} Error messages from failed lint rules.
+ */
+module.exports = async (files) => {
+  const failures = [];
+  for (file of files) {
+    const frontMatter = await getYamlFrontMatter(file)
+    checkHasProperty(file, frontMatter, 'date', failures);
+  }
+  return failures;
+};
--- a/.github/actions/lint-front-matter/rules/updated-is-current.js
+++ b/.github/actions/lint-front-matter/rules/updated-is-current.js
@@ -1,0 +1,20 @@
+/**
+ * @param {string} file Filename of the file being linted.
+ * @param {Object} yaml The yaml front matter from the file.
+ * @param {?string} yaml.updated The updated property in the yaml front matter.
+ * @param {Array<string>} failures Error messages from failed lint checks.
+ */
+module.exports = (file, {updated}, failures) => {
+  // If updated is undefined, just return early.
+  // Rather than error here, let another lint rule that explicitly checks for
+  // the presence of the updated property handle it.
+  if (!updated) {
+    return;
+  }
+
+  const today = new Date();
+  today.setUTCHours(0, 0, 0, 0);
+  if (today.getTime() > new Date(updated).getTime()) {
+    failures.push(`${file} updated field is out of date. It must be at least ${today.toISOString().split('T')[0]}.`);
+  }
+}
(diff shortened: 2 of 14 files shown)


petele:
This is awesome. Two questions (similar to issues we ran into on WebFu)...

Is there any way to tell the linter that it is OK to not update the `updated` field? For example, maybe I just fixed a typo, but haven't really made any material changes to the doc. A user seeing the `Updated Jan 10, 2020` might think it's up to date when in reality, the doc hasn't been properly updated in a while.

What happens if I create a PR with two changed files, set the `updated` field on each appropriately. Then, a few days later (but before they're merged), update one file (and set it's `updated` appropriately). Will this check only verify the newly changed file, or both files in the PR?

robdodson:
> Is there any way to tell the linter that it is OK to not update the updated field? For example, maybe I just fixed a typo, but haven't really made any material changes to the doc. A user seeing the Updated Jan 10, 2020 might think it's up to date when in reality, the doc hasn't been properly updated in a while.

I guess we could add a label, similar to setting `cla: yes`. How did y'all solve it for webfu?

> What happens if I create a PR with two changed files, set the updated field on each appropriately. Then, a few days later (but before they're merged), update one file (and set it's updated appropriately). Will this check only verify the newly changed file, or both files in the PR?

I believe it should check both files. My understanding of the action that we use to do the diffing is that it runs against all commits in the PR and keeps sending that list of files to the linter until the PR is merged.

petele:
> I guess we could add a label, similar to setting cla: yes. How did y'all solve it for webfu?

On WebFu it was just a warning that would show up in local linting and was ignorable.

> I believe it should check both files.

If that's the case, we may need some kind of way to tell it to ignore. Especially if I write multiple files and they take a while to get merged, the first one will be "out of date" even though the `updated` field will be accurate.

robdodson:
I'm not sure I understand the second point. Wouldn't you want the updated
field to reflect the date the file was actually published?

petele:
Specific example:
Jan 1 - change file1.md and file2.md, update `updated` field, create PR
Jan 5 - (same PR, not yet merged) file2.md changed and `updated` field set correctly, file1.md not touched

On Jan 5, when the tests run, will file1.md be flagged as having `updated` as Jan 1?

robdodson:
Would something like this work:

- When a pull request is submitted
  - If a file was modified
    - Verify that it has an updated field
    - Look up the original file, confirm that the file in the PR's updated date is further in the future than the original file's updated date.
    - Don't worry if the updated date is >= today. Just verify that it is further in the future than the original file's updated date.

.github/actions/lint-front-matter/rules/updated-is-current.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = (file, {updated}, failures) => {
+  // If updated is undefined, just return early.
+  // Rather than error here, let another lint rule that explicitly checks for
+  // the presence of the updated property handle it.
+  if (!updated) {
+    return;
+  }
+
+  const today = new Date();
+  today.setUTCHours(0, 0, 0, 0);
```
samthor: I think this will actually hurt folks who live far west—places like the US west coast. UTC could be a day ahead if you make a post ... before 8am. I know this is obscure, but still.

My brain is a bit confused with time math, but I think you want to pass... +12 here. (The farthest timezone _west_).

.github/actions/lint-front-matter/linters/all-files.js
```diff
@@ -0,0 +1,15 @@
+const getYamlFrontMatter = require('../utils/get-yaml-front-matter.js.js');
+const checkHasProperty = require('../rules/has-property.js/index.js.js');
```
samthor: This require path seems wrong. It has an extra suffix?

samthor: Actually all the other files are `.js.js` too. Is there a reason?

robdodson: Hm. I think I may have told VS Code that it was ok for it to auto update my imports when i dragged a folder and perhaps it flubbed it.

.github/actions/lint-front-matter/rules/updated-is-current.js
robdodson: Yeah I think we may actually scrap this per the discussion with pete.

If `updated` was added in a file, i'm going to check that it's greater than `date`.
If `updated` was modified in a file, I'm going to check that it's greater than the existing file's `updated`.

robdodson:
Adding a note to self that we can skip a step using this syntax:

`if: contains(github.event.pull_request.labels.*.name, 'some-defined-label')`

So if we want to tell the linter to be quiet when we add a label we'd use ^^

robdodson:
It looks like we may want a second piece to this, which is a precommit hook or something similar that actually revs the updated field for you, similar to what webfu does here:
https://github.com/google/WebFundamentals/blob/756469ee280006dc5014bb095a24dec7e53a2dd6/gulp-tasks/wfUpdatedOn.js#L78-L84

PR #2191, opened 2020-02-19
Add disclaimer banner to translated posts.

devnook:
Screenshot 2020-02-19 at 15 47 11


src/site/_includes/partials/post.njk
```diff
@@ -54,6 +54,32 @@
       </div>
     {% endif %}
 
+    {% if translation %}
+      {% if translation === true %}
```
robdodson: Instead of mixing types, what do you think about setting this to `translation: 'none'`?

src/site/_includes/partials/post.njk
```diff
@@ -54,6 +54,32 @@
...
+        <div class="w-banner w-banner--warning w-banner--body">
+          This page is not translated yet.
```
robdodson: This means all articles will initially have a 'this article has not been translated yet banner', right?

Earlier you mentioned that we might end up in a situation where an article hasn't been machine translated yet, but someone submits a translation, and then the machine translation gets submitted shortly thereafter.

What if we only show the gitlocalize link after the article has gotten its initial machine translation? And change the language to something like "We're currently working on an initial translation for this article."

robdodson:
Left some comments on the implementation. I think it may also be worth working with @jimoong to create a nicer design for the various states of the banner.

src/site/_filters/gitlocalize-link.js
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+const path = require("path");
+const site = require("../_data/site");
+
+module.exports = (inputPath, lang) => {
+  if (!inputPath) {
+    /* eslint-disable-next-line */
+    console.warn(
```
samthor: This is presumably about to crash. Do we expect this to happen?
If so, then we should return an empty string etc—otherwise I'm happy to remove this as the `Error` itself is notable enough.

src/site/_filters/gitlocalize-link.js
```diff
@@ -0,0 +1,29 @@
...
+      "inputPath passed to gitlocalizeLink filter was undefined or null.",
+    );
+  }
+  inputPath = inputPath.replace(`/${lang}/`, `/${site.defaultLocale}/`);
+  return `${path.join(site.gitlocalize, lang, inputPath)}`;
```
samthor: backticks and ${ are redundant since there's only one value here.

src/site/_filters/gitlocalize-link.js
```diff
@@ -0,0 +1,29 @@
...
+const site = require("../_data/site");
+
+module.exports = (inputPath, lang) => {
+  if (!inputPath) {
+    /* eslint-disable-next-line */
+    console.warn(
+      "inputPath passed to gitlocalizeLink filter was undefined or null.",
+    );
+  }
+  inputPath = inputPath.replace(`/${lang}/`, `/${site.defaultLocale}/`);
```
samthor: Can we sanity-check by adding e.g.

```js
if (!inputPath || inputPath.startsWith(`/${lang}/`)) {
  return inputPath;  // not a translated link, return early for safety
}
```

src/site/_includes/partials/post.njk
```diff
@@ -54,6 +54,32 @@
...
+          Consider improving it by contributing your translation
+          <a href="{{ page.inputPath | gitlocalizeLink(lang) }}">here</a>.
+        </div>
+      {% endif %}
+      {% if translation === 'machine' %}
+        <div class="w-banner w-banner--caution w-banner--body">
+          This page is automatically translated using Google Translation API.
```
samthor: using _the_ Google Translation API

samthor: Also, will this string itself need to be translated? 😆

src/site/_includes/partials/post.njk
samthor: Yeah, I'm very confused about what `true` means here. I think "none" is better .

I also know we have test content for now but I presume we won't include English articles in other folders at launch.

src/site/_includes/partials/post.njk
```diff
@@ -54,6 +54,32 @@
...
     {% endif %}
 
+    {% if translation %}
+      {% if translation === true %}
+        <div class="w-banner w-banner--warning w-banner--body">
+          This page is not translated yet.
+          Consider improving it by contributing your translation
+          <a href="{{ page.inputPath | gitlocalizeLink(lang) }}">here</a>.
+        </div>
+      {% endif %}
```
samthor: Nunjucks supports `elif`.

src/site/_includes/partials/post.njk
devnook: I think manual (human) translation is usually better than machine one? Should we deter someone from posting their translation because we can get machine translation later on? (Machine one does not automatically overwrite what is already present, it submits a PR just like a human would do)

src/site/_filters/gitlocalize-link.js
devnook: Empty string should be ok

src/site/_filters/gitlocalize-link.js
devnook: ooh.

src/site/_filters/gitlocalize-link.js
devnook: I moved whole locale code to a separate module, and use it for sanity check here.

src/site/_includes/partials/post.njk
devnook: All template strings will need to be translated, eventually:)

robdodson:
Restarted netlify. Not sure why it's failing.

@devnook you still need a design from Jiwoong for this, correct?

src/site/_includes/partials/post.njk
```diff
@@ -14,6 +14,25 @@
...
+      {% elif translation === 'machine' %}
+        This page is automatically translated using the Google Translation
+        API.
+      {% elif translation === 'manual' %}
+        This article was translated from
+        [the original post](/en{{ page.url | stripLanguage}}).
+      {% endif %}
+      {% if gitlocalizeUrl %}
+        Consider improving it by contributing your translation
+        [here]({{ gitlocalizeUrl }}).
```
robdodson:
```suggestion
        Consider improving it by [contributing your translation
        here]({{ gitlocalizeUrl }}).
```

robdodson: Small request to link more than the word "here". It's a good a11y practice to link phrases for when users are navigating by link in a screen reader.

src/site/_includes/partials/post.njk
```diff
@@ -14,6 +14,25 @@
   {# Make sure Actions are first in the tab order. #}
   <web-actions actions="share|subscribe" authors="{{ authors | expandContributors | join('|') }}"></web-actions>
 
+  {% set gitlocalizeUrl =  page.inputPath | gitlocalizeLink(lang) %}
```
robdodson: Would you be ok moving lines 17-35 into a nunjucks include?

src/site/content/en/handbook/markup-post-codelab/index.md
```diff
@@ -43,6 +43,7 @@ The YAML fields at the start of each piece of web.dev content define how the pie
 * `codelabs`: The slug(s) for the codelabs, if any, associated with the post.
   All indicated codelabs appear in a callout at the end of the post.
 * `tags`: Tags for the site search. The `post` tag is required for a post to appear on the blog.
+* `translation`: true/machine/manual. Set only if the post is a translation from another, original post.
```
robdodson:
```suggestion
* `translation`: none/machine/manual. Set only if the post is a translation from another, original post.
```

src/site/_includes/partials/post.njk
robdodson: I agree that a human translation is better. My concern was if someone submitted a pull request that only translated a portion of the article, and then at the same time we receive the machine translation. Which of the two pull requests would we choose to merge?

This might be an edge case though. So I think we can go with the text that you have and if it causes issues we can always tweak it.

devnook:
PTAL

robdodson:
> Are we machine translating the content every time that someone requests a page, or are we somehow storing the machine translation? Seems like the latter might be a way to save on costs, because our content doesn't change that often. At the expense of added complexity, though. Or maybe the cost of the Translate API is nominal or has some sort of built-in cache...

We're storing the translations as markdown files. So if I have an article at `/en/foo/index.md`, gitlocalize will send a PR with a machine translated copy to `/pl/foo/index.md`. You can see an example of the process over on webfu: https://github.com/google/WebFundamentals/pull/8539

robdodson:
@samthor do you want to give it one last look?

locale-handler.js
```diff
@@ -14,16 +14,8 @@
...
 const path = require("path");
-
-const isProd = Boolean(process.env.GAE_APPLICATION);
-const contentDir = isProd ? "dist" : "src/site/content";
-const dirs = fs.readdirSync(path.join(__dirname, contentDir));
-const SUPPORTED_LOCALES = dirs.filter((dir) => localeCode.validate(dir));
-const DEFAULT_LOCALE = "en";
-
-const isSupported = (locale) => SUPPORTED_LOCALES.indexOf(locale) > -1;
+const locales = require("./src/lib/utils/locale");
```
samthor: nitpick: can we make this either "locales" the file or "locale" the import? Seems weird to have it be subtly different.

PR #2198, opened 2020-02-20
Add case study template

### LINKED ISSUE #1462: Create case study guidelines and template
This came up from a discussion between @Snugug and I. 

@Snugug - can you add your thoughts.

kaycebasques:
Fixes #1462


robdodson:
I think the template may be missing some fields:

```
authors:
  - username
hero: hero.jpg
alt: A description of the hero image for screen reader users.
```

It may be worth also be worth pointing folks to the post template because it includes fields for creating custom social media cards.

PR #2201, opened 2020-02-21
Make fugu-status live

petele:
- removes `draft: true` from fugu-status page


petele:
> @petele you should have the ability to merge this now but please hold off until Monday

Awesome - will do, thanks for the **super fast** review! :)

PR #2209, opened 2020-02-24
don't include SW on bad hostnames, redirect from appengine-test

### LINKED ISSUE #2024: appengine-test.web.dev should not be crawled.
**Describe the bug**
Searching on Google returns results for the domain **appengine-test.web.dev**

Screen Shot 2019-12-19 at 9 57 05 AM

**To Reproduce**
1. Go to https://www.google.com/search?q=pwa+multi+origin+demian
2. Second result is pointing to appengine-test.web.dev

**Expected behavior**
web.dev not appengine-test.web.dev

### LINKED ISSUE #2044: Web.dev by google seems to lag navigation
I feel many pages are not ordered well as well as many of them are missing navigation like https://web.dev/web-share/ which you cannot navigate from HOME.

samthor:
Fixes #2024 

Also attempts to prevent more major failures in future by actively not including the Service Worker on unsupported domains.

We need the `server.js` change in this for two reasons:
* the literal 301 redirect for Google/search
* some poor user might have installed the Service Worker on "appengine-test.web.dev", and unless we nuke it, they'll never get an update (but the client-side change will hopefully help prevent things like this in future).

### DIFF
--- a/src/lib/bootstrap.js
+++ b/src/lib/bootstrap.js
@@ -35,5 +35,22 @@ WebComponents.waitFor(async () => {
 });
 
 if ("serviceWorker" in navigator) {
-  navigator.serviceWorker.register("/sw.js");
+  const allowedHostnames = ["www.web.dev", "localhost"];
+  if (allowedHostnames.indexOf(window.location.hostname) !== -1) {
+    navigator.serviceWorker.register("/sw.js");
+  } else {
+    console.warn(
+      "skipping SW, unsupported hostname:",
+      window.location.hostname,
+    );
+
+    // Remove previous Service Worker instances from this hostname. This should never normally
+    // happen but is here for safety.
+    navigator.serviceWorker.getRegistrations().then(async (all) => {
+      await all.map((reg) => reg.unregister());
+      if (all.length) {
+        window.location.reload();
+      }
+    });
+  }
 }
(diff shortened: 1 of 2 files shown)


src/lib/bootstrap.js
```diff
@@ -35,5 +35,22 @@ WebComponents.waitFor(async () => {
...
+      "skipping SW, unsupported hostname:",
+      window.location.hostname,
+    );
+
+    // Remove previous Service Worker instances from this hostname. This should never normally
+    // happen but is here for safety.
+    navigator.serviceWorker.getRegistrations().then(async (all) => {
+      await all.map((reg) => reg.unregister());
+      if (all.length) {
+        window.location.reload();
```
robdodson: Because this does a reload do you think folks would actually see the warning on 42?

src/lib/bootstrap.js
```diff
@@ -35,5 +35,22 @@ WebComponents.waitFor(async () => {
 });
 
 if ("serviceWorker" in navigator) {
-  navigator.serviceWorker.register("/sw.js");
+  const allowedHostnames = ["www.web.dev", "localhost"];
```
robdodson:
```suggestion
  const allowedHostnames = ["web.dev", "localhost"];
```

src/lib/bootstrap.js
samthor: Well, only if they had a SW installed already. This isn't likely because e.g., appengine-test is never going to see this code (the SW will replace itself before new JS is brought in).

This whole branch only exists in case we mess up again and serve the content from a weird domain.

PR #2210, opened 2020-02-24
strip utm_ from Workbox

### LINKED ISSUE #2033: SW should ignore utm query params
**Describe the bug**
@jeffposnick mentioned our sw should ignore `utm_medium=` on incoming navigation requests.

samthor:
Fixes #2033.

Jeffy, let me know if I'm doing this right. I'm surprised (but I'm sure there's a valid reason) that `registerRoute` doesn't take an extra arg allowing me to ignore params.

But I'm also surprised at how expressive `registerRoute` allows me to be. For instance, our `contentHrefRe` (on line 122), when passed to Workbox, matches anywhere in the URL, so I can e.g. load something like: https://web.dev/$@not-valid-slug@$/valid-slug/

Is the right idea to actually build a regexp that contains the hostname on the left (including escaping)? e.g.: `/$https?:\/\/web\/.dev\/([\\w-]+/)*(|index.html)(\\?|$)/`

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -5,12 +5,25 @@ const serviceWorkerArchitecture = "v3";
 
 const normalizeIndexCacheKeyPlugin = {
   cacheKeyWillBeUsed({request, mode}) {
+    // web.dev does not have any handlers which respond to ?search queries, so strip them (useful
+    // for ?utm_... params, which aren't stripped by Workbox).
+    // From: https://github.com/GoogleChrome/workbox/issues/1709#issuecomment-429917667
+    const u = new URL(request.url);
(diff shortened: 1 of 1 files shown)


samthor:
Oh sorry, I realise I can pass a function to the first arg of `registerRoute`. I've updated the PR.

samthor:
My 2c:

* I don't think the Service Worker will ever see the hash, but it's sort of in the same bucket as search, so it's easy to remove at the same time

* I think my confusion was around: I wanted the Service Worker to _rewrite_ the request before it happened, but fundamentally if you know it's going to give you the same result, then changing the cache key is going to work the same way.

Thanks!

PR #2213, opened 2020-02-24
Add a console dump filter.

robdodson:

- Adds a `consoleDump` filter. https://twitter.com/eleven_ty/status/1227229750747389952

@MichaelSolati @devnook I was wondering if y'all might find this useful?


MichaelSolati:
How does the dump look for objects? Does it appear as a string or `[object, object]`?

robdodson:
image

robdodson:
^^ that is logging the `page` object.

here is an example from logging `collections.all`.

image

PR #2222, opened 2020-02-24
Remove Details animation

mfriesenhahn:
The chevron animation in the Details component is janky in Firefox. Since the details panel doesn't animate, animating the chevron doesn't seem essential, so this PR removes it.

- Remove Details component chevron animation.
- Remove $BORDER_COLOR variable from details stylesheet so it uses the global variable.

**Note:** Percy is grumping because it's not seeing a lazy-loaded image on the Learn page, but this PR causes no visual changes to the page on load; it just removes the chevron animation in the Details component.

### DIFF
--- a/src/styles/components/_details.scss
+++ b/src/styles/components/_details.scss
@@ -1,11 +1,7 @@
-@import '../settings/colors';
-@import '../settings/global';
-@import '../settings/type';
 @import '../tools/mixins';
 
 $CHEVRON_SIZE: 24px;
 $CHEVRON_MARGIN: 12px;
-$BORDER_COLOR: $GREY_300;
 
 .w-details {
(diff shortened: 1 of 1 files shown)


robdodson:
> Note: Percy is grumping because it's not seeing a lazy-loaded image on the Learn page, but this PR causes no visual changes to the page on load; it just removes the chevron animation in the Details component.

I'm going to approve this (incorrect) snapshot and see if the fix I'm about to PR corrects it.

src/styles/components/_details.scss
```diff
@@ -22,6 +18,7 @@ $BORDER_COLOR: $GREY_300;
   }
 }
 
+
```
mfriesenhahn:
```suggestion
```

PR #2226, opened 2020-02-24
Add round buttons [Self-assessments]

mfriesenhahn:
The self-assessment modal will  have a round **Close** button. Since it's not the first round button we have--and likely not the last--this PR adds some general styles for square and round icon buttons. It also tidies up some tooltip and button a11y issues.

- Add aria-label to the CopyCode button and the SideNav **Close** button, for a11y.*
- Add a tooltip to the CopyCode button, also for a11y.
- Create general styles for icon buttons (`.w-button--icon`) and round icon buttons (`.w-button--round`).
- Migrate the CopyCode button to the general round icon button styles and simpify the CopyCode stylesheet.
- Update button documentation in `/handbook/web-dev-components`.
- Fix tooltips so they show on non-hover devices when using keyboard to focus the parent element.

**Question:** Should I make the SideNav hamburger and close buttons `.w-button--icon w-button--round` and/or add tooltips to them?

*We shouldn't rely on the Material icon font names because they're not always going to be semantic. And we also can't rely on tooltip content because it's only visible on focus, preventing NVDA from reliably reading it.


src/lib/components/CopyCode/index.js
```diff
@@ -21,8 +21,18 @@ class CopyCode extends BaseElement {
 
     if (!this.copyButton) {
       this.copyButton = document.createElement("button");
-      this.copyButton.className = "web-copy-code__button";
+      this.copyButton.className =
+        "w-button--icon w-button--round web-copy-code__button";
+      this.copyButton.setAttribute("data-icon", "file_copy");
+      this.copyButton.setAttribute("aria-label", "Copy code");
```
samthor: It took some reading to understand that the `span` tooltip only shows on hover. Maybe a quick comment: we set `aria-label` and not `title` because `title` would result in both the browser's tooltip and ours showing.

src/lib/components/CopyCode/index.js
```diff
@@ -21,8 +21,18 @@ class CopyCode extends BaseElement {
...
       this.copyButton.addEventListener("click", this.onCopy);
+
+      this.tooltip = document.createElement("span");
+      this.tooltip.className = "w-tooltip w-tooltip--right";
+      this.tooltip.setAttribute("role", "tooltip");
+      this.tooltip.innerHTML = "Copy code";
```
samthor: General advice: use `.textContent` unless you really need `.innerHTML`.

src/lib/components/SideNav/index.js
```diff
@@ -62,6 +62,7 @@ class SideNav extends BaseElement {
           <button
             @click=${this.onCloseSideNav}
             class="web-side-nav__hide"
+            aria-label="Close"
```
samthor: This makes sense but I also know that the `title` attribute is supported here. Is that a simpler choice that would give a11y information and _also_ provide a hover tooltip for mouse users? (Since we're not using our tooltip code)

@robdodson is probably the expert here.

src/styles/components/_buttons.scss
```diff
@@ -138,3 +108,61 @@
...
+  height: 3rem;
+  overflow: unset;
+  padding: 0;
+  transition: background $TRANSITION_SPEED ease, color $TRANSITION_SPEED ease;
+  width: 3rem;
+}
+
+@include hover() {
+  .w-button--icon:hover,
+  .w-button--icon:focus {
```
samthor: It might be worth commenting that we leave the default `outline: ...` alone (the one implicitly defined on `:focus`).

src/lib/components/CopyCode/styles.scss
```diff
@@ -1,44 +1,24 @@
 @import "../../../styles/settings/colors";
-@import '../../../styles/settings/global';
 
 .web-copy-code__button {
```
samthor: does this end up mixing in `base-button` anyway, so we keep the `-webkit-tap-highlight-color: ...` property?

src/lib/components/CopyCode/index.js
mfriesenhahn: Actually, I'm setting `aria-label` because `title` isn't accessible to sighted keyboard users. I'll note that.

src/lib/components/SideNav/index.js
mfriesenhahn: My understanding is that a tooltip is preferred over `title` because it doesn't display on keyboard focus, making it inaccessible to sighted keyboard users.

I asked in the PR notes whether I should convert this to a circular button, including a tooltip. I think that'd be better for a11y, but I wasn't sure what was intentional about this button since it's different in various respects from our other buttons (e.g., it doesn't have a `pointer` cursor). 😄 

If you're cool with that, I can go ahead and convert it!

samthor: Sorry, I missed that question. And this is just a simple controller for the side nav. I wouldn't read too much into how we built it, feel free to update it.

As a digression, I have some opinions that maybe eventually this element should instead be a `<label>` that toggles a "side nav open" checkbox (which I like as a no-JS solution...) but either way it'll still live here in the HTML tree and be either a button or `role="button"`.

src/styles/components/_buttons.scss
mfriesenhahn: Actually, we don't leave it alone. `.w-button` doesn't currently have an accessible focus style. It just shows a slightly darker version of the hover overlay that isn't dark enough to meet WCAG 2.1 minimum contrast requirements for UI (3:1).

I've fixed that for circular buttons by adding a 1px `box-shadow`, which I'm happy to note. Also happy to suggest an accessible focus style for `.w-button` in a separate PR if folks are interested. 😄

src/lib/components/CopyCode/styles.scss
mfriesenhahn: Oops, need to import mixins, not these. Good catch.

mfriesenhahn:
Hey, @samthor, thanks for catching all the nits! I really appreciate it!

I left a couple questions for you. Hoping they'll be quick :)

src/styles/components/_buttons.scss
mfriesenhahn: Something like these, maybe:

image

image

(I used a psuedo with the opacity dropped a bit so the outlines aren't so slap-you-in-the-face while still being accessible. I'm thinking this would only apply on keyboard focus, as with our other focus rings. I'd also like to turn off our current button focus style for non-keyboard focus so it doesn't linger. I know it's inherited from Material, but the drop shadow and lighter background make it look to me like something's gone wrong with the button's state.)

samthor: I'm not really sure what I'm looking at, sorry. Is the blue background the `:focus` hint?

Can `:focus` and `:hover` have the same colors, by the spec? I can't see where they're different.

src/lib/components/SideNav/index.js
mfriesenhahn: It's been updated :)

Re: the checkbox solution, I vaguely recall reading that there are a11y complications with using checkboxes for non-checkbox-y things, but I'd have to refresh my memory.

src/styles/components/_buttons.scss
mfriesenhahn: Sorry, the inset outline is the focus hint here. I'm showing both `.w-button` and `.w-button--primary` variants.

And, yeah, so far as I know there's no rule against `:hover` and `:focus` being the same, assuming your `:hover` state is contrast conformant. That's often not the case, though, since hover is a transient state. (In fact, our `:hover` isn't conformant.) So, one common strategy is to use an additive `:focus` style: `:hover` plus an additional hint. That's what I'm proposing here, with the inset outline being the additional hint.

mfriesenhahn:
Note that Percy is failing because the hamburger button and logo shifted to the right by a couple pixels. This is because I increased the size of the hamburger button to 44 x 44 px to meet [WCAG 2.1 SC 2.5.5](https://www.w3.org/TR/WCAG21/#target-size). Ditto the SideNav close button.

src/lib/components/SideNav/index.js
mfriesenhahn: So, since this issue is an aside, is this PR good to go? 😄

mfriesenhahn:
Hey, @robdodson, it's rebased!

samthor:
@mfriesenhahn Looks like it needs another rebase :)

src/styles/components/_buttons.scss
```diff
@@ -198,4 +202,3 @@
 
 .w-button--icon::before {
   margin: 0;
-}
```
mfriesenhahn: }

src/styles/components/_buttons.scss
```diff
@@ -198,4 +202,3 @@
 
 .w-button--icon::before {
   margin: 0;
```
mfriesenhahn:
```suggestion
  margin: 0;
}
```

src/styles/components/_buttons.scss
```diff
@@ -174,3 +144,61 @@
...
+  display: inline-block;
+  height: 24px;
+  margin-right: 16px;
+  // To prevent screen readers from reading icon names.
+  speak: none;
+  width: 24px;
+}
+
+.w-button--icon::before {
+  margin: 0;
```
mfriesenhahn:
```suggestion
  margin: 0;
}
```

mfriesenhahn:
@robdodson, can you poke Percy and merge this when you get a sec?

mfriesenhahn:
@robdodson, can you merge this? Percy's baseline just needs updated.

PR #2230, opened 2020-02-25
Add Assessment component [Self-assessments]

mfriesenhahn:
This PR adds the top-level Assessment component, which will house all self-assessment question types (coming in separate PRs).

- Add `Assessment` web component and accompanying shortcode so it's easy for authors to add to posts.
- Add `BaseModalElement` component, which `Assessment` extends.
- Add utility to check overflow (repeated in #2229).
- Add self-assessment template. (Won't be fully functional until assessment question types are added in later PRs.)
- Add **temporary** `md-block` filter to handle rendering of MarkDown blocks in the Assessment component.
- Add skeleton handbook page for self-assessments (to be populated as self-assessment question types are added in later PRs).
- Add self-assessment color palette and global border radius variables.
- Make white space in self-assessment callouts responsive to maximize space for questions on mobile.

**Notes:**
1. Content in the sample self-assessment set in `handbook/self-assessment-components` is currently unstyled because assessment question types have not yet been added. (Forthcoming in separate PRs.)
1. The shortcodes for the old approach to rendering self-assessments (`AssessmentCallout`, `AssessmentHint`, `Tab`, and `Tabs`) are being left in place for now  until we're ready to migrate existing self-assessments to the new approach.


src/lib/components/Assessment/index.js
```diff
@@ -0,0 +1,182 @@
...
+    this.open = false;
+  }
+
+  updated(changedProps) {
+    if (changedProps.has("open")) {
+      this.modal = this.open;
+      if (this.open) {
+        this.openModal();
+      } else {
+        this.addEventListener("animationend", this.closeModal, {
```
samthor: `closeModel` doesn't seem to be bound to `this`

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+
+  // Assumes modal on/off transitions are handled by an animation
+  // rather than a transition.
+  onAnimationEnd() {
+    this.animatable = false;
+    // Must wait until animation ends to change focus and check overflow.
+    this.manageFocus();
+    if (this.open) this.onResize();
```
samthor: general style nit: can we use `{}` everywhere?

src/lib/components/Assessment/index.js
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Element that renders a self-assessment callout.
+ * @extends {BaseModalElement}
+ * @final
+ */
+class Assessment extends BaseModalElement {
+  static get properties() {
+    return {
+      modal: {attribute: "aria-modal", reflect: true},
```
samthor: Why is `this.modal` different than `this.open`?

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  manageDocument() {
+    // Inert everything other than the modal when it's open.
+    const main = document.querySelector("main");
+    const header = document.querySelector("web-header");
+    const footer = document.querySelector("footer");
+    const overFlowClass = "web-modal__overflow-hidden";
+
+    if (this.open) {
```
samthor: So this stuff should be on if _any_ modal is open (and it might be a bad idea to allow multiple modals to be open, too). So there needs to be some static singleton state here.

samthor:
I have opinions on modals. They're hard. We actually have `<dialog>` which is supported in all Chromium-based browsers, and we have a polyfill for Safari and Firefox—the `inert`-based approach is fine but always requires a polyfill (and see a few comments on global state) and is O(n) with the number of elements that might need to be made inert.

mfriesenhahn:
> I have opinions on modals. They're hard. We actually have `<dialog>` which is supported in all Chromium-based browsers, and we have a polyfill for Safari and Firefox—the `inert`-based approach is fine but always requires a polyfill (and see a few comments on global state) and is O(n) with the number of elements that might need to be made inert.

Here's the most recent a11y guidance about this I know of (which is admittedly not very recent): https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/

I'm cool with whatever you and @robdodson prefer here; I was just following what was already in place, TBH.

src/lib/components/Assessment/index.js
mfriesenhahn: This is to toggle the `aria-modal` property when the modal is open. Normally, you'd leave it on all the time, but since the assessment is moving itself, it shouldn't be left on when not in the modal state, I believe. (This is super weird. We shouldn't be moving this thing around, as noted in the outstanding issues. I think @robdodson has an idea for how to do this better.)

src/lib/components/BaseModalElement/index.js
mfriesenhahn: Fixed

src/lib/components/Assessment/index.js
mfriesenhahn: Woops

samthor:
Sure, I would probably (regretfully, to be honest) agree that `<dialog>` is not coming.

Having said that, if we accept Shadow DOM, then [this approach](https://twitter.com/samthor/status/1232138316654362627) also works—without polyfills. Not that I necessarily want to throw more spanners into works ;-)

src/site/content/en/handbook/self-assessment-components/index.md
```diff
@@ -0,0 +1,66 @@
...
+title: Self-assessments
+date: 2020-01-28
+description: |
+  Learn how to use web.dev's self-assessment components.
+---
+
+Self-assessments provide opportunities for users
+to check their understanding of concepts covered in your post.
+
+{% Assessment page, 'self-assessment' %}
```
robdodson: On initial readthrough I thought `'self-assessment'` was a required argument to define the _type_ of assessment. But it's really a unique identifier. Maybe rename to `'my-first-assessment'` or something to signify it's just an identifier?

mfriesenhahn: Fixed

src/lib/components/Assessment/index.js
mfriesenhahn: After considering how much markup could be copied if/when you get more complex question types (a lot!), I decided to stick with (an improved version of) the moving approach. So I think we still need to toggle `aria-modal` when the assessment is visible in its unopened state (i.e., on desktop).

src/lib/components/BaseModalElement/index.js
mfriesenhahn: I added an action but didn't add logic to prevent multiple modals at once because we already have a case where we need that: the assessment is modal on mobile, as is the question feedback form. I know nested modals sounds gross in the abstract, but when you [try it out](https://codepen.io/mfriesenhahn/full/mdygdej), I think it ends up being the least-bad option.

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,184 @@
...
+    const content = this.querySelector(".web-modal__content");
+
+    if (!content) {
+      return;
+    }
+    handleOverflow(content, "height", "web-modal--overflow", this);
+  }
+
+  getTrigger() {
+    // When the modal is opened, asign a class to the triggering element
```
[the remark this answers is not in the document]
mfriesenhahn: Fixed

mfriesenhahn:
Hey, @robdodson and @samthor, I think I've addressed all feedback. Could you PTAL--a fair amount has changed.

**Re: modals,**
I'm not sure where we stand. Seems like Sam's arguing for an approach that requires Shadow DOM, while Rob's arguing for sticking with `inert`. 

My only real horse in this race is to get this PR put to bed sooner rather than later so I can submit the other self-assessment pieces that depend on it and get them reviewed before end of the month. So I guess I'd vote for `inert` for now and maybe revisit if/when we shift to Shadow DOM. Is that cool?

mfriesenhahn:
Hey, @robdodson, I've added comments in `Assessment.js` and your fix for code block rendering in the `md-block.js` filter! Never been so happy to see a `<br>` before 😆

src/lib/components/Assessment/_styles.scss
```diff
@@ -0,0 +1,119 @@
...
+      display: flex;
+      margin: 5em 0 !important; // Override .w-callout
+      padding: 0 !important; // Override .w-callout
+    }
+  }
+}
+
+@keyframes modal-slideup {
+  0% {
+    transform: translatey(40%);
```
samthor: nit: `translateY`

src/lib/components/Assessment/_styles.scss
```diff
@@ -0,0 +1,119 @@
...
+    opacity: 1;
+  }
+}
+
+// Overrides .w-button--primary so button matches self-assessment color
+.web-assessment__open,
+.web-assessment__open:hover,
+.web-assessment__open:active,
+.web-assessment__open:focus {
+  background: $SELF_ASSESSMENT_PRIMARY_COLOR;
```
samthor: Do we need all these extra states? I imagine the background wins?

src/lib/components/Assessment/index.js
```diff
@@ -0,0 +1,232 @@
...
+      question.addEventListener("request-assessment-reset", this.reset);
+    }
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    // Close modal if viewport is >481 px so document isn't inert on desktop
+    const mqString = `(min-width: 481px)`;
+
+    matchMedia(mqString).addListener(this.onAssessmentResize);
```
samthor: I think you probably want to save the result of `matchMedia()` somewhere in the constructor (I think you're getting difference instances and the current `removeListener` does nothing), and add/remove listeners on connected/disconnected. And call `this.onAssesmentResize` when connected to establish initial state.

src/lib/components/Assessment/index.js
```diff
@@ -0,0 +1,232 @@
...
+      <button
+        @click="${this.onOpenClick}"
+        class="w-button w-button--primary web-assessment__button web-assessment__open"
+      >
+        Open quiz
+      </button>
+    `;
+    // lit-element prevents children from being duplicated,
+    // so grab setLeader text content.
+    const text = this.setLeader[0] ? this.setLeader[0].innerText : "";
```
samthor: nit: don't use `.innerText` (it has weird issues and is an IE-introduced property), always `.textContent`

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,218 @@
...
+    // (e.g., adding borders to the child element handling overflow).
+    // If the client component needs to use a different class for the element
+    // handling overflow, it will need its own animationend listener.
+    const content = this.querySelector(".web-modal__content");
+
+    if (!content) {
+      return;
+    }
+
+    const hasOverflow = handleOverflow(content, "height");
```
samthor: set directly?

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,218 @@
...
+    this.inert = true;
+    this.overflow = false;
+    this.idSalt = BaseElement.generateIdSalt("web-modal-");
+
+    this.onKeyUp = this.onKeyUp.bind(this);
+    this.onResize = this.onResize.bind(this);
+    this.onAnimationEnd = this.onAnimationEnd.bind(this);
+    this.manageDocument = this.manageDocument.bind(this);
+    this.manageFocus = this.manageFocus.bind(this);
+    this.getTrigger = this.getTrigger.bind(this);
```
samthor: you're not adding this as a listener anywhere - it doesn't need to be bound?

samthor: Same as with the `manage...` methods.

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,218 @@
...
+    if (this.open_ === val) {
+      return;
+    }
+
+    const oldVal = this.open_;
+
+    this.open_ = val;
+    if (this.open_) {
+      // Must get trigger before manipulating the DOM.
+      this.getTrigger();
```
samthor: Just save `this._triggerElement = document.activeElement`. Adding a class is needless when you can just take a ref. (Yes, that element may no longer be in the DOM when the page changes. But you unref it then.)

src/lib/components/BaseModalElement/index.js
```diff
@@ -0,0 +1,218 @@
...
+  connectedCallback() {
+    super.connectedCallback();
+    this.addEventListener("click", this.onClick);
+    // Set tabindex to -1 so modal can be focused when it's opened.
+    this.tabIndex = -1;
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener("click", this.onClick);
```
samthor: Can we close the modal on disconnection? What happens otherwise - is the `inert` state retained incorrectly?

src/lib/components/Assessment/_styles.scss
mfriesenhahn: Yes, we need these to override `.w-button--primary`.

mfriesenhahn: Made a note.

src/lib/components/BaseModalElement/index.js
mfriesenhahn: yup!

src/lib/components/Assessment/index.js
mfriesenhahn: I did everything but call `this.onAssessmentResize` when connected because (a) the assessment's initial state is established by `BaseModalElement` and (b) the assessment moves itself, so closing on connect means it can never open.

mfriesenhahn:
@samthor, I think this is ready for another review. PTAL :)

I've come up with a for-now solution to the problem of the assessment modal staying open on page change. A better solution is proposed in ticket #2415, but it'd be much more efficient if someone other than me wrote that.

mfriesenhahn:
@robdodson @samthor can y'all PTAL when you have a sec? Would really love to land self-assessments this week if possible. But I understand there's a lot going on!

src/lib/app.js
```diff
@@ -1,3 +1,4 @@
+import "./components/Assessment";
```
robdodson: This should be imported in src/lib/pages/default.js, not here

robdodson: fixed it me self

PR #2231, opened 2020-02-25
Disabled indented code blocks

mfriesenhahn:
MarkDown's default behavior of converting indented text to code blocks makes writing human-readable short codes very difficult. This PR turns off indented code blocks.


mfriesenhahn:
> I presume we don't use them anywhere and encourage authors to do the triple-backtick plus language approach.

Yes, that's covered in the handbook here: https://deploy-preview-2231--web-dev-staging.netlify.com/handbook/markup-code/#code-blocks

PR #2237, opened 2020-02-26
feat(subscribe): add element to subscribe to newsletter to the bottom of posts, fixes #2109

### LINKED ISSUE #2109: Add newsletter sign-up form to the bottom of posts
Currently we use web.dev/subscribe to let folks sign up for a newsletter. This redirects to a Google service that handles collecting and storing email addresses in a privacy compliant way. However, this same service supports an ajax API. We could POST the data to the service from our own form which would give us much more control over the styling of the form.

We add our own newsletter sign-up form to the bottom of posts and deprecate web.dev/subscribe.

MichaelSolati:
Fixes #2109


src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+      submitted: {
+        type: Boolean,
+      },
+    };
+  }
+
+  constructor() {
+    super();
+    this.checkboxes = ["WebDevNewsletter", "collects-pii-spii-checkbox"];
+    this.needsDoubleOptIn = [
```
samthor: I was about to ask why we have a country list, and I think now I understand.

Also, I'm confused this isn't the entire EU.

src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+
+    form.append("LanguagePreference", "en-US");
+    form.delete(this.robotName);
+
+    return form;
+  }
+
+  submit(form) {
+    return new Promise((resolve, reject) => {
+      const request = new XMLHttpRequest();
```
samthor: can we just `fetch`?

src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+
+    if (this.errors) {
+      if (typeof this.errors === "string") {
+        errorMessage = html`
+          <div class="web-subscribe-error">
+            <p>${this.errors}</p>
+          </div>
+        `;
+      } else {
+        const errorMessages = Object.values(this.errors).map(
```
samthor: Isn't `this.errors` going to contain a single `Error`? This makes it seem like it's an `Object` containing many. But on line 91 you're just setting one.

src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+        errorMessage = html`
+          <div class="web-subscribe-error">
+            ${errorMessages}
+          </div>
+        `;
+      }
+    }
+
+    const countriesOptions = countries.map((country) => {
+      return country[1] === "United States"
```
samthor: I'd love to use [geoIP headers](https://cloud.google.com/appengine/docs/flexible/dotnet/reference/request-headers) to choose the correct initial selection (I get that a user might change it).

src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+            <option value="${country[0]}">${country[1]}</option>
+          `;
+    });
+
+    return this.submitted
+      ? ""
+      : html`
+          <form @submit="${this.handleSubmit}">
+            <div class="web-subscribe-field">
+              <input
```
samthor: Does the subscribe form require two fields for Name? Names come in many shapes and sizes.

src/lib/components/SubscribeForm/index.js
MichaelSolati: Not all EU countries require this (maybe?) If you look at the source on https://services.google.com/fb/forms/web-dev-subscribe/ you'll find the following:

```JavaScript
const needsDoubleOptIn = [
  'AT: Austria',
  'DE: Germany',
  'GR: Greece',
  'LU: Luxembourg',
  'NO: Norway'
];
document.addEventListener('DOMContentLoaded', () => {
  const countrySelect = document.querySelector('#f-Country');
  const webDevOptIn = document.querySelector('#f-WebDevNewsletter');
  const chromeNewsOptIn = document.querySelector('#f-WebDevChromeNewsUpdates');
  countrySelect.addEventListener('change', (e) => {
    const chosenCountry = e.target.value;
    if (needsDoubleOptIn.includes(chosenCountry)) {
      webDevOptIn.setAttribute('value', 'Unconfirmed');
      chromeNewsOptIn.setAttribute('value', 'Unconfirmed');
    } else {
      webDevOptIn.setAttribute('value', 'True');
      chromeNewsOptIn.setAttribute('value', 'True');
    }
  });
});
```

So I was trying to do something similar.

src/lib/components/SubscribeForm/index.js
MichaelSolati: It's also based on the current form which asks for first/last, I'm not sure who would be the right person to ask this.

src/lib/components/SubscribeForm/index.js
MichaelSolati: I don't think we can control that. It's a formbox endpoint, so I'm not sure what could be done here.

src/lib/components/SubscribeForm/index.js
MichaelSolati: So when an invalid submission is made you'll receive an object with a key for each field that's invalid. I've refactored it to take the object and get all the errors out of it. This may change so that instead of listing all the errors in the bottom of the form to show the error message for the field that was invalid.

src/lib/components/SubscribeForm/index.js
MichaelSolati: consider it done

src/lib/components/SubscribeForm/index.js
samthor: We get access to the user's geolocation inside `server.js` as a header, we can do something with it there—perhaps pipe it back to this code via cookie, HTTP endpoint, ...

src/lib/components/SubscribeForm/index.js
samthor: Sure, if we're just dialing up the form we can't change this requirement.

samthor:
So I'm concerned we're reimplementing functionality of something we fundamentally don't really control (and accepting risks of it changing from under us). Is it really such a huge deal that the form is external? Could we put it in an `<iframe>`?

src/lib/components/SubscribeForm/index.js
robdodson: This list are the only countries that require double-opt in. When they sign up we send an additional email to ask if they really meant to sign up and they have to confirm that as well. It's one of the requirements we have to follow.

src/lib/components/SubscribeForm/_styles.scss
```diff
@@ -0,0 +1,79 @@
+@import '../../../styles/settings/colors';
+@import '../../../styles/tools/breakpoints';
+
+web-subscribe-form {
+  display: block;
+  margin-top: 32px;
+  text-align: center;
+  width: 100%;
+
+  button {
```
robdodson: I think these will need BEM selectors like .web-subscribe-form__button. We try to avoid styling generic elements with descendant selectors because it can lead to specificity battles and accidental cascades.

src/site/content/en/_redirects.yaml
```diff
@@ -5,11 +5,6 @@ redirects:
 - from: /policies
   to: https://policies.google.com/
 
-# Newsletter
```
robdodson: I'm not sure if it's safe to remove this redirect. I think we have existing places where `web.dev/subscribe` has been published and if we remove it those will 404.

One option would be to keep the redirect as is for now, and if we do a dedicated newsletter sign-up page, change it to point at that.

src/lib/components/SubscribeForm/index.js
```diff
@@ -0,0 +1,242 @@
...
+          `
+        : html`
+            <option value="${country[0]}">${country[1]}</option>
+          `;
+    });
+
+    return this.submitted
+      ? ""
+      : html`
+          <form @submit="${this.handleSubmit}">
```
robdodson: To match the BEM selectors we use elsewhere in the site I think you'll want to add classes to these elements.

The block could be `.web-subscribe` (might also consider changing the custom element name to match).
And the elements would be `.web-subscribe__form, .web-subscribe__field, .web-subscribe__input, etc.`

robdodson:
> So I'm concerned we're reimplementing functionality of something we fundamentally don't really control (and accepting risks of it changing from under us).

It's a valid concern. My understanding is that the ajax form path is one of their documented/supported approaches so it (ideally) shouldn't change out from under us. Also, I believe the project is on maintenance mode and they don't have plans to modify it.

> Is it really such a huge deal that the form is external? Could we put it in an <iframe>?

The main issue I've had with the form is that we don't have much control over the HTML it renders, so styling it is always a bit wonky. Essentially we're allowed to inject code into the `<head>` of the form page via their CMS so that's how we style and add script.

Implementing the form ourselves gives us full control (which is nice) but it definitely adds some code weight. I thought that's what https://developer.android.com/subscribe had done, but only recently realized theirs was an iframe.

@MichaelSolati what're your thoughts?

MichaelSolati:
@robdodson I do think the experience ends up looking better with the form element we have rather than embedding an iframe. The iframe model would work, but it just doesn't look nice. Also super biased since I did the work, but I do really think it looks nice 😁

robdodson:
Based on that email thread with Harleen, it sounds like we might pivot the direction a bit such that we have a dedicated page with the newsletter sign up form and an archive of previous newsletters. Similar to http://developers.android.com/newsletter.

I think if we go the dedicate page route them I feel much better about shipping our own form because it means we'll only include the JS for that form on that one page.

In terms of next steps, I think @MichaelSolati should work with Jiwoong to come up with a design for the "sign up for our newsletter" button that we would include at the bottom of posts, and also to work with him on a design for the newsletter page.

MichaelSolati:
@robdodson or @samthor if I can get a new LGTM, also let me give y'all an update:

1. The form now works with JS disabled, the code for the HTML form should look and be simple enough.
2. What I mentioned before works because the form is actually a partial, which I wrap in a custom element that will handle the submission IF JS is enabled.
3. If JS is enabled I intercept the form element's submit event and cleans the form a little bit. For example, if JS is disabled then the value of checkboxes is set to `Unconfirmed` rather than `True` (double opt in by default is safer incase let's say a German submits the form with JS disabled). Also we have a check to ensure that a robot is not submitting the form which runs with JS.
4. Also if JS is enabled we will update the content inside of the element to either provide error feedback or a success message while removing the form so that they can not resubmit it.

src/site/_includes/partials/paged.njk
```diff
@@ -47,3 +47,5 @@
   {% endif %}
 
 </nav>
+<web-actions actions="subscribe"></web-actions>
+{% include 'partials/subscribe.njk' %}
```
robdodson: I think we wouldn't have the subscribe form at the bottom of posts now. Instead there would just be a button at the bottom of pages that would link to /newsletter, which is where the subscribe form would be embedded.

src/site/_includes/partials/post.njk
```diff
@@ -134,4 +134,5 @@
       {% endif %} {% endfor %}
     </ul>
   </div> #}
+  {% include 'partials/subscribe.njk' %}
```
robdodson: Same point as above.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,86 @@
+@import '../settings/colors';
+@import '../tools/breakpoints';
+
+.w-subscribe {
```
robdodson: Might need to enforce some dimensions on the heading so it doesn't touch the edges on smaller screen sizes. At around 850px it touches the edges of the screen.

robdodson: Also I think we're not quite matching the design of Jiwoong's mocks.
https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZciehgH2S_Ws/files/MCGT2l9AragxTMzZaGwrrRpy

src/site/_includes/partials/subscribe.njk
```diff
@@ -0,0 +1,66 @@
+<web-subscribe id="subscribe">
+  <div class="w-subscribe">
+    <div>
+      <h1>Become a better web developer.</h1>
+      <p id="w-subscribe__message">Stay up to date with the latest about web platform.</p>
```
robdodson: We have some CSS that we use to create spacing around elements with IDs. This is so when someone deep links to an id, the header doesn't overlap the content. In this case it's making additional space around the `p` element. So if you don't need the `id` it might be good to remove it to keep things tidy.

src/site/_includes/partials/subscribe.njk
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+    <form class="w-subscribe__form" action="{{ site.subscribeForm }}" method="post">
+      <div class="w-subscribe__fields">
+        <div class="w-subscribe__field">
+          <input class="w-subscribe__input" id="sub-firstname" name="FirstName" placeholder="First Name" required
```
robdodson: Lighthouse will grump at us because technically these elements aren't labeled for accessibility. You can create visually hidden labels (https://css-tricks.com/small-tweaks-can-make-huge-impact-websites-accessibility/#article-header-id-2) and add those to these elements.

```
<input class="w-subscribe__input" id="sub-firstname">
<label class="visually-hidden" for="sub-firstname">First name</label>
```

I'd suggest creating a new file for the visually-hidden helper class in `styles/tools/_helpers.scss`.

src/styles/components/_subscribe.scss
MichaelSolati: I don't think I had seen this design before, will update to look like this then.

src/site/_includes/partials/subscribe.njk
MichaelSolati: The ids were used for query selectors in the Subscribe custom element, do you have any recommendations as to how I should do that? (By tag or class?)

src/styles/components/_subscribe.scss
robdodson: oh sorry. I think it's linked from the slides deck where we were discussing the design.

robdodson: the things I was looking for us to match were around the styling of the input elements (padding, border color/contrast). The checkbox styles and their spacing.

src/site/_includes/partials/subscribe.njk
robdodson: I'd suggest doing it by class.

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+ * Element that renders newsletter subscription form.
+ *
+ * @extends {BaseElement}
+ * @final
+ */
+class Subscribe extends BaseElement {
+  constructor() {
+    super();
+
+    this.form = this.querySelector("form");
```
robdodson: Suggest using `.w-subscribe__form` here. I figure we won't have more than one form inside of this element but may as well play it safe.

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+    this.subscribeError = this.querySelector("div#w-subscribe__error");
```
robdodson: Don't think you need `div` in this selector. I think using either the id or a class is fine here. I tend to never use id's unless I'm using shadow dom but that's kinda just me being dogmatic :P

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+    this.subscribeMessage = this.querySelector("p#w-subscribe__message");
```
robdodson: Can remove the `p` and either go with an id or class here.

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+      body,
+    }).then((r) => r.json());
+  }
+
+  onError(errors) {
+    this.subscribeError.innerHTML = "";
+    if (errors) {
+      Object.values(errors).forEach((e) => {
+        const pTag = document.createElement("p");
+        pTag.innerText = typeof e === "string" ? e : e.join(" ");
```
robdodson: This seems like an area where having slots and shadow DOM would be helpful because then we could have LitElement render this for us...

I'm anxious about creating DOM on the fly and inserting it into the page, especially if the text we're inserting is coming from an external service. Feels like an XSS vector. I don't know if it's overkill but one option would be to make `w-subscribe__error` into a little custom element that just renders whatever array of errors we give it. @samthor — any better suggestions?

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+      "GR: Greece",
+      "LU: Luxembourg",
+      "NO: Norway",
+    ];
+    this.robotName = "is-it-just-me-or-was-this-form-filled-out-by-a-robot";
+    this.submissionUrl = this.form.action;
+    this.processing = false;
+    this.submitted = false;
+
+    this.form.addEventListener("submit", this.onSubmit.bind(this));
```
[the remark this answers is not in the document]
samthor: You in fact _have_ to do this, because each bind gives you back a new object. You need to assign it once in a ctor, as Rob's suggestion.

src/lib/components/Subscribe/index.js
samthor: Also `.innerText` is harmful. Don't use it- use `.textContent` instead.

This isn't XSS since we're not setting `.innerHTML`.

I think I don't want to block Micheal from getting this PR in. I think we revisit the Shadow DOM approach soon and render in lit with its in-built array support.

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+  postForm(body) {
+    return fetch(this.submissionUrl, {
+      method: "POST",
+      body,
+    }).then((r) => r.json());
+  }
+
+  onError(errors) {
+    this.subscribeError.innerHTML = "";
```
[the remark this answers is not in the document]
robdodson: Oops. ignore this comment. I deleted some of the related comments once I understood what you were using innerHTML for. But _do_ read Sam's comment below.

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,119 @@
...
+        pTag.innerText = typeof e === "string" ? e : e.join(" ");
+        this.subscribeError.appendChild(pTag);
+      });
+    }
+  }
+
+  onSubmit(e) {
+    e.preventDefault();
+    if (this.processing || this.submitted) {
+      return false;
```
robdodson: any reason to return false instead of just returning here?

src/lib/components/Subscribe/index.js
```diff
@@ -0,0 +1,119 @@
...
+    if (formIsRobot) {
+      this.onSuccess();
+      return false;
+    }
+    const cleanedForm = this.cleanForm(form);
+
+    this.postForm(cleanedForm)
+      .then((response) => {
+        if (response && response.result === "accepted") {
+          this.onSuccess();
```
robdodson: To be safe I would suggest binding `this.onSuccess` and `this.onError` in the constructor, similar to what you did for `this.onSubmit`. Technically you don't need it because the value of `this` inside of the `onSubmit` function is pointing at the instance that has `onSuccess` and `onError` defined on it, but if that ever changed then it might appear to break mysteriously.

src/site/content/en/fast/fast.11tydata.js
```diff
@@ -71,7 +71,7 @@ module.exports = {
           "defer-non-critical-css",
           "minify-css",
           "extract-critical-css",
-          "optimize-css-background-images-with-media-queries"
+          "optimize-css-background-images-with-media-queries",
```
robdodson: Ah I think I fixed this in a separate commit, you can reset this change.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  display: block;
+  margin-top: 32px;
```
robdodson: Let's remove this and have the newsletter page define the spacing / margining for all of this child elements. Might be a good time to start introducing [the lobotomized owl selector](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/).

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  text-align: center;
+  width: 100%;
```
robdodson:
```suggestion

```

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
+@import '../settings/colors';
+@import '../tools/breakpoints';
+
+.w-subscribe {
+  display: block;
+  margin-top: 32px;
+  text-align: center;
```
robdodson: It seems like you don't want this here. Instead you probably want to move this to the div that's the first child of w-subscribe and contains the h1. That seems to be the only place where you actually want centered text.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  color: $WARNING_COLOR;
+}
+
+.w-subscribe__field {
+  display: flex;
+  justify-content: space-between;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 8px;
+  text-align: left;
```
robdodson:
```suggestion
  margin-top: 8px;
```

robdodson: Note, I don't think you want this margin on the first set of fields which appear inside of the w-subscribe__fields element. You probably want that element to remove the top margin from its children.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.w-subscribe__form {
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 720px;
+  width: calc(100% - 32px);
```
robdodson: I'm not sure if you need this?

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 8px;
+  text-align: left;
+}
+
+.w-subscribe__form {
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 720px;
```
robdodson: I'm not sure if you need this. Based on the design, it seems like the header area gets constrained, but perhaps the fields go full width?

image

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  justify-content: space-between;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 8px;
+  text-align: left;
+}
+
+.w-subscribe__form {
+  margin-left: auto;
+  margin-right: auto;
```
robdodson: Can remove these if we decide to make the fields full bleed. Based on the design below I _think_ they're meant to be full width. We might have to fiddle with it a bit to see what we prefer.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  max-width: 720px;
+  width: calc(100% - 32px);
+}
+
+.w-subscribe__input {
+  border: 2px solid #f1f3f4;
```
robdodson:
```suggestion
  border: 1px solid $GREY_300;
```

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  border-top-left-radius: 2px;
+  border-top-right-radius: 2px;
```
robdodson:
```suggestion
```

Suggest removing these

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  font: inherit;
+  font-size: 16px;
```
robdodson: hmmmm... i wonder if we should define these in a generic/_input.scss file? Seems like we'd want all inputs to have consistent typography. We'd also probably want the font-size to follow whatever body copy is doing at different viewport sizes. This is my fault for not specifying a better default here...

Let's at least not set an explicit font-size. You can leave the font: inherit for now.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  line-height: 20px;
```
robdodson: Do you need this?

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  padding: 8px;
```
robdodson: The design seems to call for more padding than this.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,79 @@
...
+  width: 100%;
+
+  &:focus {
+    box-shadow: 0 2px 0 0 $WEB_SECONDARY_COLOR;
```
robdodson: I think the design calls for a purple inner shadow

image

Don't worry about the design of the popout style for the select. I recognize that styling select is darn near impossible :P

src/styles/components/_subscribe.scss
MichaelSolati: I won't be able to apply padding to the inner text of the select element though, so it'll look like they're not lined up properly.

src/site/_includes/components/Checkbox.js
```diff
@@ -0,0 +1,57 @@
...
+            <input
+              id="${this.id}"
+              name="${this.name}"
+              ${this.required && 'required'}
+              ${this.value && `value="${this.value}"`}
+              class="w-checkbox__input"
+              type="checkbox"
+            />
+          </div>
+          <div>
```
robdodson, line 48: This `div` doesn't seem to be doing anything. can you remove it?

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
...
+.w-checkbox__input {
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  appearance: none;
+  background: url('/images/icons/check_box_outline_blank.svg');
+  border: 2px solid $WHITE;
+  border-radius: 4px;
+  cursor: pointer;
+  height: 28px;
+  margin: 0px 30px 0px 0px;
```
robdodson, line 18: suggest changing this to something like `margin: 3px` or some default value that wraps all the way around, similar to the native checkbox.

ideally components shouldn't define their own margins, they should be set by the component or page that's using them. otherwise you end up needing a lot of overrides.

I think in this case, having a default margin that wraps all the way around, or no margin at all, seem ok as both of those would match the expectations for a checkbox. But just having it on one side feels a little too bespoke to how we're using it on this particular page.

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
+@import '../settings/colors';
+
+.w-checkbox__label {
+  cursor: pointer;
```
robdodson, line 4: Don't need this. Form labels typically don't do the pointer cursor on hover.

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
...
+  display: inline-flex;
+  position: relative;
```
robdodson, line 6: This doesn't seem to be doing anything. I think you can remove it.

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
+@import '../settings/colors';
+
+.w-checkbox__label {
```
robdodson, line 3: I'm not sure if this class is quite what we want, it seems too specific to checkbox and this one use case. Something more generic and reusable might be better.

If we want a generic class that makes all labels `inline-flex` then we should add it to `label` in the `generic/*` directory. But i'm not sure if we want that either...

Another option is to create some flexbox utility classes in the `overrides/*` directory. That may be a better approach.

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
...
+  -webkit-appearance: none;
+  appearance: none;
+  background: url('/images/icons/check_box_outline_blank.svg');
+  border: 2px solid $WHITE;
+  border-radius: 4px;
+  cursor: pointer;
+  height: 28px;
+  margin: 0px 30px 0px 0px;
+  outline: none;
+  transition-duration: .2s;
```
robdodson, line 20: Are we using this on our other focus styles? Feels a bit odd to me.

btw we do have a variable for this in `settings/gobal.scss`

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
...
+  display: inline-flex;
+  position: relative;
+}
+
+.w-checkbox__input {
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  appearance: none;
+  background: url('/images/icons/check_box_outline_blank.svg');
+  border: 2px solid $WHITE;
```
robdodson, line 14: Can we make this transparent instead of white? Is it needed because the focus style adds a border and you don't want a visual shift?

src/styles/components/_checkbox.scss
```diff
@@ -0,0 +1,30 @@
...
+  cursor: pointer;
+  height: 28px;
+  margin: 0px 30px 0px 0px;
+  outline: none;
+  transition-duration: .2s;
+  width: 28px;
+}
+
+.w-checkbox__input:checked {
+  background: url('/images/icons/check_box_tertiary.svg');
```
robdodson, line 25: One thing I noticed is that this image requires a network request so you see a flash of unstyled content the first time you click the checkbox and the svg loads. Might be better to base64 encode these and include as a data uri. @samthor wdyt?

src/site/_includes/components/Checkbox.js
```diff
@@ -0,0 +1,57 @@
...
+
+const {html} = require('common-tags');
+
+/* eslint-disable require-jsdoc,indent,max-len */
+
+/**
+ * Checkbox used to forms.
+ * @param {Object} arg Data for the checkbox.
+ */
+class Checkbox {
```
robdodson, line 25: Since you were able to style the input without needing a wrapping element I'd suggest getting rid of the shortcode entirely. The reason is because this implementation really tightly couples the label and the checkbox in a way that's not generally reusable (only works for specific layouts).

robdodson:
@MichaelSolati I can't recall if this PR contains it, but please be sure to add /newsletter/ to the percy tests. It may also be good to add a checkbox and label example to /handbook/web-dev-components.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,83 @@
...
+  &.w-subscribe__padded--major {
+    max-width: 540px;
+  }
+
+  &.w-subscribe__padded--minor {
+    max-width: 720px;
+  }
+}
+
+.w-subscribe__checkbox {
```
robdodson, line 48: I don't think this is used so you can remove it.

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,83 @@
+@import '../settings/colors';
+@import '../tools/breakpoints';
+
+.w-subscribe {
+  display: block;
+  margin-top: 32px;
+}
+
+.w-subscribe__border {
```
robdodson, line 9: If this is only used to give select a border maybe rename to .w-subscribe__select-border

src/styles/components/_subscribe.scss
```diff
@@ -0,0 +1,83 @@
...
+.w-subscribe__error {
+  color: $WARNING_COLOR;
+}
+
+.w-subscribe__field {
+  display: flex;
+  margin-top: 8px;
+}
+
+.w-subscribe__padded {
```
robdodson, line 34: Staring at this for a bit... I'm wondering if you need these classes? It seems like they're primarily used to center stuff. Could we use flex box to center align the three major sections and set a max-width on them as needed?

src/site/content/en/authors/index.njk
```diff
@@ -16,7 +16,6 @@ pagination:
 <div class="authors-landing-page">
   {# Make sure Actions are first in the tab order. #}
   <web-actions actions="subscribe"></web-actions>
-
   <div class="w-layout-container w-pb--non">
```
robdodson, line 19: I think these are just whitespace changes. Might be able to fix them to cut down on noise in the PR. Otherwise they'll show up in git blame.

PR #2244, opened 2020-02-27
clear scroll restoration before unload

samthor:
cc @philipwalton 

### an aside

This actually isn't ideal and I played with not using scroll restoration, and instead doing something more like the polyfills for this feature do—basically when the browser does a popstate, two things happen:

1. `popstate`
2. (later, not sure how far) `scroll` (to restore position)

So you can basically retain the scroll position in the first event, immediately revert that position on the scroll, then once your async loading code is done, set the position as it was intended by `scroll`.

But the behavior seems pretty inconsistent.

### DIFF
--- a/src/lib/utils/router.js
+++ b/src/lib/utils/router.js
@@ -5,6 +5,11 @@ let recentActiveUrl; // current URL not including hash
 // async, and most browsers will move the scroll position automatically for us, even on old content.
 // Instead, we call `scrollOnFrame` when the async load helper is done.
 window.history.scrollRestoration = "manual";
+window.addEventListener("beforeunload", (e) => {
+  // ... but re-enable when the page is unloaded, which happens if a user refreshes the page using
+  // their browser. This prevents this type of reload from jumping back to the top of the viewport.
+  window.history.scrollRestoration = "auto";
+});
 
 /**
  * @return {string} URL pathname plus optional search part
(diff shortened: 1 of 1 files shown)


src/lib/utils/router.js
```diff
@@ -5,6 +5,11 @@ let recentActiveUrl; // current URL not including hash
 // async, and most browsers will move the scroll position automatically for us, even on old content.
 // Instead, we call `scrollOnFrame` when the async load helper is done.
 window.history.scrollRestoration = "manual";
+window.addEventListener("beforeunload", (e) => {
```
philipwalton: I'm not sure adding an unconditional `beforeunload` handler is the right solution here (as it can [break bfcache in browsers](https://developers.google.com/web/updates/2018/07/page-lifecycle-api#the-beforeunload-event)). Would it work if you used the `pagehide` event?

samthor: TIL `pagehide`. Yes this fixes it.

philipwalton:
I can confirm that the issue I was experience is not happening when I look at the staging link, but I'm not sure familiar with the `scrollRestoration` API, so I probably can't offer much more feedback than what I already said.

samthor:
Does that count as a LGTM? :)

samthor:
yeah I need a code owner. Thanks anyway!

PR #2246, opened 2020-02-27
workbox to 5.0.0, inline

samthor:
Part of #1370.

Upgrades to Workbox 5, and inlines the code.

### DIFF
--- a/build.js
+++ b/build.js
@@ -125,6 +125,8 @@ async function build() {
     input: "src/lib/sw.js",
     plugins: [
       // Workbox uses a variable known to be undefined (!) and forces this plugin to be used.
+      // TODO(samthor): This generates statements like "if (1 !== 1)", which are NOT removed from
+      // the final bundle code. Terser/Rollup don't strip them.
       rollupPluginReplace({
         "process.env.NODE_ENV": JSON.stringify("production"),
       }),
--- a/build.js
+++ b/build.js
@@ -23,6 +23,7 @@ const log = require("fancy-log");
 const rollupPluginNodeResolve = require("rollup-plugin-node-resolve");
 const rollupPluginCJS = require("rollup-plugin-commonjs");
 const rollupPluginVirtual = require("rollup-plugin-virtual");
+const rollupPluginReplace = require("rollup-plugin-replace");
 const rollup = require("rollup");
 const terser = isProd ? require("terser") : null;
 const {getManifest} = require("workbox-build");
@@ -123,6 +124,10 @@ async function build() {
   const swBundle = await rollup.rollup({
     input: "src/lib/sw.js",
     plugins: [
+      // Workbox uses a variable known to be undefined (!) and forces this plugin to be used.
+      rollupPluginReplace({
+        "process.env.NODE_ENV": JSON.stringify("production"),
+      }),
       rollupPluginVirtual({
         "cache-manifest": `export default ${JSON.stringify(
           manifest,
--- a/package.json
+++ b/package.json
@@ -44,12 +44,18 @@
     "iso-639-1": "^2.1.0",
     "js-yaml": "^3.13.1",
     "lit-element": "^2.2.0",
+    "rollup-plugin-replace": "^2.2.0",
     "rollup-plugin-virtual": "^1.0.1",
     "terser": "^4.3.9",
     "unistore": "^3.4.1",
     "wicg-inert": "^3.0.1",
     "workbox-build": "^5.0.0",
-    "workbox-expiration": "^5.0.0"
+    "workbox-cacheable-response": "^5.0.0",
+    "workbox-expiration": "^5.0.0",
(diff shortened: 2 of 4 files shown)


src/lib/sw.js
```diff
@@ -161,7 +163,7 @@ self.addEventListener("fetch", (event) => {
 
   const p = Promise.resolve().then(async () => {
     // First, check if there's actually something in the cache already.
-    const cacheKey = workbox.precaching.getCacheKeyForURL(
+    const cacheKey = workboxPrecaching.getCacheKeyForURL(
```
jeffposnick: https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v4#reading_precached_responses

In v5, you can simplify these multiple statements to

```
import {matchPrecache} from 'workbox-precaching';

// ...

const cached = await matchPrecache(u.pathname + "/index.html");
```

src/lib/sw.js
```diff
@@ -187,16 +189,14 @@ self.addEventListener("fetch", (event) => {
...
   // Destination is set by loading this content normally; it's not set for fetch(), so look for our
   // custom header.
   const isDocumentRequest =
     event.request.destination === "document" ||
     event.request.headers.get("X-Document");
   if (isDocumentRequest) {
-    const cacheKey = workbox.precaching.getCacheKeyForURL(
-      "/offline/index.html",
-    );
+    const cacheKey = workboxPrecaching.getCacheKeyForURL("/offline/index.html");
```
jeffposnick: You might be able to use https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-v4#reading_precached_responses here as well.

package.json
```diff
@@ -48,7 +48,8 @@
     "terser": "^4.3.9",
     "unistore": "^3.4.1",
     "wicg-inert": "^3.0.1",
-    "workbox-build": "^4.3.1"
+    "workbox-build": "^5.0.0",
+    "workbox-expiration": "^5.0.0"
```
jeffposnick: `workbox-build` results in a transitive dependency on all the Workbox runtime libraries, so they all end up in `node_modules/`, but I think it's better practice to explicitly list all the dependencies here. (Right now you're only listing `workbox-expiration`, so that would be `workbox-precaching`, `workbox-routing`, etc.)

src/lib/sw.js
samthor: This doesn't work as we're not _precaching_ offline in development. We're just checking the runtime cache—it looks like `matchPrecache` does what it says on the tin, only matches precache, not runtime :)

jeffposnick: Sorry, I'm not clear what you mean here—I commented on a line that was previously calling `workbox.precaching.getCacheKeyForURL()`, which leads me to believe that you were checking for a precached entry, not a runtime cached entry?

If your point was that it might be a cache miss, that's okay—`matchPrecache()` will just return `null` in that scenario, and you can check for that.

jeffposnick:
CC: @philipwalton regarding the Rollup `process.env.NODE_ENV` feedback. (`webpack` [automatically replaces](https://webpack.js.org/configuration/mode/) `process.env.NODE_ENV` for you.)

I've seen a number of folks get caught up by this in Rollup so far. It's called out in https://developers.google.com/web/tools/workbox/guides/using-bundlers#keeping_dev-only_code_out_of_the_bundle, but maybe we can do more to explain it to folks before it becomes a bigger frustration? Or alternatively, encourage the usage of the Rollup Workbox plugins that the community has developed (which, I haven't checked if they handle the replacement automatically, but we should tell them to if they don't).

In case it helps, this is the Rollup config I've been using for production minification and splitting off the runtime code for Workbox into its own file: https://github.com/jeffposnick/jeffposnick.github.io/blob/active/rollup.config.js

samthor:
Yup, that's what I've done. Sorry for any confusion. PTAL at the source for
a better explanation.

samthor:
I'm going to call this a practical review and mege.

build.js
```diff
@@ -123,6 +124,12 @@ async function build() {
   const swBundle = await rollup.rollup({
     input: "src/lib/sw.js",
     plugins: [
+      // Workbox uses a variable known to be undefined (!) and forces this plugin to be used.
+      // TODO(samthor): This generates statements like "if (1 !== 1)", which are NOT removed from
```
philipwalton: Do you have a repro for this? I don't see this issue in our production builds, so I'm wondering what's different between your Rollup builds and ours.

samthor: So I actually worked out the issue, but it's still fairly confusing.

* Rollup ignores the `if` statements and spits them out regardless.
* Terser on the _command line_ seems to retain the statements (which is what I was testing and was confused by).
* Terser, programatically, deletes them.

It seems like programmatically, Terser has the "conditionals" boolean set, but on the CLI, it does not.

Let me know if that matches your understanding. I can remove or update this comment.

PR #2253, opened 2020-02-28
fix broken sw

samthor:
I broke the SW in #2246.

I returned an undefined variable at the bottom of the catch handler.


samthor:
I'm going to TBR this once the tests pass.

PR #2258, opened 2020-02-28
feat(canonical): add support for canonical tags on pages, fixes #1035, fixes #2204

### LINKED ISSUE #1035: Enable blog post syndication.
Sometimes folks on our team publish on other platforms (Medium, personal blogs, etc). It would be great if we could syndicate those posts to web.dev so folks subscribed to our RSS feed see that content.

Someone should be able to add a link[rel=canonical] tag to a blog post so they can syndicate it on our site.

### LINKED ISSUE #2204: non-canonical pages with i18n
All content can now be accessed at both:

https://web.dev/en/about/
https://web.dev/about/

And we don't have any `rel="canonical"` tags.

MichaelSolati:
Fixes #1035 #2204

By adding a `canonical` field to the head of a file will add a `rel=canonical` to the head of the HTML.

@devnook @samthor PTAL, it's a small feature

### DIFF
--- a/src/site/_includes/components/Meta.js
+++ b/src/site/_includes/components/Meta.js
@@ -108,11 +108,12 @@ module.exports = (locale, page, collections, renderData = {}) => {
   }
 
   function renderCanonicalMeta() {
-    if (pageData.canonical) {
-      return html`
-        <link rel="canonical" href="${pageData.canonical}" />
-      `;
-    }
+    return html`
+      <link
+        rel="canonical"
+        href="${pageData.canonical ? pageData.canonical : site.url + pageUrl}"
+      />
+    `;
   }
 
   // prettier-ignore
--- a/src/site/_includes/components/Meta.js
+++ b/src/site/_includes/components/Meta.js
@@ -107,11 +107,20 @@ module.exports = (locale, page, collections, renderData = {}) => {
     `;
   }
 
+  function renderCanonicalMeta() {
+    if (pageData.canonical) {
+      return html`
+        <link rel="canonical" href="${pageData.canonical}" />
+      `;
+    }
+  }
+
   // prettier-ignore
   return html`
     <title>${pageData.title || pageData.path.title}</title>
     <meta name="description" content="${pageData.description || pageData.path.description}" />
 
+    ${renderCanonicalMeta()}
     ${renderGoogleMeta()}
     ${renderFacebookMeta()}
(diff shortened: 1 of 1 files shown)


samthor:
> By adding a `canonical` field to the head of a file 

We don't have any pages with a "canonical" field in the front matter yet, right? (this is what you mean? There's the HTML head, the front matter...)

If so, I don't think we should merge this, since it won't be used implicitly, and we'd have to add that field to _every_ page. I'd rather do something automated since we know the target URL.

Or am I misunderstanding?

MichaelSolati:
> > By adding a `canonical` field to the head of a file
> 
> We don't have any pages with a "canonical" field in the front matter yet, right? (this is what you mean? There's the HTML head, the front matter...)
> 
> If so, I don't think we should merge this, since it won't be used implicitly, and we'd have to add that field to _every_ page. I'd rather do something automated since we know the target URL.
> 
> Or am I misunderstanding?

I believe the intention is for authors to be able to add their own canonical links if we're hosting a post for an author. So if I posted on medium and wanted to cross-post and web.dev I could use the medium link as the canonical url.

samthor:
Ah, I understand now. I *also* want an implicit default canonical URL
(solving #2204).

This is fine, but maybe we should also update the handbook or something to
make it clear this is how it works.

robdodson:
Should we go ahead and add a default `canonical` to pages in this PR? I think `Meta.js` knows the page's url so it could use that if one isn't provided in the front matter.

robdodson:
looks like it might need to be rebased against master to fix the conflict

MichaelSolati:
Removed the conflict with the loader. We should be good now.

mfranzke:
@MichaelSolati I can't find the rel="canonical" tags being rendered on the website (like e.g. the homepage https://web.dev/ or https://web.dev/contact-picker/) at all (whereas it is being rendered on the preview for this issue https://deploy-preview-2258--web-dev-staging.netlify.app/).

Before opening an issue for that, is there (a known and unknown to me) reason that for, like e.g. that the code change out of this ticket hasn't made it to PROD so far?

MichaelSolati:
@mfranzke when you browse through the site with JS enabled tags do t get rendered, but it should work with JS disabled (for crawlers)

mfranzke:
@michaelsolati, thanks for the clarification, I just wasn‘t aware of this differentiation.

PR #2270, opened 2020-03-02
Make caches named & bounded

samthor:
We were growing our caches forever. Let's not do that.

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -91,11 +91,15 @@ initializeGoogleAnalytics();
 
 const externalExpirationPlugin = new ExpirationPlugin({
   maxAgeSeconds: 60 * 60 * 24 * 365, // 1 yr
-  maxEntries: 30,
 });
 
-const localExpirationPlugin = new ExpirationPlugin({
-  maxAgeSeconds: 60 * 60 * 24 * 21, // 21 days
+const contentExpirationPlugin = new ExpirationPlugin({
+  maxEntries: 50, // store the most recent ~50 articles
+});
+
+const assetExpirationPlugin = new ExpirationPlugin({
+  maxAgeSeconds: 60 * 60 * 24 * 7, // 1 wk
+  maxEntries: 100, // allow a large number of images, but expire quickly
 });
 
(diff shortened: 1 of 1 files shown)


src/lib/sw.js
```diff
@@ -89,11 +89,21 @@ self.addEventListener("activate", (event) => {
 
 initializeGoogleAnalytics();
 
+const externalExpirationPlugin = new ExpirationPlugin({
```
jeffposnick: Here's how I think about configuring `ExpirationPlugin`:

- Is the runtime cache something that could potentially grow quickly, because there are a fairly unique set of URLs that might be requested as the user navigates around my web app? If so, set `maxEntries`, even if it's to something fairly high, just as a safeguard.

- Is the runtime cache storing data that might not be relevant or inaccurate after a specific amount of time had passed? And is the associated strategy one that would end up preferring the cache over the network? If so, set `maxAgeSeconds` to whatever the "natural" lifetime of the content would be.

Using both `maxAgeSeconds` and `maxEntries` makes sense for some types of requests, as long as you've thought through both of those points.

Given that, I wonder if using `maxEntries` in the `externalExpirationPlugin` is necessary—given that you're going to be using it to control a cache of Google Fonts URLs that are likely not to increase in size very frequently, it could be overkill. (It's also unlikely to hurt anything, so if you want to keep it, go for it.)

More to the point, though, I'd recommend adding in a `maxEntries` configuration to the `localExpirationPlugin` config, and potentially remove `maxAgeSeconds` from there. The goal, I think, for this config is to prevent the total number of cached entries from growing past a certain point, rather than to make guarantees around freshness.

The main point is to make sure that whatever config you end up with, you've thought about the characteristics of your anticipated cache growth.

samthor: Thanks for the detailed writeup. I've removed the entries limit from our external cache, but split local caches into content (50 entries max) and images/assets (100 entries max, expire after a week).

I wanted to make sure that we separately consider images and assets vs the content itself (although using the same plugin probably would have been fine since we named the caches differently?). Content is more important vs assets, I'm more eager to remove assets if a user hasn't read an article because they'll take up more space on a user's device.

samthor:
PTAL.

PR #2277, opened 2020-03-03
Add missing meta theme-color

beaufortfrancois:
This PR adds missing meta theme-color element as suggested in Audits "Progressive Web App" section.

The value is the same as the one found in the web manifest.

image


samthor:
So this is the same as we have in the manifest this seems fine. I'm slightly concerned that "white" in the tab switcher looks a bit odd though.

PR #2281, opened 2020-03-04
Add linter for wiki MDN links

### LINKED ISSUE #2142: Certain types of MDN links should throw a linter error or warning.
We need to be sure that links to MDN are to the proper domain. 

MDN is now split between a content domain (https://developer.mozilla.org/) and an editing domain (https://wiki.developer.mozilla.org/). The former is the default domain for MDN display and the latter is intended for MDN contributors writing or editing pages. Readers initially see the content domain. If they click "Edit in wiki" they will be taken to the editing domain and prompted to sign in. It is possible to link to the editing domain.

MDN made this change to lay groundwork for moving from using a wiki for content maintenance to using a GitHub repo with pull requests. Sometime in the next 18 months, links to https://wiki.developer.mozilla.org/ are likely to be broken. Hopefully, MDN will set up a redirect mechanism. It would be prudent to not depend on that.

Links to MDN should only go to the content domain only.

The content linter should flag links with https://wiki.developer.mozilla.org/ and instructor contributors to change them to https://developer.mozilla.org/.

(I suggest writing a general solution that we could plug different urls into as needed.)

beaufortfrancois:
This PR adds a linter that checks for links to "wiki.developer.mozilla.org" and warn to change to "developer.mozilla.org". 

It also updates some articles that are not compliant with this rule.

Fixes #2142

### DIFF
--- a/.remarkrc.js
+++ b/.remarkrc.js
@@ -40,10 +40,10 @@ exports.plugins = [
   require('remark-lint-no-duplicate-headings-in-section'),
 
   // Custom plugins.
+  require('./tools/linting/bad-urls.js'),
   require('./tools/linting/fenced-code-flag.js'),
   require('./tools/linting/no-dash-spaces.js'),
   require('./tools/linting/no-repeat-punctuation.js'),
   require('./tools/linting/no-smart-quotes.js'),
-  require('./tools/linting/no-wiki-mdn-urls.js'),
   require('./tools/linting/no-unescaped-template-tags.js'),
 ];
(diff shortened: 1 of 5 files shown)


jpmedley:
👏👏👏👏👏

kaycebasques:
We need a review from @robdodson or @samthor because it touches JS files.

kaycebasques:
The idea SGTM. Thank you @beaufortfrancois for automating the process of improving content quality.

beaufortfrancois:
If you feel strongly about it or we already have some known URLs to block, I'm happy to update this PR.

If not, how about we start with this simple PR and update it when more things come?

beaufortfrancois:
@samthor As discussed offline, please have a look at updated PR.

PR #2320, opened 2020-03-09
Add accessible focus ring to w-button

mfriesenhahn:
The existing `.w-button` states had two issues:
- Hover and focus states were sticky on touch devices.
- The focus state wasn't conformant with [WCAG 2.1 SC 1.4.11](https://www.w3.org/TR/WCAG21/#non-text-contrast) (contrast for visual indicators of focus state was below 3:1).

- Using `@hover()` mixin, only show hover and existing focus state on devices that support hover.
- Using `focus-visible`, add contrast conformant focus ring* that appears only on keyboard navigation.

*Note: I would have preferred the focus ring to have rounded corners, but `box-shadow` and both pseudos are already in use.

### DIFF
--- a/src/styles/components/_buttons.scss
+++ b/src/styles/components/_buttons.scss
@@ -33,9 +33,7 @@
   border: 0;
 }
 
-.w-button:hover,
-.w-button:focus,
-.w-button:active {
+@mixin button-hover() {
   background: $WHITE; // DevSite override
   box-shadow:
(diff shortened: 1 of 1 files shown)


mfriesenhahn:
Another issue with the current hover/focus behavior on mobile: changes to other states (e.g., disabled) aren't visible until focus moves to another element, which can cause problems (e.g., for the **Check** button in self-assessments).

PR #2338, opened 2020-03-10
Add redirect for /pwa/ to PWA content

### LINKED ISSUE #2337: Redirect from /pwa/ to /progressive-web-apps/ doesn't work
**Describe the bug**
Redirects for `/pwa` works, but `/pwa/` doesn't 

**To Reproduce**
1. Go to https://web.dev/pwa/
2. See 404

**Expected behavior**
Redirects to https://web.dev/progressive-web-apps/

petele:
Fixes #2337 

- Adds redirect from `/pwa/` to `/progressive-web-apps/`

We already redirect from `/pwa`, this just ensures that whether you include the trailing slash, we redirect you to the right place.

### DIFF
--- a/src/site/content/en/_redirects.yaml
+++ b/src/site/content/en/_redirects.yaml
@@ -168,6 +168,9 @@ redirects:
 - from: /pwa
   to: /progressive-web-apps/
 
+- from: /pwa/
+  to: /progressive-web-apps/
+
 - from: /installable/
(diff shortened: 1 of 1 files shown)


samthor:
I'm going to accept this but I'd like to make the redirects handler support both trailing and non-trailing slashes. Would you happy with that fix too?

PR #2340, opened 2020-03-11
Fix external dead URLs

beaufortfrancois:
By running a custom version of [remark-lint-no-dead-urls](https://github.com/davidtheclark/remark-lint-no-dead-urls), I was able to find dead external URLs in web.dev. This PR updates all of them to the correct ones or removes them when it doesn't exist or is not relevant anymore.

### DIFF
--- a/src/site/content/en/blog/ar-hit-test/index.md
+++ b/src/site/content/en/blog/ar-hit-test/index.md
@@ -25,7 +25,7 @@ that some of the work is finished. In Chrome 81, two new features have arrived:
 * [Hit testing](https://www.chromestatus.com/features/4755348300759040)
 
 This article covers the [WebXR Hit Test
-API](https://github.com/immersive-web/hit-test/blob/master/explainer.md), a
+API](https://github.com/immersive-web/hit-test/blob/master/hit-testing-explainer.md), a
 means of placing virtual objects in a real-world camera view.
 
 In this article I assume you already know how to create an augmented reality
@@ -289,7 +289,7 @@ reticle gives you a constant source of hit tests, the simplest way to place an
 object is to draw it at the location of the reticle at the last hit test. If you
 need to, say you have a legitimate reason not to show a reticle, you can call
 `requestHitTest()` in the select event [as shown in the
-sample](https://github.com/immersive-web/webxr-samples/blob/master/proposals/phone-ar-hit-test.html#L187).
+sample](https://github.com/immersive-web/webxr-samples/blob/429aeb7cd46f1009d2e529e69854418a5a0903d5/proposals/phone-ar-hit-test.html#L189).
 
 ```js
 function onSelect(event) {
@@ -304,7 +304,7 @@ function onSelect(event) {
 ## Conclusion
 
 The best way to get a handle on this is to step through the [sample
-code](https://immersive-web.github.io/webxr-samples/proposals/phone-ar.html) or
+code](https://github.com/immersive-web/webxr-samples/blob/4c0646414ebb55eb741838772ad4e80e901391a6/proposals/phone-ar.html) or
 try out the
 [codelab](https://codelabs.developers.google.com/codelabs/ar-with-webxr). I hope
 I've given you enough background to make sense of both.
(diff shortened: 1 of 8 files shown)


src/site/content/en/blog/ar-hit-test/index.md
```diff
@@ -289,7 +289,7 @@ reticle gives you a constant source of hit tests, the simplest way to place an
 object is to draw it at the location of the reticle at the last hit test. If you
 need to, say you have a legitimate reason not to show a reticle, you can call
 `requestHitTest()` in the select event [as shown in the
-sample](https://github.com/immersive-web/webxr-samples/blob/master/proposals/phone-ar-hit-test.html#L187).
+sample](https://github.com/immersive-web/webxr-samples/blob/429aeb7cd46f1009d2e529e69854418a5a0903d5/proposals/phone-ar-hit-test.html#L189).
```
jpmedley: @toji What's the updated location for the Phone AR Hit Test? We should probably do something better than linking to a commit blob.

src/site/content/en/blog/ar-hit-test/index.md
```diff
@@ -304,7 +304,7 @@ function onSelect(event) {
 ## Conclusion
 
 The best way to get a handle on this is to step through the [sample
-code](https://immersive-web.github.io/webxr-samples/proposals/phone-ar.html) or
+code](https://github.com/immersive-web/webxr-samples/blob/4c0646414ebb55eb741838772ad4e80e901391a6/proposals/phone-ar.html) or
```
jpmedley: @toji Same question.

kaycebasques:
So cool! Thank you @beaufortfrancois. Really appreciate your voluntary ownership here. 

Looks like there's some ongoing discussion on where some of the links should point to now.

src/site/content/en/blog/ar-hit-test/index.md
toji: Live Page: https://immersive-web.github.io/webxr-samples/hit-test.html
Source: https://github.com/immersive-web/webxr-samples/blob/master/hit-test.html

src/site/content/en/blog/ar-hit-test/index.md
toji: Live Page: https://immersive-web.github.io/webxr-samples/immersive-ar-session.html
Source: https://github.com/immersive-web/webxr-samples/blob/master/immersive-ar-session.html

src/site/content/en/blog/ar-hit-test/index.md
jpmedley: @beaufortfrancois Please use Brandon's links.

beaufortfrancois: @jpmedley I've linked to a commit blob because the article references `requestHitTest` which doesn't exist in the updated live page/source pages shared by @toji.
Unless you're willing to update the article, we should use those links.

src/site/content/en/blog/ar-hit-test/index.md
beaufortfrancois: @jpmedley Same as above

src/site/content/en/blog/ar-hit-test/index.md
jpmedley: @toji It's called `requestHitTestSource()` now, right?

toji: Correct, used here: https://github.com/immersive-web/webxr-samples/blob/master/hit-test.html#L137

But this does bring up a good point, in that you'll need to update the surrounding text as well because the method isn't called in the select event any more. The equivalent call in frame callbacks or select events is now `frame.getHitTestResults()` https://github.com/immersive-web/webxr-samples/blob/master/hit-test.html#L199

kaycebasques:
@toji @jpmedley @beaufortfrancois it looks like this is just waiting on an agreement between you 3 on which links are most appropriate.

jpmedley:
Merge this with the commit blob urls now so they're not broken. I need to update the article for real, which I'll do today.

jpmedley:
In the future, I'd rather avoid using commit blobs for sample URLs.

src/site/content/en/blog/ar-hit-test/index.md
jpmedley: I opened #2394 to address these links and surrounding text for real.

PR #2354, opened 2020-03-12
Fix search widget accessibility issues

beaufortfrancois:
This PR fixes some accessibility issues in the search widget discovered in the Audits accessibility section:
- Search input role needs to be ["searchbox"](https://www.w3.org/TR/wai-aria-1.1/#searchbox) instead of "search" to be compliant
- The search input wrapper aria-owns values was invalid because the `id` didn't exist in the page when search was not triggered. This is now fixed.

This change allows web.dev to reach the Accessibility score of 100.

image

### DIFF
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -86,7 +86,7 @@ class Search extends BaseElement {
         <input
           class="web-search__input"
           type="text"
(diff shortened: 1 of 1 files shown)


robdodson:
ah so great!! thank you!

Issue #1941, opened 2019-11-25
Not properly tracking signed in state.


robdodson:
I noticed when we moved off of DevSite that we stopped tracking when the user is signed in/out.

We have the ability to track signed in state by setting an analytics custom dimension:
```
ga.set(site.analytics.dimensions.SIGNED_IN, true);
```
This means all subsequent pings will contain this info so we can discern actions that occurred from signed in versus signed out users.

I'd suggest that we move the code we currently use to set the signed in state out of fb.js, and into actions.js. Then we can set the custom dimension as a side effect.

Phil suggested that we also consider delaying the send event for our initial pageview, and wait until firebase auth has returned so we can track whether the user was signed in/out for the initial page view. He suggested we set a timeout to ensure we capture the pageview event if fb is taking too long to return.

robdodson:
@samthor what do you think about the above proposal?

samthor:
I'm not an expert re: custom dimensions but it sounds good. Remember that users can transition in both directions signed in <=> signed out on the same page.

We could also add a `beforeunload` handler and log pageview on timeout via beacon. Just to track bounces.

robdodson:
> Remember that users can transition in both directions signed in <=> signed out on the same page.

Yeah I think we currently have a `clearSignedInState` action and I was wondering if we should have an action for signing in. Then we would set the custom dimension in those two actions to indicate either signed in or signed out.

This also got me thinking about analytics middleware and if we may want to have an easier way of setting analytics events/states/whatever as side effects of actions.

robdodson:
Spoke with Sam today.

We discussed setting the dimension and firing an event when the client side js boots up and we are able to see that the user is signed in. @devnook happy to discuss this more over VC.

robdodson:
@samthor @devnook I've assigned this to both of you. I was thinking, whoever is free first can claim this issue and work on it. Just be sure to leave a comment here and remove the other assignee.

I spoke with Addy today related to other projects and it sounds like having this stat would be useful so we should prioritize adding this back to the site soon if we can.

samthor:
There's a pending PR (which to be fair is superseded). Keeping this alive. 👨‍🎤

PR #2358, opened 2020-03-12
Load dynamically Firebase

beaufortfrancois:
This PR improves loading performance by loading dynamically Firebase after images have been requested.

Note that the diff doesn't reflect well at a first glance what actually changed in the first part of  `src/lib/fb.js`. See help below

```js

// Load dynamically Firebase
firebaseLoader()
  .then(() => {
   /* Unchanged */
  })
  .catch((err) => {
    console.warn("failed to load Firebase library");
  });

### DIFF
--- a/src/lib/fb.js
+++ b/src/lib/fb.js
@@ -1,120 +1,128 @@
 import config from "webdev_config";
 import {store} from "./store";
 import {clearSignedInState} from "./actions";
+import firebaseLoader from "./firebase-loader";
 import firestoreLoader from "./firestore-loader";
 
 /* eslint-disable require-jsdoc */
 
-const {firebaseConfig} = config;
+// Load dynamically Firebase
+firebaseLoader()
+  .then(() => {
+    const {firebaseConfig} = config;
 
-firebase.initializeApp(firebaseConfig);
+    firebase.initializeApp(firebaseConfig);
 
-// Initialize performance monitoring
-firebase.performance();
+    // Initialize performance monitoring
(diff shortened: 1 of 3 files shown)


samthor:
FWIW I feel like there was a reason why we load Firebase in the 'critical path'.

I wonder if it was just reducing complexity, but Firestore was large enough to justify delaying.

So, I think my opinion is that at some point this becomes needless - none of the user stuff works until it's loaded, that's a critical user journey, so delaying it in favour of SPA routing etc doesn't help us. I'm totally happy to be disagreed with, if you have strong opinions.

beaufortfrancois:
I naively thought that web.dev site was reached more often by non-logged-in users, therefore I wanted to improved user journey for the majority of users.

The added complexity is not that high in my opinion.

robdodson:
> I naively thought that web.dev site was reached more often by non-logged-in users, therefore I wanted to improved user journey for the majority of users.

Until we build out more of the profile, this assumption is probably correct. All the more reason to fix https://github.com/GoogleChrome/web.dev/issues/1941 so we know for sure :)

@samthor would you be ok with us trying this change?

beaufortfrancois:
(gentle ping) @samthor

robdodson:
Since our PSI score is down into the 80s for some reason I'm pretty keen to give this a shot to see if it helps. @samthor wdyt?

samthor:
Fixed in #2767. Thanks for the prework!

PR #2363, opened 2020-03-13
Deeplinks

### LINKED ISSUE #2336: Deep links seem broken
**Describe the bug**
https://web.dev/style-focus/#when-to-display-a-focus-indicator should scroll to the section with the corresponding header but it doesn't. :(

samthor:
I hate routing. Fixes #2336.

I've changed the way the router works and disabled manual `scrollRestoration`. Notably DevSite has the "flash" of incorrect scroll positions (which is an older issue) because they still are where we were a few PRs ago.

* Allows multiple loads to occur at once (but we abort via `AbortController`); previously, we tried to complete them in series (confusing for users rapidly hitting back/forward)
* Cache page contents for back/forward and don't go to the network unless we need to (matches what normal pages do)
* Creates a faux-partial for the first loaded page (since it's prerendered HTML)
  * we could also put this into the Cache API for caching the first load (but that's out of scope)
* Allows our loader to inform the router it's `ready()` early, swapping the URL, but still showing a loading bar for the JS payload

I might factor this out into a nicer library tonight, that's better documented.

### DIFF
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -22,21 +22,29 @@ async function loadEntrypoint(url) {
 }
 
 /**
- * Gets the partial content of the target normalized URL.
+ * Gets the partial content of the target normalized URL. Returns null if aborted.
  *
  * @param {string} url of the page to fetch.
- * @return {{raw: string, title: string, offline: (boolean|undefined)}}
+ * @param {!AbortSignal=} signal
+ * @return {?{raw: string, title: string, offline: (boolean|undefined)}}
  */
-export async function getPartial(url) {
+export async function getPartial(url, signal) {
   if (!url.endsWith("/")) {
     throw new Error(`partial unsupported for non-folder: ${url}`);
   }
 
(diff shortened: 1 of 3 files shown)


robdodson:
Thank you for working on this! Definitely seems to improve the deep linking. I did notice one small issue:

1. Visit https://deploy-preview-2363--web-dev-staging.netlify.com/ads-case-study-stale-while-revalidate
2. Click on a heading to get the hash anchor
3. Refresh the page
4. Scroll down and click a different heading

It'll jump back to the top of the page.

samthor:
Thanks, that's a weird issue but I'll take a look.

samthor:
So, this actually happens when the JS hasn't even _run_. (I noticed this since locally the Algolia dependency hadn't updated, so the JS was failing to load)

Still worth looking at though.

samthor:
I found the issue. I was incorrectly recording the recent active URL to _include_ the hash in some situations. PTAL!

PR #2369, opened 2020-03-14
feat(stripTitle): add filter to remove forbidden characters from string, fixes #1993

### LINKED ISSUE #1993: Backticks appear in title elements
**Describe the bug**
When an author uses backticks in a page title to display code font, the backticks are included in the page's `<title>` element. For example:

image

**To Reproduce**
1. View source for the page and expand the `<title>` element.
1. See error.

**Expected behavior**
Backticks are stripped from `<title>` elements.

MichaelSolati:
Fixes #1993

Add filter that removes forbidden characters/whatever from titles.

@samthor if I can get a LGTM that'd be dope!

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -75,9 +75,9 @@ const pagedNavigation = require(`./${filtersDir}/paged-navigation`);
 const postsLighthouseJson = require(`./${filtersDir}/posts-lighthouse-json`);
 const prettyDate = require(`./${filtersDir}/pretty-date`);
 const removeDrafts = require(`./${filtersDir}/remove-drafts`);
+const strip = require(`./${filtersDir}/strip`);
 const stripBlog = require(`./${filtersDir}/strip-blog`);
 const stripLanguage = require(`./${filtersDir}/strip-language`);
-const stripTitle = require(`./${filtersDir}/strip-title`);
 
 const buildPartial = require('./src/site/_utils/build-partial');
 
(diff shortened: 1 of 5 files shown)


samthor:
I don't mean to rain on your parade (not that parades are allowed anymore) but we already include a module called `remove-markdown` which could do just this. Could you make this filter just use that?

MichaelSolati:
> I don't mean to rain on your parade (not that parades are allowed anymore) but we already include a module called `remove-markdown` which could do just this. Could you make this filter just use that?

Testing the following code on [runkit](https://npm.runkit.com/remove-markdown):

```js
const removeMd = require('remove-markdown');
const markdown = '# Title \n\n `<div>`';
const plainText = removeMd(markdown);

console.log(plainText)
```

It doesn't seem to remove `, the result looks like this:
```js
Title\n\n ``
```

samthor:
That's weird and if that package is failing at it's one task, maybe we
remove it elsewhere too.

On my phone, your original change is fine.

robdodson:
I bet we may need a general strip/replace function for other parts of the site in the future. Rather than name this `stripTitle`, perhaps we rename it to `strip` and allow it to also take arguments. That way it will either strip a default set of things, or allow the developer to pass in additional characters in the future.

robdodson:
btw, nunjucks includes its own replace function (https://mozilla.github.io/nunjucks/templating.html#replace). Wanted to highlight that in case it's useful.

PR #2373, opened 2020-03-17
Revert "upgrade to Algolia SDK 4+ (#2348)"

robdodson:
This reverts commit 6586f7fd8b4f5f77ffc29762092e7c98efd601bc.

- Undo changes in #2348. It looks like those changes were causing index-algolia.js to break so the site wasn't deploying anymore. You can see an example of the build output here: https://github.com/GoogleChrome/web.dev/runs/514705638?check_suite_focus=true

Screen Shot 2020-03-17 at 3 34 34 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -310,121 +310,6 @@
         "prismjs": "^1.17.1"
       }
     },
-    "@algolia/cache-browser-local-storage": {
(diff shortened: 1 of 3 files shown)


samthor:
Whoops.

PR #2375, opened 2020-03-17
upgrade Algolia in search and index

samthor:
Revert revert in #2373 of #2348.

Adds fix for indexing pipeline.

### DIFF
--- a/build.js
+++ b/build.js
@@ -115,6 +115,16 @@ async function build() {
       rollupPluginPostCSS(postcssConfig),
       ...defaultPlugins,
     ],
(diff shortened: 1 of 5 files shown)


robdodson:
Do you need to add this line back?
https://github.com/GoogleChrome/web.dev/pull/2348/files#diff-3549ac7324c1b1a52b2f376e7bceaee1R10

Seems like Percy is failing because the JS bundle isn't working.

samthor:
I've fixed this and made Rollup complain in future if you try to import something it can't find.

PR #2377, opened 2020-03-18
reorder caching so images are cached after 1st load

samthor:
This works around the issue raised [over in Workbox](https://github.com/GoogleChrome/workbox/issues/2402).

This isn't specifically tracked but I noticed it while working on JS issues. Images that we aggressively precache (anything in /images/*) weren't be served while offline, because the runtime handler was winning and claiming nothing was found.

As per the other bug:

- images live under /images/ and include both core site images as well as profile pictures etc
- we only want to precache the things that appear on the home page
- but we want to runtime cache any extended images that the user happens to load.

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -10,6 +10,11 @@ import {matchPrecache, precacheAndRoute} from "workbox-precaching";
 import {cacheNames} from "workbox-core";
 import {matchSameOriginRegExp} from "./utils/sw-match.js";
 
+/**
+ * Configure default cache for standard web.dev files: the offline page, various images etc.
+ */
+precacheAndRoute(manifest);
+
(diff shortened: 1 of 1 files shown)


robdodson:
hmmmmmmmmmmmmmmmmm did you notice the percy failure? Not sure why that is...

samthor:
@robdodson I kicked Percy. I reckon it's random whether the lower image loads in time.

robdodson:
Rebasing so I can verify this works on netlify. I think the deploy was running the old (busted) loader.js code.

robdodson:
Seems to be working fine on netlify. I tried increasing the wait time on percy to 5 seconds but it still didn't load that image so I'm thinking we'll have to try something else. Maybe we don't load that footer image when we're in a test environment. Percy's great, but if it's flaky folks will complain.

PR #2378, opened 2020-03-18
Add try/catch blocks for better cross-browser support

### LINKED ISSUE #2078: Ensure all code examples under /metrics work cross-browser
I discovered today that some of the code examples for the docs under [web.dev/metrics](https://web.dev/metrics/) will now error in Safari. This is due to the fact that Safari recently added support for the `PerformanceObserver` API, but it doesn't support entry types such as `paint`, `first-input`, or `longtask`, nor does it support the new `type` observe option. Furthermore, when using these options Safari will throw an error rather than silently failing.

It's probably worth updating all the code examples to add proper feature detection and error handling for such cases.

philipwalton:
Fixes #2078

This PR adds a try/catch block around every code example that uses `PerformanceObserver` with the `type` option to avoid errors due to [this Safari bug](https://bugs.webkit.org/show_bug.cgi?id=209216).

Note: it feels a bit excessive to add the comment and bug link to each code example, but I did so because I was concerned that, without it, developers might think they need to use a try/catch *any* time they ever use `PerformanceObserver`, which isn't true.

An alternative way would be to define a helper function that wraps the try/catch and then use that helper function in each example, but that's also awkward to define for each metric.

If anyone has any thoughts on how to do this better, please let me know.

### DIFF
--- a/src/site/content/en/metrics/cls/index.md
+++ b/src/site/content/en/metrics/cls/index.md
@@ -251,13 +251,19 @@ how to create a
 that listens for `layout-shift` entries and logs them to the console:
 
 ```js
-const observer = new PerformanceObserver((list) => {
- for (const entry of list.getEntries()) {
-   console.log(entry);
- }
-});
+// Catch errors since some browsers throw when using the new `type` option.
+// https://bugs.webkit.org/show_bug.cgi?id=209216
+try {
+  const observer = new PerformanceObserver((list) => {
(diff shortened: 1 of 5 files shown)


kaycebasques:
@robdodson can you restart netlify and see if the staging build passes... doesn't seem like there's any reason this update would break the build

robdodson:
try now :)

PR #2383, opened 2020-03-18
Account for index.json partial files when serving localized content

devnook:
Since we now serve partials in SPA via index.json files, we have to redirect json requests to appropriate language just like index.html


locale-handler.js
```diff
@@ -38,27 +38,32 @@ const isSupported = (locale) => SUPPORTED_LOCALES.indexOf(locale) > -1;
  * @return {!Function}
  */
 module.exports = (req, res, next) => {
+  const isNav = req.url.endsWith("/");
+  const isJson = req.url.endsWith(".json");
```
samthor: Every partial ends with "/index.json".

locale-handler.js
```diff
@@ -38,27 +38,32 @@ const isSupported = (locale) => SUPPORTED_LOCALES.indexOf(locale) > -1;
...
   // Exit early if the url is not navigational.
-  if (!req.url.endsWith("/")) {
+  if (!isNav && !isJson) {
     return next();
   }
 
-  const pathParts = req.path.split("/");
-  // Check if language is specified in the url.
+  const fileType = isJson ? "index.json" : "index.html";
+  const normalizedPath = req.path.replace("index.json", "");
```
samthor: can we replace `/\/index\.json$/`, so we're sure we have the right path?

PR #2368, opened 2020-03-13
Podcast lander page

argyleink:
[web.dev/podcasts](https://deploy-preview-2368--web-dev-staging.netlify.com/podcasts/)

<details>
<summary>Screenshot</summary>

localhost_8080_podcasts_ (1)

</details>

**todo list:**
- [x] show list needs verified (did i miss any?)
- [ ] content needs verified by show owners
- [x] images should probably not be remotely owned
- [x] overall copy needs looked at
- [x] new data source could be added to 11ty (shows|podcasts), then looped over in the template (v2?)
- [x] unique/custom art made for the page (v2?)
- [x] consider not using `<a>` around the whole card so that talent names can link to profiles, other shows, etc
- [x] analytics tracking of links clicked and page visits (is this a todo? or free?)

PR #2384, opened 2020-03-18
feat(authors-and-tags): create landing page for authors and tags, fixes #2275 & fixes #2159

### LINKED ISSUE #2159: It's hard to find the tags from the homepage
I love looking at the content based on tags. I would love to 

1) discover all the tags more easily - https://web.dev/tags/ 404's. I would love to get a list of the tags if possible.
2) See tags linked to from the homepage (low pri), but if https://web.dev/tags/ doesn't 404 then it would be nice to make it more discoverable.

### LINKED ISSUE #2275: We should have an Authors collection
I wanted to see a complete list of authors on the site, yet https://web.dev/authors/ 404's

That page should have all of our authors on the site.

MichaelSolati:
Adds `/authors` and `/tags` pages to start to address #2159 

@robdodson or @samthor PTAL


src/site/_data/postTags.js
```diff
@@ -189,10 +189,17 @@ const tags = {
...
 
-Object.keys(tags).forEach((key) => {
-  postTags[key] = {
-    ...tags[key],
-    href: `/tags/${key}/`,
+Object.keys(tags).forEach(($key) => {
+  const tag = tags[$key];
+  const description = tag.description
+    ? tag.description
+    : `Our latest news, updates, and stories about ${tag.title}.`;
```
samthor: Random thought: strings like this are going to be painful when we start doing i18n.

src/site/_data/postTags.js
```diff
@@ -189,10 +189,17 @@ const tags = {
 
 const postTags = {};
 
-Object.keys(tags).forEach((key) => {
-  postTags[key] = {
-    ...tags[key],
-    href: `/tags/${key}/`,
+Object.keys(tags).forEach(($key) => {
```
samthor: Why `$key`? I know it's a valid JS variable name, just...  it seems like a random change.

src/site/_data/contributors.js
```diff
@@ -1727,19 +1727,20 @@ const contributors = {
 
 const processedContributors = {};
 
-Object.keys(contributors).forEach((key) => {
-  const contributorData = contributors[key];
+Object.keys(contributors).forEach(($key) => {
```
samthor: same comment here re: `$key`

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,76 @@
...
+
+  const elements = authors.map((author) => {
+    author.url = `/en${author.href}`;
+    author.data = {
+      tags: [],
+      title: author.title,
+      subhead: author.description,
+    };
+
+    if (
```
samthor: can we do

```js
for (const size of [3, 2, 1]) {
  if (fs.existsSync(...)) {
    author.data.hero = ...;
    break;
  }
}
```

src/site/_utils/get-image-path.js
```diff
@@ -26,7 +26,9 @@ const site = require("../_data/site");
  * it's a production environment.
  */
 module.exports = function getImagePath(src, pageUrl) {
-  let imagePath = path.join(pageUrl, src);
+  let imagePath = (src || "").startsWith("~/")
+    ? src.replace("~/", "/")
```
samthor:
```
src.replace(/^~\//, "/")
```

to make sure we only catch the left.

By the way, if we're just replacing the "root" with "/" anyway—why do we need the tilde?

src/site/content/en/tags/index.njk
```diff
@@ -22,7 +23,7 @@ pagination:
   <div class="w-layout-container w-pb--non">
     <header class="w-page-header">
       <h1 class="w-page-header__headline">{{ renderData.title }}</h1>
-      <p class="w-page-header__copy">{{ renderData.description }}</p>
+      <p class="w-page-header__copy">{{ renderData.description | safe }}</p>
```
samthor: Do some tags' description now have HTML in it?

src/site/_utils/get-image-path.js
MichaelSolati: The tilde is just used to make the distinction it should be from root. For posts the photo is relative to the folder the photo is in, so this function joins the directory to the image src. Adding the `~` just says not to do that. Otherwise I'll make the regex change.

robdodson: I don't want to boil the ocean, but it feels like we're working around the fact that the PostCard and PathCard elements aren't very extensible. I see this happening on the podcast page as well (https://github.com/GoogleChrome/web.dev/pull/2368#issuecomment-601855753).

I'm concerned about us shipping workarounds that create technical debt in the long run. I keep wondering if we should instead look at writing an [eleventy transform](https://www.11ty.dev/docs/config/#transforms) that converts _all_ image paths to be from the CDN if it's in the prod environment and hopefully our components wouldn't need to manually call and request CDN paths.

Also I noticed the images that are being served on the authors page seem to have an intrinsic size of 192 x 192 which seems quite small. Basically they all look pixelated and I'm assuming that's because we never asked anyone to upload an image in these larger dimensions. So I'm kind of wondering if we want to go with this design at all. We could omit the hero image and go with their name and try and work in the smaller profile image..? Or we'll want to ask everyone to upload a higher res headshot.

MichaelSolati:
@robdodson I do think the `~` solution is the easiest implementation for now (in regards to author images). Also we can look to get higher quality photos, but the page looks barren without them. I'd rather not remove the code, but I'll comment it out.

robdodson:
I think something like this could look nice. But it would require tweaks to the PostCard which I'm pretty ok with. I think we may be overly DRYing the code in this case and making it harder for us to style these unique page types. I'll take a look to see if I can offer ideas on how to get the functionality we want.

image

robdodson:
It seems like we could break `paged.njk` apart in the following way:
- Move `w-pagination` to its own include.
- Let authors.njk do its own loop and include the pagination at the end:

```njk
<section class="w-grid">
    {% for element in paged.elements %}
      {% AuthorCard {author: element} %}
    {% endfor %}
  </div>
</section>

{% include 'pagination.njk' %}
```

This proposes creating a new element called AuthorCard which is basically the same as the PostCard but some slightly different classes on the figure and the img. The img has an explicit width/height of 192px (the max resolution we have for author photos), a border radius of 50% to round it, and it removes the pseudo element in __desc that generates the little blue line at the top left corner.

robdodson:
If we're making an `AuthorCard` shortcode then we could add a parameter to `getImagePath()` to tell it to look in the root instead of relative directory. `AuthorCard` would pass this param when calling that function, instead of using special file path syntax.

robdodson:
I think percy caught an issue with the blue line at the top of the card

image

You can ignore the second percy failure. It's barking about the glitch embed that'll be fixed when https://github.com/GoogleChrome/web.dev/pull/2498 lands.

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,71 @@
...
+    author.data = {
+      tags: [],
+      title: author.title,
+      subhead: author.description,
+    };
+
+    for (const size of ["@3x", "@2x", ""]) {
+      if (
+        fs.existsSync(
+          `src/site/content/en/images/authors/${author.key}${size}.jpg`,
```
robdodson, line 56: use path.join here as well. `path.join('src', 'site', ...)`

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,71 @@
...
+        )
+      ) {
+        author.data.hero = `/images/authors/${author.key}${size}.jpg`;
```
robdodson, line 59: use path.join here as well

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,71 @@
...
+ * embedded loop O^2.
+ *
+ * @param {any} collections Eleventy collections object
+ * @return {Array<{ title: string, href: string, description: string, elements: Array<object>, index: number, pages: number }>} An array where each element is a page with some meta data and n authors for the page.
+ */
+module.exports = (collections) => {
+  const authorsWithPosts = {};
+  const posts = collections.getAll().filter(livePosts);
+
+  posts.forEach((post) => {
```
robdodson, line 36: I was wondering if we actually need this step. What if we just use the array generated by this step to do a one time cleanup of the contributors.json file? Then in this colleciton we would essentially just require contributors.json, add the author url for each person, and hand it to addPagination.

I realize looping over all of the posts is a bit more accurate, but I'd love to look for any opportunity we can to cut down on looping over the entire collection because it directly impacts build times.

src/site/_collections/paginated-tags.js
```diff
@@ -0,0 +1,59 @@
...
+    const tags = post.data.tags || [];
+    tags.forEach((tag) => (tagsWithPosts[tag] = true));
+  });
+
+  const tags = Object.values(postTags).filter(
+    (postTag) => tagsWithPosts[postTag.key],
+  );
+
+  const elements = tags.map((tag) => {
+    tag.url = `/en${tag.href}`;
```
robdodson, line 46: use path.join

src/site/_collections/paginated-tags.js
```diff
@@ -0,0 +1,59 @@
...
+ * information of that tag to pump into `_includes/partials/paged.njk`. This is because we cannot
+ * paginate something already paginated... Pagination is effectively a loop, and we can't have an
+ * embedded loop O^2.
+ *
+ * @param {any} collections Eleventy collection object
+ * @return {Array<{ title: string, href: string, description: string, elements: Array<object>, index: number, pages: number }>} An array where each element is a paged tag with some meta data and n posts for the page.
+ */
+module.exports = (collections) => {
+  const tagsWithPosts = {};
+  const posts = collections.getAll().filter(livePosts);
```
robdodson, line 34: use collections.getFilteredByGlob

src/site/_collections/paginated-tags.js
```diff
@@ -0,0 +1,59 @@
...
+
+  posts.forEach((post) => {
```
robdodson, line 36: I think another way to go about this would be to loop over postTags.json and compare it against eleventy's collections object.

```
for (tag of postTags) {
  if (collections[tag]) {
    // add it to the list of pages we generate
  }
}
```

I think this'll be a smaller loop than iterating over all of our posts.

src/styles/components/_card-author.scss
```diff
@@ -0,0 +1,13 @@
...
+
+.w-card-author {
+  text-align: center;
+}
+
+.w-card-author__figure {
+  background-color: $WHITE;
+}
+
+.w-card-author__image {
```
robdodson, line 11: add a width: 192px to these. I think that's the highest resolution image we have of folks.

src/styles/components/_card-author.scss
```diff
@@ -0,0 +1,13 @@
+@import '../settings/colors';
+
+.w-card-author {
```
robdodson, line 3: on mobile we end up with the blue line at the top:

image


maybe we should always turn that off for author cards.

src/site/_includes/components/AuthorCard.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * AuthorCard used to preview authors.
+ * @param {Object} post An eleventy collection item with post data.
+ * @return {string}
+ */
+class AuthorCard extends BaseCard {
+  constructor(arg) {
+    super({...arg, className: "w-card-author"});
+    super.renderThumbnail = this.renderThumbnail.bind(this);
```
robdodson, line 27: I think you can delete this line. The presence of `renderThumbnail` on the class acts as an override.

src/site/content/en/authors/index.njk
```diff
@@ -2,14 +2,13 @@
 layout: layout
 title: Authors
 permalink: /en/{{ paged.href }}{% if paged.index > 0 %}{{ paged.index + 1 }}/{% endif %}index.html
-override:tags: []
 description: Our latest news, updates, and stories for developers
-date: 2020-02-12
+noindex: true
```
robdodson, line 6: any reason to hide these pages from Google Search?

src/site/content/en/tags/index.njk
```diff
@@ -2,14 +2,13 @@
 layout: layout
 title: Tags
 permalink: /en/{{ paged.href }}{% if paged.index > 0 %}{{ paged.index + 1 }}/{% endif %}index.html
-override:tags: []
 description: Our latest news, updates, and stories for developers
-date: 2020-01-30
+noindex: true
```
robdodson, line 6: any reason to hide these from Google Search?

src/site/_utils/get-image-path.js
```diff
@@ -20,13 +20,14 @@ const site = require("../_data/site");
 /**
  * Takes a path to an image and converts it to an image CDN path if we're in
  * a production environment.
- * @param {!string} src A path to an image asset.
+ * @param {!string} src A path to an image asset.\
```
robdodson, line 23:
```suggestion
 * @param {!string} src A path to an image asset.
```

src/site/_utils/get-image-path.js
```diff
@@ -20,13 +20,14 @@ const site = require("../_data/site");
...
  * @param {!string} pageUrl The url for the current page.
  * @return {string} An image path. May be converted to an image CDN path if
  * it's a production environment.
  */
 module.exports = function getImagePath(src, pageUrl) {
   let imagePath = path.join(pageUrl, src);
+
```
robdodson, line 30: i think if you remove this newline then this file won't have any changes and you can remove it from the PR

robdodson:
It would be good for us to also think about how we can add /tags/ and /authors/ to percy.

You can check for `process.env.PERCY` to generate different collections. For example, if you look at `recent-posts.js` you can see how it filters down to only posts that have the `test-post` tag. Perhaps you could come up with six authors and six tags and only allow those items in the collections when PERCY is true.

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @return {Array<{ title: string, href: string, description: string, elements: Array<object>, index: number, pages: number }>} An array where each element is a page with some meta data and n authors for the page.
+ */
+module.exports = () => {
+  const authors = Object.values(contributors).sort((a, b) =>
+    a.title.localeCompare(b.title),
+  );
+
+  const elements = authors.map((author) => {
+    author.url = path.join("/en", author.href);
```
robdodson, line 37: why is the `/en` needed?

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,74 @@
...
+    author.data = {
+      tags: [],
```
robdodson, line 39: hm this feels a bit like a code smell if we're passing fake data into a UI element

src/site/_collections/paginated-tags.js
robdodson: @MichaelSolati what do you think about this suggestion?

src/styles/components/_card-base.scss
```diff
@@ -10,7 +10,7 @@
 //
 // =============================================================================
```
robdodson, line 11: The comment at the top of this file is still for Post Card

robdodson:
can we add /tags/ and /authors/ to snapshots.js

src/site/_collections/paginated-authors.js
MichaelSolati: Unfortunately needed because of the language filter.

src/site/_collections/paginated-tags.js
MichaelSolati: Sorry, I thought you said it was fine to leave this yesterday. I'll change it now

src/site/_collections/paginated-authors.js
```diff
@@ -0,0 +1,83 @@
...
+    "samthor",
+    "surma",
+    "egsweeny",
+    "addyosmani",
+    "adamargyle",
+  ];
+  const authors = Object.values(contributors)
+    .sort((a, b) => a.title.localeCompare(b.title))
+    .filter((item) =>
+      process.env.PERCY ? testAuthors.includes(item.key) : true,
```
robdodson, line 43: nit: what do you think about doing this as:

```
if (process.env.PERCY) {
  authors = authors.filter(...)
}
```

Otherwise I think we're needlessly looping over ever item on most builds.

src/site/_collections/paginated-tags.js
```diff
@@ -0,0 +1,60 @@
...
+    "lighthouse",
+    "seo",
+    "progressive-web-apps",
+    "webxr",
+  ];
+
+  const elements = Object.values(postTags)
+    .filter((tag) => {
+      const posts = collections.getFilteredByTag(tag.key).filter(livePosts);
+      return process.env.PERCY ? testTags.includes(tag.key) : posts.length > 0;
```
robdodson, line 46: same point as above, is it possible to do this in an `if` so we only run the filter if we're in a percy env?

tools/percy/snapshots.js
```diff
@@ -28,6 +28,14 @@ const pagesToTest = [
     url: "handbook/web-dev-components",
     title: "Components page",
   },
+  {
+    url: "authors",
+    title: "Authors",
```
robdodson, line 33:
```suggestion
    title: "Authors page",
```

tools/percy/snapshots.js
```diff
@@ -28,6 +28,14 @@ const pagesToTest = [
...
+  },
+  {
+    url: "tags",
+    title: "Tags",
```
robdodson, line 37:
```suggestion
    title: "Tags page",
```

PR #2389, opened 2020-03-19
hide Glitch on codelabs on mobile

### LINKED ISSUE #2344: We should hide Glitch on codelabs on mobile
**Describe the bug**
Currently we still load the Glitch for codelabs on small screens even though it's not visible or usable (<700px).

**To Reproduce**
open e.g., https://web.dev/codelab-remove-unused-code/, make screen small

**Expected behavior**
Something more useful than happens now.

Let's add a blurb saying this is unsupported at the top of the codelab.

beaufortfrancois:
This PR hides Glitch for codelabs on mobile devices as it's not usable.

Live preview: https://deploy-preview-2389--web-dev-staging.netlify.com/codelab-remove-unused-code/

Fixes #2344

### DIFF
--- a/src/lib/components/Codelab/index.js
+++ b/src/lib/components/Codelab/index.js
@@ -19,6 +19,8 @@ class Codelab extends BaseElement {
       glitch: {type: String},
       // The file to show when the Glitch renders.
       path: {type: String},
+      // Whether to show the Glitch iframe or not.
+      iframeEnabled: {type: Boolean},
     };
   }
 
@@ -27,6 +29,7 @@ class Codelab extends BaseElement {
 
     this.glitch = "";
     this.path = "index.html";
+    this.iframeEnabled = false;
   }
 
   createRenderRoot() {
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
1. Do you already have something in mind for the notice?
1. Shall we load conditionally the iframe with media queries listener in javascript?

samthor:
1. I don't have specific text in mind, but even having a place for it would
be great.

2. Yes, that sounds good, since ResizeObserver isn't quite as supported as
we'd like.

beaufortfrancois:
2. Is there another way to not use `865px`, but the scss value in https://github.com/GoogleChrome/web.dev/pull/2389/commits/d987c2fb7204002021295d52aad1510ee93c43bc?

src/lib/components/Codelab/index.js
```diff
@@ -45,8 +45,8 @@ class Codelab extends BaseElement {
   }
 
   get src() {
-    if (!this.glitch) {
-      return;
+    if (!this.glitch || window.matchMedia('(max-width: 865px)').matches) {
```
samthor: I appreciate this but I'd rather have a listener that maintains/updates a property on this element, so we can add or remove the codelab if the user changes the window size.

beaufortfrancois: @samthor WDYT of https://github.com/GoogleChrome/web.dev/pull/2389/commits/e0c08fae0a23ddeb0e84cca66aa68faf1c97c4eb?

beaufortfrancois: @samthor New commit at https://github.com/GoogleChrome/web.dev/pull/2389/commits/84e05226ccde4dc366b3add4e065bc368713b6b6

PR #2393, opened 2020-03-19
[Async Clipboard article] Remove superfluous try...catch block

tomayac:
- Removes a superfluous try...catch block.
- Uses err for error names instead of e.


robdodson:
LGTM! going to merge as this seems like a chill content change

PR #2394, opened 2020-03-19
update the AR Hit Test article.

jpmedley:


jpmedley:
@toji Can you please sanity check this. It might not hurt to glance at [the whole article](https://web.dev/ar-hit-test/) too.

PR #2397, opened 2020-03-19
Fix warning in Firefox related to querying null

robdodson:

- I noticed Firefox was logging a warning because when we navigate to a page without a hash in the url we still try to querySelector for one.

### DIFF
--- a/src/lib/utils/router.js
+++ b/src/lib/utils/router.js
@@ -179,10 +179,13 @@ export function route(url) {
     // element (if any) and scroll to it.
     const hash = u.hash.substr(1);
     if (hash) {
-      document.getElementById(hash).scrollIntoView();
-    } else {
-      document.documentElement.scrollTop = 0;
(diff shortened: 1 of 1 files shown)


src/lib/utils/router.js
```diff
@@ -177,9 +177,9 @@ export function route(url) {
 
     // Since we're loading this page dynamically, look for the target hash-ed
     // element (if any) and scroll to it.
-    const target = document.getElementById(u.hash.substr(1)) || null;
-    if (target) {
-      target.scrollIntoView();
+    const hash = u.hash.substr(1);
+    if (hash) {
```
samthor: We should make sure the target exists:

```js

if (hash) {
  const target = document.getElementById(hash);
  if (target) {
    target.scrollIntoView();
    return;
  }
}
document.documentElement.scrollTop = 0;

```

PR #2398, opened 2020-03-19
Fix formatting of 11ty file to make prettier happy.

robdodson:
Small change to make prettier chill.


robdodson:
going to merge because it's trivial

samthor:
I've seen this happen once before in another file. I wonder if we can automate this as part of the linter.

PR #2399, opened 2020-03-20
simplify var

samthor:
Updates use of `content` var, since otherwise we're doing a `querySelector` multiple times.

Also clarifies how `ready` is used.

### DIFF
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -98,9 +98,8 @@ function forceFocus(el) {
  * @param {!Object} partial
  */
 function updateDom(partial) {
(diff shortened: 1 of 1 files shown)


robdodson:
What do you think about working on a separate PR to add a Percy test for this? I think we'd tell Percy to load web.dev/blog, then tell it to click on an article, then tell it to screenshot that page. I'm curious if we could use that as a kind of lazy end-to-end test.

samthor:
I think that sounds good and will test our SPA logic. I don't know much about Percy.

PR #2401, opened 2020-03-20
update web-codelab and include notice on mobile

samthor:
This updates @beaufortfrancois' work, but tries to add/remove the listener as the element appears and is removed from the page.

It also (somewhat awkwardly) adds a warning notice when the screen is small. (This will be infinitely better when we switch to Shadow DOM).

### DIFF
--- a/src/lib/components/Codelab/index.js
+++ b/src/lib/components/Codelab/index.js
@@ -5,7 +5,10 @@
 
 import {html} from "lit-element";
 import {BaseElement} from "../BaseElement";
+import config from "webdev_config";
 import "./_styles.scss";
+
+const {env} = config;
 
 /**
  * Render codelab instructions and Glitch
@@ -93,6 +96,14 @@ class Codelab extends BaseElement {
             </p>
           </div>
         </div>
+      `;
+    }
+
+    // Disable the Glitch in dev environments, as it takes a while to load.
+    // Used for screenshot testing.
+    if (env === "test") {
+      return html`
+        <div class="w-sizer w-test"></div>
       `;
     }
 
--- a/src/lib/components/Codelab/index.js
+++ b/src/lib/components/Codelab/index.js
@@ -19,8 +19,8 @@ class Codelab extends BaseElement {
       glitch: {type: String},
       // The file to show when the Glitch renders.
       path: {type: String},
-      // Whether to show the Glitch iframe or not.
-      iframeEnabled: {type: Boolean},
+      // Whether we are a mobile browser or not.
+      _isMobile: {type: Boolean},
     };
   }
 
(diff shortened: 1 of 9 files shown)


beaufortfrancois:
I think there's something wrong with CSS in this new PR. See below
`Contribute | File a bug | ...` are not aligned properly

localhost_8080_codelab-remove-unused-code_

src/lib/components/Codelab/index.js
```diff
@@ -47,47 +61,51 @@ class Codelab extends BaseElement {
...
+    if (embed) {
+      url += `#!/embed/${encodeURI(this.glitch)}`;
+    }
 
     return url;
   }
 
   render() {
-    /* eslint-disable indent */
+    const loadGlitch = !this._isMobile && this.glitch;
```
beaufortfrancois: Maybe we should switch to `isDesktop` as the positive side of things is easier to read. That will remove plenty of `!`

src/lib/components/Codelab/index.js
```diff
@@ -47,47 +61,51 @@ class Codelab extends BaseElement {
...
+    let iframePart = "";
+
+    if (loadGlitch) {
```
beaufortfrancois: How about structure like this instead?

```
if (!this.glitch) {
  return '';
}

if (!this.isDesktop) {
  return html`<div class="w-sizer"><div class="w-aside w-aside--warning">...</div></div>`
}

return html`<div class="w-sizer"><iframe ....></div>`
```

src/lib/components/Codelab/index.js
samthor: Done

src/lib/components/Codelab/index.js
```diff
@@ -47,47 +61,51 @@ class Codelab extends BaseElement {
...
+        </iframe>
+      `;
+    } else if (this.glitch) {
+      iframePart = html`
+        <div class="w-aside w-aside--warning">
+          <p>
+            <strong>Warning:</strong> This Glitch isn't available on small
+            screens,
+            <a target="_blank" rel="noopener" href=${this.glitchSrc(false)}>
+              open it in a new tab</a
```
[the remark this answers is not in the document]
samthor: Done

src/lib/components/Codelab/index.js
samthor: I like it. Done.

samthor:
I've pushed a change which makes the codelab itself sticky, and removes the weird scrolling of the codelab page content.

samthor:
The Percy failure here is a random image on the home page. I think this is lazyloading striking again.

samthor:
So this now "works" but we have an amusing problem.

AFAICT, [Percy snapshots the DOM on page](https://docs.percy.io/docs/responsive-visual-testing), and then resizes it later. This means that our media listener never gets a chance to run and never inserts the iframe. Percy renders the larger output with a big warning saying the Glitch isn't available on mobile.

I'm going to tweak the CSS so we still test the sticky behavior in this case. I've pushed a change.

src/lib/components/Codelab/index.js
```diff
@@ -29,7 +32,21 @@ class Codelab extends BaseElement {
 
     this.glitch = "";
     this.path = "index.html";
-    this.iframeEnabled = false;
+    this._isDesktop = false;
```
beaufortfrancois: Shouldn't this be initialized based on `this._mql.matches` instead of calling `this._toggleDesktop();` in `connectedCallback`?

samthor: `render()` is only called between `connectedCallback` and `disconnectedCallback`, so the value of this variable doesn't matter except during that time. I'll add a comment.

beaufortfrancois: Good!

PR #2403, opened 2020-03-21
test percy nav

samthor:
Follow-on from #2399. Adds a navigation test.

Also adds an easier way to do local Percy testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "stage:personal": "ELEVENTY_ENV=dev npm run build && gcloud app deploy --project web-dev-staging --quiet --no-promote",
     "deploy": "ELEVENTY_ENV=prod npm run build && node index-algolia.js && gcloud app deploy --project web-dev-production-1 --quiet",
     "snapshots": "percy exec -- node tools/percy/snapshots.js",
+    "dev:snapshots": "node tools/percy/snapshots-dev.js",
     "test:snapshots": "ELEVENTY_ENV=test npm-run-all --serial build --parallel --race start snapshots"
   },
   "license": "Apache-2.0",
--- a/tools/percy/snapshots.js
+++ b/tools/percy/snapshots.js
@@ -1,6 +1,9 @@
 const PercyScript = require("@percy/script");
 const puppeteer = require('puppeteer');  // for types only
 
+// Puppeteer is imported so that Code and friends can provide types.
+const puppeteer = require('puppeteer');  // eslint-disable-line no-unused-vars
+
 const isDev = Boolean(process.env.WEBDEV_PERCY_DEV)  // set by snapshots-dev.js
 
 const pagesToTest = [
--- a/tools/percy/snapshots.js
+++ b/tools/percy/snapshots.js
@@ -3,7 +3,7 @@ const PercyScript = require("@percy/script");
 // Puppeteer is only imported so that Code and friends can provide types.
 const puppeteer = require("puppeteer"); // eslint-disable-line no-unused-vars
 
-const isDev = Boolean(process.env.WEBDEV_PERCY_DEV)  // set by snapshots-dev.js
+const isDev = Boolean(process.env.WEBDEV_PERCY_DEV); // set by snapshots-dev.js
 
 const pagesToTest = [
   {
--- a/tools/percy/snapshots.js
+++ b/tools/percy/snapshots.js
@@ -2,7 +2,7 @@ const PercyScript = require("@percy/script");
 const puppeteer = require('puppeteer');  // for types only
 
 // Puppeteer is imported so that Code and friends can provide types.
-const puppeteer = require('puppeteer');  // eslint-disable-line no-unused-vars
+const puppeteer = require("puppeteer"); // eslint-disable-line no-unused-vars
 
 const isDev = Boolean(process.env.WEBDEV_PERCY_DEV)  // set by snapshots-dev.js
 
--- a/tools/percy/snapshots.js
+++ b/tools/percy/snapshots.js
@@ -1,7 +1,6 @@
 const PercyScript = require("@percy/script");
-const puppeteer = require('puppeteer');  // for types only
 
-// Puppeteer is imported so that Code and friends can provide types.
+// Puppeteer is only imported so that Code and friends can provide types.
 const puppeteer = require("puppeteer"); // eslint-disable-line no-unused-vars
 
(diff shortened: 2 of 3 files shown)


samthor:
This is failing Percy because we have a new snapshot. :)

tools/percy/snapshots.js
```diff
@@ -25,15 +31,23 @@ const pagesToTest = [
...
+      await percySnapshot(title);
     }
+    // Navigate back to home page by clicking on a link. We expect most assets
+    // to already be lazyloaded by this point so don't scroll.
+    // This is a canary for problems with the SPA routing code.
+    const navigationPromise = page.waitForNavigation();
+    await page.click(".web-header__logo-link");
+    await navigationPromise;
+    await page.waitFor(200);
+    await percySnapshot("Home page, navigation");
```
robdodson: I actually don't know if this page would have shown the failure we were seeing before. It seemed like it was only happening to content images inside of blog posts. I'm curious if it's worth clicking through a few pages and snapshotting them since the snapshots are fairly cheap. Could go Home > Blog > Post

tools/percy/snapshots.js
```diff
@@ -1,5 +1,11 @@
 const PercyScript = require("@percy/script");
 const scrollToBottom = require("scroll-to-bottomjs");
+
+// Puppeteer is imported so that Code and friends can provide types.
+const puppeteer = require('puppeteer');  // eslint-disable-line no-unused-vars
+
+const isDev = Boolean(process.env.WEBDEV_PERCY_DEV)  // set by snapshots-dev.js
```
robdodson: Why do you need this? I was able to run Percy locally without needing to set environment variables (though, you do have to export the PERCY_TOKEN first). Also, I think you'll want the `ELEVENTY_ENV=test` to be set if you're doing percy snapshots because that forces the blog to display the same cards https://github.com/GoogleChrome/web.dev/blob/master/src/site/_collections/post-descending.js#L20. Otherwise every new post would cause the snapshot to fail because it would show up on /blog

tools/percy/snapshots.js
robdodson: Come to think of it... we could just click through the site and snapshot the pages, instead of going directly to the urls. That way we'd test SPA and snapshot all the pages at the same time.

tools/percy/snapshots.js
samthor: This lets me watch Percy run. I just want to pass a flag letting me see it work.

robdodson: You mean it opens a chrome window and you can see it scrolling around and whatnot?

tools/percy/snapshots.js
samthor: So I think you might be right, in that clicking this will still actually work if the JS isn't running. I need to think about this a bit more.

tools/percy/snapshots.js
samthor: Yup. Is there another way to make Percy run that behavior?

I thought about maybe, if this script is running (rather than being `require()`'d) but it seems like Percy just runs it directly too. I thought maybe it was being imported by a Percy harness, but alas.

tools/percy/snapshots.js
robdodson: That's true but I think clicking through the pages would have caught the failures we had the other day where images were being request wrong. I think it's worth adding these tests and continuing to improve on them. But I would like to click through the blog and at least one post to snapshot those pages as well because they're so critical.

package.json
```diff
@@ -28,6 +28,7 @@
     "stage:personal": "ELEVENTY_ENV=dev npm run build && gcloud app deploy --project web-dev-staging --quiet --no-promote",
     "deploy": "ELEVENTY_ENV=prod npm run build && node index-algolia.js && gcloud app deploy --project web-dev-production-1 --quiet",
     "snapshots": "percy exec -- node tools/percy/snapshots.js",
+    "dev:snapshots": "node tools/percy/snapshots-dev.js",
```
robdodson: I think there might be a couple of issues here. You may want this command to also build the site in test mode (which ensures that the blog always has the same 4 posts on it) and start the local server. Rather than have a separate script (snapshots-dev) could the npm script be written to look like `test:snapshots` but have it set both environment variables, ELEVENTY_ENV=test and PERCY_ENV=dev?

tools/percy/snapshots.js
robdodson:
```suggestion
const isDev = Boolean(process.env.WEBDEV_PERCY_DEV);  // set by snapshots-dev.js
```

tools/percy/snapshots.js
```diff
@@ -1,5 +1,11 @@
 const PercyScript = require("@percy/script");
 const scrollToBottom = require("scroll-to-bottomjs");
+
+// Puppeteer is imported so that Code and friends can provide types.
+const puppeteer = require('puppeteer');  // eslint-disable-line no-unused-vars
```
robdodson:
```suggestion
const puppeteer = require("puppeteer");  // eslint-disable-line no-unused-vars
```

package.json
samthor: So since this is literally for development to _see_ what is happening 👀, it actually doesn't matter that it has the same 4 posts on it.

I have a thought. Instead of extra env vars, could we overload `ELEVENTY_ENV=dev` to mean, show the display?

samthor:
So this now "works" but we have an amusing problem.

AFAICT, [Percy snapshots the DOM on page](https://docs.percy.io/docs/responsive-visual-testing), and then resizes it later. This means that our media listener never gets a chance to run and never inserts the iframe. Percy renders the larger output with a big warning saying the Glitch isn't available on mobile.

I'm going to tweak the CSS so we still test the sticky behavior in this case.

package.json
robdodson: > So since this is literally for development to see what is happening 👀, it actually doesn't matter that it has the same 4 posts on it.

I could be wrong but I think percy will still try to upload the screenshots won't it?

> I have a thought. Instead of extra env vars, could we overload ELEVENTY_ENV=dev to mean, show the display?

We can but then we wouldn't get the test posts. I guess I'm ok with us trying it and if it's annoying we'll think of another solution.

robdodson:
btw this shows an example of the gcloud stage thing failing.

samthor:
@robdodson I feel like your testing work supersedes this. I'm going to close.

PR #2410, opened 2020-03-23
add basic bundle for Analytics

### LINKED ISSUE #1941: Not properly tracking signed in state.

samthor:
Fixes #1941 by setting the logged-in dimension.

Creates an additional JS target that is "basic", and run by all browsers. We need this because otherwise the initial page load won't be correctly tagged with signed-in status. (And having the build target lets us share JS resources).

Alternatives considered:
* We could continue to inline the Analytics config and JS in the layout—however, that would make @beaufortfrancois' suggestion to enable CSP difficult to enforce (hash generation for inline scripts)

This also fixes:
* We now wrap `localStorage` as Firefox in Incognito can throw a `SecurityError` while trying to access it (we saw the same problem in Santa Tracker). There's possibly other browsers that also prevent access.


robdodson:
Given the two rather large outages we've had over the past two weeks I think I'd like to put a pause on this and other PRs that modify the guts of the site until we can increase our testing surface area and create more shared knowledge amongst the team. I can think of a number of areas where we need to increase redundancy and it might be good to spend Q2 shoring things up so folks can have more confidence in these changes.

samthor:
Rebased PR.

PR #2412, opened 2020-03-23
actually test WC, switch to karma TDD

samthor:
Felt like working on tests for a bit. Just tests a single WC right now.


robdodson:
ok so i'm going to _not_ approve the percy snapshot because it's a flake and we don't want to approve one if it changes our golden snapshot. So we can ignore that fail while we work on how to make percy more resilient.

PR #2425, opened 2020-03-25
Check for 200 on staging deploy

samthor:
Deploy to App Engine but check for a 200 afterwards.


samthor:
After literally five fives and reruns, this works now. It's making requests to the deployed server and confirming that it 200's.

robdodson:
it seems like this PR contains codelab changes. Is it possible to split them into separate PRs?

Also, the previous action was bombing out because of permission issues. I'm not sure why that is tho...

samthor:
I targeted the wrong branch here. I'm not sure why. I'll recreate this PR.

samthor:
Actually, we dropped staging to App Engine for now, so I'll leave it.

PR #2427, opened 2020-03-25
fix SW redirect and normalizing URLs

### LINKED ISSUE #2422: Wrong link in "A long cache lifetime can speed up repeat visits to your page" section of Lighthouse report.
#### Provide the steps to reproduce
1. Run LH on <www.profissaocanada.com>
1. Go to the Developer Tools
1. Run the Lighthouse tool to generate the report
1. Go to the Performance section.
1. Go to the section "Serve static assets with an efficient cache policy" (10 resources found)
1. Expand it and click on the learn more from the text "A long cache lifetime can speed up repeat visits to your page. Learn more."
1. You just went to an INVALID URL.
  

#### What is the current behavior?
When you follow the steps described above you go to an invalid URL
(`https://web.dev/uses-long-cache-ttl?utm_source=lighthouse&utm_medium=devtools///////////////////`)

#### What is the expected behavior?
It should take the visitor to a valid URL
(`https://web.dev/uses-long-cache-ttl`)

#### Environment Information
* Affected Channels: 
* Lighthouse version: 5.7.1
* Chrome version: Version 83.0.4094.0 (Official Build) canary (64-bit)
* Node.js version:
* Operating System: Windows 10 Version 1903 (OS Build 18362.657)

**Related issues**

samthor:
Fixes #2422

Loads to pages like "/foo?bar" (i.e., with no trailing slash on "/foo" but including a search param) resulted in redirects forever. This was a problem in both the SW and foreground pages.

This also fixes a couple of other issues:

* adds a test for normalizing
* fixes a likely `null` in the modal assessment cleanup code

### DIFF
--- a/src/lib/components/LighthouseGauge/test.js
+++ b/src/lib/components/LighthouseGauge/test.js
@@ -18,7 +18,7 @@
 import {assert} from "../../test/assert";
 import "./index";
 
-suite("web-lighthouse-gauge", async () => {
+suite("element: web-lighthouse-gauge", async () => {
   await customElements.whenDefined("web-lighthouse-gauge");
 
   test("basic", async () => {
--- a/src/lib/loader.js
+++ b/src/lib/loader.js
@@ -6,6 +6,7 @@
  */
 
 import {store} from "./store";
+import {normalizeUrl} from "./urls";
 import "./utils/underscore-import-polyfill";
 
 /**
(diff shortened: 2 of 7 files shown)


robdodson:
Merging this as it's need for a p0 fix but I did notice that it changes the current behavior as hashes now persist in urls across navigation. So if you click on a heading (#foo) in a blog post, then click the top nav to go to /measure, you'll end up with web.dev/measure/#foo. I don't think this will break anything (unless two posts us the same hash) but we should fix it.

robdodson:
oh also I renamed url_test to url-test because I think we use dashes through the code.

PR #2428, opened 2020-03-25
Fix typo in flag name

beaufortfrancois:


jpmedley:
Actually, this flag got renamed at some point.

image

PR #2430, opened 2020-03-25
Fix broken HTML

christianliebel:
Fixes a broken closing `figure` tag.


robdodson:

PR #2431, opened 2020-03-25
Link actual Web Share Target demo

christianliebel:

- Web Share Target demo link led to a Web Share demo, now leads to an actual Web Share Target demo


robdodson:

kaycebasques:
@jpmedley or @petele is this where you want to link the demos?

PR #2432, opened 2020-03-25
Set origin trial URL for Web NFC

beaufortfrancois:
The OT URL was missing somehow.


robdodson:

PR #2438, opened 2020-03-25
Migrate the "Read local files" guide from H5R

### LINKED ISSUE #2439: content: Migrate the "Read local files" guide from H5R
* author: @kaycebasques 
* deadline: N/A
* background
  * update and migrate [read local files](https://www.html5rocks.com/en/tutorials/file/dndfiles/) guide on H5R because it's one of the most popular pages on the topic and having the guide on web.dev will provide a good foundation for us to discuss more modern approaches alongside the old approaches

kaycebasques:
Fixes #2439

Staged page: https://deploy-preview-2438--web-dev-staging.netlify.com/read-files/

### DIFF
--- a/src/site/content/en/blog/read-files/index.md
+++ b/src/site/content/en/blog/read-files/index.md
@@ -9,6 +9,7 @@ authors:
  - kaycebasques
 tags:
   - post
+draft: true
 ---
 
 This guide shows you how to:
@@ -31,7 +32,9 @@ at the time of writing.
 This guide assumes you're familiar with:
 
 * [Beginner web development concepts](https://developer.mozilla.org/docs/Learn)
-* [Events](https://developer.mozilla.org/docs/Learn/JavaScript/Building_blocks/Events)
+* [Events](https://developer.mozilla.org/docs/Learn/JavaScript/Building_blocks/Events), in
+  particular [`Event.stopPropagation()`][propagation] and
+  [`Event.preventDefault()`][preventDefault]
 * [Feature detection][detection]
 * Modern JavaScript syntax like
   [arrow functions](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Functions/Arrow_functions),
@@ -43,17 +46,20 @@ This guide assumes you're familiar with:
 
 ```html
 <!-- The `multiple` attribute lets users select multiple files. -->
-<input type="file" id="input" multiple/>
+<input type="file" id="file-selector" multiple>
 <script>
   if (window.FileList && window.File) {
-    document.getElementById('input').addEventListener('change', event => {
-      const files = event.target.files; // This is a `FileList` object.
-      // `forEach()` won't work unless you convert `files` to an `Array`.
-      for (let i = 0; i < files.length; i++) {
-        const file = files[i]; // This is a `File` object.
+    document.getElementById('file-selector').addEventListener('change', event => {
+      // `file` is a `File` object.
+      // `event.target.files` is a `FileList` object.
+      for (const file of event.target.files) {
+        // Not supported in Safari for iOS.
         const name = file.name ? file.name : 'NOT SUPPORTED';
+        // Not supported in Firefox for Android or Opera for Android.
         const type = file.type ? file.type : 'NOT SUPPORTED';
-        console.log({name, type});
+        // Unknown cross-browser support.
+        const size = file.size ? file.size : 'NOT SUPPORTED';
+        console.log({file, name, type, size});
       }
     }); 
   }
@@ -63,16 +69,12 @@ This guide assumes you're familiar with:
 This example lets a user select multiple files using their operating system's built-in
 file selection UI and then logs out the name and file type of each selected file to the Console.
 
-[`<input type="file">`][input] is the recommended way to enable users to select files because
-browser vendors do a lot of work to make sure that their built-in elements are accessible.
-Some browsers also treat `<input type="file">` elements as automatic drag-and-drop targets.
 After the user clicks the `<input type="file">` element the operating system's built-in
 file selection UI appears. Add the `multiple` attribute to your `<input type="file">` element
 to instruct the operating system to allow the user to select multiple files. When the
 user finishes selecting a file or files the browser fires the `<input type="file">` element's
 `change` event. You can access the list of files from `event.target.files`, which is a [`FileList`][filelist]
-object, not an `Array`. If you want to iterate over `files` with `forEach()` you'll need to convert
-it to an `Array`. Each item in the `FileList` is a [`File`][file] object.
+object. Each item in the `FileList` is a [`File`][file] object.
 
 <div class="glitch-embed-wrap" style="height: 480px; width: 100%;">
   <iframe src="https://glitch.com/embed/#!/embed/input-type-file?previewSize=100"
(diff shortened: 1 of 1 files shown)


kaycebasques:
Requests for code/technical reviewers:

* Are the code samples using good cross-browser compatibility design patterns?
* Are there good error-handling design patterns that should be added to the code samples (without making them extremely long or harder to follow)?
* Are any of code samples **egregiously** outdated/deprecated? I'm aware that these patterns probably aren't modern but they seem to be supported across all browsers. If the more modern stuff is supported across all browsers then we can remove these old patterns, otherwise we should just discuss the modern patterns alongside the old patterns with the disclaimer that the modern patterns are not supported on all browsers.

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+[`FileList`](https://developer.mozilla.org/docs/Web/API/FileList#Browser_compatibility)
+
+### Select files with a custom drag-and-drop UI (not recommended) {: #drag-and-drop }
+
+{% Aside 'caution' %}
+  [Selecting files with an `<input type="file">` element](#input) is the recommended
+  approach because browser vendors do a lot of work to ensure that their elements
+  are accessible. However, MDN doesn't currently state how many browsers treat their
+  `<input type="file">` elements as built-in drag-and-drop targets.
+{% endAside %}
```
kaycebasques: @robdodson it seems like the usual rationale for using built-in elements also applies here? Don't use a custom drag-and-drop because it will take a lot of work to handle all the a11y edge cases.

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+UI you access the selected files from `event.dataTransfer.files` in your `drop` event listener,
+which is a `FileList` object.
+
+{# This example doesn't work as an embed. #}
+
+Check out [Custom drag-and-drop](https://custom-drag-and-drop.glitch.me/)
+for a live demonstration.
+
+Browser compatibility data for the APIs used in this section:
+[`<input type="file">`](https://developer.mozilla.org/docs/Web/HTML/Element/input/file#Browser_compatibility),
```
mkruisselbrink: nit: I don't think this section uses input type=file?

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+[`File`](https://developer.mozilla.org/docs/Web/API/File#Browser_compatibility),
+[`FileList`](https://developer.mozilla.org/docs/Web/API/FileList#Browser_compatibility),
+[`DataTransfer`](https://developer.mozilla.org/docs/Web/API/DataTransfer#Browser_compatibility)
+
+## Read a file's content {: #read }
```
mkruisselbrink: Using FileReader is defintiely the most cross-browser compatible way of reading a file's contents, however we should probably also mention the more modern way where you can now call `file.text()`, `file.arrayBuffer()` and `file.stream()` to read files as well? Currently supported in recent versions of chrome and firefox (and trivially polyfillable)

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+
+[compat]: https://developer.mozilla.org/docs/Web/API/File#Browser_compatibility
+[file]: https://developer.mozilla.org/docs/Web/API/File
+[filereader]: https://developer.mozilla.org/docs/Web/API/FileReader
+[filelist]: https://developer.mozilla.org/docs/Web/API/FileList
+[input]: https://developer.mozilla.org/docs/Web/HTML/Element/input/file
+[detection]: https://developer.mozilla.org/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection
+[events]: https://developers.google.com/analytics/devguides/collection/analyticsjs/events
+[buffer]: https://developer.mozilla.org/docs/Web/API/FileReader/readAsArrayBuffer
+[binary]: https://developer.mozilla.org/docs/Web/API/FileReader/readAsBinaryString
```
mkruisselbrink: you're not actually referencing this anywhere?

kaycebasques:
@petele I followed up in email but this is the plan and rationale:

The best approach is to add a section about the Native File System (NFS) API to the guide that I migrated (which will live at https://web.dev/read-files/). The NFS API section should just show the basic usage of the NFS API and then link out to the existing guide (https://web.dev/native-file-system/). This will dramatically increase awareness of the NFS API while still allowing web developers to decide which approach is right for them. From the MDN survey we know that cross-browser compatibility is their #1 frustration so we should expect most to still use the old APIs.

https://web.dev/native-file-system/ is still good to have because it goes into a level of detail that some (but not all) developers will want. So I definitely see these as complementary efforts/guides, not overlapping.

Long story short, this migrated content is the foundational entry point to the topic which will eventually cover all available approaches. The dedicated NFS guide provides additional detail for the subsection of web developers who decide that NFS is the right approach for them.

src/site/content/en/blog/read-files/index.md
robdodson: that seems ok to me. I don't know much about accessible drag and drop except that everyone says it's hard/impossible. So having an actual element that someone can just tab to sounds good imo.

kaycebasques:
Sam's feedback: https://docs.google.com/document/d/1qse23g_37ex8svpxpKg1zBJagI36pFGcAUilutykXBQ/edit

src/site/content/en/blog/read-files/index.md
kaycebasques: Copy-paste error. Good catch.

src/site/content/en/blog/read-files/index.md
kaycebasques: Can you give me the polyfill code for each?

kaycebasques: And these exist on the `File` interface? I'm not seeing any of them on MDN: https://developer.mozilla.org/en-US/docs/Web/API/File

src/site/content/en/blog/read-files/index.md
kaycebasques: Right, I omitted it because MDN says that it was once deprecated and then put back for backwards compatibility and recommends using an array buffer instead: https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsBinaryString

src/site/content/en/blog/read-files/index.md
mkruisselbrink: `File` inherits from `Blob`, so these are listed on https://developer.mozilla.org/en-US/docs/Web/API/Blob

The easiest way to polyfill these would be using `Response`:
`new Response(file).text()` is equivalent to `file.text()`, `new Response(file).arrayBuffer()` is equivalent to `file.arrayBuffer()` and `new Response(file).body` is equivalent to `file.stream()`. I think at least all modern browsers implement Response, so that should work pretty much everywhere today.

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+
+[`<input type="file">`][input] is the recommended way to enable users to select files because
+browser vendors do a lot of work to make sure that their built-in elements are accessible.
+Some browsers also treat `<input type="file">` elements as automatic drag-and-drop targets.
+After the user clicks the `<input type="file">` element the operating system's built-in
+file selection UI appears. Add the `multiple` attribute to your `<input type="file">` element
+to instruct the operating system to allow the user to select multiple files. When the
+user finishes selecting a file or files the browser fires the `<input type="file">` element's
+`change` event. You can access the list of files from `event.target.files`, which is a [`FileList`][filelist]
+object, not an `Array`. If you want to iterate over `files` with `forEach()` you'll need to convert
```
pwnall: Is `for ... of` the thing to recommend these days? I prefer it to `forEach()`.

src/site/content/en/blog/read-files/index.md
```diff
@@ -0,0 +1,253 @@
...
+</div>
+
+Browser compatibility data for the APIs used in this section:
+[`<input type="file">`](https://developer.mozilla.org/docs/Web/HTML/Element/input/file#Browser_compatibility),
+[`File`](https://developer.mozilla.org/docs/Web/API/File#Browser_compatibility),
+[`FileList`](https://developer.mozilla.org/docs/Web/API/FileList#Browser_compatibility)
+
+### Select files with a custom drag-and-drop UI (not recommended) {: #drag-and-drop }
+
+{% Aside 'caution' %}
```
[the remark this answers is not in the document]
kaycebasques: @petele had similar feedback. I clarified the section:

image

src/site/content/en/blog/read-files/index.md
kaycebasques: Yeah I vaguely recall people not liking `forEach()` so I'll just remove that outright. The code samples are now using `for ... of` BTW... @samdutton gave me the same feedback in his Doc.

src/site/content/en/blog/read-files/index.md
kaycebasques: Created #2462 for a reminder to do this later.

kaycebasques:
@petele and I agreed to the following plan:

* I'm going to merge this now as a draft, which means it will be in the codebase but won't be published to the web.dev production site.
* Pete's team will have a month to update the content.
* If they haven't made significant progress on the update within a month then I can publish this post as-is (I'll work on #2462 before publishing).

The rationale for waiting is that Pete's team has a better progressive enhancement story that incorporates modern APIs like Native File System (Pete explained the progressive enhancement story and I agree that it's better than what I've provided here). The rationale for putting a hard deadline around the waiting period is that the updated content in this PR is better than having web developers continue to rely on the outdated H5R guide from 2010.

PR #2442, opened 2020-03-26
Update Web OTP for origin trial graduation

jpmedley:

PR #2443, opened 2020-03-26
Update fugu-status for Web OTP.

jpmedley:


jpmedley:
@agektmr Here's the update of the OTP article. Can you please update your samples the new version of the API?

agektmr:
There's already one but I'm working on more public friendly version. Please allow me a few days.

agektmr:
The same change is in #2442. Should we close this pull request?

jpmedley:
I forgot that was in there. Thanks.

PR #2454, opened 2020-03-27
Correct window to navigator in supporting section

NOtherDev:

- fixes the invalid call of `setAppBadge` and `clearAppBadge` - they are available within `Navigator`, as the article itself informs.


NOtherDev:
@googlebot I signed it!

kaycebasques:
@petele WDYT of this change?

jpmedley:
@kaycebasques This is correct. The spec was change, but what we care about in terms of an article is whether it's landed in Chrome. You can [see that here](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/badging/navigator_badge.idl).

kaycebasques:
Thanks @NOtherDev & @jpmedley

PR #2460, opened 2020-03-27
Add node_modules caching to speed up Actions.

robdodson:

- Add cache action to speed up our build times ([see cache Action explainer](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows))
- Change percy task to only run on pull_requests. It seems duplicative to run it when we merge into master as well.

### DIFF
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -4,6 +4,14 @@ name: Percy
 # configure each event separately. You must append a colon (:) to all events,
 # including events without configuration.
 on:
+  push:
+    branches:
+      - master
+    paths:
+      - '**.js'
+      - '**.njk'
+      - '**.scss'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/percy-workflow.yml
```diff
@@ -4,13 +4,6 @@ name: Percy
 # configure each event separately. You must append a colon (:) to all events,
 # including events without configuration.
 on:
-  push:
```
MichaelSolati: So we wouldn't run percy on push to master? I guess we never really push straight to master... So there's no need to check how the site looks, the push wouldn't be blocked. But incase there is a push to master, couldn't percy notify us if the pages don't look right?

robdodson: It certainly could, though at the moment we're not doing a good job of seeing those kinds of alerts. But that seems like a good enough reason to leave it in for now. I'll update the PR!

PR #2463, opened 2020-03-27
Feat/scheduled posts

### LINKED ISSUE #2698: Let authors schedule a post to go live on a specific date.

robdodson:
Fixes #2698 

- Let folks schedule posts in the future. To schedule a future post set `scheduled: true` and set the `date` to some date in the future. Scheduled posts will be deployed when the site does its usual build at 7am PST/8am PDT. Prior to going live the post will show up as a draft while in dev mode.
- Adds tests for the live-post filter and isScheduledForTheFuture helper function.
- Adds a visual indicator for draft post cards.
Screen Shot 2020-04-18 at 11 38 16 AM


jpmedley:
Yeah! 

Do I need to do anything special to invoke this? Or do I just need to add a future date?

robdodson:
Adding the future date should do it. The site gets rebuilt and deployed every day at 7am PST. After this lands we should try it with a test post to verify.

jpmedley:
Feel free to hit me up for that. I think I'm one of the main advocates of this feature.

src/site/_filters/live-posts.js
```diff
@@ -6,13 +6,11 @@ const {env} = require("../_data/site");
...
 module.exports = function livePosts(post) {
-  // If we ever wanted to enable featured posts we could uncomment these lines.
-  // See https://github.com/GoogleChrome/web.dev/pull/1222#issuecomment-516621561
-  // for an explanation on why this feature is currently disabled.
-  // const now = new Date();
-  // if (post.date > now) {
-  //   post.data.draft = true;
-  // }
+  const today = new Date();
+  today.setUTCHours(12, 0, 0, 0);
```
samthor, line 10: Is the point of this line that it matches 7am PST? What about DST and friends?

robdodson: yeah. i added it based on one of your prior suggestions. idk, what do you recommend we do here?

https://twitter.com/rob_dodson/status/1121928275410866177

robdodson:
If I'm not mistaken, I think the logic will work like this:

- Eleventy calls `new Date()` using the `date` field. [The date will be at midnight UTC time](https://www.11ty.dev/docs/dates/#dates-off-by-one-day).
  - e.g. `new Date('2020-04-18')` will be midnight, UTC time on April 18th, 2020.
- The filter will also call `new Date()` at the time that it runs. This time will be in UTC hours.
  - e.g. calling `new Date()` at 7am PST will return a date with hours set to 15 (UTC).
- Finally, we compare these two dates.

**Important notes**:
- Our GitHub Action [uses UTC time for its schedule](https://help.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events-schedule). Currently it's set to run at 15:00 UTC / 07:00 PST (or 08:00 PDT) ([source](https://github.com/GoogleChrome/web.dev/blob/c8557c2869afe3a2699c60e43a84d5aa1a16d5c6/.github/workflows/continuous-integration-workflow.yml#L7)).
- If someone manually deploys the site at 4pm PST / 5pm PDT on April 18th it will be midnight April 19th in UTC time so those scheduled posts would go live (probably not intended).
  - Because we deploy on merge it means we'd probably be launching posts intended to go live the next day.

**Thoughts**:
When we do the comparison of dates in this filter, should we offset the eleventy date so it uses something like 07:00 UTC instead of 00:00?

src/site/_collections/algolia-posts.js
```diff
@@ -42,7 +42,7 @@ function limitText(fulltext, limit = 7500) {
 module.exports = (collection) => {
   const validTags = ['post', 'pathItem'];
   const eleventyPosts = collection
-    .getAll()
+    .getFilteredByGlob('**/*.md')
```
samthor, line 45: This seems unrelated but probably fine. I presume this fixes the issue that we're e.g. indexing our empty partial.

test/site/_filters/live-posts.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('should return false if the post has a past date but is a draft', function() {
+    // set the post date to 1 day in the past.
+    post.date.setDate(post.date.getDate() - 1);
+    post.data.draft = true;
+    const actual = livePosts(post);
+    assert.strictEqual(actual, false);
+  });
+});
```
samthor, line 84: I wonder if it's worth adding a param to `livePosts`, which defaults to `new Date()`, that lets us test time just before and just after 15:00 UTC.

We have so many tests already and it seems like low-hanging fruit.

src/site/_collections/algolia-posts.js
robdodson: yeah I added this because I noticed we were running all sorts of stuff like sitemap.xml through liveposts and that seemed...bad.

test/site/_filters/live-posts.js
robdodson: added it!

src/site/_filters/live-posts.js
```diff
@@ -1,23 +1,87 @@
...
+ * If a post has a future date it will automatically be set to `draft: true`.
+ * When the date arrives, our daily GitHub Action that publishes the site
+ * should pickup the new post and publish it.
+ * This action runs at around 7am PST / 15:00 UTC.
+ * Because Eleventy sets the post.date to midnight, UTC time, we offset it
+ * to be at 15:00.
+ * If we did not do this, then deploying the site at 4pm PST / 00:00 UTC
+ * would launch posts intended for the next day.
+ * @param {*} post An eleventy post object.
+ * @param {*} now A Date object representing the current time. You shouldn't
```
samthor, line 33: `@param {!Date=}` ?

! means non-null, = means there's a default

PR #2465, opened 2020-03-28
Redirect focus on incorrect question submit

mfriesenhahn:

- Currently, when a user submits an incorrect response to a self-assessment question, focus remains on the **Check** button even though it becomes disabled. To improve UX for keyboard users, this redirects focus to the currently active tab


src/lib/components/AssessmentQuestion/index.js
```diff
@@ -120,6 +120,9 @@ class AssessmentQuestion extends BaseElement {
         this.updateResponseComponents();
         this.state = "unanswered";
         this.ctaLabel = "Recheck";
+
+        const tabs = this.closest("web-tabs");
+        tabs.focusTab(tabs.activeTab);
```
samthor: Can we guard this with a null check?

PR #2466, opened 2020-03-29
Auto-update updated field on commits.

robdodson:
Related to https://github.com/GoogleChrome/web.dev/issues/1959

- Adds a precommit script that will rev the `updated` field on modified posts. If a post does not have an `updated` field it will log a warning to the console but it won't block the commit.

I'm separately working on a linter that will run as a GH Action when someone does a PR. That linter will also flag the `updated` field if it hasn't been updated. That way the reviewer can decide if it's worth updating it or not (e.g. maybe deciding not to update it for small typo fixes).

### DIFF
--- a/tools/update-updated/index.js
+++ b/tools/update-updated/index.js
@@ -1,0 +1,81 @@
+/**
+ * @fileoverview Task for updating `updated` YAML field.
+ *
+ * @author Matt Gaunt & Rob Dodson 💕
+ */
+
+// This task only runs against files in this directory.
+const contentDir = "src/site/content";
+
+const chalk = require("chalk");
+const fs = require("fs").promises;
+const moment = require("moment");
+const util = require("util");
+const exec = util.promisify(require("child_process").exec);
+const RE_UPDATED = /^updated:\s?(.*?)\n/m;
+const MSG_UPDATE = `Updated ${chalk.bold("updated")} in`;
+
+/**
+ * Gets the list of modified and renamed files that have been staged.
+ * This will ignore files that were added or deleted.
+ * @return {Promise<Array<string>>} Returns array of changed files.
+ */
+async function getChangedFiles() {
+  const cmd = `git diff --cached --name-only --diff-filter=MR`;
+  try {
+    const {stdout} = await exec(cmd, ".");
+    return stdout.split("\n");
+  } catch (err) {
+    throw err;
+  }
+}
+
+const run = async () => {
+  // List of all files that have changed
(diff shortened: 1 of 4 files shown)


robdodson:
fyi @kaycebasques @jpmedley

tools/update-updated/index.js
```diff
@@ -0,0 +1,81 @@
...
+    throw err;
+  }
+}
+
+const run = async () => {
+  // List of all files that have changed
+  const changedFiles = await getChangedFiles();
+
+  for (const changedFile of changedFiles) {
+    if (changedFile.indexOf(contentDir) === -1) {
```
samthor: should this be `startsWith`?

tools/update-updated/index.js
```diff
@@ -0,0 +1,81 @@
...
+      // File isn't a content file, skip it.
+      continue;
+    }
+
+    if (!changedFile.endsWith(".md")) {
+      // File isn't a Markdown file, skip it.
+      continue;
+    }
+
+    const fileContents = (await fs.readFile(changedFile)).toString();
```
samthor: `fs.readFile(changedFile, 'utf8')` skips the `toString` requirement

tools/update-updated/index.js
```diff
@@ -0,0 +1,81 @@
...
+        `${chalk.black.bgYellow(
+          `Warning:`,
+        )} Could not find updated field in ${changedFile}.`,
+      );
+      continue;
+    }
+
+    const originalUpdated = matched[0];
+    const originalTimestamp = matched[1];
+    const momentNow = moment();
```
samthor: Moment is how many kb and you pull it in purely to do.. what could be a string comparison? :(

```js
const now =  new Date().toISOString().substr(0, 10);
if (now < originalTimestamp) {
  continue;
}
```

robdodson: I'm mostly copying over what was in the webfu script which uses moment: https://github.com/google/WebFundamentals/blob/6cf700da9c2e9f5b3b3f271e27f2ad1e5b2e6a28/gulp-tasks/wfUpdatedOn.js#L71

I thought about removing it but chose to leave it for two reasons:
- I know folks shouldn't ship moment to the client, but this is running in node as part of the build process. The kb are irrelevant to me.
- I liked moment's interface for this and the formatting on line 71.

robdodson:
thanks for the review! 🙌

jpmedley:
Woohoo!

Issue #2468, opened 2020-03-30
content: update contents mentioning Chrome 82


myakura:
### What page(s) need to be updated?

From: https://github.com/GoogleChrome/web.dev/search?q=%22Chrome+82%22&type=Code

- [x] `blog/content-indexing-api/index.md`
- [x] `blog/get-installed-related-apps/index.md`
- [x] `blog/module-workers/index.md`
- [x] `blog/shape-detection/index.md`
- [x] `blog/preload-optional-fonts/index.md`

### Why is this update needed?

[Chrome 82 is cancelled](https://chromereleases.googleblog.com/2020/03/chrome-and-chrome-os-release-updates.html) and no longer exists.

jpmedley:
@kaycebasques I add a checklist to the description.

PR #2480, opened 2020-03-31
[Shape Detection] Update launch info: 82 → 83

tomayac:

- Update launch info: 82 → 83


jpmedley:
Addresses #2468

PR #2490, opened 2020-04-01
Update release timing information.

jpmedley:


jpmedley:
@rowan-m PTAL

jpmedley:
@rowan-m PTAL

FYI. my language around this is not as strong as the Chromium release. The director of Chrome engineering may be free to ignore the PR advice around the virus, but I am not.

PR #2492, opened 2020-04-01
Module Workers release: m82 -> m83

developit:
Changes Module Workers release version from m82 to m83.


jpmedley:
Addresses #2468

PR #2493, opened 2020-04-01
Add a workflow to catch the DO NOT MERGE label.

robdodson:

- Adds a GitHub Action to catch the DO NOT MERGE label. Should be set as a required check for branch protection in repo settings.

### DIFF
--- a/.github/workflows/required-checks-workflow.yml
+++ b/.github/workflows/required-checks-workflow.yml
@@ -6,8 +6,8 @@ on:
 
 jobs:
   do_not_merge:
+    name: Do not merge
     runs-on: ubuntu-latest
     steps:
-      - name: Do not merge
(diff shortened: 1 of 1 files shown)


robdodson:
I think this PR is hitting this issue with duplicate checks: https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request

I've emailed GitHub support about it.

jpmedley:
Good call!

robdodson:
GitHub folks confirmed this is a bug in GitHub Actions and said they'd add it to their backlog.

robdodson:
adding the do not merge label as this PR shouldn't be merged until gh fixes the issue cited above.

robdodson:
yay github fixed their issue so we can land this now.

PR #2498, opened 2020-04-02
Clean up percy.

robdodson:

- Disable the glitch embed when percy is running against the components page.

I think at some point we should probably rename the `site.env` variable since it's specific to eleventy at this point but I didn't want to do that in this PR for fear of breaking something unseen.


robdodson:
@MichaelSolati it's set in package.json when you do `npm run percy`.

MichaelSolati:
Kk, I don't think I saw it in the changes, but kk

robdodson:
ah yeah it was done in a previous pr

PR #2502, opened 2020-04-02
Task/add integration test

robdodson:

- Begins to add integration tests. This first batch just asserts that key files show up in our build.

Kinda curious what folks think about the npm scripts.


robdodson:
I realized that we can just do `npm run build` because eleventy is what actually generates the `algolia.json` file. No need to do a prod build since the only real change there is the image urls are different.

robdodson:
I guess the next question is to figure out where this fits in our github actions.

robdodson:
looks like Percy caught a CSS issue with the blue line at the top of the cards:

image

You can ignore the second percy failure, it's just barking about the glitch embed which will be fixed when https://github.com/GoogleChrome/web.dev/pull/2498 lands.

package.json
```diff
@@ -23,12 +24,14 @@
...
-    "test": "npm-run-all clean build:rollup karma",
+    "test": "npm run test:unit && npm run test:integration",
+    "test:integration": "mocha ./test",
+    "test:unit": "npm-run-all clean rollup karma",
     "watch:eleventy": "eleventy --watch --quiet",
     "watch:gulp": "npx gulp build && gulp watch",
     "watch:rollup": "chokidar \"src/lib/**/*.{js,scss}\" -c \"npm run rollup\"",
     "watch:sass": "chokidar \"src/styles/**/*.scss\" -c \"npm run sass\"",
-    "watch:test": "TEST_MODE=dev npm-run-all clean build:rollup --parallel watch:rollup karma"
+    "watch:test": "TEST_MODE=dev npm-run-all clean rollup --parallel watch:rollup karma"
```
samthor, line 34: I hate to nitpick but this is now "watch:test:unit" rather than integration tests, right?

robdodson:
> So this is fine for now but I still think the npm commands are a bit confusing.

yeah I agree. Because the integration tests will probably be a lot slower than the component specific unit tests I wanted to separate them out. But naming leaves a lot to be desired.

> A semi-serious alternative proposal: we just have one test target, but that itself spawns Karma or (even better) Puppeteer. Internally that test runner might support e.g. "--type=unit" or "--type=integration". Just my 2c.

hm yeah maybe we could do something like `npm run test -- unit` which passes the arg `unit` to a test.js that does an `exec` based on what was passed in? However, we still have tasks like `watch:test` or whatever that need to set env vars and run rollup and karma. Not sure if we want to also put those into the script or keep them as npm tasks..?

PR #2526, opened 2020-04-07
Feat/cdn all the things

robdodson:
Related to #2446 

- Adds an [eleventy transform](https://www.11ty.dev/docs/config/#transforms) which converts _all_ image paths to use our image CDN. This transform gets run against the HTML for the pages right before eleventy writes them to disk (so at that point all shortcodes and markdown rendering have happened).
- Adds tests for the transform.

Note the transform is called responsive-images because I would like to do a follow up that adds srcset/sizes support to it. But step one is just getting all of our images to consistently be served by our CDN.

TODOs:

- [x] Need to change build-partial.js over to being a transform that runs _after_ responsive-images. Otherwise our service worker partials will use the original image paths.
- [x] Need to add tests for the new build-partial transform.
- [x] Need to remove usage of `get-image-path.js` as its now subsumed by this transform.
- [x] Ugh, now srcset urls are being skipped over :P


src/site/_transforms/responsive-images/helpers.js
```diff
@@ -0,0 +1,42 @@
...
+  // false. In this scenario we should also just return the src as is.
+  if (!isLocal || !outputPath) {
+    return {
+      src,
+      isLocal,
+    };
+  }
+
+  if (path.isAbsolute(src)) {
+    return {src: new URL(src, site.imageCdn), isLocal};
```
samthor, line 25:
```js
const u = new URL(src, site.imageCdn);
return {src: u.toString), isLocal};
```

src/site/_transforms/responsive-images/helpers.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  // At this point we've determined that the image has a relative path.
+  let base = path
+    .dirname(outputPath)
+    .split(path.sep)
+    .pop();
+
+  // The handbook is the only place on the site where we nest paths.
```
samthor, line 34: But it's pretty easy to make more of these, even accidentally. Can we just set the path based on the input? I don't know what "outputPath" here is exactly.

test/README.md
```diff
@@ -0,0 +1,58 @@
...
+certain files (like our sitemap or RSS feed).
+
+These tests run using mocha.
+
+We would like to eventually add puppeteer integration tests to check that
+clicking through the site works as expected.
+
+## Percy screenshot tests
+In addition to the automated tests mentioned above, we also use
+[Percy](percy.io) to do screenshot testing of key pages. These tests are run as
```
samthor, line 52: https://percy.io

.eleventy.js
```diff
@@ -229,6 +230,10 @@ module.exports = function(config) {
     config.addTransform('disable-lazy-load', disableLazyLoad);
   }
 
+  if (process.env.ELEVENTY_ENV === 'prod') {
+    config.addTransform('responsive-images', responsiveImages);
```
samthor, line 234: I feel like if we add more cheerio things we'll want to do them at once (parsing overhead). But this is probably fine for now.

src/site/_transforms/responsive-images/index.js
```diff
@@ -0,0 +1,12 @@
+const cheerio = require("cheerio");
+const {determineImagePath} = require("./helpers");
+
+module.exports = (content, outputPath) => {
```
samthor, line 4: So this won't run on partials. Would it be useful making this a shortcode instead, so we parse/update the content at the right stage of the pipeline? (i.e., at the same place we steal the partial right now)

I also don't think we should CDN images that are part of our site's UI.

.eleventy.js
devnook: Do you mean to have one prod-transform taht combines all necessary sub-transforms in one pass?

MichaelSolati: I'm not opposed to having one transform, but as a small complaint... The way we do imports currently is by creating the path with a template literal, which is a nuisance to dig into the source (can't click through in vscode). Not a necessary bit, but is there any chance we can consider using plain strings?

src/site/_transforms/responsive-images/helpers.js
MichaelSolati:
```js
return {src: u.toString(), isLocal};
```

or

```js
return {src: u.href, isLocal};
```

src/site/_transforms/responsive-images/index.js
robdodson: > So this won't run on partials. Would it be useful making this a shortcode instead, so we parse/update the content at the right stage of the pipeline? (i.e., at the same place we steal the partial right now)

I think you're referring to the service worker partials, and not nunjucks partials, correct?
I converted build-partial over to being a transform that will run after this transform.

> I also don't think we should CDN images that are part of our site's UI.
Why not?

.eleventy.js
robdodson: > I feel like if we add more cheerio things we'll want to do them at once (parsing overhead). But this is probably fine for now.

Btw i don't think we necessarily can do these all at once. For instance, this transform mutates the html then pipes it on to the next transform. The next transform in the chain needs to send the new html to cheerio or it might be working with outdated data.

src/site/_transforms/responsive-images/index.js
robdodson: discussed over chat. Sam was concerned about CDN'ing these images as they're already cached by the SW and they're fundamentally different from content images. Sam was nice and let me keep it for now but we agreed if imgix goes down i will have to wear the shame cube.

robdodson:
I think I'm happy with this initial change and it _seems_ to work as expected. Unfortunately imgix is having uptime issues today so it's a little hard to be 100% confident in my test. Let's hold off on merging and give it another run through next week.

src/site/_transforms/responsive-images/helpers.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  if (path.isAbsolute(src)) {
+    const url = new URL(src, site.imageCdn);
+    return {src: url.href, isLocal};
+  }
+
+  // At this point we've determined that the image has a relative path.
+  // outputPath will be something like dist/en/some-article/index.html
+  // so remove the first two items and use the rest to get the url directory.
```
samthor, line 52: It might be sane to check for a two-letter prefix before just removing it; things might change in i18n-land fast.

Something like:

```js
const parts = path.dirname(outputPath).split(path.sep);
if (parts[0] !== 'dist') {
  throw new ...
}
parts.shift();
if (parts[0].length === 2) {
  parts.shift();
}
```

... but I don't have a strong opinion.

samthor: Or maybe a regexp.

```js
const expectedPrefix = /^dist\/(|\w\w\/)/;
path = path.replace(expectedPrefix, '');
```

src/site/_transforms/responsive-images/index.js
```diff
@@ -0,0 +1,49 @@
...
+  const $ = cheerio.load(content);
+  const $img = $("img");
+  $img.each((i, elem) => {
+    const $elem = $(elem);
+    const originalSrc = $elem.attr("src");
+    const newSrc = determineImagePath($elem.attr("src"), outputPath).src;
+    $elem.attr("src", newSrc);
+    // If the image already has srcset defined,
+    // update that to use the image CDN as well.
+    // Note that this assumes the srcset paths use the same src as the image.
```
samthor, line 35: can we not?
I feel like we should tell our authors _not_ to use `srcset`, you update it here, and we put the default web.dev-hosted image at the end (we talked about the fallback).

test/site/_transforms/service-worker-partials/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+      const hash = Math.random().toString(36);
+      const outputPath = path.join(".", ".tmp", hash, "index.html");
+      await serviceWorkerPartials($.html(), outputPath);
+      const actual = JSON.parse(
+        await fs.readFile(outputPath.replace(".html", ".json"), "utf-8"),
+      );
+      assert.deepStrictEqual(actual, expected);
+      // Cleanup .tmp dir
+      fs.rmdir(path.join(".", ".tmp"), {recursive: true});
```
samthor, line 40: `await`

Actually, can we add this to the `afterEach` section? Even `fs.rmdirSync` if we can't await there.

```js
const fsBase = require("fs");
const {promises: fs} = fsBase;

describe('service-worker-partials', function() {

  afterEach(function() {
    $ = null;
    fsBase.rmdirSync(path.join(".", ".tmp"), {recursive: true});
  });
```

src/site/_transforms/responsive-images/index.js
robdodson: chatted over video about this. expanding the comment to mention that some components like Hero and PostCard rely on this because they do srcsetting ahead of time. Eventually we'd like to remove that behavior from those components and have all the srcsetting done in this transform. That'll be phase 2.

PR #2536, opened 2020-04-09
wrap localStorage to deal with Firefox issues

samthor:
Wraps access to `window.localStorage`. This works around Firefox throwing a `SecurityException` in its private mode when you access this object.


robdodson:
oh looks like it needs a rebase too. I changes the project over to using single quotes which touches basically every file :P

samthor:
Percy looks flaky. Going to commit.

PR #2537, opened 2020-04-09
build script cleanup + virtual imports

samthor:
Extracted from #2410 which will get closed.

* Adds virtual imports to all our build bundles (since you can just _not_ import them if you don't need them in your code)
  * Exports individual keys (so our JS bundles aren't all including the whole config each)
  * Update some code to use these keys

* Adds comments to the build script
* Disallows all external imports—i.e., those which Rollup can't find—Rollup would normally allow mistypings or missing deps (for example, if you haven't run `npm install` in a while), now throw

### DIFF
--- a/build.js
+++ b/build.js
@@ -36,6 +36,9 @@ process.on('unhandledRejection', (reason, p) => {
   process.exit(1);
 });
 
+/**
+ * Virtual imports made available to all bundles. Used for site config and globals.
+ */
 const virtualImports = {
   webdev_config: {
     prod: isProd,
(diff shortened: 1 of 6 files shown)


src/lib/bootstrap.js
```diff
@@ -5,14 +5,15 @@
...
-import config from 'webdev_config';
+import {version} from 'webdev_config';
 import './webcomponents-config'; // must go before -loader below
 import '@webcomponents/webcomponentsjs/webcomponents-loader.js';
 import {swapContent, getPartial} from './loader';
 import * as router from './utils/router';
 import {store} from './store';
 
-console.info('web.dev', config.version);
+// TODO(samthor): This forces a rehash of the JS on every deploy. Remove this.
```
robdodson, line 15: should this be removed in this PR?

build.js
```diff
@@ -29,29 +29,36 @@ const rollup = require('rollup');
...
-      .toISOString()
-      .replace(/[\D]/g, '')
-      .slice(0, 12),
-  firebaseConfig: isProd ? site.firebase.prod : site.firebase.staging,
+/**
+ * Virtual imports made available to all bundles. Used for site config and globals.
+ */
+const virtualImports = {
+  webdev_config: {
+    prod: isProd,
```
robdodson, line 44: any reason to use `prod` instead of `isProd` here?

PR #2539, opened 2020-04-09
Domesticmouse/gitpod setup

domesticmouse:
Hey @samthor,

Any thoughts on adding configuration to make this easily usable from gitpod.io?

brett


robdodson:
I think I have mixed feelings about this change. I like the idea of making it easier for folks to hack on the site, but I don't like the idea of having to maintain files for a tool that (to my knowledge) none of our team uses or is familiar with. If folks file bugs against the dockerfile for instance, I don't think I we want to spend time fixing it.

domesticmouse:
Cool, thanks Rob

Issue #2542, opened 2020-04-09
content: Update Native File System article 


tomayac:
handle.createWriter is now handle.createWritable. 

I’m on this, but @petele FYI.

petele:
Ah- sorry, I was already on it. Just submitted PR #2543

PR #2543, opened 2020-04-09
Update Native File System article

petele:
- Updates `WritableStream`
- Adds serializability of FileHandles

**Still TODO:**

* [x] Add comment about `getFile` only valid while the file hasn't changed
* [x] Update images
* [x] Add some sample code for saving file handle to indexed db
* [x] Verify permissions section has info about permission being for the domain, not the tab.


tomayac:
Fixes #2542

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +286,40 @@ butDir.addEventListener('click', async (e) => {
 });
 ```
 
+## What's new/changed? {: #whats-new }
+
+Chrome 82
+
+* Support for [writable streams](#save-to-disk) was added, and the previous
+  method for writing to disk (`FileSystemWriter`) was deprecated.
+* File handles can now be serialized and stored in IndexedDB, or `postMessage`ed
```
tomayac, line 295:
```suggestion
* File handles can now be serialized and stored in IndexedDB, or sent via `postMessage`
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -216,17 +255,18 @@ takes a string, which is what we want for a text editor. But it can also take a
...
+you could offer to re-open the last file when the app is opened, or show a
+list of recently edited files within your app. You can also `postMessage`
+file handles between the same top-level origin, for example between tabs or
+from tabs to workers.
 
-#### Open a directory and enumerate its contents
+### Open a directory and enumerate its contents
 
 To enumerate all files in a directory, call `chooseFileSystemEntries()`
 with the `type` option set to `'openDirectory'`. The user selects a directory
```
mkruisselbrink, line 272: `'openDirectory'` is now `'open-directory'`

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +286,40 @@ butDir.addEventListener('click', async (e) => {
 });
 ```
 
+## What's new/changed? {: #whats-new }
+
+Chrome 82
+
+* Support for [writable streams](#save-to-disk) was added, and the previous
+  method for writing to disk (`FileSystemWriter`) was deprecated.
```
mkruisselbrink, line 294: not sure if "deprecated" is the right word here, we actually removed the previous method.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +286,40 @@ butDir.addEventListener('click', async (e) => {
...
+* File handles can now be serialized and stored in IndexedDB, or sent via `postMessage`
+  to other windows or workers within the same origin. Note that permissions
+  are not retained between browser sessions. For example, when a browser tab
+  is re-opened, and a file handle is obtained from IndexedDB, the user will
+  need to grant permission to read/write to the file again.
```
mkruisselbrink, line 299: it is probably important to mention somewhere that currently chrome won't automatically prompt for read access if you try to do a operation a handle where read permission expired (for example restored from IndexedDB in a new browser session). If `await handle.queryPermission()` returns "prompt" you need to explicitly call `await handle.requestPermission()` before you might be able to use the handle.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +286,40 @@ butDir.addEventListener('click', async (e) => {
...
+* Added support for `isSameEntry()`, which returns true if two entries
+  represent the same file or directory.
+* Updated usage indicators to indicate whether the user has granted the domain
+  permission to files, including a new read-only indicator.
+
```
mkruisselbrink, line 304: we also added `resolve()` to figure out if a file/directory is a descendant of a directory, and if so return the relative path from the directory to the entry. (envisioned use case: editor with a "project" opened (a directory) shown in a side bar. Now user opens a file in a file picker, after which the editor might want to highlight the opened file in the project side bar).

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +286,40 @@ butDir.addEventListener('click', async (e) => {
...
 (or [explainer][explainer]) has been completed.
 
-As of Chrome 78, the following functionality is not available, or
+As of Chrome 82, the following functionality is not available, or
 doesn't match the spec:
 
-* Handles are not serializable, meaning they cannot be passed via
-  `postMessage()`, or stored in IndexedDB.
-* Non-atomic writes (i.e. calls to `FileSystemFileHandle.createWriter()`
+* Non-atomic writes (i.e. calls to `FileSystemFileHandle.createWritable()`
```
mkruisselbrink, line 314: I'm not sure if we should even still mention this, or if we do we should definitely rephrase it. There is no mention of "inPlace" anymore in the spec (other than a note mentioning that there isn't currently any support for non-atomic writes).

We still don't really have a story/solution for use cases where we do need non-atomic writes though, so not sure what we should mention here.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -184,12 +185,50 @@ async function getNewFileHandle() {
...
+the user for permission. If permission isn't granted, `createWritable()`
+will throw a `DOMException`, and the app will not be able to write to the
+file. In the text editor, these `DOMException`s are handled in the
+[`saveFile()`][text-editor-app-js] method.
+
+The `write()` method takes a string, which is what we want for a text editor.
+But it can also take a [BufferSource][buffersource], or a [Blob][blob].
+Unlike most `WritableStreams`, changes are **not** written to disk until the
+stream is closed by calling `close()`.
+
```
mkruisselbrink, line 221: is it worth having some example that actually uses the fact that this is a WritableStream as well? The current example doesn't really use any stream specific bits, instead it just uses our convenience wrappers around what is provided by WritableStream. Not sure what such an example would look like (perhaps something that pipes the result of fetch() to a file?)

Other thoughts on this area is that perhaps we should call out that this is effectively a seekable writable stream? I.e. it's a normal writable stream, but you can also seek in it.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -21,20 +21,21 @@ origin_trial:
 ---
```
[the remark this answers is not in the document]
petele: Fixed.

src/site/content/en/blog/native-file-system/index.md
petele: Changed to removed.

src/site/content/en/blog/native-file-system/index.md
petele: Added an item for `resolve()`

src/site/content/en/blog/native-file-system/index.md
petele: Removed this whole section for the time being.

src/site/content/en/blog/native-file-system/index.md
petele: Added an example using fetch and piping the response, and a sentence about `seek` and `truncate`.

src/site/content/en/blog/native-file-system/index.md
petele: D'oh, thought I caught all of these, thanks!

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -184,12 +191,70 @@ async function getNewFileHandle() {
...
+  const writable = await fileHandle.createWritable();
+  // Make an HTTP request for the contents.
+  const response = await fetch(url);
+  // Stream the response into the file.
+  await response.body.pipeTo(writable);
+  // pipeTo() closes the destination pipe by default, no need to close it.
+}
+```
+
+You can also [`seek()`][spec-seek], or [`truncate`][spec-truncate] within the
```
jpmedley, line 240:
```suggestion
You can also [`seek()`][spec-seek], or [`truncate()`][spec-truncate] within the
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -184,12 +191,70 @@ async function getNewFileHandle() {
...
+stream to update the file at a specific position, or resize the file.
+
+{% Aside 'caution' %}
+Changes are **not** written to disk until the stream is closed, either by
+calling `close()` or if the stream is closed by the pipe.
```
jpmedley, line 245:
```suggestion
Changes are **not** written to disk until the stream is closed, either when
the script calls `close()` or the stream is closed by the pipe.
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -184,12 +191,70 @@ async function getNewFileHandle() {
...
+{% endAside %}
+
+{% Details %}
+{% DetailsSummary 'h5' %}
+
+Chrome 81 and earlier
+
+Support for writable streams was added in Chrome 82, and the previous
```
jpmedley, line 253:
```suggestion
Support for writable streams was added in Chrome 83, and the previous
```

jpmedley: Check this throughout.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -216,27 +281,28 @@ takes a string, which is what we want for a text editor. But it can also take a
...
-  the `close()` method is called.
-{% endAside %}
+{% endDetails %}
 
-### What else is possible?
+### Storing file handles in IndexedDB
 
-Beyond reading and writing files, the Native File System API provides
-several other new capabilities.
+Starting in Chrome 82, you can save a file handle to IndexedDB. This makes
```
jpmedley, line 288:
```suggestion
Starting in Chrome 83, you can save a file handle to IndexedDB. This makes
```

jpmedley, line 288: Either write out Indexed database or enclose it in backticks.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -216,27 +281,28 @@ takes a string, which is what we want for a text editor. But it can also take a
...
+it possible to keep a list of recently opened or edited files. For example
+you could offer to re-open the last file when the app is opened, or show a
+list of recently edited files within your app. You can also `postMessage`
+file handles between the same top-level origin, for example between tabs or
+from tabs to workers.
```
jpmedley, line 293:
```suggestion
list of recently edited files within your app. You can also call `postMessage()`
to send file handles between the same top-level origin, for example between
tabs or from tabs to workers.
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
-
-* Handles are not serializable, meaning they cannot be passed via
-  `postMessage()`, or stored in IndexedDB.
-* Non-atomic writes (i.e. calls to `FileSystemFileHandle.createWriter()`
-  with `inPlace: true`).
-* Writing to a file using a [`WritableStream`][writablestream].
-* The [`FileSystemDirectoryHandle.resolve()`][fs-dir-handle] method.
+## What's new/changed? {: #whats-new }
+
+Chrome 82
```
jpmedley, line 317:
```suggestion
Chrome 83
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+
+* Support for [writable streams](#save-to-disk) was added, and the previous
+  method for writing to disk (`FileSystemWriter`) was removed.
+* File handles can now be serialized and stored in IndexedDB, or sent via
```
jpmedley, line 321: IndexedDB again.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+  `postMessage` to other windows or workers within the same origin. Note that
```
jpmedley, line 322:
```suggestion
  `postMessage()` to other windows or workers within the same origin. Note that
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+  permissions are not retained between browser sessions. For example, when a
```
jpmedley, line 323: It seems like information about permissions should be gathered together under its own heading.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+  browser tab is re-opened, and a file handle is obtained from IndexedDB, the
+  user will need to grant permission to read/write to the file again.
```
jpmedley, line 325:
```suggestion
  user will need to grant permission to read and write to the file again.
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+* Added support for [`isSameEntry()`][spec-issameentry], which returns `true`
```
jpmedley, line 326: A best practice is for bulleted list to follow the same grammatical structure.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -246,27 +312,32 @@ butDir.addEventListener('click', async (e) => {
...
+  if two entries represent the same file or directory.
+* Updated usage indicators to indicate whether the user has granted the domain
+  permission to files, including a new read-only indicator.
+* Added support for [`resolve()`][spec-resolve], which will return the
+  relative path from one entry to another. Especially helpful for multi-file
```
jpmedley, line 331:
```suggestion
  relative path from one entry to another. This is especially helpful for multi-file
```

src/site/content/en/blog/native-file-system/index.md
petele: Added some content to cover that around line 300

src/site/content/en/blog/native-file-system/index.md
petele: I believe IndexedDB is the correct term - from MDN https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API, or is there something else?

src/site/content/en/blog/native-file-system/index.md
petele: Good catch, will fix.

src/site/content/en/blog/native-file-system/index.md
petele: Suggestions on how to fix this one?

src/site/content/en/blog/native-file-system/index.md
petele: This kind of felt like it flowed into one thought to me. If you separate them out, the permission comment doesn't fit in context.

src/site/content/en/blog/native-file-system/index.md
petele: I re-wrote this section before I saw your feedback. I think I've addressed it, but maybe?

src/site/content/en/blog/native-file-system/index.md
petele: Can't we just called it IndexedDB? It's not actually the API interface, but the API name.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API):
> IndexedDB is a low-level API for client-side storage of significant amounts of structured data, including files/blobs.

src/site/content/en/blog/native-file-system/index.md
jpmedley: I'm not sure why it's not showing against this page, but I've done a PR: https://github.com/GoogleChrome/web.dev/pull/2569/files

petele: AWESOME - that makes more sense now, got it. I fixed all the but permissions one.

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -216,27 +287,62 @@ takes a string, which is what we want for a text editor. But it can also take a
...
+  // Request permission, if the user grants permission, return true.
+  if (await fileHandle.requestPermission(opts) === 'granted') {
+    return true;
+  }
+  // The user did nt grant permission, return false.
+  return false;
+}
+```
+
+By requesting write permission with the read request, I was reduced
```
kaycebasques, line 331:
```suggestion
By requesting write permission with the read request, I reduced
```

src/site/content/en/blog/native-file-system/index.md
```diff
@@ -216,27 +287,62 @@ takes a string, which is what we want for a text editor. But it can also take a
...
+Starting in Chrome 83, file handles are serializable, which means that you
+can save a file handle to IndexedDB, or `postMessage` them between the same
+top-level origin.
 
-### What else is possible?
+Saving file handles to IndexedDB means that you can store state, or remember
+which files a user was working on. This makes it possible to keep a list of
+recently opened or edited files, offer to re-open the last file when the app
+is opened, et cetera. In the text editor, I store a list of the 5 most recent
+files the user has  opened, making it easy to access those files again.
```
kaycebasques, line 302:
```suggestion
files the user has opened, making it easy to access those files again.
```

Extra space

PR #2557, opened 2020-04-11
optimize text for speed versus legibility.

robdodson:
Random tidbit I came across while reading [A Modern CSS Reset](https://hankchizljaw.com/wrote/a-modern-css-reset/).

https://marco.org/2012/11/15/text-rendering-optimize-legibility

- Changes our CSS to use `text-rendering: optimizeSpeed` instead of `optimizeLegibility`.

### DIFF
--- a/src/styles/tools/_mixins.scss
+++ b/src/styles/tools/_mixins.scss
@@ -199,7 +199,7 @@
   -webkit-font-smoothing: antialiased; // @alternate
   font: normal normal normal 24px / 1 'Material Icons';
   font-feature-settings: 'liga';
-  text-rendering: optimizeLegibility;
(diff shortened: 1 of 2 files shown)


src/styles/tools/_mixins.scss
```diff
@@ -199,7 +199,7 @@
   -webkit-font-smoothing: antialiased; // @alternate
   font: normal normal normal 24px / 1 'Material Icons';
   font-feature-settings: 'liga';
-  text-rendering: optimizeLegibility;
+  text-rendering: optimizeSpeed;
```
MichaelSolati, line 202: Why aren't we just using the `$FONT_SMOOTHING_RENDERING` variable?

PR #2569, opened 2020-04-13
Make list all of same grammatical structure.

jpmedley:


petele:
AH! I get it now, thanks. I've merged this in via a manual PR.

Issue #2529, opened 2020-04-07
Redirect from /pwa/ fails


petele:
**Describe the bug**
Redirects are failing.

**To Reproduce**
1. Go to https://web.dev/pwa/ 
2. Go to https://web.dev/pwa 

**Expected behavior**
Both: Redirect to https://web.dev/progressive-web-apps/

**Actual behavior**
1. User sees a 404 page (does not work)
2. Works as expected and user is redirect to https://web.dev/progressive-web-apps/

Redirects are set-up at https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/_redirects.yaml#L168

kaycebasques:
@robdodson any idea why this is borking?

robdodson:
It looks like the service worker is messing up. Try the above links in new incognito windows.

robdodson:
same bug as #2488 I believe

PR #2580, opened 2020-04-14
make 404s always go to network

### LINKED ISSUE #2488: Service worker can prevent redirects from working
**Describe the bug**
If a post accidentally goes live and we remove it and ship a redirect the service worker will cause the user to see a 404 instead of the redirect.

These are the steps I observed today:

- We accidentally shipped /digging-into-the-privacy-sandbox/
- We moved the post to `_drafts` and add a redirect to `_redirects.yaml`
- If you visit the site in a new incognito window and type https://web.dev/digging-into-the-privacy-sandbox/ into the url bar you'll get the redirect.
- However, if you first visit http://web.dev and then type https://web.dev/digging-into-the-privacy-sandbox/ into the URL bar you'll get a 404.

samthor:
Fixes #2488 and #2529.

This (hopefully...) simplifies our SW a bit by treating 404's as unrecoverable errors that always go to the network. This occurs in the Service Worker as well as in the SPA routing (so a 404 will "break" the SPA). We just don't know if there's really a redirect backing whatever page 404'ed otherwise.

I've also added a comment to the redirects YAML to make it clear that trailing slashes there are unnecessary.

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -104,56 +104,17 @@ workboxRouting.registerRoute(
 );
 
 /**
- * This is a special handler for requests without a trailing "/". These requests _should_ go to
- * the network (so that we can match web.dev's redirects.yaml file) but fallback to the normalized
- * version of the request (e.g., "/foo" => "/foo/").
- */
-const untrailedContentPathRe = new RegExp('^(/[\\w-]+)+$');
-
-workboxRouting.registerRoute(
-  matchSameOriginRegExp(untrailedContentPathRe),
-  async ({url, event}) => {
-    const {pathname} = url;
-
-    // First, check if there's actually a partial node in the cache already. If so, there's no need
-    // to do a real network fetch, as we know the modified request will be satisfied.
-    const cachedResponse = await caches.match(
-      new Request(pathname + '/index.json'),
-    );
-    if (!cachedResponse) {
-      // If there's not, then try the network.
-      try {
-        return await fetch(event.request);
-      } catch (e) {
-        // If fetch fails, just redirect below.
-      }
-    }
-
-    // Either way, redirect to the updated Location.
-    const headers = new Headers();
-    headers.append('Location', `${pathname}/${url.search}`);
-    const redirectResponse = new Response('', {
-      status: 301,
-      headers,
-    });
-    return redirectResponse;
-  },
-);
-
-/**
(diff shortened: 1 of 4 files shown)


src/lib/sw.js
```diff
@@ -104,56 +104,19 @@ workboxRouting.registerRoute(
...
-    return redirectResponse;
-  },
-);
-
-/**
- * Match fetches for patials, for SPA requests. Matches "/foo-bar/index.json" and
- * "/foo-bar/many/parts/index.json", for partial SPA requests.
- */
-const partialPathRe = new RegExp('^/([\\w-]+/)*index\\.json$');
+const partialPathRe = new RegExp('^/([\\w-]+/)*\\w+\\.json$');
```
robdodson, line 112: what happens if the site requests a `.json` file that isn't related to partials. Will this send it to the network?

src/site/content/en/_redirects.yaml
```diff
@@ -1,3 +1,16 @@
...
+#
+#  (1) Simple redirects, e.g., "from: /foo"
+#      - These don't match anything under this URL, like "/foo/bar"
+#      - No trailing slash is required, as this matches "/foo", "/foo/" and
+#        "/foo/index.html"
+#
+#  (2) Group redirects, e.g., "from: /foo/...", which will match the top-level
+#      path (with the same rules as (1)) as well as any descendant paths
+#
+# WARNING! Redirects DO NOT take priority over static content.
```
robdodson, line 12: Can you add an example so folks can quickly grok what this means.

robdodson: Like, we're trying to say "If you try to redirect /foo, but there's a file at /foo/index.html, then we're going to still serve /foo/index.html. You have to move the file."

src/site/content/en/_redirects.yaml
```diff
@@ -1,3 +1,16 @@
+# Master redirects file for web.dev. This mirrors the old DevSite file.
+# There are two types of redirects:
+#
+#  (1) Simple redirects, e.g., "from: /foo"
+#      - These don't match anything under this URL, like "/foo/bar"
```
robdodson, line 5: Maybe rephrase to something like:

These do not match **subpaths**. i.e., "from: /foo" will match "/foo" and "/foo/" but not "/foo/bar"

PR #2581, opened 2020-04-14
add service-worker kill switch

samthor:
Adds a quick boolean that we can use to nuke the SW if necessary.


robdodson:
thanks for working on this :D

PR #2597, opened 2020-04-15
Content Indexing update: launchUrl is now url

jeffposnick:
This addresses a change to the Content Indexing API that will go into effect after the Origin Trial period ends, along with guidance about compatibility during the Origin Trial.

CC: @rayankans to confirm the details.


rayankans:
LGTM, thanks!

src/site/content/en/blog/content-indexing-api/index.md
```diff
@@ -212,6 +214,12 @@ await registration.index.add({
 });
 ```
 
+{% Aside 'note' %}
+  The `add()` method previously took a `launchUrl` parameter, which has been renamed to `url`.
+  During the Origin Trial period, you should include both `url` and `launchUrl`, set to the same
+  value. Following the Origin Trial, only `url` will be supported.
+{% endAside %}
+
```
kaycebasques, line 222:
```suggestion


```

PR #2608, opened 2020-04-16
refactor(collections): create collections for `tags` and `authors` that derivative collections reference

MichaelSolati:
I moved most of the logic for the `paginated-tags`, `paginated-authors`, `paginated-posts-by-tags`,  and `paginated-posts-by-authors` into these "parent" collections (`authors` and `tags`). Since they had somewhat redundant logic/behavior I figured it would be good to apply all of that in one place.

Two reasons why this may be useful is:
1) We can now use the `authors` and `tags` collections to generate RSS feeds for each author and tag.
2) When we move to 11ty 0.11.0 there's a chance we can use computed data in order to do the pagination for the tags/authors landing pages and the individual tags/authors pages local data rather than creating multiple collections with similar data.

So it requires adding two collections for now, but it simplifies refactoring code for the derivative collections, allows for RSS feed, and may be used to remove the paginated collections down the road.

### DIFF
--- a/src/site/_collections/authors.js
+++ b/src/site/_collections/authors.js
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const fs = require('fs');
+const path = require('path');
+const contributors = require('../_data/contributors');
+const livePosts = require('../_filters/live-posts');
+const setdefault = require('../_utils/setdefault');
+
+/**
+ * Generate map the posts by author's username/key
+ *
+ * @param {*} posts
+ * @return {Map<string, Array<any>>} Map of posts by author's username/key
+ */
+const findAuthorsPosts = (posts) => {
+  const authorsMap = new Map();
+  posts.forEach((post) => {
+    const authors = post.data.authors || [];
+    authors.forEach((author) => {
+      const postsByAuthor = setdefault(authorsMap, author, []);
+      postsByAuthor.push(post);
+      authorsMap.set(author, postsByAuthor);
+    });
+  });
+  return authorsMap;
+};
+
+/**
+ * Finds image of author, returns path.
+ *
+ * @param {string} key
+ * @return {string} Path for image.
+ */
+const findAuthorsImage = (key) => {
+  for (const size of ['@3x', '@2x', '']) {
+    if (
+      fs.existsSync(
+        path.join(
+          'src',
+          'site',
+          'content',
+          'en',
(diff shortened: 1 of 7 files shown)


src/site/_collections/authors.js
```diff
@@ -0,0 +1,107 @@
...
+const fs = require('fs');
+const path = require('path');
+const contributors = require('../_data/contributors');
+const livePosts = require('../_filters/live-posts');
+const setdefault = require('../_utils/setdefault');
+
+/**
+ * Generate map the posts by author's username/key
+ *
+ * @param {*} posts
```
samthor, line 25: This is an array of something? It can be `Array<*>`, that's fine too if it's a complex Eleventy type.

src/site/_collections/authors.js
```diff
@@ -0,0 +1,107 @@
...
+ * @return {Map<string, Array<any>>} Map of posts by author's username/key
+ */
+const findAuthorsPosts = (posts) => {
+  const authorsMap = new Map();
+  posts.forEach((post) => {
+    const authors = post.data.authors || [];
+    authors.forEach((author) => {
+      const postsByAuthor = setdefault(authorsMap, author, []);
+      postsByAuthor.push(post);
+      authorsMap.set(author, postsByAuthor);
```
samthor, line 35: I thought the point of setdefault is that you don't need to set it again :)

samthor: So this is now a oneliner:

```js
setdefault(authorsMap, author, []).push(post);
```

src/site/_collections/authors.js
```diff
@@ -0,0 +1,107 @@
...
+ * Finds image of author, returns path.
+ *
+ * @param {string} key
+ * @return {string} Path for image.
+ */
+const findAuthorsImage = (key) => {
+  for (const size of ['@3x', '@2x', '']) {
+    if (
+      fs.existsSync(
+        path.join(
```
samthor, line 51: can you break this into

```js
const jpegPath = path.join(...);
if (fs.existsSync(jpegPath)) {
  ... 
}
```

src/site/_collections/authors.js
```diff
@@ -0,0 +1,107 @@
...
+    .getFilteredByGlob('**/*.md')
+    .filter(livePosts)
+    .sort((a, b) => b.date - a.date);
+
+  const authorsPosts = findAuthorsPosts(posts);
+
+  const authors = Object.values(contributors)
+    .sort((a, b) => a.title.localeCompare(b.title))
+    .reduce((accumulator, author) => {
+      author.url = path.join('/en', author.href);
```
samthor, line 85: So I think in this and the other place you use an accumulator, you're modifying an existing collection, right? It's not really a problem but maybe worth calling it out. (We probably did this before, did I complain then?)

I think calling the method a second time gives you the same result (you're not actually modifying each individual object, just _adding_ properties).

So maybe just add a comment above `Object.values`, like:

```js
// This updates the shared contributors object with meta information and is safe to be called multiple times.
```

MichaelSolati: It's also used to to prevent needing to `map` and `filter` collections. Like, there's no need to filter through them twice.

src/site/_collections/authors.js
MichaelSolati: Ok, so I would have liked to do this, but 11ty isn't liking it 😤:

Screen Shot 2020-04-17 at 11 44 00 AM

MichaelSolati: Just realized `push` wouldn't return the array, so we're assigning a number to `postsByAuthor`

PR #2612, opened 2020-04-16
fix(newsletter): update copy

MichaelSolati:
Update copy based on feedback from Harleen.

### DIFF
--- a/src/site/_includes/partials/subscribe.njk
+++ b/src/site/_includes/partials/subscribe.njk
@@ -1,10 +1,12 @@
 <web-subscribe id="subscribe">
   <div class="w-subscribe">
-    <div class="w-display--flex w-text--center">
-      <h1 class="w-align-items--center">Developer Newsletter</h1>
+    <div class="w-display--flex w-justify-content--center">
+      <h1>Developer Newsletter</h1>
     </div>
-    <div class="w-display--flex w-subscribe--padded w-align-items--center w-text--center">
-      <p class="w-mt--non w-align-items--center w-subscribe__message">Get the latest news, techniques and updates straight to your inbox.</p>
+    <div class="w-justify-content--center">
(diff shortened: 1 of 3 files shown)


src/site/_includes/partials/subscribe.njk
```diff
@@ -1,10 +1,10 @@
 <web-subscribe id="subscribe">
   <div class="w-subscribe">
-    <div class="w-display--flex w-subscribe--padded w-text--center">
-      <h1 class="w-headline--two">Helping you build powerful web experiences</h1>
+    <div class="w-display--flex w-text--center">
+      <h1 class="w-subscribe--margin">Developer Newsletter</h1>
```
robdodson, line 4: instead of using a unique class here, what do you think about adding a class to `_display.scss` that lets you set the flexbox alignment/justified mode.

`.w-justify-content--center`
`.w-align-items--center`

then you could put them on the parent of this element.

src/site/_includes/partials/subscribe.njk
```diff
@@ -41,18 +43,22 @@
       <div class="w-subscribe__field w-text--center w-display--flex">
         <span class="w-subscribe__notice">Don't worry, no spam here! Your information will only be used for.... We share updates every 1-2 months and you can unsubscribe anytime.</span>
       </div>
-      <div class="w-subscribe__field w-display--flex w-subscribe--padded">
-        <div class="w-display--inline-flex">
-          <input id="sub-newsletter" name="WebDevNewsletter" required value="Unconfirmed" class="w-checkbox" type="checkbox" />
-          <label for="sub-newsletter" class="w-ml--l">Add me to the web.dev mailing list.</label>
+      <div class="w-justify-content--center">
```
robdodson, line 46: wouldn't this need display--flex as well?

PR #2618, opened 2020-04-17
Fixes for prefers-color-scheme article

farskid:

- use `MediaQueryList.matches` instead of `MediaQueryList.media !== 'not all'` to detect support for `prefers-color-scheme`

### DIFF
--- a/src/site/content/en/blog/prefers-color-scheme/index.md
+++ b/src/site/content/en/blog/prefers-color-scheme/index.md
@@ -4,7 +4,7 @@ subhead: Overhyped or necessity? Learn everything about dark mode and how to sup
 authors:
   - thomassteiner
 date: 2019-06-27
-updated: 2019-11-25
+updated: 2020-04-17
 hero: hero.jpg
 hero_position: bottom
 alt: |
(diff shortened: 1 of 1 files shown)


farskid:
@googlebot I signed it!

kaycebasques:
@tomayac WDYT?

tomayac:
This is not correct. IE11 does not know the media query, but still reports `false` since something it doesn't know about doesn't match. This makes it indistinguishable from a correctly reported `false` on a browser that supports it. For more details on this, please read my [blog post](https://blog.tomayac.com/2019/09/16/not-all-is-nothing-finding-out-if-the-browser-supports-a-given-media-query/). Here's a screenshot from IE11: 

Screen Shot 2020-04-21 at 09 32 44

Regarding `addListener()`, please see the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener#wikiArticle:~:text=This%20is%20basically%20an%20alias%20for,inherits%20from%20EventTarget%20in%20newer%20browsers.):

> This is basically an alias for EventTarget.addEventListener(), for backwards compatibility purposes. Older browsers should use addListener instead of addEventListener since MediaQueryList only inherits from EventTarget in newer browsers.

PR #2620, opened 2020-04-17
feat(rss): create feed for tags and authors, fixes #2155

### LINKED ISSUE #2155: Create RSS feeds for various tags

MichaelSolati:


test/site/_transforms/service-worker-partials/index.js
```diff
@@ -12,7 +12,7 @@ describe('service-worker-partials', function() {
   let $;
 
   beforeEach(function() {
-    const fixture = `<html lang="en"><head><title>Test title</title></head><body><div id="content"><div class="guide-landing-page"><div>Hello</div></div><div>Bonus sibling</div></div></body></html>`;
+    const fixture = `<html lang="en"><head><title>Test title</title><link rel="alternate" href="/feed.xml" type="application/atom+xml" data-title="web.dev feed"></head><body><div id="content"><div class="guide-landing-page"><div>Hello</div></div><div>Bonus sibling</div></div></body></html>`;
```
robdodson, line 15: would you mind adding a test that verifies that it returns a unique rss for the author and/or tags page? I think testing either is fine. Just something to assert that if the <link> has a different value we will not return /feed.xml

src/lib/sw.js
```diff
@@ -192,13 +192,15 @@ workboxRouting.registerRoute(normalMatch, async ({url}) => {
...
-    '%_CONTENT_REPLACE_%',
-    meta + `<title>${partial.title || ''}</title>` + partial.raw,
-  );
+  const title = `<title>${partial.title || 'web.dev'}</title>`;
+  const rssHref = partial.rss || '/feed.xml';
+  const rssTitle = partial.rss ? `${partial.title} on web.dev` : 'web.dev feed';
+  const rss = `<link rel="alternate" href="${rssHref}" type="application/atom+xml" data-title="${rssTitle}"></link>`;
+
+  const output = layoutTemplate
+    .replace('%_HEAD_REPLACE_%', meta + title + rss)
```
samthor, line 202: I honestly did only one replacement out of laziness and the fact that I couldn't find any browser we supported that cared _where_ header-like things existed on the page.

I thought if we ever switched to proper 'streams' it would be faster/easier/simpler to just split the page into "start", "our content", and "end".

I still don't really see a need for things to _actually_ be in the `<head>`, but there's nothing wrong with this approach.

src/site/_includes/layout.njk
```diff
@@ -13,11 +13,12 @@
       <meta name="offline" content="true" />
     {%- endif %}
 
-    {% if not nometa %}
+    {% if nometa %}
```
robdodson, line 16: Would you mind adding a comment to this block to explain what's happening. Essentially that for most posts we use Meta to give them their metadata but nometa is used by the sw partial and we want to put a token in there that we can replace at runtime.

test/integration/partials-test.js
```diff
@@ -0,0 +1,50 @@
+const assert = require('assert');
+const path = require('path');
+const dist = path.resolve(__dirname, '..', '..', 'dist');
+
+describe('Partials test', function() {
```
robdodson, line 5:
```suggestion
describe('Service worker partials', function() {
```

test/integration/partials-test.js
```diff
@@ -0,0 +1,50 @@
...
+  it('Partials include RSS urls', async function() {
```
robdodson, line 6:
```suggestion
  it('includes RSS urls', function() {
```

Suggest using this style of test message throughout.

Also, do these need to be `async` functions? I don't think they use `await`.

test/integration/partials-test.js
robdodson: I realize I used "Build test" in the other file but was thinking we probably want to rename that to just "Build".

test/integration/partials-test.js
```diff
@@ -0,0 +1,50 @@
...
+    const expected = '/feed.xml';
+    const partialPath = path.join(dist, 'en', 'index.json');
+    const partial = require(partialPath);
+
+    assert.ok(
```
robdodson, line 11: If you use `assert.deepStrictEqual(actual, expected)` then you don't need to write your own message because it will log the difference between the actual and expected.

robdodson:
not sure why percy is showing the breakage with the magnifying glass in the search field. Could you try sending in an empty commit, `git commit -m empty --allow-empty && git push` to poke it and see if it still shows up

src/lib/loader.js
```diff
@@ -104,6 +104,9 @@ function updateDom(partial) {
   // Update the page title.
   document.title = partial.title || '';
 
+  const rss = document.querySelector('link[type="application/atom+xml"]');
+  rss.href = partial.rss || rss.href;
```
robdodson, line 108: need to add a check, `if (rss)` juuuuust in case

src/lib/sw.js
```diff
@@ -192,13 +192,18 @@ workboxRouting.registerRoute(normalMatch, async ({url}) => {
...
+  const title = `<title>${partial.title || 'web.dev'}</title>`;
+  const rssHref = partial.rss || '/feed.xml';
+  const rssTitle =
+    partial.rss !== '/feed.xml'
+      ? `${partial.title} on web.dev`
+      : 'web.dev feed';
+  const rss = `<link rel="alternate" href="${rssHref}" type="application/atom+xml" data-title="${rssTitle}" />`;
+
+  const output = layoutTemplate
+    .replace('<title>web.dev</title>', `${meta}\n${title}\n${rss}`)
```
robdodson, line 205: Would it be safer to use a regex to replace the first title tag, regardless of its text content? My thinking is that if we ever changed this text to `<title>web.dev | learn new things!</title>` then this would (silently?) start failing.

@samthor should double check this my regex here, but I was thinking something like:

`\<title\>.*\<\/title\>`

robdodson: This is also a really good reason to write some service worker tests, to catch those potential failures :D

src/site/_includes/layout.njk
```diff
@@ -13,11 +13,13 @@
...
+      {# Area on page that service worker will load title, rss url and if offline #}
+      <title>web.dev</title>
```
robdodson, line 18: How about using the title from `site.js` here?
Could also add a comment to `site.js` explaining that changing the title in that file will affect the service worker.

src/site/content/en/authors/feed.njk
```diff
@@ -0,0 +1,47 @@
+---
```
robdodson, line 1: Note that we may want to do this (https://github.com/GoogleChrome/web.dev/issues/2652) in this PR. Or do a follow up PR to do it for all of the feeds and other files.

test/integration/partials-test.js
```diff
@@ -0,0 +1,41 @@
+const assert = require('assert');
+const path = require('path');
+const dist = path.resolve(__dirname, '..', '..', 'dist');
+
+describe('Service worker partials', function() {
```
robdodson, line 5: Should these test check all of the expected values for the partials, as opposed to just checking the RSS field?

In other words, should we use this file as the general integration test for the entire service work partial.

src/lib/sw.js
```diff
@@ -192,13 +192,18 @@ workboxRouting.registerRoute(normalMatch, async ({url}) => {
...
+  const title = `<title>${partial.title || 'web.dev'}</title>`;
+  const rssHref = partial.rss || '/feed.xml';
+  const rssTitle =
+    partial.rss !== '/feed.xml'
+      ? `${partial.title} on web.dev`
+      : 'web.dev feed';
+  const rss = `<link rel="alternate" href="${rssHref}" type="application/atom+xml" data-title="${rssTitle}" />`;
+
+  const output = layoutTemplate
+    .replace(/<title\>.*\<\/title\>/, `${meta}\n${title}\n${rss}`)
```
MichaelSolati, line 205: @samthor does this look good for replacing the content in the head? It seems like we need to use a proper tag in the head rather than the `%_HEAD_REPLACE_%` in order to add the tags to the `sw-partial-layout.partial`.

samthor: Something something don't parse HTML with regexp.

This is kind of going down the rabbit hole here. What was the problem with just inserting `<title>` and friends inline in the document? Every browser we ship to doesn't seem to care if we include these at the "wrong" place.

MichaelSolati: @robdodson should we just put it in the wrong place?

PR #2625, opened 2020-04-17
Update the HTTP Cache guide

### LINKED ISSUE #2522: content: migrate the webfu http caching guide to web.dev
Background: I want to migrate [the webfu guide](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching) into [the web.dev guide](https://web.dev/http-cache). But to do that we need to make sure that there isn't any important information that's getting lost in the migration. The only way for me to be confident about that is to do a paragraph-by-paragraph analysis of the old content.

@jeffposnick please **edit this comment** and explain inline whether you agree or disagree about migrating each idea. Like this:

Example

I can do the PR myself (unless you want to do it) once we get broad agreement on what ideas should get migrated.

<h3 id="item1">item 1</h3>

> Fetching something over the network is both slow and expensive. Large responses require many roundtrips between the client and server, which delays when they are available and when the browser can process them, and also incurs data costs for the visitor. As a result, the ability to cache and reuse previously fetched resources is a critical aspect of optimizing for performance.

I think the webfu guide does a slightly better job at "selling" the feature. Based on pageviews HTTP caching seems to be a hugely popular technique. I think we should update the web.dev intro to "sell" the feature a bit more.

Agreed. — @jeffposnick

<h3 id="item12">item 2</h3>

> The good news is that every browser ships with an implementation of an HTTP cache. All you need to do is ensure that each server response provides the correct HTTP header directives to instruct the browser on when and for how long the browser can cache the response.

We should add a cross-browser compatibility section to the web.dev guide. We know that cross-browser compatibility is a huge pain point for web developers, so when something has 100% support that's a huge plus.

Agreed. — @jeffposnick 

### skip

> Note: If you are using a WebView to fetch and display web content in your application, you might need to provide additional configuration flags to ensure that the HTTP cache is enabled, its size is set to a reasonable number to match your use case, and the cache is persisted. Check the platform documentation and confirm your settings.

It might be worth adding a one-sentence heads-up that WebViews are weird and you'll have to do more work.

I'm not sure about that. There are probably other content areas on web.dev where we describe features behave differently in WebViews, and we don't explicitly call them out. It might give the wrong impression if we call out WebViews once in this article. — @jeffposnick

### skip

> When the server returns a response, it also emits a collection of HTTP headers, describing its content-type, length, caching directives, validation token, and more. For example, in the above exchange, the server returns a 1024-byte response, instructs the client to cache it for up to 120 seconds, and provides a validation token ("x234dff") that can be used after the response has expired to check if the resource has been modified.

We can skip this. I'll just say `Skip.` from here on out. Whenever I say `Skip.` it means that the information isn't crucial or the web.dev guide already covers the idea.

Agreed. — @jeffposnick

<h3 id="item3">item 3</h3>

> Assume that 120 seconds have passed since the initial fetch and the browser has initiated a new request for the same resource. First, the browser checks the local cache and finds the previous response. Unfortunately, the browser can't use the previous response because the response has now expired. At this point, the browser could dispatch a new request and fetch the new full response. However, that’s inefficient because if the resource hasn't changed, then there's no reason to download the same information that's already in cache! That’s the problem that validation tokens, as specified in the ETag header, are designed to solve. The server generates and returns an arbitrary token, which is typically a hash or some other fingerprint of the contents of the file. The client doesn't need to know how the fingerprint is generated; it only needs to send it to the server on the next request. If the fingerprint is still the same, then the resource hasn't changed and you can skip the download.

I like this explanation of ETag. I think we should expand the web.dev guide's explanation of ETag a bit.

Agreed. — @jeffposnick

### skip

> In the preceding example, the client automatically provides the ETag token in the "If-None-Match" HTTP request header. The server checks the token against the current resource. If the token hasn't changed, the server returns a "304 Not Modified" response, which tells the browser that the response it has in cache hasn't changed and can be renewed for another 120 seconds. Note that you don't have to download the response again, which saves time and bandwidth.

Skip.

Agreed. — @jeffposnick

### skip (previously item 4)

> As a web developer, how do you take advantage of efficient revalidation? The browser does all the work on our behalf. The browser automatically detects if a validation token has been previously specified, it appends the validation token to an outgoing request, and it updates the cache timestamps as necessary based on the received response from the server. The only thing left to do is to ensure that the server is providing the necessary ETag tokens. Check your server documentation for the necessary configuration flags.

I like that this explains what the browser does automatically and what you need to make sure you do. I think we should add that idea to the guide.

I think that's already covered to some extent via https://web.dev/http-cache/#response-headers:-configure-your-web-server, so this would entail adding to that section as needed, I guess — @jeffposnick

Update: I think Jeff was right in that the web.dev guide pretty much covers this already, so I'm going to change this to `skip`.

### skip

> Note: Tip: The HTML5 Boilerplate project contains sample configuration files for all the most popular servers with detailed comments for each configuration flag and setting. Find your favorite server in the list, look for the appropriate settings, and copy/confirm that your server is configured with the recommended settings.

Skip.

Agreed. — @jeffposnick

<h3 id="item5">item 5</h3>

> From a performance optimization perspective, the best request is a request that doesn't need to communicate with the server: a local copy of the response allows you to eliminate all network latency and avoid data charges for the data transfer. To achieve this, the HTTP specification allows the server to return Cache-Control directives that control how, and for how long, the browser and other intermediate caches can cache the individual response.

Adding a quick explanation of `Cache-Control` to the web.dev guide would be good. 

Agreed. — @jeffposnick

### skip

> Note: The Cache-Control header was defined as part of the HTTP/1.1 specification and supersedes previous headers (for example, Expires) used to define response caching policies. All modern browsers support Cache-Control, so that's all you need.

Skip.

Agreed. — @jeffposnick

### skip

> "no-cache" indicates that the returned response can't be used to satisfy a subsequent request to the same URL without first checking with the server if the response has changed. As a result, if a proper validation token (ETag) is present, no-cache incurs a roundtrip to validate the cached response, but can eliminate the download if the resource has not changed.

Skip.

Agreed. — @jeffposnick

<h3 id="item6">item 6</h3>

> By contrast, "no-store" is much simpler. It simply disallows the browser and all intermediate caches from storing any version of the returned response—for example, one containing private personal or banking data. Every time the user requests this asset, a request is sent to the server and a full response is downloaded.

This seems like an important thing to at least mention in a sentence so that web developers are aware of it. I think it's important because of the private personal data use case that Ilya mentions.

Agreed. — @jeffposnick

### skip

> If the response is marked as "public", then it can be cached, even if it has HTTP authentication associated with it, and even when the response status code isn't normally cacheable. Most of the time, "public" isn't necessary, because explicit caching information (like "max-age") indicates that the response is cacheable anyway.

Skip.

Agreed. — @jeffposnick

<h3 id="item7">item 7</h3>

> By contrast, the browser can cache "private" responses. However, these responses are typically intended for a single user, so an intermediate cache is not allowed to cache them. For example, a user's browser can cache an HTML page with private user information, but a CDN can't cache the page.

Seems like we should add a sentence on this to make web developers aware of it.

Agreed. — @jeffposnick

### skip

> This directive specifies the maximum time in seconds that the fetched response is allowed to be reused from the time of the request. For example, "max-age=60" indicates that the response can be cached and reused for the next 60 seconds.

Skip.

Agreed, in that I think it's already covered in the web.dev guide. — @jeffposnick

<h3 id="item8">item 8</h3>

> Flowchar

This seems useful. Should we add?

Agreed on adding. — @jeffposnick

### skip

> Follow the decision tree above to determine the optimal caching policy for a particular resource, or a set of resources, that your application uses. Ideally, you should aim to cache as many responses as possible on the client for the longest possible period, and provide validation tokens for each response to enable efficient revalidation.

Skip.

Agreed. — @jeffposnick

<h3 id="item9">item 9</h3>

> Table

Skip.

It wouldn't hurt to include this as a summary, maybe at the end? — @jeffposnick

### skip

> According to HTTP Archive, among the top 300,000 sites (by Alexa rank), the browser can cache nearly half of all the downloaded responses, which is a huge savings for repeat pageviews and visits. Of course, that doesn’t mean that your particular application can cache 50% of the resources. Some sites can cache more than 90% of their resources, while other sites might have a lot of private or time-sensitive data that can’t be cached at all.

Skip.

Agreed. — @jeffposnick

<h3 id="item10">item 10</h3>

> All HTTP requests that the browser makes are first routed to the browser cache to check whether there is a valid cached response that can be used to fulfill the request. If there's a match, the response is read from the cache, which eliminates both the network latency and the data costs that the transfer incurs.

This is a pretty concise explanation of the basic idea that we might consider adding to the intro.

Agreed. — @jeffposnick

<h3 id="item11">item 11</h3>

> However, what if you want to update or invalidate a cached response? For example, suppose you've told your visitors to cache a CSS stylesheet for up to 24 hours (max-age=86400), but your designer has just committed an update that you'd like to make available to all users. How do you notify all the visitors who have what is now a "stale" cached copy of your CSS to update their caches? You can't, at least not without changing the URL of the resource. After the browser caches the response, the cached version is used until it's no longer fresh, as determined by max-age or expires, or until it is evicted from cache for some other reason— for example, the user clearing their browser cache. As a result, different users might end up using different versions of the file when the page is constructed: users who just fetched the resource use the new version, while users who cached an earlier (but still valid) copy use an older version of its response. How do you get the best of both worlds: client-side caching and quick updates? You change the URL of the resource and force the user to download the new response whenever its content changes. Typically, you do this by embedding a fingerprint of the file, or a version number, in its filename—for example, style.x234dff.css.

I think we can trim this down a lot, but I like that the webfu guide explains *why* versioned URLs are a good practice and therefore might motivate web developers to implement versioning. I think we should add the motivation behind versioning to the web.dev guide.

Agreed. — @jeffposnick

### skip

> The ability to define per-resource caching policies allows you to define "cache hierarchies" that allow you to control not only how long each is cached for, but also how quickly visitors see new versions. To illustrate this, analyze the above example...

Skip.

Agreed. — @jeffposnick

<h3 id="item12">item 12</h3>

> Checklist

I like the idea of summarizing the actions that the developer needs to take as a checklist. But obviously the web.dev guide's checklist would be different.

Agreed. — @jeffposnick

kaycebasques:
Fixes #2522

Hey @igrigorik & @jeffposnick, here's the updated HTTP Cache guide that includes Ilya's content from the old developers.google.com/web page. I can go through the PR and leave comments showing how I addressed every item in #2522 but I'm pretty sure they're all there. So it might be faster for you two to just review this PR first and if it all looks good then we can just leave it at that.

Staged page: https://deploy-preview-2625--web-dev-staging.netlify.app/http-cache/


src/site/content/en/reliable/http-cache/index.md
```diff
@@ -1,52 +1,77 @@
 ---
 layout: post
-title: 'The HTTP cache: your first line of defense'
+title: 'The HTTP Cache: Avoiding unnecessary network requests with `Cache-Control` and `ETag`'
 authors:
   - jeffposnick
+  - ilyagrigorik
```
jeffposnick, line 6: Feel free to add your name here, too, Kayce!

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -1,52 +1,77 @@
...
+This guide shows you the basics of an effective HTTP caching implementation.
+
+## Browser compatibility {: #browser-compatibility }
+
+There isn't actually a single API called the HTTP Cache. It's the general name
+for a collection of web platform APIs. Those APIs are supported in all browsers:
+
+* [`Cache-Control`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Cache-Control#Browser_compatibility)
+* [`ETag`](https://developer.mozilla.org/docs/Web/HTTP/Headers/ETag#Browser_compatibility)
+* [`Last-Modified`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Last-Modified#Browser_compatibility)
```
jeffposnick, line 41: It's a little weird that `Cache-Control` and `ETag` are specifically called out in the article's title, but there are actually three things listed here.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -55,44 +80,76 @@ with specific
...
+* [`Cache-Control`][cache-control].
+  The server can return a `Cache-Control` directive to specify how, and for how
+  long, the browser and other intermediate caches should cache the individual
+  response.
+* [`ETag`][etag]. When
+  the browser finds an expired cached response, it can send a small token 
+  (usually a hash of the file's contents) to the server to check if the file has
+  changed. If the server returns the same token, then the file is the same, and there's
+  no need to re-download it.
+* [`Last-Modified`][last-modified].
```
jeffposnick, line 98: I think this language comes from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified, but I'm not in favor of it. I'd prefer to think of it as there being two ways of the server identifying the version of an HTTP response: content-based (`ETag`) or time-based (`Last-Modified`).

If we're talking about, e.g., a response for an API feed that updates once every 15 minutes, then using `Last-Modified` and including the time that the API feed was last generated should be just as "accurate" as using an `ETag`. It's also conceptually more straightforward.

So, different options that each could perform a similar role.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -55,44 +80,76 @@ with specific
...
+    cached responses. 
+  {% endDetailsSummary %}
+  Suppose your server instructs browsers to cache a CSS file
+  for 1 year (<code>Cache-Control: max-age=31536000</code>) but your designer just made an
+  emergency update that you need to roll out immediately. How do you notify browsers
+  to update the "stale" cached copy of the file?
+  You can't, at least not without changing the URL of the resource. After the
+  browser caches the response, the cached version is used until it's no longer
+  fresh, as determined by <code>max-age</code> or <code>expires</code>, or until it is evicted from the cache
+  for some other reason— for example, the user clearing their browser cache. As a
```
jeffposnick, line 143: That looks like an extra space after the `—` character.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -118,65 +175,178 @@ ignored](https://www.keycdn.com/blog/cache-control-immutable#browser-support) in
...
+
+* `no-cache`. This instructs the browser that it must revalidate with the
+  server every time before using a cached version of the URL.
+* `no-store`. This instructs the browser and other intermediate caches (like CDNs) to never
+  store any version of the file. `no-store` is often used for files that have sensitive or
+  personal data.
+* `private`. Browsers can cache the file but intermediate caches cannot.
+  For example, `private` would make sense for a server-side rendered HTML file
+  that contains private user information.
+* `public`. The response can be stored by any cache, even if it's usually not cacheable.
```
jeffposnick, line 205: I'm not clear on what "even if it's usually not cacheable" means.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -118,65 +175,178 @@ ignored](https://www.keycdn.com/blog/cache-control-immutable#browser-support) in
...
+are cached:
+
+* `no-cache`. This instructs the browser that it must revalidate with the
+  server every time before using a cached version of the URL.
+* `no-store`. This instructs the browser and other intermediate caches (like CDNs) to never
+  store any version of the file. `no-store` is often used for files that have sensitive or
+  personal data.
+* `private`. Browsers can cache the file but intermediate caches cannot.
+  For example, `private` would make sense for a server-side rendered HTML file
+  that contains private user information.
```
jeffposnick, line 204: The distinction between "sensitive or personal data" (don't cache it anywhere, as per the previous section) and "private user information" (okay to cache in the browser, as per this section) is probably going to confuse folks.

src/site/content/en/reliable/http-cache/index.md
kaycebasques:
```suggestion
  for some other reason—for example, the user clearing their browser cache. As a
```

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -1,52 +1,77 @@
 ---
 layout: post
-title: 'The HTTP cache: your first line of defense'
+title: 'The HTTP Cache: Avoiding unnecessary network requests with `Cache-Control` and `ETag`'
```
kaycebasques, line 3:
```suggestion
title: 'Prevent unnecessary network requests with the HTTP Cache'
```

@jeffposnick WDYT? I wanted to include `Cache-Control`, `ETag`, and `Last-Modified` in the title to improve the page's SEO, but I think it's just too long.

src/site/content/en/reliable/http-cache/index.md
kaycebasques: If you're OK with the new title then this should be resolved now.

src/site/content/en/reliable/http-cache/index.md
kaycebasques: Yes, that is where I got that language. Your explanation makes sense. I'll re-word.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -55,44 +80,76 @@ with specific
...
+  The server can return a `Cache-Control` directive to specify how, and for how
+  long, the browser and other intermediate caches should cache the individual
+  response.
+* [`ETag`][etag]. When
+  the browser finds an expired cached response, it can send a small token 
+  (usually a hash of the file's contents) to the server to check if the file has
+  changed. If the server returns the same token, then the file is the same, and there's
+  no need to re-download it.
+* [`Last-Modified`][last-modified].
+  This header serves the same purpose as `ETag`, but is less accurate.
```
kaycebasques, line 99:
```suggestion
  This header serves the same purpose as `ETag`, but uses a time-based strategy
  to determine if a resource has changed, as opposed to the content-based strategy
  of `ETag`.
```
@jeffposnick WDYT?

src/site/content/en/reliable/http-cache/index.md
kaycebasques: Thanks, I thought about it, but it bugs me when there's a long list of authors on a page.

src/site/content/en/reliable/http-cache/index.md
kaycebasques:
```suggestion
* `public`. The response can be stored by any cache.
```

Good point. I'm also not sure what it means. I think we can just leave it out.

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -118,65 +175,178 @@ ignored](https://www.keycdn.com/blog/cache-control-immutable#browser-support) in
...
+as possible.
+
+The following `Cache-Control` values can help you fine-tune where and how unversioned URLs
+are cached:
+
+* `no-cache`. This instructs the browser that it must revalidate with the
+  server every time before using a cached version of the URL.
+* `no-store`. This instructs the browser and other intermediate caches (like CDNs) to never
+  store any version of the file. `no-store` is often used for files that have sensitive or
+  personal data.
```
kaycebasques, line 201:
```suggestion
  store any version of the file.
```

src/site/content/en/reliable/http-cache/index.md
kaycebasques, line 204:
```suggestion
```

kaycebasques: Maybe let's just remove them both? I think it's good to make readers aware of their existence but maybe we can just leave it at that.

src/site/content/en/reliable/http-cache/index.md
kaycebasques: I also wanted to lead off with `HTTP Cache` because it's such a household name but `HTTP Cache: Avoid unnecessary network requests` seems weird to me.

jeffposnick: 👍

kaycebasques:
@igrigorik PTAL

src/site/content/en/reliable/http-cache/index.md
```diff
@@ -118,65 +177,175 @@ ignored](https://www.keycdn.com/blog/cache-control-immutable#browser-support) in
...
 
-For most developers, though, either `Cache-Control: no-cache` or `Cache-Control:
-max-age=31536000` should be fine.
+[crp]: https://developers.google.com/web/fundamentals/performance/critical-rendering-path
+[304]: https://developer.mozilla.org/docs/Web/HTTP/Status/304
+[if-modified-since]: https://developer.mozilla.org/docs/Web/HTTP/Headers/If-Modified-Since
+[if-none-match]: https://developer.mozilla.org/docs/Web/HTTP/Headers/If-None-Match
+[etag]: https://developer.mozilla.org/docs/Web/HTTP/Headers/ETag
+[last-modified]: https://developer.mozilla.org/docs/Web/HTTP/Headers/Last-Modified
+[cache-control]: https://developer.mozilla.org/docs/Web/HTTP/Headers/Cache-Control
```
kaycebasques, line 351:
```suggestion
[cache-control]: https://developer.mozilla.org/docs/Web/HTTP/Headers/Cache-Control

```

kaycebasques:
Ilya's probably busy with more pressing stuff and I'm pretty confident that this migration is good so I'm just going to merge. We can always iterate with subsequent PRs if needed.

PR #2629, opened 2020-04-20
Rename 'post' tag to 'blog'

### LINKED ISSUE #2645: Remove `pathItem` tag. Normalize usage of `post` and `blog`.
The current tag system we use to define learning path posts versus blog posts is a bit confusing.
Everything tagged `post` appears in the blog, and things tagged `pathItem` can appear in a learning path.

It may be cleaner to say that every piece of content is a `post`. Then we introduce a new tag, `blog`, to indicate if a post should appear in the blog feed.

This would also be nice as we currently use eleventy's directory data files to inherit the `pathItem` tag, but we'd like to move to a model where all posts live in a single directory. It seems like the only thing that relies on `pathItem` is one collection dedicated to gathering all the posts that /measure references. But we could just as easily have that collection look at all of the `post`'s.

devnook:
Fixes #2645

This is intended to normalise the usage of phrase "post" in our setup, and to enable us to tag all articles with layout "post" with this tag.


robdodson:
This is a big enough change that i think I may need to review it on Thursday when I can give it a few hours. I want to check every page and collection to make sure nothing is broken.

src/site/content/en/blog/ar-hit-test/index.md
```diff
@@ -4,14 +4,14 @@ subhead: The Hit Test API lets you position virtual items in a real-world view.
 authors:
   - joemedley
 date: 2020-02-13
-updated: 2020-03-19
+updated: 2020-04-20
```
robdodson, line 7: I'd recommend not changing the `upated` field for any of theses posts because it will show up on the post itself and make it seem like content was updated.

robdodson: I think these were revved by our precommit hook. You can bypass it with `git commit --no-verify`. I'm updating the instructions in https://github.com/GoogleChrome/web.dev/blob/master/.huskyrc to mention this.

devnook:
PTAL

src/site/content/en/blog/vr-comes-to-the-web/index.md
```diff
@@ -12,10 +12,10 @@ description:
   this launch is the foundation for both augmented and virtual reality. This
   article is the first in a series, exploring basic concepts and describing how
   to enter an XR session. Other browsers will soon be supporting the WebXR
-  Device API, including Firefox Reality, Oculus Browser, Edge and Magic Leap's
+  Device API, including Firefox Reality, Oculus Browser, Edge and Magic Leap'sgit diff
```
samthor, line 15: `git diff` snuck in here

src/site/_collections/posts-with-lighthouse.js
```diff
@@ -21,7 +21,7 @@ const livePosts = require('../_filters/live-posts');
 
 module.exports = (collection) => {
   return collection
-    .getFilteredByTag('pathItem')
+    .getFilteredByTag('post')
```
samthor, line 24: Is this a restriction in the number of posts that will appear in this collection?

IIRC from our conversation, `pathItem` is everything, and `post` is just the blog posts.

devnook: The idea is to change our tags that "everyting" is `post` and blog posts are `blog`. So `s/pathItem/post `and `s/post/blog`.

devnook:
PTAL

.eleventy.js
```diff
@@ -150,9 +150,9 @@ module.exports = function(config) {
   // COLLECTIONS
   // ----------------------------------------------------------------------------
   config.addCollection('authors', authors);
-  config.addCollection('posts', postDescending);
+  config.addCollection('blogposts', blogpostDescending);
```
robdodson, line 153:
```suggestion
  config.addCollection('blogPosts', blogPostsDescending);
```

.eleventy.js
```diff
@@ -150,9 +150,9 @@ module.exports = function(config) {
...
   config.addCollection('postsWithLighthouse', postsWithLighthouse);
-  config.addCollection('recentPosts', recentPosts);
+  config.addCollection('recentBlogposts', recentBlogposts);
```
robdodson, line 155:
```suggestion
  config.addCollection('recentBlogPosts', recentBlogPosts);
```

src/site/_collections/posts-with-lighthouse.js
robdodson: If we get rid of `pathItem` and blog posts get a `blog` tag added to them, then do we need a `post` tag at all? Could we instead do `collection.getFilteredByGlob("**/*.md")` to get all of the posts?

Also, a tiny optimization might be to move the filter that happens on line 26 so that it happens before the filter that happens on line 25. Then we'd be sending a much smaller subset of posts through `livePosts`.

This actually brings up a bigger point which is that I think we do `filter(livePosts)` in a number of places, and it might make sense to just do that once, and then have other collections use that data. But we can worry about that in a separate PR.

src/site/_includes/components/Meta.js
```diff
@@ -86,7 +86,7 @@ module.exports = (locale, page, collections, renderData = {}) => {
     // These tags are only used internally to determine which layout a page
     // should use.
     let tags = (type === 'article' ? pageData.tags : null) || [];
-    tags = tags.filter((tag) => tag !== 'pathItem' && tag !== 'post');
+    tags = tags.filter((tag) => tag !== 'post');
```
robdodson, line 89: I think you want to filter out `blog`

src/site/_includes/post.njk
```diff
@@ -1,6 +1,7 @@
 ---
 layout: layout
 permalink: "/{{lang}}/{{page.fileSlug}}/index.html"
+tags: [ "post" ]
```
robdodson, line 4: I could be wrong but I think we usually use the yaml style of:

```
tags:
  - post
```

src/site/_collections/posts-with-lighthouse.js
devnook: We need "post" tag to distinguish it from the other pages, e.g. landing pages.

To me livePosts seems less expensive than filter in line 26, so it is better to perform it first (less items have to go through filter in 26). If we feel a need to optimize it (do we? this call happens only at build time) we should combine both filters to do only one  pass through posts, e.g.:

```
  return collection
    .getFilteredByTag('post')
    .filter((post) => {
      if (!livePosts(post)) {
         return false;
      }
      const audits = post.data.web_lighthouse;
      if (typeof audits === 'string' && audits !== 'N/A') {
        return true;
      }
      return audits instanceof Array && audits.length;
    });
};
``` 
IMO a bit less readable though. Let me know wdyt.

devnook:
PTAL

src/site/_collections/posts-with-lighthouse.js
robdodson: I like your suggestion. I do like to optimize these loops because if you're writing an article on the site, it tries to rebuild every time you hit save. It used to do so pretty quickly but we've slowly been increasing that time (both by adding more docs and adding more collections that loop over those docs). Although maybe the collection looping isn't the real bottleneck (it may just take a long time to write the files to disk)

.eleventy.js
```diff
@@ -150,9 +150,9 @@ module.exports = function(config) {
   // COLLECTIONS
   // ----------------------------------------------------------------------------
   config.addCollection('authors', authors);
-  config.addCollection('posts', postDescending);
+  config.addCollection('blogposts', blogPostsDescending);
```
robdodson, line 153:
```suggestion
  config.addCollection('blogPosts', blogPostsDescending);
```

src/site/_includes/partials/post.njk
```diff
@@ -98,7 +98,7 @@
     {% ArticleNavigation {
       back: '/blog',
       backLabel: 'Return to all articles',
-      collection: collections.posts,
+      collection: collections.blogposts,
```
robdodson, line 101:
```suggestion
      collection: collections.blogPosts,
```

robdodson:
Need to also add the blog tag to `src/site/_drafts/_template-case-study` and `src/site/_drafts/_template-case-post`

robdodson:
@kaycebasques heads up that we're working on this change. I'd like to catch up with you over VC before we move forward with landing it to make sure you understand what's changing and why we want to do this. Also to make sure it doesn't disrupt things for you or the writers.

devnook:
PTAL

robdodson:
Fixed the merge conflict. Just waiting on tests to pass before I merge.

PR #2635, opened 2020-04-20
fix(meta): remove double quote from text in meta attributes, fixes #2627

### LINKED ISSUE #2627: OGP doesn't pickup double quote
**Describe the bug**
When posting a link on Twitter, quote is stripped form the page title
For example, [`Understanding "same-site" and "same-origin"`](https://web.dev/same-site-same-origin/) is displayed as `Understanding`.
https://twitter.com/agektmr/status/1251404023342587906

This is probably caused by the OGP not being html-escaped.

MichaelSolati:
Fixes #2627

Replaces double quote's in attributes of meta tags using the `strip` function's `additionallyForbidden` argument.

### DIFF
--- a/src/site/_filters/strip.js
+++ b/src/site/_filters/strip.js
@@ -18,15 +18,15 @@
  * Removes any characters that should not exist in a string.
  *
  * @param {(string|number)} [text] Text to clean
- * @param {(Array<string|RegExp>)} [additionallyForbidden] Additional rules to be removed.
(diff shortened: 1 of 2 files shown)


src/site/_filters/strip.js
```diff
@@ -18,15 +18,15 @@
  * Removes any characters that should not exist in a string.
```
samthor, line 18: can you update this comment say "for presentation", rather than for any other reason?

PR #2637, opened 2020-04-21
Add code coverage.

### LINKED ISSUE #2678: Add code coverage to our unit and integration tests.
It's great that we're doing testing but we need coverage to help identify areas that we've overlooked or under tested.

robdodson:
Fixes #2678 

Work in progress, adds code coverage.

- Adds a new `c8` command which will generate code coverage reports for us. ([article](https://medium.com/the-node-js-collection/rethinking-javascript-test-coverage-5726fb272949))
- Adds istanbul to the rollup test build. c8 is built by the creator of istanbul so the reports they generate are basically the same. We have to do the extra step of teaching rollup and karma about istanbul if we want to get code coverage for our front end stuff.


karma.conf.js
```diff
@@ -26,8 +26,7 @@ module.exports = function(config) {
     browsers: ['ChromeHeadless'],
     files: ['dist/test/index.js'],
     frameworks: ['mocha'], // use the mocha test frameworkl
-    plugins: ['karma-chrome-launcher', 'karma-mocha', 'karma-sourcemap-loader'],
-    reporters: ['dots'], // report results in these formats.
+    reporters: ['dots', 'coverage'], // report results in these formats.
```
robdodson, line 29: note to reviewers — you don't need to use the `plugins` field in karma.conf because karma knows to look for anything that's `karma-*` in node_modules.

samthor:
This is a surprisingly small PR. I get that c8 is doing some magic. LGTM.

PR #2639, opened 2020-04-21
Fix closing figure tags

samdutton:
Closing tags for the two figure elements are missing a slash.

Currently:

image

Fixed: 

image


robdodson:
good catch, thank you!

PR #2640, opened 2020-04-21
Add tests for postsWithLighthouse collection

devnook:


robdodson:
Instead of running eleventy I think it might be better to mock the `collection` object [that eleventy passes to the collection function](https://github.com/GoogleChrome/web.dev/blob/master/src/site/_collections/posts-with-lighthouse.js#L22).

test/site/_collections/posts-with-lighthouse/pages/collection.md
```diff
@@ -0,0 +1,3 @@
+{%- for post in collections.postsWithLighthouse -%}
```
robdodson, line 1: What do you think about renaming this directory to `fixtures` so folks can tell at a glance what these are

test/site/_collections/posts-with-lighthouse/dev.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  '..',
+  '.tmp',
+  path.basename(__dirname),
+  'pages',
+  'collection',
+  'index.html',
+);
+
+describe('posts-with-lighthouse', function() {
+  describe('postsWithLighthouse', function() {
```
robdodson, line 16: not sure if you need this second level of nesting, you could probably do:

```
describe('posts-with-lighthouse', function() {
  it('...')
});
```

test/site/_collections/posts-with-lighthouse/dev.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('creates postsWithLighthouse collection in dev env', async function() {
```
robdodson, line 17: I think the message may be redundant because mocha will be showing:

```
_collections
  in DEV env
    posts-with-lighthouse
      creates postsWithLighthouse collection in dev env
```

One question to ask: what aspect of this output do we care about? Did it include some files and exclude others? Maybe we would say that it output the expected number of files with the `pathItem` tag and the `web_lighthouse` field?

You could also add test fixtures that have `web_lighthouse: N/A` and maybe junk stuff like `web_lighthouse: true` or null, etc.

test/site/_collections/posts-with-lighthouse/prod.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  '..',
+  '.tmp',
+  path.basename(__dirname),
+  'pages',
+  'collection',
+  'index.html',
+);
+
+describe('posts-with-lighthouse', function() {
+  describe('postsWithLighthouse', function() {
```
robdodson, line 16: See above about nesting these `describe()` calls.

test/site/_collections/posts-with-lighthouse/prod.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('does not include drafts in the postsWithLighthouse collection in prod', async function() {
```
robdodson, line 17:
```suggestion
    it('does not include drafts', async function() {
```

test/site/_collections/index.js
```diff
@@ -0,0 +1,33 @@
...
+    });
+
+    after(function() {
+      fs.rmdirSync(path.join('.', '.tmp'), {recursive: true});
+    });
+
+    require('./posts-with-lighthouse/prod.spec');
+    require('./recent-posts/prod.spec');
+  });
+});
```
robdodson, line 33: Note there's also a `PERCY` env that changes the behavior of `livePosts`. We should test that as well.

test/site/_collections/recent-posts/dev.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  '.tmp',
+  path.basename(__dirname),
+  'pages',
+  'collection',
+  'index.html',
+);
+
+describe('recent-posts', function() {
+  describe('recentPosts', function() {
+    it('creates recentPosts collection in dev env', async function() {
```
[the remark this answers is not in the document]
robdodson: Also might be worth highlighting that it excludes posts if they don't have a hero or thumbnail. Might also state that it limits it to the correct number of posts (3)

test/site/_collections/recent-posts/prod.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  '..',
+  '.tmp',
+  path.basename(__dirname),
+  'pages',
+  'collection',
+  'index.html',
+);
+
+describe('recent-posts', function() {
+  describe('recentPosts', function() {
```
robdodson, line 16: see note above about nesting

test/site/_collections/recent-posts/prod.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('excludes drafts in the recentPosts collection in prod', async function() {
```
robdodson, line 17:
```suggestion
    it('excludes drafts', async function() {
```

test/site/_collections/eleventy.js
```diff
@@ -0,0 +1,14 @@
+const path = require('path');
+const postsWithLighthouse = require(`../../../src/site/_collections/posts-with-lighthouse`);
+const recentPosts = require(`../../../src/site/_collections/recent-posts`);
+
+module.exports = function(config) {
+  config.addCollection('postsWithLighthouse', postsWithLighthouse);
+  config.addCollection('recentPosts', recentPosts);
```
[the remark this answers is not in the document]
devnook: wow!

test/site/_collections/index.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
+const path = require('path');
+const runEleventy = require('./runEleventy');
+
+describe('_collections', function() {
+  describe('_collections in DEV env', function() {
```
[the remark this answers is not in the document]
devnook: done

test/site/_collections/posts-with-lighthouse/pages/collection.md
devnook: done

test/site/_collections/index.js
devnook: Noted. Will try to test in in livePosts test then.

test/site/_collections/runEleventy.js
```diff
@@ -0,0 +1,14 @@
+const assert = require('assert');
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
+
+module.exports = async function(env) {
+  console.log('Running Eleventy...');
+  try {
+    await exec(
+      `ELEVENTY_ENV=${env} npx @11ty/eleventy --config="./eleventy.js"`,
```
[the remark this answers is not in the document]
devnook: npx should pick up eleventy from our node_modules dir, so should be compatible with our main run. If I run `npm run eleventy` here, it is run from the / root dir i picks up the evelenty.js config located there, even if I provide the correct path later on (which is weird). I can also run directly `../../../../../node_modules/.bin/eleventy --config="./eleventy.js" ` but slightly prefer npx as it will get dependencies in case some modules are missing. Wdyt?

devnook:
PTAL

paulirish:
Just curious, what does this collection do?  I'm somewhat of an 11ty noob.

robdodson:
@paulirish it looks at all of the markdown files that have `web_lighthouse` in the YAML front matter and puts them into an array. Later on that array gets dumped into a big json that's at the bottom of the /measure/ page. When the measure tool runs it uses that json to know which guide to link to. It's a super old system from the earliest days of the site, before we had plans to just write audit guides for all of LH.

Now that we have audit guides for LH it would probably make sense to redo this but it works for the moment.

paulirish:
@robdodson super helpful thanks!

PR #2643, opened 2020-04-21
Migrate cache api intro doc

petele:
- Migrates [Using the Cache API](https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/cache-api) article from WebFu


src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+
+## Adding to a cache
+
+There are three ways to add an item to a cache - `add`, `addAll`, and `put`.
+All three methods return a `Promise`.
+
+### `cache.add`
+
+First, there is `cache.add()`. It takes one parameter, either a `Request`
+or a URL (`string`). It makes a request to the network and stores to response
```
kaycebasques, line 68:
```suggestion
or a URL (`string`). It makes a request to the network and stores the response
```

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+in the cache. If the
+fetch fails, or if the status code of the response is not in the 200 range,
+then nothing is stored and the `Promise` rejects. Note that cross-origin
+requests not in CORS mode cannot be stored because they return a `status` of
+`0`, such requests can only be stored with `put`.
```
kaycebasques, line 73:
```suggestion
`0`. Such requests can only be stored with `put`.
```

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+cache.add(new Request('/data.json'));
+
+// Retreive data.json from the server and store the response.
+cache.add('/data.json');
+```
+
+### `cache.addAll`
+
+Next, there is `cache.addAll()`. It works similarly to `add()`, but takes an
+array of `Request` objects or URLs (`strings`). This works similarly to
```
kaycebasques, line 86:
```suggestion
array of `Request` objects or URLs (`string`s). This works similarly to
```

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+For example
+
+```js
+const response = new Response('Hello world');
+const buffer = await response.arrayBuffer();
+console.log(new Uint8Array(buffer));
+// Uint8Array(11) [72, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100]
+```
+
+## Retrieving from a cache {: #retreiving }
```
kaycebasques, line 222:
```suggestion
## Retrieving from a cache {: #retrieving }
```
Typo fix

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+rejects if any single request is not cached.
+
+```js
+const urls = ['/weather/today.json', '/weather/tomorrow.json'];
+cache.addAll(urls);
+```
+
+In each of these cases, a new entry overwrites any matching existing entry.
+This uses the same matching rules described in the section on
+[retrieving](#retreiving).
```
kaycebasques, line 97:
```suggestion
[retrieving](#retrieving).
```
Typo fix

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+## Retrieving from a cache {: #retreiving }
+
+To find an item in a cache, you can use the `match` method.
+
+```js
+const response = await cache.match(request);
+console.log(request, response);
+```
+
+If `request` is a string it is first be converted to a `Request` by calling
```
kaycebasques, line 231:
```suggestion
If `request` is a string it first needs to be converted to a `Request` by calling
```
@petele I'm unsure if you meant that the browser converts this or if it's something the developer needs to do

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+`new Request(request)`. The function returns a `Promise` that resolves to
+a `Response` if a matching entry is found, or `undefined` otherwise.
+
+To determine if two `Requests` match, more than just the URL is used. Two
```
kaycebasques, line 235:
```suggestion
To determine if two `Requests` match, the browser uses more than just the URL. Two
```
I think we're talking about the browser's logic but it's not clear to me because we don't specifically mention the subject of the sentence

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+requests are considered different if they have different query strings,
+`Vary` headers and/or methods (`GET`, `POST`, `PUT`, etc.).
```
kaycebasques, line 237:
```suggestion
`Vary` headers, or HTTP methods (`GET`, `POST`, `PUT`, etc.).
```
I think just using `or` and removing `and` is more appropriate here because we're talking about boolean "any" logic (if any of the conditions are true, then the 2 requests are considered different)

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+
+The Cache API does not provide a way to search for requests or responses
+except for matching entries against a `Response` object. However, you can
+implement your own search using filtering or by creating an index.
+
+### Filtering
+
+One way to implement your own search is to iterate over all entries and
+filter down to the ones that you want. Let's say that you want to find all
+items that have URLs ending with '.png'.
```
kaycebasques, line 281:
```suggestion
items that have URLs ending with `.png`.
```

src/site/content/en/blog/cache-api-quick-guide/index.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+This way you can use any property of the `Request` and `Response` objects to
+filter the entries. Note that this is slow if you search over large sets of
+data.
+
+### Creating an index
+
+The other way to implement your own search is to maintain a separate index of
+entries that can be searched, stored in IndexedDB. Since this is the kind of
```
kaycebasques, line 313:
```suggestion
entries that can be searched and store the index in IndexedDB. Since this is the kind of
```
Suggested rephrasing because I stumbled over this sentence and it took me a couple reads to figure out what it was saying

src/site/content/en/blog/cache-api-quick-guide/index.md
petele: Good catch - the browser does it. Will fix.

PR #2646, opened 2020-04-21
Updated to match the latest W3C specification

simon-robertson:
Updated the example code for "in-app installation flow" to match the latest W3C specifcation.

It appears the `prompt()` call now returns a promise that resolves with an object containing the `userChoice` value.

https://www.w3.org/TR/appmanifest/#beforeinstallpromptevent-interface


simon-robertson:
@googlebot I signed it!

simon-robertson:
I am closing this pull request due to the fact the Web App Manifest API still appears to be undergoing radical changes. The `beforeinstallprompt` event and the `appinstalled` event have both been removed from the spec.

https://github.com/w3c/manifest/pull/836
https://w3c.github.io/manifest/#incubations

It looks like a few web.dev articles need to be updated to reflect these changes.

PR #2653, opened 2020-04-22
Workbox service worker tweaks

jeffposnick:
R: @samthor 
CC: @robdodson @MichaelSolati 

In the course of looking into something else service worker-related, I found a few tweaks that can optimize a your service worker a bit more (or are just best practices).

- Bumping your Workbox version dependencies in `package.json` so that they all match.

- Adding in a `dontCacheBustURLsMatching` config to your precache manifest generation, so that precaching requests for your versioned JavaScript URLs can be fulfilled from the HTTP cache.

- Changing your service worker's Rollup bundling to use [`@surma/rollup-plugin-off-main-thread`](https://github.com/surma/rollup-plugin-off-main-thread), which allows you to code split your service worker and then pull in everything you need at runtime via an AMD-style loader. This means that every time the, e.g., precache manifest changes (as part of the `virtual` chunk), users won't have to download the `workbox` or `idb-keyval` chunks.

### DIFF
--- a/build.js
+++ b/build.js
@@ -118,20 +118,6 @@ function disallowExternal(source, importer, isResolved) {
 }
 
 /**
- * Ensures that the passed Rollup result only contains a single chunk.
- *
- * @param {!Promise<rollup.RollupOutput>} rollupPromise
- * @return {!rollup.RollupChunk} the single output chunk
- */
-async function singleOutput(rollupPromise) {
-  const result = await rollupPromise;
-  if (result.output.length !== 1) {
-    throw new Error(`expected single output, was: ${result.output.length}`);
-  }
-  return result.output[0];
-}
-
-/**
  * Performs main site compilation via Rollup: first on site code, and second
  * to build the Service Worker.
  */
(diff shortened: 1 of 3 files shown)


build.js
```diff
@@ -179,7 +168,14 @@ async function build() {
 
   const swBundle = await rollup.rollup({
     input: 'src/lib/sw.js',
-    external: disallowExternal,
+    manualChunks: (id) => {
+      const chunkNames = ['idb-keyval', 'virtual', 'workbox'];
```
samthor, line 172: This is probably fine but I always feel awkward matching paths against this.

What about:

```js
if (!path.isAbsolute(id)) {
  return id.startsWith('virtual:');
}
const chunks = ['node_modules/idb-keyval/', 'node_modules/workbox/'];
const rel = path.relative(__dirname, id);
return chunks.includes(rel);
```

I'm happy to be disagreed with.

samthor:
This sounds good.

This is probably obvious, but I want to spell it out anyway: point 3 works around [old browsers without this](https://developers.google.com/web/updates/2019/09/fresher-sw) because fundamentally the main SW references the other files by their hash name.

build.js
jeffposnick: Yeah, feel free to change this logic based on your desired chunk granularity.

Except that [`manualChunks`](https://rollupjs.org/guide/en/#manualchunks) expects you to return either `undefined` (keep things in the top-level output) or a `string` (move the code to the chunk with that name), and in your sample code, you're returning `boolean`.

Given that, what would you like this code to look like?

samthor:
I'm happy for you to submit this and I'll take a pass at it later.

jeffposnick:
👍

PR #2654, opened 2020-04-22
feat(author): add image and social to author pages, fixes #2274

### LINKED ISSUE #2274: Author Page should have more information about the author

I wanted to find the social information of some of the authors, so  I went to the author page and there was nothing there, yet the article page had a link to their social accounts. (see https://web.dev/authors/philipwalton/ as an example)

Add links to their configured meta data (like Github, Twitter etc) to the top of the author page.

### LINKED ISSUE #2317: Title text wraps on tag pages
Screenshot_20200307-185023

MichaelSolati:
Fixes #2274, fixes #2317

Add conditional links to author's social media profiles as well as a photo of themselves


robdodson:
one quick observation — we'll need mobile styles:

image

robdodson:
Could you also choose an individual author page and add a percy test for it?

robdodson:
I think the design may be a bit crowded on desktop. I might suggest having the author's photo above the text:
image

MichaelSolati:
Yeah stacked would be easier for styling, I was just worried it'd take up too much space on top. I'll fix it in a lil bit and add a test for a page (however I'll have to filter posts as well.... ☹️

Would it be ok to test a page without any posts?

src/site/content/en/authors/individual.njk
```diff
@@ -19,9 +19,39 @@ pagination:
...
+    <header class="w-page-header w-author-page-header">
+      {% if paged.data.hero %}
+        <div class="w-author-page__image--container">
+          <img src="{{paged.data.hero}}" alt="{{paged.data.alt}}" class="w-author-page__image">
+        </div>
+      {% endif %}
+      <div class="w-author-page__description">
+        <h1 class="w-page-header__headline">{{ renderData.title }}</h1>
+        <p class="w-page-header__copy">{{ renderData.description | safe }}</p>
+        {% if paged.homepage %}
```
robdodson, line 31: I think you may want to put all of these anchors into a containing div and apply the margin to the anchors instead of the images within the anchors. You'll probably want to use flexbox on the container. Something like:

```
<div class="icon-list">
  <a class="icon-list__icon"><img/></a>
</div>
```

src/styles/components/_all.scss
```diff
@@ -4,6 +4,7 @@
 @import 'article-header';
 @import 'article-navigation';
 @import 'asides';
+@import 'author-page';
```
robdodson, line 7: should go in src/site/styles/pages instead

robdodson:
For someone with a longer name, it looks like the heading breaks the word on mobile and the author image is quite large.

image

I think it's the same bug as #2317.

I think what we want is something more like:
image

A short term fix would be to add a class to tags and authors headings and size their type down on smaller displays. Note in the second screenshot i also reduced the author image to 128 x 128, and set the line-height to 1.2 on the heading text.

src/site/content/en/authors/individual.njk
```diff
@@ -20,8 +20,36 @@ pagination:
 
   <div class="w-layout-container w-pb--non">
     <header class="w-page-header">
+      <img src="{{paged.data.hero}}" alt="{{paged.data.alt}}" class="w-author-page__image">
       <h1 class="w-page-header__headline">{{ renderData.title }}</h1>
       <p class="w-page-header__copy">{{ renderData.description | safe }}</p>
+      <div class="w-author-page__links">
+        <div>
```
robdodson, line 27: This extra div makes it so your flexbox CSS doesn't really do anything. I think it should be removed.

robdodson:
I think we should have hover and focus styles for the icons. Need to think on the hover style for a bit. I'm going to noodle on the CSS to see if I can make the anchor properly wrap the contained image. The current focus style looks a bit off:

image

robdodson:
Removing the anonymous div and relying on flexbox fixes the focus outline issue.

image

robdodson:
off the top of my head I can't come up with a good svg hover effect so we can skip it for now

MichaelSolati:
@robdodson made some changes, please take a look when you can

PR #2661, opened 2020-04-23
Example noop collection test

robdodson:
**Edit**: I'd love to hop on a call to discuss this stuff. The notes below are a bit of a brain dump but I think I ran out of cognitive function before writing them so they ramble quite a bit 🧠😴

---

This example has both unit tests and an integration test for posts-with-lighthouse.js. Because there are a lot of files I recommend cloning it and running it locally.

Here's what I'm thinking and would like to get folks' opinions on:

## Unit tests
The [Eleventy MockCollection](https://github.com/GoogleChrome/web.dev/pull/2661/files#diff-fd796e5c9c88e279862d6aaf6b957ea8R1) returns whatever data you give it, regardless of the methods that are called. This is done to remove Eleventy from the test scenario. An analogy might be if you had a module that used `fetch()`. You'd probably write a mock that immediately returns a JSON response. It's true that you might pass the wrong URL to fetch, but in that scenario you would probably write code in your module to verify the url is a valid URL ahead of time—you wouldn't rely on fetch() to tell you that. And if you wanted added security you could write an integration test which _does_ call fetch() and ensures that all of the pieces work when they're joined together.

So I think the question is: what are we trying to test in these unit tests?

Using posts-with-lighthouse.js as an example, I think what we care about is verifying [the logic we wrote on lines 26-32](https://github.com/GoogleChrome/web.dev/blob/2b55278cc97f60dbb26e3bbb0a249f7b19c08023/src/site/_collections/posts-with-lighthouse.js#L26-L32). It's probably safe to say that if we tell Eleventy to filter by `pathItem` that it will only return collection items with the `pathItem` tag. Similarly [if we pass something to `livePosts`](https://github.com/GoogleChrome/web.dev/blob/2b55278cc97f60dbb26e3bbb0a249f7b19c08023/src/site/_collections/posts-with-lighthouse.js#L25) it will exclude items with `draft: true` in prod ([and we have tests to verify this](https://github.com/GoogleChrome/web.dev/pull/2463/files#diff-f602edd9695998b58761da030891834fR1)). Though, we could mock `livePosts` to be a noop as well if we wanted extra security. That leaves the custom logic on lines 26-32 as the smallest unit of code worth testing.

I think the key benefit of this approach is that [we can very easily pass different collection items](https://github.com/GoogleChrome/web.dev/pull/2661/files#diff-77ba308581c674c150500b141edca7b9R40-R48) to our custom logic using plain javascript objects. This might be easier than having to create markdown files that we feed to Eleventy's collection classes and hopefully makes the tests easier to read and understand.

It's true that one could cheat the tests by writing something like this:

```js
const posts = [
  {
    tags: ['foo'], // <------ this isn't right!
    data: {
      web_lighthouse: 'aria-hidden',
    },
    date: new Date(),
  },
];
```

but that would imply that calling `collection.getFilteredByTag('pathItem')` doesn't work as it says it does.

I think this is [white box testing](http://softwaretestingfundamentals.com/white-box-testing/) — we know the internals of the module we're testing, and we know the exact input it's receiving, so we can assert against all of the module's conditions and states.

## Integration tests
The counterpoint to the unit tests are [the integration tests](https://github.com/GoogleChrome/web.dev/pull/2661/files#diff-7a39f858a0933be2d9742ba572edc3e3R14). These run eleventy using markdown files and assert that the collection outputs the correct files. I think this is an example of [black box testing](http://softwaretestingfundamentals.com/black-box-testing/). We're setting up the initial conditions, running the black box, and asserting against the output on the other side. These tests give us confidence that when we call collection.getFilteredByTag('pathItem'), followed by livePosts and our custom logic, that the results are what we intend. These tests are less granular but ensure that when we join all of the pieces that they work as expected.

---

I think that posts-with-lighthouse.js isn't a very complicated file so these tests aren't super high stakes, but it may be good for us to talk through our testing methodology now while it's still early so we can collectively be on the same page moving forward 🙌


test/integration/site/_collections/index.js
```diff
@@ -0,0 +1,22 @@
...
+const runEleventy = require('./runEleventy');
+
+describe.only('_collections', function() {
+  describe('in DEV env', function() {
+    before(async function() {
+      // Disable the timeout as it'll take build a little while to finish.
+      // eslint-disable-next-line
+      this.timeout(0);
+
+      await runEleventy('dev');
```
devnook, line 12: Please note that this is implemented in https://github.com/GoogleChrome/web.dev/pull/2640

test/site/_collections/mocks/collection.js
```diff
@@ -0,0 +1,21 @@
...
+  getAll() {
+    return this.items;
+  }
+
+  getFilteredByGlob() {
+    return this.items;
+  }
+
+  getFilteredByTag() {
+    return this.items;
```
devnook, line 15: Instead of removing the call to this method from the test's scope (by mocking it here), I'd rather see us extract the relevant part of the logic we want to test to a separate function and test that.

robdodson:
keep alive

PR #2662, opened 2020-04-24
 Add some tests to _filters

devnook:


src/site/_filters/contains-tag.js
```diff
@@ -17,7 +17,7 @@
 // Check to see if the post contains a specific tag.
```
robdodson, line 17: Since we're editing this file, what do you think about changing this comment to a `@fileoverview`? I want to see if we can start enforcing that via linting so may as well start now :D

test/site/_filters/capitalize.js
```diff
@@ -0,0 +1,25 @@
+const assert = require('assert');
+const capitalize = require('../../../src/site/_filters/capitalize');
```
robdodson, line 2: do you know if we only use `capitalize` inside of njk templates? I realized [nunjucks has a built-in filter for capitalize](https://mozilla.github.io/nunjucks/templating.html#capitalize) that we could be using. So if we're not using `capitalize` in any of our .js files we could remove it?

devnook: We use it mainly in our .js files in custom components.

src/site/_filters/contains-tag.js
devnook: I removed the contains-tag filter as it was mostly u used by now.

test/site/_filters/capitalize.js
devnook: We use it mostly in out js files for. ustom component, not njk templates.

devnook:
PTAL

PR #2664, opened 2020-04-24
Add synchronize to required checks event types.

robdodson:

- Try to be more aggressive about running this check. For whatever reason it seems to be missing some new PRs.

### DIFF
--- a/.github/workflows/required-checks-workflow.yml
+++ b/.github/workflows/required-checks-workflow.yml
@@ -1,5 +1,7 @@
 name: Required checks
 
+# Run this check any time a PR is opened, synchronized, reopened,
+# labeled or unlabeled.
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


samthor:
I'm confused because I thought this list specifically restricts the times we run checks. I thought leaving it out ran it in all cases?

Is there a missing event we're trying to avoid? Maybe add that as a comment.

robdodson:
> I'm confused because I thought this list specifically restricts the times we run checks. I thought leaving it out ran it in all cases?

No it's the opposite. This list specifies which events should trigger the workflow. In theory, telling this to run when a PR is `opened` should make it run for all new PRs, but that doesn't seem to be happening so I decided to add `synchronize` as a test to see if that kicks it into action.

robdodson:
i should add a comment anyway :P

PR #2670, opened 2020-04-27
Fix #2669

### LINKED ISSUE #2669: Fails when a preferred lang is set in cookie
**Describe the bug**
When a `preferred_lang` is set in cookie, "Internal Server Error" will be returned.

**To Reproduce**
1. Using DevTools, manually set `preferred_lang` as `en` as a cookie.
2. Load https://web.dev/

**Expected behavior**
Switch to the preferred language.

agektmr:
Returns the languages string specified in the cookie instead of the boolean.


robdodson:
@devnook would you mind giving this a review as well?

PR #2676, opened 2020-04-27
Adds page to handbook on resolving common glitch issues

housseindjirdeh:

- Adds page to handbook on resolving common glitch issues


housseindjirdeh:
@samthor that's a great point. Added a little note for that :)

robdodson:
It seems like web.dev would want to have a boosted team account versus asking individuals to pay for it themselves. Happy to leave the note in and if folks bring it up we can discuss having a team policy.

PR #2683, opened 2020-04-27
Task/organize test dir

robdodson:
This brings the `lib` tests into the `test` directory and updates karma to run them correctly. I think this is a incremental step in the larger refactoring of our project structure. I think eventually we probably want our project structure to be something like:

```
/
  eleventy/
  client/
  server/
  test/
    eleventy/
    client/
    server/
```

But we're not there yet. This is a step in that direction and allows us to write tests which will help us ensure as we _do_ move and rename directories that we don't break the site.

- Moves lib tests into test/unit.
- Changes lib tests to use `describe()/it()` instead of `suite()/test()`. This change was made to match the other existing tests.
- Explicitly adds node resolve, commonjs, and virtual plugins to buildTest(). Not sure why but doing `[...defaultPlugins, rollupPluginPostCSS()]` or `[rollupPluginPostCSS(), ...defaultPlugins]` didn't work.
- Updates package.json to run lib tests as part of `npm run test:unit`.
- Updates README in `test/`


robdodson:
percy change can be ignored, it's just being flaky

samthor:
I feel like this might make us switching to `import` for the codebase in future a bit trickier (we might need to do dynamic `import()` in tests).

And unfortunately I don't have any better suggestions for cleaner `require()` statements, short of a Rollup plugin.

Not approving yet since I want to know why the Rollup config doesn't work.

robdodson:
> I feel like this might make us switching to import for the codebase in future a bit trickier (we might need to do dynamic import() in tests).

Yeah I wasn't stoked about the change but wanted to at least make the test dir consistent. I think this refactoring should be followed by another refactoring where we actually break the client out and at that point it would be less weird if some tests were using import and others weren't.

> Not approving yet since I want to know why the Rollup config doesn't work.

¯\_(ツ)_/¯ but since this is blocking the code coverage stuff i'd like to land it today if we can

robdodson:
chatted with sam. we're going to go ahead and merge and try to sort out the rollup plugins issue afterwards.

PR #2711, opened 2020-04-28
Fix masthead on mobile.

### LINKED ISSUE #2483: /lean mastheads have too much margin on mobile.
image

Removing the margin on the h1 seems to make things look better.

robdodson:
Fixes #2483 

- Clean up the learning path mastheads so they look better on mobile. Trying to be better about using relative units like `em` to (slowly) nudge us in the right direction for an eventual typography/writespace refactor.

### DIFF
--- a/src/styles/components/_mastheads.scss
+++ b/src/styles/components/_mastheads.scss
@@ -75,22 +75,20 @@
 
 .w-masthead-path__container {
   align-items: center;
-  box-sizing: content-box;
   display: flex;
-  flex-direction: column-reverse;
-  margin: 0 auto;
-  max-width: 960px;
-  overflow: visible;
-  padding: 0 24px;
+  flex-direction: column;
+  padding: 2em 2em 0;
 
   @include bp(md) {
+    box-sizing: content-box;
     flex-direction: row;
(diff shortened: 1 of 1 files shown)


robdodson:
Failing percy tests are too be expected because I changed the orientation of the image and the text (text first, then image). And the whitespace around the text changed a bit with some padding changes.

Unfortunately percy also lagged when it snapshotted the /about/ page so it look like its blue indicator didn't show up in time :\

src/styles/components/_mastheads.scss
```diff
@@ -75,22 +75,20 @@
 
 .w-masthead-path__container {
   align-items: center;
-  box-sizing: content-box;
   display: flex;
-  flex-direction: column-reverse;
```
samthor, line 80: did `-reverse` actually manifest in a serious way before?

robdodson: > did -reverse actually manifest in a serious way before?

it used to make the image go before the text on mobile. That was the original design but I changed it in this PR

robdodson:
pushing an empty commit to see if i can make that one flaky percy snapshot do the right thing

PR #2727, opened 2020-04-29
add new metrics, show only if available

### LINKED ISSUE #2164: Add new metrics to /measure
I believe the new metrics would go in this table:

image

@samthor and I discussed having him work on this because he's familiar with the measure tool. I've pinged @addyosmani and the LH peeps to give us a copy of the json containing the new metrics.

samthor:
Attempts to fix #2164.

I've labelled this DO NOT MERGE because I don't quite understand if this is the goal yet. Waiting on email replies.


samthor:
@robdodson this is ready for review now.

PR #2740, opened 2020-04-29
feat(newsletters): add archived newsletters to `/newsletter`, fixes #2725

### LINKED ISSUE #2725: Add archived newsletters to /newsletter
We can get the html out of the newsletter emails that are sent out and embed it into the site as an iframe.

MichaelSolati:
Fixes #2725


MichaelSolati:
@robdodson PTAL, also we need a better way to rip the featured image from a post of the newsletter. Right now the html/css is a mess to dig through, so if we could as the team to maybe add a unique class/id to the image(s) in the "Featured posts" area we can better find them.

.eleventy.js
```diff
@@ -150,11 +152,13 @@ module.exports = function(config) {
   // COLLECTIONS
   // ----------------------------------------------------------------------------
   config.addCollection('authors', authors);
+  config.addCollection('newsletters', newsletters);
```
robdodson, line 155: it seems weird to add this as a collection because we never use it (IOW we never loop over `collections.newsletters`). I think you can remove this.

src/site/_collections/newsletters.js
```diff
@@ -0,0 +1,25 @@
+/*
```
robdodson, line 1: Do we need this file or could paginatedNewsletters import _data/newsletters?

src/lib/components/Newsletter/index.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * @fileoverview Element that renders copyable code.
```
robdodson, line 2: ?

src/lib/components/Newsletter/index.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+import {BaseElement} from '../BaseElement';
+import {debounce} from '../../utils/debounce';
+
+/**
+ * Renders code block that can easily be copied.
```
robdodson, line 9: I think this is a left over comment

src/lib/components/Newsletter/index.js
```diff
@@ -0,0 +1,40 @@
...
+    window.addEventListener('resize', this.adjustIFrame);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    window.removeEventListener('resize', this.adjustIFrame);
+  }
+
+  adjustIFrame() {
+    if (this.iframe.scrolling !== 'no') {
```
robdodson, line 33: since we control adding the iframe could we add it with scrolling already set to no?

src/site/_data/newsletter/archive/2020/03/index.html
```diff
@@ -0,0 +1,770 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
robdodson, line 1: Eleventy's data folder will merge content into the data cascade. I don't think we should use it as a place to store static content. I think we'd want to instead put these files into src/site/content.

Would it make more sense to treat these like any other piece of content in the site and just have them be part of a regular eleventy collection? In other words we'd make them into .njk files and use a `- newsletter` tag on them.

Since we already have to manually check these files in, it means we could also just write down the important info like title, hero image, etc. in their frontmatter. Essentially all the stuff that newsletter.js is trying to dynamically do right now.

Instead of using srcdoc we would set the iframe src to the generated file.

src/site/_includes/components/BaseCard.js
```diff
@@ -56,7 +57,7 @@ class BaseCard {
   }
 
   renderThumbnail(url, img, alt) {
-    const imagePath = path.join(url, img);
+    const imagePath = img.match(/^http(s?):\/\//) ? img : path.join(url, img);
```
robdodson, line 60: This seems like the BaseCard knows too much about how it's being used. It would probably make more sense to override this method in a subclass and add this. But see my comments above about treating the newsletters like any other piece of content. We may not need to do this if we copy their feature image into the same directory as the newsletter's html.

src/site/_includes/components/BaseCard.js
```diff
@@ -80,7 +81,8 @@ class BaseCard {
   }
 
   renderAuthorImages(authors) {
-    if (!Array.isArray(authors) || authors.length > 2) return;
+    if (!Array.isArray(authors) || authors.length > 2 || authors.length === 0)
```
robdodson, line 84:
```suggestion
    if (!Array.isArray(authors) || !authors.length || authors.length > 2)
```

src/site/_includes/components/BaseCard.js
```diff
@@ -105,7 +107,7 @@ class BaseCard {
   }
 
   renderAuthorNames(authors) {
-    if (!Array.isArray(authors)) return;
+    if (!Array.isArray(authors) || authors.length === 0) return;
```
robdodson, line 110:
```suggestion
    if (!Array.isArray(authors) || !authors.length) return;
```

src/site/_includes/components/BaseCard.js
```diff
@@ -32,11 +32,12 @@ const postTags = require('../../_data/postTags');
  */
 class BaseCard {
   constructor({post, featured = false, className = ''}) {
-    this.post = post;
+    this.element = post;
```
robdodson, line 35: I'm not sure I see the value in renaming `post` to `element`, especially if we're keeping the `{post}` argument in the constructor. I think it makes the code less obvious, IOW, it's introducing ambiguity but we don't gain anything from it. Essentially we know that 'post' is referring to a post item in a collection, and this abstracts that. I'd vote to just keep it as 'post'.

robdodson:
also please be sure to add tests for the collection that gets added

.eleventy.js
MichaelSolati: We use it in `/site/content/en/newsletter/individual.njk` but we can come back to it's use later.

src/lib/components/Newsletter/index.js
MichaelSolati: I thought it would be good to default scrolling to yes and then with js set it to no. The resizeing of the height wont work if JS is disabled, that way users can still scroll through the content with JS off.

src/site/_includes/components/BaseCard.js
MichaelSolati: I was actually thinking that the BaseCard should use `element` instead of `post` only because it ends up being the root/base for not just `PostCard` but also `AuthorCard`. So it thought a more generic name would be appropriate...

MichaelSolati:
> also please be sure to add tests for the collection that gets added

So writing a test for the collection that get's used is fine. But now it's only `paginated-newsletter`, and we don't modify the collection much anymore. Uhm, so all I can really test for is pagination for just one element...

Guidance and PTAL @robdodson would be appreciated.

src/site/_includes/components/AuthorCard.js
```diff
@@ -23,6 +23,7 @@ const BaseCard = require('./BaseCard');
  */
 class AuthorCard extends BaseCard {
   constructor(arg) {
+    arg.collectionItem = arg.author;
```
robdodson, line 26: mentioned in Chat. Let's explore just getting rid of the named arguments and make BaseCard take three arguments instead of an object

src/lib/components/Newsletter/index.js
```diff
@@ -0,0 +1,41 @@
...
+class Newsletter extends BaseElement {
+  constructor() {
+    super();
+    this.resizeIFrame = debounce(this.resizeIFrame.bind(this), 100);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.iframe = this.querySelector('iframe');
+    this.iframe.scrolling = 'no';
```
robdodson, line 23: if (this.iframe) {
  this.iframe.scrolling...
}

src/lib/components/Newsletter/index.js
```diff
@@ -0,0 +1,41 @@
...
+    window.addEventListener('resize', this.resizeIFrame);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    window.removeEventListener('resize', this.resizeIFrame);
+  }
+
+  resizeIFrame() {
+    this.iframe.height = this.iframe.contentWindow.document.body.scrollHeight;
```
robdodson, line 37: if (this.iframe) {
  this.iframe.height = ...
}

src/lib/components/Newsletter/index.js
robdodson: I think this `@fileoverview` comment is not correct

test/unit/src/site/_utils/add-pagination/index.js
```diff
@@ -0,0 +1,35 @@
+const assert = require('assert');
+const addPagination = require('../../../../../../src/site/_utils/add-pagination');
+
+describe('add-pagination', function() {
+  describe('addPagination', function() {
```
robdodson, line 5: mind adding a test for the `additionalData` argument that `addPagination()` takes?

PR #2742, opened 2020-04-30
track signed-in state via Analytics

### LINKED ISSUE #1941: Not properly tracking signed in state.

samthor:
Fixes #1941.

Adds an additional "basic" script which is run by everyone. This lets us share the dimensions rather than hard-coding them inline.


samthor:
I don't know. I'll revert it.

src/site/_includes/layout.njk
```diff
@@ -44,14 +44,12 @@
...
-      ga('set', 'transport', 'beacon');
-      ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version }}');
-      ga('send', 'pageview');
-    </script>
     <script async src="//www.google-analytics.com/analytics.js"></script>
+    {#
+      The basic script is "defer" as it should run before the normal bootstrap module code (which
+      is defer by default).
+    #}
+    <script defer src="/basic.js"></script>
```
robdodson, line 52: wdyt about renaming the file to something more descriptive like `analytics`? although i think we already have an analytics.js elsewhere, but `basic` seems like folks wouldn't be able to tell at a glance what it does.

build.js
```diff
@@ -42,6 +42,11 @@ process.on('unhandledRejection', (reason, p) => {
  * Virtual imports made available to all bundles. Used for site config and globals.
  */
 const virtualImports = {
+  webdev_analytics: {
```
robdodson, line 45: why put these here instead of in `webdev_config`?

src/lib/basic.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * @fileoverview Run by all browsers in `script defer`. Used to set up
+ * Analytics, including with our known previous signed-in state that was cached
+ * in `localStorage`.
+ *
+ * This file is built with Rollup, but separately to our core bundle.
```
robdodson, line 6: Is it worth mentioning explicitly that this file should never import anything from our core bundle?

Also might be good to explicitly state it's for IE11 etc.

build.js
samthor: You don't need to import everything if you only want a subset of config. Importing `{* as foo}` can bring all values in, so it's nice to break it apart into different configs.

PR #2747, opened 2020-04-30
Properly feature-detect StorageManager

NOtherDev:
navigator.storage might not be defined


kaycebasques:
@petele WDYT?

petele:
@kaycebasques - yep this is a good call.

@NOtherDev - thanks for catching this!

PR #2758, opened 2020-04-30
Add custom frontmatter linter

robdodson:
Related to #1959, supersedes #2189 

## TODO

- [x] Refactor main.js to simplify calling added/modified/all files per Sam's feedback.
- [x] Get all tests passing again.
- [x] Add fileoverview comments to everything.

## What this PR does

In a nutshell, this PR adds a new set of GitHub Actions to lint the frontmatter for our posts and publish a comment to the PR that lists any failures or warnings. It checks for the following things:

- All `.md` files in the PR
  - Verify the `date` property is present.
  - If the file has a `hero`, verify the `alt` property is present
- Added `.md` files in the PR
  - Verify the url matches the sluggified title. The reviewer can add a `lint: url approved` label to ignore this this rule.
  - Verify the url contains at least one dash. The reviewer can add a `lint: url approved` label to ignore this this rule.
- Modified `.md` files in the PR
  - Verify the `updated` property is present.

**Because rules are functions, we can just keep adding more**. For example, one to verify that the updated date is current, or that the tags used in the post match the list of tags in postTags.js.

## How to review this PR

- Start by looking at the changes to the lint_and_test.workflow.yml. I've added a series of steps at the end that run this custom linter.
- Take a look at `.github/actions/lint-front-matter/main.js`, that's the entry point to the linter.
- Look at each of the linters in `/linters` to see how they pull in their rules and iterate over the markdown files.
- Look at the rules in `/rules` to see how they're implemented. They're pretty simple functions.
- Look at the tests in `/test/.github/actions/lint-front-matter` to see what the expected behavior is.

## Limitations

- This linter only looks at markdown files (and their associated YAML frontmatter).
- This linter can only be run as a GitHub Action. I wrote this before the [update-updated pre-commit hook](https://github.com/GoogleChrome/web.dev/tree/master/tools/update-updated) and my mind wasn't thinking about wanting to run this locally. So folks would need to push to the repo to get these checks to run. It's a bit of a bummer, but I figured if the writers find these checks useful then it's better to have them now, and refactor to a version that can be run locally later versus just continuing to put this off.


jpmedley:
>If the writers find these checks useful then it's better to have them now, and refactor to a version that can be run locally later versus just continuing to put this off.

+1

test/unit/.github/actions/lint-front-matter/linters/modified-files-test.js
```diff
@@ -0,0 +1,46 @@
...
+      },
+    };
+
+    const files = [path.join(__dirname, 'fixtures', 'my-post', 'index.md')];
+    const actual = await modifiedFiles(files);
+    assert.deepEqual(actual.length, 1);
+    // Verify properties exist on results object.
+    assert.ok(actual[0].passes);
+    assert.ok(actual[0].failures);
+    assert.ok(actual[0].warnings);
```
jpmedley, line 44: I didn't realize you could have multiple asserts in a Mocha test.

robdodson:
oh hey look the bot is doing its thing :D

jpmedley:
👏🎉

.github/actions/lint-front-matter/rules/tags-are-valid.js
```diff
@@ -0,0 +1,35 @@
+const id = 'tags-are-valid';
+const Rule = require('./rule');
+const {ok, fail, warn} = require('../utils/status');
+const postTags = require('../../../../src/site/_data/postTags');
```
jpmedley, line 4: I wonder if this should be renamed `possibleTags` or something like that. This name caused me to think the inclusion logic at line 18 was inverted. In other words, `postTags` sounds more like a name for tags in a single post than it does for the list of all tags.

robdodson: mmmm maybe. But we should do that in a separate PR. There's a lot of things depending on `postTags` and we need to update all those files.

.github/actions/find-comment/action.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Find comment'
+description: 'Finds the id of the first comment by the GitHub Actions bot.'
+runs:
+  using: 'node12'
+  main: 'main.js'
+inputs:
+  author:
+    description: |
+      The name of the author to search for in the list of comments to update.
+      This task will find the id of the first comment by this author.
```
samthor, line 10: ... ", or the blank string if it's not made a post yet."

(to distinguish from an error case)

.github/actions/lint-front-matter/rules/url-length.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {string} file Filename of the file being linted.
+ * @return {Object} The result of the check and any failure/warning messages.
+ */
+const test = (file) => {
+  const dirname = path
+    .dirname(file)
+    .split('/')
+    .pop();
+
+  if (dirname.split('-').length === 1) {
```
samthor, line 16: So I'm allowed to have an article as long as it looks like a Custom Element name?

.github/actions/lint-front-matter/rules/url-matches-title.js
```diff
@@ -0,0 +1,29 @@
...
+const test = (file, yaml) => {
+  const dirname = path
+    .dirname(file)
+    .split('/')
+    .pop();
+  const sluggifiedTitle = slugify(yaml.title, {lower: true, strict: true});
+
+  if (dirname !== sluggifiedTitle) {
+    return fail('URL does not match title.', {
+      expected: `${sluggifiedTitle}`,
```
samthor, line 21: huge nit: ``'s are redundant since the var can just be used

.github/actions/lint-front-matter/rules/url-matches-title.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {string} file Filename of the file being linted.
+ * @param {Object} yaml The yaml front matter from the file.
+ * @return {Object} The result of the check and any failure/warning messages.
+ */
+const test = (file, yaml) => {
+  const dirname = path
+    .dirname(file)
+    .split('/')
+    .pop();
+  const sluggifiedTitle = slugify(yaml.title, {lower: true, strict: true});
```
samthor, line 17: This will fail on a bunch of existing posts I think. Not a huge deal, just a note.

.github/actions/find-comment/main.js
```diff
@@ -0,0 +1,41 @@
+const github = require('@actions/github');
```
samthor, line 1: I feel like since you're using an external dependency for _updating_ the comment, there was no good equivalent here. Is that a reasonable thing to say?

.github/actions/lint-front-matter/main.js
```diff
@@ -0,0 +1,33 @@
...
+    const added = core.getInput('added').length
+      ? core.getInput('added').split(' ')
+      : [];
+    const modified = core.getInput('modified').length
+      ? core.getInput('modified').split(' ')
+      : [];
+    const all = [...added, ...modified];
+
+    const results = [
+      ...(await lintAllFiles(all)),
```
samthor, line 20: I really think this would be much simpler if these methods just existed in this file. The three other files end up with tons of boilerplate. It looks scary to add another one since that seems fragile.

samthor: I'm aware this ties into how you test them. But I really think each _group_ of tests could look like:

```js
const allFilesRun = (frontMatter) => {
  return [
    hasProperty.test(frontMatter, 'date'),
    hasAltText.test(frontMatter),
  ];
};

// maybe later
module.exports.allFilesRun = allFilesRun;
```

.github/actions/lint-front-matter/rules/url-length.js
robdodson: yep. The intent is to catch things like `web.dev/javascript` — someone accidentally squatting on a high value path. I can update the comment to clarify that better.

.github/actions/lint-front-matter/rules/url-matches-title.js
robdodson: yeah i think this is only being run against newly added posts in added-files. And we may decide to scrap this rule entirely.

.github/actions/find-comment/main.js
robdodson: I'm not sure I follow... are you asking if we should use that `peter-evans/create-or-update-comment` action in the workflow because this action is already finding the comment so it may as well update it?

samthor: Yeah, I just want to know (potentially as a side note in `@fileoverview`) why we have our own. It might not be obvious to an external reader that this is here mostly because we couldn't find a useful external module.

.github/actions/lint-front-matter/rules/url-matches-title.js
robdodson: @kaycebasques @jpmedley do y'all want this rule? It basically fails if the url doesn't match the sluggified title. A reviewer can add a label to skip this test if they approve of the title and url. Another option would be to warn, or not bother flagging it at all.

.github/actions/find-comment/main.js
robdodson: yep i'll add a note (that is the reason). We could probably just ditch the other action and write the code ourselves to update or create the comment but it was kind of nice/convenient to just pipe things along to the next workflow step that "does the thing."

MichaelSolati:
@googlebot I consent

robdodson:
I think this is ready for another round of review but folks should feel free to ignore this until we get past web.dev/live if folks already feel like they have enough on their plates.

The tests are failing (on purpose) because this PR adds some failing files as test fixtures.

.github/actions/find-comment/main.js
peter-evans: Hi! I'm the author of `find-comment` and I stumbled across this PR while searching to see if anyone was _actually using_ it. 😆 Have no problem with you hacking on it to build your own, by the way. 👍 

I was just intrigued, though, why the action didn't work for you as-is. It looks like you are searching for the first comment by `github-actions[bot]` in PR comments? I think the action can do that already. Just curious, because if it's something I've missed or a bug I would like to fix it.

.github/actions/lint-front-matter/rules/url-matches-title.js
jpmedley: I don't think this is useful. It will result in either very long slugs or very short titles. I don't think either serves our readers.

PR #2762, opened 2020-05-01
fix scroll-position for in-page links relative to fixed header

### LINKED ISSUE #2598: Viewports off for internal links
In #2567, I had to split information regarding the three shape detection APIs between two locations in the document. I provided links between the locations so that readers could easily move between them. When I placed the target anchors in locations that seemed logical from a code standpoint I found that following the links lead to a page view where the linked-to content is scrolled off the top of the page. 

For example, I placed an anchor in the Shape Detection label under "Behind a flag" and linked to it from elsewhere in the document. When I followed that link, I was brought to a view that looked like this: 

image

What I expect to see is this:

image

I tried several anchor placements that seemed logical from a code standpoint. None of them produced the outcome I wanted. The only thing that seemed to work was placing the anchors in illogical locations: the end of the introductory paragraph, for example. This is hacky but I accepted the approach because the linking is temporary. As soon as all three portions of the Shape Detection API are enabled by default, the links will go away.

I am under the impression that there are techniques for addressing this that are less hacky, but I'm not familiar with them. Since they may involve changing site code and not something I can implement as an author, I've opened this ticket.

paulirish:
THIS CSS IS AMAZING

```css
html {
  scroll-padding-top: 70px; /* height of sticky header */
}
```  

https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/

Fixes #2598

cc @brendankenny 
adding some keywords because i always have trouble refinding this property: `scroll under overlap offset sticky fixed header named anchors inpage hash`


src/styles/generic/_shared.scss
```diff
@@ -26,6 +27,10 @@
   outline: none !important;
 }
 
+// https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/
+html {
+  scroll-padding-top: calc(#{$TOPNAV_HEIGHT} + 24px);
```
paulirish, line 32: 24 was arbitrary but it felt nice to give it a tad more space. otherwise the text is right up against the bottom of the header.

samthor:
IIRC we're putting a margin on headings to emulate this experience (`margin-top: -$TOPNAV_HEIGHT; padding-top: $TOPNAV_HEIGHT`). Can we remove that if this works?

robdodson:
> IIRC we're putting a margin on headings to emulate this experience (margin-top: -$TOPNAV_HEIGHT; padding-top: $TOPNAV_HEIGHT). Can we remove that if this works?

Yeah I would love to remove that and swap in this CSS. I had thought that this feature wasn't well supported but [it seems to be](https://caniuse.com/#search=scroll-padding). I'm not sure I understand the bit about Safari not supporting it on "fragment targets".. isn't that what we're linking to? But we can test it in Safari and see :D

robdodson:
hm yeah it doesn't seem to work in safari :(

robdodson:
to test, remove lines 74+ in _shared.scss

robdodson:
What if we do something like this?

https://twitter.com/simevidas/status/1257376146523652098

robdodson:
Technically Safari does support `scroll-padding-top` but only inside of a scroll snap container. Unfortunately that means we can't do:

```
@supports(...) {}
@supports not (...) {}
```

robdodson:
Sooooooo I rebased the PR to just make our anchor padding hack have a bit more space because I can't figure out a good way to feature detect scroll-padding. But I would like to merge the change I made at least.

Issue #2710, opened 2020-04-28
LH perf score is down.


robdodson:
**Describe the bug**
Running web.dev through PSI or web.dev/measure shows a perf score in the 80s. I'm not sure when we regressed but we should fix this.

**Expected behavior**
We need to be as close to 100 on perf as possible.

robdodson:
keep alive. Seems like it's actually gotten worse with LH v6 :(

PR #2767, opened 2020-05-01
load site code async, additional async firebase

samthor:
Supersedes @beaufortfrancois' (sorry) PR #2358 and lazily loads Firebase as needed.

This goes one step further though, and actually lazily loads our whole site's code. It's not required to read our content, and we push it after 'load' now.

Is this cheating on performance reviews? Yes. Yes it is.


src/lib/pageview.js
```diff
@@ -11,6 +11,7 @@
  */
```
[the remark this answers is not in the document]
samthor: The comment is sort of accurate but I've updated it.

I've also renamed this file to 'bootstrap', and our previous one to 'entrypoint'. I've updated all the comments to reflect that.

I've added a comment here about blocking loading site code unless Shadow DOM is supported. This is kind of my intention with that; I want to remove the WC loader code and block loading unless there's native support here. (Expect a follow-up CL).

src/lib/fb.js
```diff
@@ -1,36 +1,33 @@
...
-import firestoreLoader from './firestore-loader';
+import firebaseLoader from './utils/firebase-loader';
 
-/* eslint-disable require-jsdoc */
-
-firebase.initializeApp(firebaseConfig);
-
-// Initialize performance monitoring
-firebase.performance();
+const firebasePromise = firebaseLoader('app', 'auth', 'performance');
```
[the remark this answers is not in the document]
robdodson: I mean rename `const firebasePromise` to `const loadFirebase` but maybe a better name would be `const isFirebaseLoaded` so we could do `isFirebaseLoaded.then(...`

build.js
```diff
@@ -145,14 +154,14 @@ async function build() {
...
   }
 
-  // Rollup bootstrap to generate graph of source needs. This eventually uses
-  // dynamic import to bring in code required for each page (see router.js).
-  // Does not hash "bootstrap.js" entrypoint, but hashes all generated chunks,
-  // useful for cache busting.
+  // Rollup "entrypoint.js" to generate graph of source needs. This eventually
+  // uses dynamic import to bring in code required for each page (see router).
+  // The entrypoint itself is generated with a dynamic hash, so we can import
+  // it via bootstrap (which is run as ES5).
```
robdodson, line 160: bootstrap is run as es5?

src/lib/utils/firebase-loader.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+const libs = {};
+
+const firebasePrefix = '//www.gstatic.com/firebasejs/6.6.1';
+
+/**
+ * @param {string} library to preload
+ * @return {!Promise} for library
+ */
+function internalPreload(library) {
```
robdodson, line 16: is this preloading or is it loading?

src/lib/entrypoint.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * @fileoverview Site bootstrap code.
```
robdodson, line 2: Should this line say something else? I do worry that entrypoint and bootstrap are synonyms so it's still unclear at a glance which comes first...

robdodson: I know 'bundle' is a bit of a noise word, but it is idiomatic in something like webpack to say that bundle is the thing everything has been packed into. Kinda wondering if that would be a better name for this file... I realize I'm bikeshedding but it is confusing at a glance

test/unit/src/lib/utils/firebase-loader.js
```diff
@@ -0,0 +1,29 @@
+const {assert} = require('../assert');
+const loader = require('../../../../../src/lib/utils/firebase-loader');
+
+describe('firebase-loader', function() {
+  describe('loader', function() {
```
robdodson, line 5: Can you write a test to assert that on the initial call it returns a function and another test to verify that calling that returned function returns a promise. Since we're not using typescript compiler yet I think the tests would be the best way for us to enforce the contract.

test/unit/src/lib/utils/firebase-loader.js
```diff
@@ -0,0 +1,29 @@
+const {assert} = require('../assert');
+const loader = require('../../../../../src/lib/utils/firebase-loader');
+
+describe('firebase-loader', function() {
```
robdodson, line 4: is it worth exporting `internalPreload` so we can test that it does the things we expect it to do?

src/lib/fb.js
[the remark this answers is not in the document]
robdodson: It may also be worth renaming `firebaseLoader` to `queue()` or something. At first glance `firebaseLoader` seems like it's going to load something, but it's just a function that _gives you_ a loader which you still have to call yourself.

src/lib/fb.js
```diff
@@ -1,36 +1,33 @@
...
+firebasePromise.then(initialize).catch((err) => {
+  console.error('failed to load Firebase', err);
+});
 
-// Listen for the user's signed in state and update the store.
-let firestoreUserUnsubscribe = null;
-firebase.auth().onAuthStateChanged((user) => {
-  store.setState({checkingSignedInState: false});
+const firestoreLoader = (() => {
```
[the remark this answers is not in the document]
robdodson: > So we've always dynamically loaded Firestore, because we don't need it unless the user is signed in.

That makes sense.

> And the existing model does await whenever we need it (to write/clear the user's measure row), but that will always be a no-op after we sign in.

I do wonder if that means we're going to have a lot of boilerplate `await firebase` calls as we start to use fb more. Not a huge issue for this PR but worth considering.

build.js
samthor: I've updated this a bit just to explain the flow from entrypoint => bootstrap.

src/lib/utils/firebase-loader.js
samthor: I've removed references to 'preload' as that clearly has a strict meaning on the web, one which we're not using (i.e., `modulepreload` and friends).

src/lib/entrypoint.js
samthor: I didn't catch this comment. I've updated it a whole bunch to be clearer about what we do here.

test/unit/src/lib/utils/firebase-loader.js
samthor: I could be convinced, but I don't think so.

test/unit/src/lib/utils/firebase-loader.js
samthor: I've done this.

samthor:
This might help #2710.

PR #2773, opened 2020-05-01
Ignore paths that were triggering percy.

robdodson:
Some files like contributors.js get touched when an article gets published, and then percy runs every time someone pushes a change to that article during review. This limits the number of files percy cares about to just the subset that I think would actually affect the appearance of the site.

### DIFF
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -8,19 +8,21 @@ on:
       - master
     paths:
       - '**.js'
-      - '**.json'
+      - '!src/site/_data/contributors.js'
+      - '!src/site/_data/countries.js'
+      - '!src/site/_data/postTags.js'
+      - 'package.json'
       - '**.njk'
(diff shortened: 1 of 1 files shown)


.github/workflows/percy-workflow.yml
```diff
@@ -8,19 +8,21 @@ on:
...
     paths:
       - '**.js'
-      - '**.json'
+      - '!src/site/_data/contributors.js'
+      - '!src/site/_data/countries.js'
+      - '!src/site/_data/postTags.js'
+      - 'package.json'
       - '**.njk'
       - '**.scss'
-      - '**.yml'
```
samthor, line 14: did you mean to remove these?

robdodson: I did yeah. I don't think we're using them in the project for anything other than configuring github actions or app engine or some build bot thing. We don't use them for content it seems.

PR #2775, opened 2020-05-02
Fix mistake in the example

littleli:
- field `used` does not exist on resolved value, I belive it's a type and `usage` should be used

usage


littleli:
@googlebot I signed it!

samthor:
Thanks!

PR #2780, opened 2020-05-04
removes webcomponents loader, start Shadow DOM

samthor:
This does a few things:

* removes Web Components loader code
* adds a Rollup plugin which compiles ".scss.js" files to the css-tag required by lit-element (this could be later extended just to export the compiled style text, too)
* uses that plugin for `lighthouse-gauge`.

This is basically the first step in flipping us all to Shadow DOM.

This can go in now, but should probably wait until #2767 so we can gate loading site code on `window.ShadowRoot` support.


compile-css.js
```diff
@@ -20,70 +20,17 @@ const isProd = process.env.ELEVENTY_ENV === 'prod';
...
-    return require('sass');
-  }
-})();
-
-/**
- * @param {string} input filename to read for input
- * @param {string} output filename to use for output (but does not write)
- * @return {{css: !Buffer, map: !Buffer}}
- */
-function compileCSS(input, output) {
```
samthor, line 39: The change here is that I've moved this helper into "src/build/compile-css.js" so we can share with both the top-level CSS compilation and CSS compilation for components.

samthor:
So FWIW Percy is failing as it's not showing the element I modified. Running the site locally, even in prod, shows it working fine.

I'm not 100% sure what's going on here; perhaps lit is rendering a frame later in the SD case.

If you're happy with the _approach_ I'll explore why Percy isn't showing this.

samthor:
As an aside, while looking at our Analytics data from Jan-Apr this month, I believe that about 99%^ of our users support WC and Shadow DOM. 

^I've extrapolated only last month of Edge use, since the Chromium-based versions are much more prevalent now than in Jan.

samthor:
So I don't think Percy has any interest in supporting SD. We can't switch if we want to continue to use their testing platform.

I do actually think we could hack their API to make it so—they were generally unresponsive as to how it worked so it would be a lot of work from our end.

robdodson:
Doing some house cleaning. I think I'm going to close this PR but will revisit it if we decide to move forward with SD.

PR #2781, opened 2020-05-04
covid banner

### LINKED ISSUE #2709: Add covid-19 banner to the front page of the site.
We have a design for this from Harleen, just need to add it to the site. We could dig up the banner we wrote for CDS and reuse that.

samthor:
Resolves #2709.

I've taken the simplest approach and reused the CDS banner CSS. But honestly this just seems like it's a featured post card. Is there something I'm missing?

also, the designs have a cute stylized graphic, but that doesn't match the hero image we've used for the blog post. I've used the hero but I'm happy to change it.

Screen Shot 2020-05-04 at 12 57 32


robdodson:
> But honestly this just seems like it's a featured post card. Is there something I'm missing?

I think there are some small differences. The post card will do a big card hover effect and I think it needs to be inside of a grid container. We could try it and see how it looks? I'm fine with either.

> also, the designs have a cute stylized graphic, but that doesn't match the hero image we've used for the blog post. I've used the hero but I'm happy to change it.

Good question. I suspect we should use the graphic they gave us and maybe swap out the graphic on the post itself. I think that's what Harleen wanted.

robdodson:
Looks kinda squashed on mobile.

image

kaycebasques:
Yes we should replace the hero image on the post with the provided graphic per Harleen's request

LMK if you want to do it here or if I should create a separate PR

robdodson:
I think we can do it here. I think Sam was going to tweak the CSS a bit first.

samthor:
I've updated the banner while still making it just a regular inline link (not a card). It looks like this:

Screen Shot 2020-05-12 at 11 59 41
Screen Shot 2020-05-12 at 11 59 47

samthor:
My main concerns were around the copy so if Kayce has approved I'll get this in.

PR #2782, opened 2020-05-04
add missing localStorage import

samthor:
Fixes a tiny missing import.


samthor:
I'm going to TBR this to check the gcloud stuff, which is why I did this PR.

PR #2788, opened 2020-05-05
Add RUM for all Core Web Vitals metrics

philipwalton:
This PR adds the [web-vitals](https://github.com/GoogleChrome/web-vitals) JS library and measurement code for each of the three Core Web Vitals metrics to web.dev.

The logic added is taken exactly from the [Google Analytics example in the README](https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-analytics), which I'm doing intentionally because I plan to build a tool that will allow anyone to view a custom dashboard showing how their site currently scores on these metrics based on their own Google Analytics data (for the last 28 days). And it will work as long as they're sending their data in the same format as shown in the README example.

Also, potentially we could add this tool to the `/measure` page, but that's a separate discussion :)


src/lib/pageview.js
```diff
@@ -10,13 +10,42 @@
...
+    // Use a non-interaction event to avoid affecting bounce rate.
+    nonInteraction: true,
+  });
+}
+
 // eslint-disable-next-line
 window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
 ga('create', id);
 ga('set', 'transport', 'beacon');
+ga('set', 'page', window.location.pathname);
```
philipwalton, line 43: Note, I added this change for consistency with the logic [here](https://github.com/GoogleChrome/web.dev/blob/fb3d450613ecd5ec68f84b3407aaeeca4675746d/src/lib/loader.js#L170) (which will remove query params from the URLs of pages in your GA reports).

I can remove this though if you want to keep this PR just to the scope of adding `web-vitals`.

robdodson:
This seems pretty neat. @samthor would you mind giving it a review?

robdodson:
also @philipwalton i think it may need to be rebased. I think a previous PR renamed pageview to bootstrap

philipwalton:
Hmmm, from the description is sounds like the `bootstrap` file is not transpiled and needs to be ES5-friendly? If so then it might make sense to move this logic elsewhere (the metrics don't work in older browsers anyway).

@samthor what do you think? The `web-vitals` lib is <1K, but not sure if you want it included in the bootstrap logic. I'm happy to update this PR to whatever you think is best.

samthor:
Sorry about these files being renamed. The head version should be called "bootstrap.js".

Bootstrap isn't transpiled, it just doesn't use modern features (not really enforced though). It's a best-effort for older browsrs.

* Does web-vitals make sense on unsupported browsers? Consider, by the way, that 99% of our users (according to my rough Analytics estimations) support our bundle.

* It's running last in bootstrap, so can we just move it to the first part of entrypoint.js? (this is lazy-loaded though, will this effect reporting?)

philipwalton:
> Bootstrap isn't transpiled, it just doesn't use modern features (not really enforced though). It's a best-effort for older browsrs.

The `web-vitals` source code includes things that would be a syntax error in IE11 (e.g. arrow functions), so including it and not transpiling would likely break all of it for older browsers.

> Does web-vitals make sense on unsupported browsers? Consider, by the way, that 99% of our users (according to my rough Analytics estimations) support our bundle.

I don't think it does. I think it'd be totally find to add it to the `analytics.js` module and have it not execute in nomodule contexts.

> (this is lazy-loaded though, will this effect reporting?)

Lazy-loading shouldn't affect reporting at all, unless the user abandons the load (which I don't think we can really do much about anyway).

philipwalton:
@samthor I've updated the PR to add that code to the `analytics.js` file. Can you PTAL?

philipwalton:
@robdodson @samthor polite ping, can we add this?

robdodson:
i'll merge after checks pass

PR #2791, opened 2020-05-05
firebase loading fix

samthor:
I inadvertently introduced two issues to the Firebase code in #2767.

* We weren't returning the `firestore` object from its associated `Promise`
* We weren't setting `.async = false` on the dynamic scripts, so if e.g. Firebase's performance script loaded _before_ app, the load would fail.
(This didn't really show up in testing since cached files executed in element order.)


samthor:
I'm going to TBR this because our /measure page is currently broken for logged-in users.

PR #2794, opened 2020-05-05
Add analytics events to measure tool.

robdodson:

- Add missing analytics events for measure tool. Should now track when someone switches url, runs an audit, or clears the url.


robdodson:
going to self merge becasue this is needed for an ad campaign running today

PR #2796, opened 2020-05-05
chore(redirect): send users from `/subscribe` to `/newsletter`, fixes #2724

### LINKED ISSUE #2724: Redirect web.dev/subscribe to web.dev/newsletter
For the newsletter page to go live we need to redirect /subscribe to it, and we should make sure to remove its `noindex` flag.

MichaelSolati:
Fixes #2724

Add redirects from `/subscribe` to `/newletter`.

### DIFF
--- a/src/lib/components/Actions/index.js
+++ b/src/lib/components/Actions/index.js
@@ -105,8 +105,7 @@ class Actions extends BaseElement {
         data-category="web.dev"
         data-label="subscribe, newsletter"
         data-action="click"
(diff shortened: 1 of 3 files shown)


robdodson:
btw should we remove `noindex` from the newsletter page as part of this PR?

MichaelSolati:
@robdodson the newsletter page already has noinfex removed... So we should be good?

robdodson:
ok let's give it a go! 🚀

PR #2798, opened 2020-05-05
Use the Content Indexing API

### LINKED ISSUE #2031: Try out the Content Indexing API
The [Content Indexing API](https://web.dev/content-indexing-api/) is running as an Origin Trial in Chrome 80.

Since all web.dev posts under `/blog/` that a user has visited are cached and should work offline, web.dev should consider trying out the Content Indexing API.

Using the API when you already have an offline-capable PWA is pretty straightforward. [Here are some examples](https://twitter.com/jeffposnick/status/1207764802271154178) of adding it to some PWAs that I maintain. If you're open to the idea but there's no engineering time from the core team available, I could submit a PR to add support.

CC: @petele

jeffposnick:
Fixes #2031

See https://web.dev/content-indexing-api/ for a description of the API.

Moving some discussion from the issue:

I speculatively refactored the service worker partial JSON output, so that it includes extra description, url, and imageSrc fields, when those values are available. This should be backwards compatible with the existing cache storage entries, so I didn't bump the versioning in the cache name.

That makes using the Content Indexing API a lot easier, as I can just do a sweep through all of the indexed JSON once during startup, and use the values directly from cache storage to keep the index up to date. Otherwise, I had to put some code inside of `loader.js` that would react to every page load and pull out the required metadata, using CSS class names to figure out the `description` and/or `imageSrc` values, and that was messy.


robdodson:
Looks like there's some test failures. Might need to update things in `test/unit/src/site/_transforms/service-worker-partials`

src/lib/content-indexing.js
```diff
@@ -0,0 +1,98 @@
...
+  url.searchParams.set('fm', 'webp');
+
+  return {
+    sizes: `${PREFERRED_ICON_SIZE}x${PREFERRED_ICON_SIZE}`,
+    src: url.href,
+    type: 'image/webp',
+  };
+}
+
+// This method syncs the currently cached media with the Content Indexing API
```
samthor, line 41: I feel like this is ripe for a helper library.

src/lib/entrypoint.js
```diff
@@ -35,6 +36,9 @@ WebComponents.waitFor(async () => {
 
 if (serviceWorkerIsSupported(window.location.hostname)) {
   ensureServiceWorker();
+  syncContentIndex().catch((error) => {
+    console.error('Content Indexing error:', error);
```
samthor, line 40: analytics.js exports a `trackError` helper which pushes this to Analytics.

src/lib/content-indexing.js
```diff
@@ -0,0 +1,98 @@
...
+// (on browsers that support it). The Cache Storage is the source of truth.
+export async function syncContentIndex() {
```
samthor, line 43: Should this also have an "insert" method which is run on SPA navigation? Right now, the index is only updated when I load the site anew.

e.g.:

- I load https://web.dev/, the SW is installed, content is synced (almost nothing)
- I navigate to a few pages
- Content isn't synced

- I load https://web.dev a day later, now those pages get synced

jeffposnick: It could, yeah. That's how I initially had it implemented, but I moved to a one-off sync at startup in the name of... efficiency, I guess?

I can refactor to do on-the-fly indexing in addition to the one-off startup sync, and I guess if that looks like it's too heavyweight in the real world, we can adjust.

src/lib/content-indexing.js
jeffposnick: There is a decent amount of boilerplate code here that I've reused from other integrations. But at the same time, the specific details that differ from site to site—around URL schemes and cache storage—are pretty closely tied to the implementation. So it's been hard to abstract the reusable bits out.

If you've got any specific ideas, though, we can follow-up on that.

src/lib/content-indexing.js
jeffposnick: See my follow-up comment about where to run that on-the-fly indexing code.

src/lib/utils/router.js
```diff
@@ -193,6 +195,9 @@ export function route(url) {
   globalHandler(candidateUrl, u.hash).then((aborted) => {
     if (!aborted) {
       scrollToHashOrTop(u.hash);
+      addPageToContentIndex(u.href).catch((error) => {
```
jeffposnick, line 198: @samthor, following up from https://github.com/GoogleChrome/web.dev/pull/2798/files#r420448496, this is my attempt to add to the content index on the fly.

I am not confident that this is the right place to put this code, but _somewhere_ within `router.js` seemed like a reasonable place to put some code that is meant to be run following user navigation. If there's a better location for it, though, I'm happy to put it elsewhere.

I'm also not 100% sure that at this execution point the new page's JSON is guaranteed to be in the Cache Storage API, so if you have doubts about that, we should move it to someplace where it would run later in the navigation lifecycle, when everything is fully cached.

src/lib/content-indexing.js
jeffposnick: FWIW, I did some internal refactoring when adding in support for on-the-fly single URL indexing.

jeffposnick:
If we get to the point where folks are happy with the new code, I'll write some unit test for `content-index.js` and add them to this PR.

(One question I had about testing: are folks okay with adding in a `devDependency` for [`sinon`](https://sinonjs.org/)? It's really useful for mocking out some of the service worker interfaces inside of unit tests.)

robdodson:
> (One question I had about testing: are folks okay with adding in a devDependency for sinon? It's really useful for mocking out some of the service worker interfaces inside of unit tests.)

yeah that may be a good call. FWIW we have been using [proxyquire](https://www.npmjs.com/package/proxyquire) in other places when we need to stub out something that a module requires.

jeffposnick:
...

kaycebasques:
@jeffposnick are you still waiting on review from @samthor?

robdodson:
@jeffposnick is this waiting on feedback from Sam? I think we're all heads down working on either /live/ or other OKRs at the moment, but I did want to figure out who's move it is so I can set an assignee for the issue.

jeffposnick:
Yes, feedback from @samthor would be appreciated (e.g. on https://github.com/GoogleChrome/web.dev/pull/2798#discussion_r421085069)

I know that some tests would be required before merging this, but I am holding off on writing those until I get the basic "yes, this approach looks reasonable.:

jeffposnick:
Ping @samthor.

samthor (commented):
I'm sorry, I completely missed that you pinged this PR.

I'm generally happy with this but can you bring it up to HEAD?

src/lib/content-indexing.js
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Adds a given pageURL to the content index.
+ *
+ * Metadata for the indexed page is read from the Cache Storage API, by
+ * translating the page's URL into a cache key for the JSON metadata.
+ *
+ * @param {string} pageURL The URL of the page being indexed, or the cache key
+ * corresponding to that page's metadata.
+ * @param {Cache} [cache] If set, an open Cache object that the metadata can
```
samthor, line 73: nit: I don't think `[cache]` needs to be in square brackets

src/lib/content-indexing.js
```diff
@@ -0,0 +1,158 @@
...
+ * be read from.
+ */
+export async function addPageToContentIndex(pageURL, cache) {
+  const index = await getContentIndexInterface();
+  //  Bail early if the Content Indexing API isn't supported.
+  if (!index) {
+    return;
+  }
+
+  const cacheKey = pageURL.endsWith('index.json')
```
samthor, line 83: I think this will always be fine, but a more extreme option could be:

```js
import {normalizeUrl} from './urls';

  // later
  pageURL = normalizeURL(pageURL);
  if (!pageURL.endsWith('/')) {
    return;  // abort
  }
  pageURL += 'index.json';
```

samthor (approved):
Thanks for the refactor. Amusingly now it feels heavy handed to redo the entire cache on every load, but I can't really see a better option.

samthor:
so I massaged this PR a bit to make it work. Basically it was failing to insert entries because the domain is always wrong in testing, and the Content Indexing API was unhappy with that.

PR #2799, opened 2020-05-05
feat(feedback): add custom element to take feedback on posts, fixes #2386

### LINKED ISSUE #2386: Create a "Was this page helpful?" widget

We need to make it possible for web.dev users to give us feedback on whether or not our content is actually helpful.

https://codepen.io/mfriesenhahn/full/QWwVzoj

image

I don't know if we discussed this in the meeting, but I think the widget should be open (expanded) by default if possible. cc @robdodson 

N/A

**Additional context**

One question that might be very helpful to add: "Is it easy to map this guidance back to your codebase?"

@robdodson, @mfriesenhahn, @MichaelSolati, and I have already met and reached consensus on this design.

MichaelSolati:
Fixes #2386


src/site/_includes/partials/feedback.njk
```diff
@@ -0,0 +1,80 @@
+<web-feedback>
```
samthor, line 1: Can we put this inside Feedback instead? It doesn't feel like to me that it belongs in the generated source (although I'll admit the line is a bit fuzzy).

We're moving to SD so this will be less awkward soon, but right now maybe just set `.innerHTML` on the `<web-feedback>` on startup?

src/styles/generic/_input.scss
```diff
@@ -14,3 +14,27 @@ input[type='text'] {
     outline: none;
   }
 }
+
+
+input[type='radio'] {
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  appearance: none;
+  background: url('data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAyNCAyNCIgd2lkdGg9IjI0Ij4KICA8cGF0aCBkPSJNMTIgMkM2LjQ4IDIgMiA2LjQ4IDIgMTJzNC40OCAxMCAxMCAxMCAxMC00LjQ4IDEwLTEwUzE3LjUyIDIgMTIgMnptMCAxOGMtNC40MiAwLTgtMy41OC04LThzMy41OC04IDgtOCA4IDMuNTggOCA4LTMuNTggOC04IDh6Ii8+CiAgPHBhdGggZD0iTTAgMGgyNHYyNEgweiIgZmlsbD0ibm9uZSIvPgo8L3N2Zz4K');
```
robdodson, line 23: I think we discussed doing this purely with CSS to see if it avoids pixelation. Just adding a note about that.

src/site/_includes/partials/feedback.njk
robdodson: Had a chat with Sam about this.

I think what I'd recommend is moving all of this content into the `render()` function for `web-feedback`. I think this element probably wouldn't use Shadow DOM even if we had that option because it's not doing any slotting so there isn't much benefit of obscuring the styles and having to use adoptedStyleSheets. Instead, it can call render, splat out all this content as light dom children, and take advantage of the site's already existing stylesheets.

src/styles/components/_feedback.scss
```diff
@@ -0,0 +1,17 @@
+@import '../settings/colors';
+
+#w-feedback__dropdown {
```
robdodson, line 3: I would suggest using classes here. Not a huge deal that we're using IDs but I also don't see much value in it. If (for whatever reason) we did have more than one feedback element on a page then this would technically be wrong (though I think it would still work).

src/styles/components/_feedback.scss
```diff
@@ -0,0 +1,17 @@
...
+  display: none;
+}
+
+#w-feedback__dropdown + .w-feedback__label + .w-feedback__form {
```
robdodson, line 7: I think if you're using the existing summary/details element (https://web.dev/handbook/web-dev-components/#details) you may not need to do this.

robdodson:
If it's helpful at all I did a version of the layout using grid, https://codepen.io/robdodson/pen/yLYprLb

I think you could also do it using flexbox but I love me some grid :)

src/styles/components/_feedback.scss
MichaelSolati: Don't think the web-dev-component will help if its being returned in the `render` function.

src/styles/components/_feedback.scss
kaycebasques: I'll spare you the details unless helpful but I could imagine more than one feedback component on a page

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,193 @@
...
+        category: 'Feedback',
+        action: 'submit',
+        label: entry[0],
+        value: Number(entry[1]),
+      });
+    });
+  }
+
+  render() {
+    return html`
```
samthor, line 44: Add a comment here saying that we intentionally want to be in the light DOM.

MichaelSolati: good point, and doone

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,194 @@
...
+        </div>
+        <form
+          class="w-feedback__form ${this.submitted && 'hidden'}"
+          @submit=${this.submit}
+        >
+          <small>All fields optional</small>
+          <div class="web-feedback__rows">
+            <div class="web-feedback__row">
+              <div></div>
+              <div>Yes</div>
```
robdodson, line 68:
```suggestion
              <div id="yes-label">Yes</div>
```

Need this for the `aria-labelledby` attributes to work.

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,194 @@
...
+              <div>No</div>
```
robdodson, line 69:
```suggestion
              <div id="no-label">No</div>
```

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,194 @@
...
+      <details class="w-details">
+        <summary class="w-details__summary">
+          <h2 class="w-details__header">Give feedback</h2>
+        </summary>
+        <div
+          class="w-display--flex w-justify-content--center ${!this.submitted &&
+            'hidden'}"
+        >
+          <div class="w-subscribe--padded">
+            <p class="w-text--center w-mt--non">
```
robdodson, line 55:
```suggestion
            <p class="w-feedback__confirmation w-text--center w-mt--non w-force-focus" tabindex="-1">
```

Adding `tabindex=-1` here will allow us to call `focus()` on this element after the user hits submit so a screen reader will announce the confirmation message. The `force-focus` class will suppress the focus outline so sighted users don't see a blue ring on the paragraph.

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,194 @@
...
+    return html`
+      <details class="w-details">
+        <summary class="w-details__summary">
+          <h2 class="w-details__header">Give feedback</h2>
+        </summary>
+        <div
+          class="w-display--flex w-justify-content--center ${!this.submitted &&
+            'hidden'}"
+        >
+          <div class="w-subscribe--padded">
```
robdodson, line 54: You can delete the div with w-subscribe--padded. I don't think it's contributing anything.

I did notice that the confirmation message is a little iffy on mobile.
image


Maybe we'd want it to be vertically centered?

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,194 @@
...
+    formIterable.forEach((entry) => {
+      trackEvent({
+        category: 'Feedback',
+        action: 'submit',
+        label: entry[0],
+        value: Number(entry[1]),
+      });
+    });
+  }
+
```
robdodson, line 42:
```suggestion
updated() {
    if (this.submitted) {
      this.querySelector('.w-feedback__confirmation').focus();
    }
  }
```

Use this to focus the confirmation text

src/styles/generic/_input.scss
```diff
@@ -1,5 +1,19 @@
 @import '../settings/colors';
 
+@keyframes ripple {
```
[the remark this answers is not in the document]
MichaelSolati: Going to remove it for now, it's an area i'm to weak in and will take up way too much time =/

MichaelSolati:
@robdodson PTAL

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,202 @@
...
+          <div class="web-feedback__rows">
+            <div class="web-feedback__row">
+              <div></div>
+              <div id="yes-label">Yes</div>
+              <div id="no-label">No</div>
+            </div>
+
+            <div class="web-feedback__row">
+              <div id="helpful-label">
+                <label>Was this page helpful?</label>
```
robdodson, line 82: Using a `<label>` isn't necessary here because it's not associated with anything. It would need to either wrap a form input, or use its `for=""` attribute to reference the form input to be associated with it. Instead, the outer div's `id` + the use of `aria-lablledby` is what's providing the label for the form inputs. https://web.dev/labels-and-text-alternatives/

IOW, you can remove the `<label>` and leave the text:

```html
<div id="helpful-label">Was this page helpful?</div>
```

src/lib/components/Feedback/index.js
```diff
@@ -0,0 +1,202 @@
...
+                  type="radio"
+                  aria-labelledby="readable-label no-label"
+                  name="Readability"
+                  value="0"
+                />
+              </label>
+            </div>
+          </div>
+
+          <div class="w-text--center">
```
robdodson, line 191: Maybe give this element `.5em` of `padding-top` so it has a bit of breathing room on mobile.

src/lib/components/Feedback/_styles.scss
```diff
@@ -0,0 +1,30 @@
...
+web-feedback {
+  display: block;
+
+  .web-feedback__row {
+    align-items: center;
+    border-bottom: 1px solid $GREY_300;
+    display: grid;
+    font-family: "Google Sans", sans-serif;
+    font-size: .875em;
+    font-weight: 500;
```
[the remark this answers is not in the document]
robdodson: need to import styles/settings/type to use that var.

robdodson:
🙌  awesooooome!
I'll add this to our dashboard so we can start visualizing the data 📊

mfriesenhahn:
It lives!⚡ Awesome!

robdodson:
@mfriesenhahn 👋  yep, it's finally arrived 🛥

PR #2800, opened 2020-05-05
fix(web-newsletter): ensure iframe has loaded before modifying it

MichaelSolati:
Add `this.iframe.addEventListener('load', this.onIFrameLoad)` and/or check if `this.iframe.contentWindow.document.readyState === 'complete'` before removing scroll, setting height and modifying the `<a>` tags.


robdodson:
LGTM, feel free to merge when you're ready

PR #2802, opened 2020-05-06
remove lib/entrypoint.js, merge with lib/app.js

samthor:
- removes lib/entrypoint.js (previously loaded when page components loaded) and merge with lib/app.js—I think at first reading it wasn't obvious where our code _began_

- moves 'focus-visible' polyfill to where it's needed

- adds/updates some comments

- removes disused "urls-test.js" (this was already copied to the test/ folder)


robdodson:
random half-baked thought — should we instead rename entrypoint to app?

samthor:
Yes, that sounds good. I'll update and submit.

PR #2818, opened 2020-05-07
redirect /foo => /foo/

### LINKED ISSUE #2807: web.dev/promises flowchart 404's
**Describe the bug**
The promises article includes a `promise-flow.svg` which 404's.

**To Reproduce**
1. Go to web.dev/promises
2. Open the console
3. See error

**Expected behavior**
Flow graph loads

```
promise-flow.svg:1 Failed to load resource: the server responded with a status of 404 ()
```

samthor:
Fixes #2807.

I think I was too aggressive last time I changed this stuff. I can't see a reason not to reintroduce this blind redirect, fundamentally because our server that generates redirects (which will be hit if e.g., /foo/ 404's) _also_ doesn't care (i.e., normalizes) all URLs to end with a slash, too.

### DIFF
--- a/src/lib/sw.js
+++ b/src/lib/sw.js
@@ -164,7 +164,20 @@ workboxRouting.registerRoute(normalMatch, async ({url}) => {
   // it's otherwise missing.
   if (!pathname.endsWith('.html')) {
     if (!pathname.endsWith('/')) {
-      pathname += '/';
+      // Special-case a URL that doesn't end with "/". We must redirect before
+      // returning content as otherwise relative URLs don't work.
+      // This is not a problem as:
+      //   * this is occuring almost immediately (no network traffic)
+      //   * our webserver treats "/foo" and "/foo/" the same for other folders
+      //     and for the _redirects.yaml handler (i.e., a 404 on /foo and /foo/
+      //     will be treated the same way)
+      //   * a _real_ network request to "/foo" gets 301'ed to "/foo/"
+      const headers = new Headers();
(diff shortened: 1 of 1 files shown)


src/lib/sw.js
```diff
@@ -164,7 +164,20 @@ workboxRouting.registerRoute(normalMatch, async ({url}) => {
...
+      // Special-case a URL that doesn't end with "/". We must redirect before
+      // returning content as otherwise relative URLs don't work.
+      // This is not a problem as:
+      //   * this is occuring almost immediately (no network traffic)
+      //   * our webserver treats "/foo" and "/foo/" the same for other folders
+      //     and for the _redirects.yaml handler (i.e., a 404 on /foo and /foo/
+      //     will be treated the same way)
+      //   * a _real_ network request to "/foo" gets 301'ed to "/foo/"
+      const headers = new Headers();
+      headers.append('Location', pathname + '/');
```
robdodson, line 176: Dumb question... remember that time we ended up with web.dev/foo////////////////?bar=baz. Is there risk of doing the same thing here? Just want to make sure we don't hit that one again.

samthor: This is explicitly looking at the pathname so I don't believe so (and a quick test confirms that). Actually, we _weren't_ forwarding the search param, so I've fixed that and now:

* /foo => /foo/
* /foo?bar => /foo/?bar
* /foo?bar#zing => /foo/?bar#zing

PR #2820, opened 2020-05-07
Fix date.

jpmedley:


jpmedley:
@argyleink I got carried away and merged your flexbox gap article. This is the date fix.

PR #2835, opened 2020-05-08
fix(postTags): set title in default description to lower case, fixes #2785

### LINKED ISSUE #2785: Tag pages description should be lowercase
https://web.dev/tags/performance/

image

The word "Performance" in the description should be lowercase.

MichaelSolati:
Fixes #2785

Call `.toLowerCase()` on `tag.title` when generating a tag's description.

### DIFF
--- a/src/site/_data/postTags.js
+++ b/src/site/_data/postTags.js
@@ -202,7 +202,7 @@ Object.keys(tags).forEach((key) => {
   const tag = tags[key];
   const description = tag.description
     ? tag.description
-    : `Our latest news, updates, and stories about ${tag.title}.`;
+    : `Our latest news, updates, and stories about ${(tag.title).toLowerCase()}.`;
 
   postTags[key] = {
     ...tag,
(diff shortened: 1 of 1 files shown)


robdodson:
hmmmm thinking about this a bit... what if the title is an acronym (3D, WASM, AMP) or a proper noun (Web Vitals).

MichaelSolati:
Should I really be writing conditional flags for each title though? At that point we should just have meaningful descriptions written for the tags (IMHO).

robdodson:
> Should I really be writing conditional flags for each title though?

Nope, that wouldn't be very scaleable.

> At that point we should just have meaningful descriptions written for the tags (IMHO).

Thinking out loud here...

- If an item doesn't have a description field then we use the standard text + calling toLowercase on the title.
- If an item _does_ have a description field then we use that instead.

Then we do a pass on the postTags file and add a description to any acronyms or proper nouns. We can do that in this PR.

Would that work?

MichaelSolati:
@robdodson can i get a lgtm

robdodson:
Nice! 🚀

PR #2836, opened 2020-05-08
Remove unused Actions shortcode.

robdodson:

- Removes unused Actions shortcode. We use the web-actions web component now.


robdodson:
> Was it require()'d anywhere?

Doesn't seem like it. I think the only place it would have been required would be `.eleventy.js` and I think we must have removed it when we added the web component but forgot to delete the file. I think the site wouldn't build (and tests would fail) if we were requiring it elsewhere

PR #2848, opened 2020-05-12
basic live page

samthor:
v0 for web.dev/live. Just includes top content for the first PR. Works responsively.

Doesn't take CSS directly from the mock, because I thought it might be nice to keep our fonts mostly consistent.


samthor:
Direct staging link:
https://deploy-preview-2848--web-dev-staging.netlify.app/live/

src/site/content/en/live/index.njk
```diff
@@ -0,0 +1,83 @@
...
+  <p>
+These are unprecedented times, with people worldwide looking to stay connected and informed.
+The Web plays a special role here, and we're proud to see how the community has stepped up.
+  </p>
+  <p>
+Join Google’s Web Platform team to celebrate our community's actions, learn modern web techniques, and connect with each other.
+  </p>
+
+  <div class="w-masthead-home__buttons">
+    <a class="w-button w-button--primary w-masthead-home__button--primary gc-analytics-event" data-category="web.dev" data-label="subscribe, newsletter" data-action="click" href="#subscribe">
```
robdodson, line 38: sadly our `::before` trickery to create scroll-padding isn't working on the #subscribe element so instead of scrolling to just before the header, it scrolls to about halfway over the header.

image

src/site/content/en/live/index.njk
```diff
@@ -0,0 +1,83 @@
...
+      Get updates
+    </a>
+    <a class="w-button w-button--secondary gc-analytics-event" data-category="web.dev" data-label="live, add to calendar" data-action="click">
```
robdodson, line 41: Missing calendar icon.

image

We have `.w-button--with-icon` to support icon buttons.

src/styles/pages/_live.scss
```diff
@@ -0,0 +1,137 @@
+@import '../settings/colors';
+@import '../tools/breakpoints';
+
+.w-live-landing-page {
```
robdodson, line 4: type sizes are a bit off and rather than redo the typography for the site, I'd suggest the easiest path is to do:

```scss
.w-live-landing-page {
  p {
    font-size: 14px;
    line-height: 1.5;
    @include bp(sm) {
      font-size: 16px;
    }
  }
}
```

src/styles/pages/_live.scss
```diff
@@ -0,0 +1,137 @@
...
+
+.w-event-section {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: center;
+  margin: 64px auto;
+  max-width: 960px;
+  text-align: center;  // for mobile
+
+  .w-event-heading {
```
robdodson, line 27: I'm not sure if it's worth it to nest in this way. It's increasing selector specificity and bytes over the wire without producing a benefit since .w-event-heading already scopes the styles pretty well on its own.

src/site/content/en/live/index.njk
robdodson, line 38: Buttons on mobile should not be full width.

Our version:
image


Compared to the mock:
image

robdodson: I think you can do this by making the container `display: inline-flex` and `align-items: center`.

samthor: I agree, but I'm leaving this for a follow-up.

samthor: fixed. I've had to do CSS like this to make the padding work though:

```css
.w-event-buttons {
  display: inline-flex;
  flex-flow: column wrap;

  .w-button--primary {
    margin: 0 0 16px 0;
  }
}
```

I could add another class. Let me know if this isn't good BEM.

src/site/content/en/live/index.njk
samthor: I've used "playlist_add" as there's no "add to calendar" button (there's a calendar one). I think it works for now.

src/styles/pages/_live.scss
```diff
@@ -0,0 +1,173 @@
...
+  @include bp(sm) {
+    flex-direction: row;
+
+    .w-button--primary {
+      margin: 0 16px 0 0;
+    }
+  }
+}
+
+.w-event-heading {
```
robdodson, line 54: oh if the parent block isn't used you could also just make the two children into blocks of their own.

.w-event-logo
.w-event-subhead

src/site/content/en/live/index.njk
robdodson: let's use `calendar_today` instead.

image

src/site/content/en/live/index.njk
robdodson: meh that's fine :D

src/lib/loader.js
```diff
@@ -16,10 +16,20 @@ import './utils/underscore-import-polyfill';
  * @return {!Promise<?>}
  */
 async function loadEntrypoint(url) {
-  if (url.startsWith('/measure/')) {
-    return import('./pages/measure.js');
-  } else if (url.startsWith('/newsletter/')) {
-    return import('./pages/newsletter.js');
+  const prefixTo = url.indexOf('/', 1);
```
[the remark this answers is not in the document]
robdodson: ah ok.

Why switch to this instead of `url.startsWith('/measure')`?

robdodson:
Replied to a few comments. Very minor stuff. LGTM after swapping the icon and if you feel like renaming the event-logo/subhead classes.

src/lib/loader.js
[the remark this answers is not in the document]
samthor: We just have more entrypoints now and a `switch` statement makes it easy to read.

src/styles/pages/_live.scss
samthor: I think it's cleaner to leave it since these two things really only make sense in context. I've added a longer comment.

PR #2849, opened 2020-05-12
IE11 article syndication

### LINKED ISSUE #2750: content: Syndication of "What To Expect When Dropping IE11" [May 29 2020]
* Author: @samthor

samthor:
Fixes #2750.

This is the first article we've syndicated; while the meta stuff shows up in the header, there's no hint on the page itself that this is from another source. Is it worth worrying about doing this properly, or should I just add e.g. a link "Originally posted on <...>"?

I'm also happy for content review (although it is mostly a list) since I originally wrote this on my own blog. I don't mind tweaking it (I already did, just for updated stats).


samthor:
Also, it's worth noting that since the publish date was last year, this isn't showing up at the top of the blog or on any index page. (Which is probably fine.)

kaycebasques:
I do feel like we need some kind of UI indicating that this has been syndicated. People assume web.dev guidance to be canonical best practice. If we don't have syndication UI then people will assume this to be canonical, just like every other page. Another option would be to have the syndicated material clearly separated from the blog.

I haven't reviewed the content, but I feel like I can't and shouldn't touch the content whatsoever because it would no longer be a syndication. That puts me in a tough spot for ensuring content quality across the site. I anticipate some people would come to me and say "this IE11 page doesn't follow any of our guidelines, yet you allowed that one to get published to the blog...?" Like I said I haven't actually reviewed the content and I'm not implying that your content doesn't follow our content quality guidelines, Sam. I'm just trying to explain a difficult scenario that syndication might create for me.

I was expecting some kind of feature that automatically pulls in the content from the source, rather than us manually copy-pasting it into the blog. I suppose that syndication will encourage more contributions because it's less work. But there will still be some tedious work for contributors.

Sorry for not being able to express this feedback earlier. I didn't realize all of the implications until I saw the staged post. In short, I'm skeptical of this now, but I am willing to continue discussing.

robdodson:
> I do feel like we need some kind of UI indicating that this has been syndicated.

Yes. I actually thought the template _did_ do that. I'm surprised it doesn't at least have a blurb saying "This post originally appeared on [insert url]". We should add that (maybe in this PR?)

>  I anticipate some people would come to me and say "this IE11 page doesn't follow any of our guidelines, yet you allowed that one to get published to the blog...?"

I think we could justifiably say that content we syndicate follows a looser process, specifically because we can't ask someone to rewrite their original post. I think folks would understand that.

> I was expecting some kind of feature that automatically pulls in the content from the source

Technically we could probably fetch the source during our prod build step and try to reverse their html into markdown that we then render. But we'd almost certainly have rendering issues. Even when dev.to "imports" your syndicated posts, it lets you get in and change the imported markdown to fix any issues. Then it gets committed to their database.

> But there will still be some tedious work for contributors.

yep I think that part is somewhat unavoidable. FWIW, most folks on the team seem to author their blog in markdown anyway so usually they're just copying that and pasting it into a PR.

> In short, I'm skeptical of this now, but I am willing to continue discussing.

I think we should add the UI clearly indicating something is syndicated and try it out. Again, it's ultimately not that dissimilar from what dev.to provides. There may even be tools out there that you can feed a URL into and it'll get you the markdown for it. We could offer that to folks who don't originally author in markdown.

samthor:
So firstly, I am very happy this PR has spurred this discussion. That was entirely the point of creating a PR for my random post, even if we don't end up _actually_ syndicating it!

To echo Rob's comments:

* I don't think it makes sense to blindly mirror external posts, even despite the inherent risks in e.g., the source changing. There's always going to be different features in each platform—we have a shortcode for `Aside`, for example. 

* dev.to does the kind of disgusting (but quite practical) thing of starting you off with _generated_ Markdown derived from the rendered HTML source (that's contained within RSS). It's not always perfect and I have to tweak it a bit.

* I don't think we should ship posts unless it's hugely clear that these are syndicated from author's personal blogs. I'd almost even go as far to say that we label them "opinion" (or something similar).

kaycebasques:
> I'm surprised it doesn't at least have a blurb saying "This post originally appeared on [insert url]". We should add that (maybe in this PR?)

+1

> I think we could justifiably say that content we syndicate follows a looser process

I'm worried that people would use this workflow to bypass our content quality control processes but we can cross that bridge if we get there

> I don't think we should ship posts unless it's hugely clear that these are syndicated from author's personal blogs

+1

Re: taking in content from other sources and automagically converting it, I now understand and agree that it's not feasible. The current workflow of copy-pasting Markdown works for me. So we can set that discussion aside.

PR #2878, opened 2020-05-15
webdev-live show sessions

samthor:
Shows sessions for web.dev/live.

Disables property-sort-order as it doesn't play nice with CSS Variables, see [issue](https://github.com/sasstools/sass-lint/issues/1224) (which has been going for ~2 years, has a PR, and seems to be being ignored).

Adds just a couple of sessions for now. We can fill them in.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -37,6 +37,7 @@ const Compare = require(`./${componentsDir}/Compare`);
 const CompareCaption = require(`./${componentsDir}/CompareCaption`);
 const Details = require(`./${componentsDir}/Details`);
 const DetailsSummary = require(`./${componentsDir}/DetailsSummary`);
+const EventTable = require(`./${componentsDir}/EventTable`);
 const Hero = require(`./${componentsDir}/Hero`);
 const Instruction = require(`./${componentsDir}/Instruction`);
 const Label = require(`./${componentsDir}/Label`);
@@ -49,6 +50,7 @@ const YouTube = require(`./${componentsDir}/YouTube`);
 
 const collectionsDir = 'src/site/_collections';
 const authors = require(`./${collectionsDir}/authors`);
+const eventSchedule = require(`./${collectionsDir}/event-schedule`);
 const paginatedAuthors = require(`./${collectionsDir}/paginated-authors`);
 const paginatedBlogPosts = require(`./${collectionsDir}/paginated-blog-posts`);
 const paginatedNewsletters = require(`./${collectionsDir}/paginated-newsletters`);
@@ -151,6 +153,7 @@ module.exports = function(config) {
   // COLLECTIONS
   // ----------------------------------------------------------------------------
   config.addCollection('authors', authors);
+  config.addCollection('eventSchedule', eventSchedule);
   config.addCollection('blogPosts', blogPostsDescending);
   config.addCollection('postsWithLighthouse', postsWithLighthouse);
   config.addCollection('recentBlogPosts', recentBlogPosts);
@@ -213,6 +216,7 @@ module.exports = function(config) {
   config.addPairedShortcode('CompareCaption', CompareCaption);
   config.addPairedShortcode('Details', Details);
   config.addPairedShortcode('DetailsSummary', DetailsSummary);
+  config.addShortcode('EventTable', EventTable);
   config.addShortcode('Hero', Hero);
   config.addShortcode('Instruction', Instruction);
   config.addPairedShortcode('Label', Label);
--- a/src/site/_collections/event-schedule.js
+++ b/src/site/_collections/event-schedule.js
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const data = require('../_data/event');
+const contributors = require('../_data/contributors');
+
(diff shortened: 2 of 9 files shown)


src/styles/pages/_live.scss
```diff
@@ -153,6 +153,28 @@
...
 }
 
+web-tabs.w-event-tabs {
+  --primary-color: #{$WEB_PRIMARY_COLOR};
+  --hover-color: #{$GOOGLE_BLUE_50};
+  --active-color: #{$GOOGLE_BLUE_100};
+}
+
+table.w-event-schedule {
+  // TODO(robdodson): The generic table has `min-width: 512px`, which extends over the side of
```
robdodson, line 163: @samthor I think we usually enforce tables be wrapped in an outer `<div class="w-table-wrapper">` so they can overflow-x properly on small devices. But we can also see about removing the mind-width.

.eleventy.js
```diff
@@ -49,6 +50,7 @@ const YouTube = require(`./${componentsDir}/YouTube`);
 
 const collectionsDir = 'src/site/_collections';
 const authors = require(`./${collectionsDir}/authors`);
+const eventSchedule = require(`./${collectionsDir}/event-schedule`);
```
robdodson, line 53: This doesn't need to be an eleventy collection. You can access items in the `_data` folder inside of any page template by using its directory name. For example, the `event.json` file is available to all pages as `{{ event }}`.

https://www.11ty.dev/docs/data-global/

src/site/_includes/components/EventTable.js
```diff
@@ -0,0 +1,64 @@
...
+// const constants = require('../../_utils/constants');
+// const getSrcsetRange = require('../../_utils/get-srcset-range');
+// const postTags = require('../../_data/postTags');
+
+const Author = require('./Author');
+
+module.exports = (collections) => {
+  const schedule = collections.eventSchedule;
+
+  // TODO(MichaelSolati): "Author" needs a post but doesn't use it, we pass an empty object.
```
robdodson, line 31: As an alternative you could use the `w-author` HTML / classes directly. The shortcode is syntactic sugar but if you're having to pass it bogus arguments that feels off.

src/site/_includes/components/EventTable.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+// const path = require('path');
+const {html} = require('common-tags');
+// const prettyDate = require('../../_filters/pretty-date');
+// const stripLanguage = require('../../_filters/strip-language');
+// const md = require('../../_filters/md');
+// const constants = require('../../_utils/constants');
+// const getSrcsetRange = require('../../_utils/get-srcset-range');
+// const postTags = require('../../_data/postTags');
```
robdodson, line 24: should these be deleted?

src/site/_collections/event-schedule.js
```diff
@@ -0,0 +1,46 @@
+/*
```
robdodson, line 1: You may want to move this file up to the _data directory. Alternatively, since stuff in _data is in the global namespace, it might make sense to make these [directory data files](https://www.11ty.dev/docs/data-template-dir/) that live in the `/live/` directory.

Since they're not using the eleventy `collections` object, it feels a little off to put them in the `collections` directory.

src/site/content/en/live/index.njk
```diff
@@ -65,15 +65,7 @@ draft: true
...
-      <web-tabs>
-        <div data-label="Day 1">
-        </div>
-        <div data-label="Day 2">
-        </div>
-        <div data-label="Day 3">
-        </div>
-      </web-tabs>
-
+{% EventTable collections %}
```
robdodson, line 68: Mind indenting this?

src/site/_includes/components/EventTable.js
```diff
@@ -0,0 +1,64 @@
...
+          <tbody>
+            ${sessions.map(renderSession)}
+          </tbody>
+        </table>
+      </div>
+    `;
+  };
+
+  return html`
+    <web-tabs class="w-event-tabs">
```
robdodson, line 60: It seems like the tabs component has a required `label` attribute. Otherwise it'll set an `aria-label` on itself of `undefined`. Not sure why it was written that way, but would you mind adding a `label="schedule"` here? That way it'll announce "selected tab, 1 of 3, schedule, tab group".

.eleventy.js
```diff
@@ -213,6 +216,7 @@ module.exports = function(config) {
   config.addPairedShortcode('CompareCaption', CompareCaption);
   config.addPairedShortcode('Details', Details);
   config.addPairedShortcode('DetailsSummary', DetailsSummary);
+  config.addShortcode('EventTable', EventTable);
```
robdodson, line 219: Would you mind adding a comment that we should remove this shortcode if we take down the live page, or if the live page no longer uses it. I think the post-event mocks don't use it so at that point it'd just be dead weight.

src/styles/pages/_live.scss
samthor: Thanks for the clarification. In this case, there's no reason to let it extend—although cramped, the visual design works well at 320px.

src/site/content/en/live/index.njk
samthor: I'll fix on Monday.

src/site/_includes/components/EventTable.js
samthor: How odd. I'll fix on Monday.

src/site/_includes/components/EventTable.js
samthor: The Author shortcode is actually quite convenient here, I really like it.

This comment (cc @MichaelSolati) was mostly observing that I think the Author shortcode has `post` left over from an older time, it's just not used.

Another bit of feedback might be that it could be nice to automatically (yet again I'm confused about Eleventy's data) load the "author" information from the `id`. Since I've had to do a bunch of work to load it up correctly.

src/site/_includes/components/EventTable.js
samthor: Yup. I did it in the `web-event-time` PR.

src/site/_collections/event-schedule.js
samthor: Thanks for the suggestion. This was a bit ambitious and I'll move it to "_data" on Monday.

PR #2880, opened 2020-05-15
web.dev/live local time component

samthor:
Adds time component to web.dev/live. Supports start time (assumed in UTC) and optional duration in hours.

This displays in local timezone. Looks like this:

Screen Shot 2020-05-15 at 17 52 32

Matches mocks.

### DIFF
--- a/src/lib/components/EventTime/_styles.scss
+++ b/src/lib/components/EventTime/_styles.scss
@@ -6,7 +6,7 @@ web-event-time {
     display: flex;
   }
 
-  .et-date-feat {
+  .date {
     text-align: left;
     align-items: center;
     display: flex;
@@ -16,21 +16,21 @@ web-event-time {
     width: 68px;
   }
 
-  .et-date-feat__month {
+  .date__month {
     text-transform: uppercase;
   }
 
-  .et-date-feat__day {
+  .date__day {
     font-size: 44px;
     line-height: 1;
     text-transform: uppercase;
   }
 
-  .et-time-feat {
+  .time {
     text-align: left;
   }
 
-  .et-time-feat__timezone {
+  .time__timezone {
     color: $GREY_500;
   }
 }
--- a/src/lib/components/EventTime/_styles.scss
+++ b/src/lib/components/EventTime/_styles.scss
@@ -7,13 +7,13 @@ web-event-time {
   }
 
   .date {
-    text-align: left;
     align-items: center;
     display: flex;
     flex-flow: column;
     font-family: 'Google Sans', sans-serif;
     padding: 16px;
-    width: 68px;
+    text-align: left;
+    width: 80px;
   }
 
   .date__month {
--- a/src/lib/components/EventTime/_styles.scss
+++ b/src/lib/components/EventTime/_styles.scss
@@ -1,0 +1,36 @@
+@import '../../../styles/settings/colors';
+
+web-event-time {
+  time {
+    align-items: center;
+    display: flex;
+  }
+
+  .et-date-feat {
+    text-align: left;
+    align-items: center;
+    display: flex;
+    flex-flow: column;
+    font-family: 'Google Sans', sans-serif;
+    padding: 16px;
+    width: 68px;
+  }
+
+  .et-date-feat__month {
+    text-transform: uppercase;
+  }
+
+  .et-date-feat__day {
+    font-size: 44px;
+    line-height: 1;
+    text-transform: uppercase;
+  }
+
+  .et-time-feat {
+    text-align: left;
+  }
+
+  .et-time-feat__timezone {
+    color: $GREY_500;
+  }
+}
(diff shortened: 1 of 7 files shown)


src/lib/components/EventTime/index.js
```diff
@@ -0,0 +1,134 @@
...
+      const end = new Date(this._date);
+      end.setHours(end.getHours() + this.duration);
+      durationPart = html`
+        &mdash; ${format(end, {hour: 'numeric', minute: '2-digit'})}
+      `;
+    }
+
+    return html`
+      <time datetime=${this._date.toISOString()}>
+        <div class="et-date-feat">
```
robdodson, line 112: it seems like you could get rid of the `-feat` suffix on all of these classes.

src/lib/components/EventTime/index.js
```diff
@@ -0,0 +1,134 @@
...
+      `;
+    }
+
+    const format = (d, options) => {
+      return new Intl.DateTimeFormat('default', options).format(d);
+    };
+    const options = new Intl.DateTimeFormat().resolvedOptions();
+
+    let durationPart = '';
+    if (this.duration > 0) {
```
robdodson, line 102: does lit provide a default value for `duration`? Just curious if it might be `null` here.

src/lib/components/EventTime/index.js
robdodson: I'm kind of wondering if we should also just drop the `et-` prefix and adopt a policy where we use one level of CSS nesting for custom elements that use light dom. Might make the html and css tidier.

```scss
web-event-time {
  .date {}
  .date__month {}
  .date__day {}
  .time {}
  .time__value {}
  .time__timezone {}
}
```

src/lib/components/EventTime/index.js
```diff
@@ -0,0 +1,134 @@
...
+    super();
+    this._date = null;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    if (connected.size === 0) {
+      // Every 10 minutes, check if we're in the same timezone.
+      timezoneCheckInterval = window.setInterval(checkTimezone, 10 * 60 * 1000);
```
robdodson, line 60: Why does an element need to do this on an interval as opposed to checking once when it's connected?

src/lib/components/EventTime/index.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  shouldUpdate(changedProperties) {
+    if (changedProperties.has('datetime')) {
+      this._date = null;
+
+      if (this.datetime) {
+        const d = new Date(Date.parse(this.datetime));
+        if (+d) {
```
robdodson, line 80: Can you add a comment explaining that this is checking the validity of the parsed date?

src/lib/components/EventTime/index.js
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+
+    return super.shouldUpdate(changedProperties);
+  }
+
+  render() {
+    if (!this._date) {
+      return html`
+        <time>?</time>
```
robdodson, line 92: I was curious if it's better to fail silently here and return nothing at all.

src/lib/components/EventTime/_styles.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  .et-time-feat {
+    text-align: left;
+  }
+
+  .et-time-feat__timezone {
+    color: $GREY_500;
+  }
+}
```
robdodson, line 36: i think the linter grumps if we don't end this with a newline.

src/lib/components/EventTime/index.js
samthor: No. But also, `null > 0` is false, so this won't run.

src/lib/components/EventTime/index.js
samthor: If the user's browser remains open but they change timezone, we want to re-render the data.

src/lib/components/EventTime/index.js
samthor: Sound good.

src/lib/components/EventTime/index.js
samthor: I've done this. I was worried that this would collide with other generic styles, but I guess the point is that we generally don't have those (we only have `.w_...`), so it is safe to use simple names when nested.

samthor: (This is especially try as we're not reparenting children)

samthor:
I've pushed an updated and merged version. This now shows the correct data for each event day.

src/lib/components/EventTime/index.js
robdodson: lol. ain't nobody gettin' on planes anymore.

but ok i understand now 👍

PR #2881, opened 2020-05-15
web.dev/live share

samthor:
Adds a sharing button and section to the web.dev/LIVE page.

This is the path of least resistance but I'm also not proud. I just forked the `<web-actions>` element, which is implicitly a FAB ("floating action button", fixed in bottom right).

I think getting this in now makes sense but we should fix it up soon after—the right way is probably to add more options to the element, and make the floating-ness part of the top-level page.

This also fixes the hero image by giving it a known w/h.

### DIFF
--- a/src/lib/components/EventShare/index.js
+++ b/src/lib/components/EventShare/index.js
@@ -1,0 +1,93 @@
+/**
+ * @fileoverview Element that renders configurable per-page actions.
+ */
+
+import {html} from 'lit-element';
+import {BaseElement} from '../BaseElement';
+import {isWebShareSupported} from '../../utils/web-share';
+
+/**
+ * Renders configurable per-page actions. This is expected to be created by
+ * page content.
+ *
+ * @extends {BaseElement}
+ * @final
+ */
+class EventShare extends BaseElement {
+  static get properties() {
+    return {
+      // Whether the Web Share API is supported
+      _webShareSupported: {type: Boolean},
+    };
+  }
+
+  constructor() {
+    super();
+    this._webShareSupported = isWebShareSupported();
+  }
+
+  onWebShare() {
+    navigator.share({
+      url: this.shareUrl,
+      text: this.shareText,
+    });
+  }
+
+  onShare(e) {
+    e.preventDefault();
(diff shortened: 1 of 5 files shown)


MichaelSolati:
Yea, if we're talking about just getting it out the door this is the solution... But I'm thinking web actions should just use this component for the share action.

MichaelSolati:
Accidentally pushed up some stuff i needed to remove. @samthor created a pr where i fixed the twitter share function and set the `Actions` component to use `EventShare` (now just `Share`)

https://github.com/GoogleChrome/web.dev/pull/2888

robdodson:
It seems like the best course would be to create separate web components for twitter share, facebook share, and instagram share (is that a thing...?) since the mock calls for all three.

image

These would be custom elements that don't render anything and just wrap a light dom child that is a styled `<button>` and listen for its click event. If web share is supported it uses the web share url, if not it opens a new window with the fallback url.

I'm not sure if it's worth it to merge this PR as is because we still have to make those other components anyway and we're probably going to have to hack around the existing FAB styling that the PR is using. It may be better to just do this refactoring now.

@MichaelSolati I think the work you did in your PR is cool but I have some reservations around calling render on a custom element without attaching it to the DOM (https://github.com/GoogleChrome/web.dev/pull/2888/files#diff-842d5dc7651bf35561a7880cc1441242R34). It means the element's `connectedCallback` will never fire and I could imagine someone trying to put stuff in `connectedCallback` someday in the future and not understanding why it doesn't work.

MichaelSolati:
> if it's worth it to merge this PR as is because we still have to make those other components anyway and we're probably going to have to hack around the existing FAB styling that the PR is using. It may be better to just do this refactoring now.
> 
> @MichaelSolati I think the work you did in your PR is cool but I have some reservations around calling r

So, why not add the twitter/facebook/insta if webshare is unavailable?

Also instead of calling the render, i could return it as some like `html\`<web-share></web-share>\`;`

robdodson:
It sounded like Sam wanted to combine twitter/facebook/insta into a single share component.

samthor:
Dropped in favor of @MichaelSolati 's PR #2888

PR #2888, opened 2020-05-15
refactor(Actions): use EventShare, now Share, in Actions for sharing

MichaelSolati:
@samthor PTAL

### DIFF
--- a/src/lib/components/Actions/index.js
+++ b/src/lib/components/Actions/index.js
@@ -4,7 +4,8 @@
 
 import {html} from 'lit-element';
 import {BaseElement} from '../BaseElement';
-import {isWebShareSupported} from '../../utils/web-share';
+import {Share} from '../Share';
+import {splitPipes} from '../../utils/split-pipes';
 
 /**
  * Renders configurable per-page actions. This is expected to be created by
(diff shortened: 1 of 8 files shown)


src/lib/components/Actions/index.js
```diff
@@ -20,82 +21,17 @@ class Actions extends BaseElement {
...
-        data-action="click"
-        href="${url}"
-        target="_blank"
-        rel="noreferrer"
-        @click=${this.onTwitterShare}
-      >
-        <span>Share</span>
-      </a>
-    `;
+    const element = new Share();
```
samthor, line 32: This is a little odd (instantiating an element class just to render it) but stranger things have happened.

As long as the two modes work (Twitter, a link, _and_ Web Share, a button) then I'm fine with this!

src/lib/components/Share/index.js
```diff
@@ -26,24 +29,42 @@ class EventShare extends BaseElement {
     this._webShareSupported = isWebShareSupported();
   }
 
+  onShare(e) {
```
samthor, line 32: I don't think this is used right now. Can we remove it?

robdodson:
Originally I was thinking of something like:

```
<web-share fallback="twitter">
  <button><!-- style this button however you want --></button>
</web-share>

<web-share fallback="facebook">
  <button><!-- style this button however you want --></button>
</web-share>

<web-share fallback="instagram">
  <button><!-- style this button however you want --></button>
</web-share>
```

I realize that's a bit odd because if web share is supported then clicking any of those buttons will open the share dialog where you have to, yet again, choose your share target.

Sam, were you thinking that if web share is supported the visual UI would only show a single share button and if it's not supported there would be three buttons?

Issue #2896, opened 2020-05-18
Author component requires post and author info


samthor:
The `Author` and `AuthorInfo` components require passing through a post and the actual author's expanded information (as opposed to just the ID of the author).

For the post argument, we don't actually use this at all, but authors were previously only shown adjacent to their posts.

For the expanded information, we plumb it through but it could be provided directly by the contributors data via a key lookup.

PR #2897, opened 2020-05-18
remove post/author info requirement, uses titles for authors

### LINKED ISSUE #2895: Authors require both first and last names
**Describe the bug**
Authors require both first and last names.

**To Reproduce**
https://web.dev/authors/surma/ shows that there's a blank space after "Surma", since we expect a last name.

**Expected behavior**
No blank space.

samthor:
Fixes #2895 and #2896.

* Allows easier `Author` component use by removing `post` (not used) and `author` (we can look it up) requirements

* Only uses valid name parts for contributor `title`, and use `title` instead of constructing a full name throughout the site


samthor:
The Percy change here correctly detects that we now display "Surma." instead of "Surma ." (removed the space)

Issue #2900, opened 2020-05-18
web.dev/live doesn't support multiple speakers


samthor:
The event data and generation code currently only supports a single speaker.

While this might remove our need to use the `Author` component, and we generate the HTML ourselves in `EventTable`, I still think it's worth fixing in #2897 (so it's easier to use).

robdodson:
We do have a design for multiple authors but the overlapping is kinda 'meh'. It's necessary because cards are pretty tight.
![image](https://user-images.githubusercontent.com/1066253/82360787-be136700-99be-11ea-8686-45d25a33ed9b.png)

PR #2905, opened 2020-05-18
web.dev/live content update

samthor:
Adds most sessions for web.dev/live.

* Skips daily keynotes (for now)
* Titles may not be final (drawn from multiple sources)
* Missing a few images (authors haven't provided them yet)

TODOs have been left in-place for when #2900 is fixed (for multiple speakers).


samthor:
I actually _somewhat_ disagree about the event thing because `EventTable`, the component, is also slated for removal once we're done with this. For now I am happy to leave this event as a singleton.

PR #2914, opened 2020-05-19
multiple authors for web.dev/live

### LINKED ISSUE #2900: web.dev/live doesn't support multiple speakers

samthor:
Fixes #2900.

This actually extracts a new component `AuthorsDate` from `BaseCard`, and swaps the event code to use it. (We were previously using the `Author` card, which _also_ supports showing information like social media, suitable for the top of blog posts or the Author page itself.)

This fixes a small issue that isn't tracked in that the _images_ of each card had no focus styles (you can see this literally on the homepage now). That's now restored.

This also attempts to make the event table usable on a tiny screen (320px iPhone), see screenshot. It's basically palatable but if we had more information I'd suggest ditching the table (or at least restyling it completely with CSS to be purely a list of author, talk, author, talk ...).

Screen Shot 2020-05-19 at 13 09 49


src/styles/components/_author.scss
```diff
@@ -117,21 +117,31 @@ $AUTHOR_IMAGE_SIZE_SMALL: 40px;
...
   height: $AUTHOR_IMAGE_SIZE_SMALL;
   margin-left: calc(-#{$AUTHOR_IMAGE_SIZE_SMALL} / 2);
-  overflow: hidden;
   position: relative;
   width: $AUTHOR_IMAGE_SIZE_SMALL;
-}
 
-.w-author__image--row-item > .w-author__image {
-  border: solid white 2px;
+  // > a {
```
robdodson, line 138: should this be uncommented?

PR #2930, opened 2020-05-20
web.dev/live data update

### LINKED ISSUE #2922: Add missing speaker images.
If speakers are missing images we can just grab them from their twitter or internal teams page.

samthor:
Fixes #2922.

- Adds all missing speaker images
- Adds keynote (just Dion, we can display all speakers but it would get cramped)
- Adds some missed paired-up talks
- Updates event times: each day is only 3 hours, and starts at different times to cater to different parts of the world


src/site/_data/event.js
```diff
@@ -21,9 +21,13 @@
 const eventData = [
   {
     title: 'Day 1',
-    when: '2020-06-30T16:00Z',
-    duration: 8,
+    when: '2020-06-30T16:00Z', // 9am PST (-7)
```
robdodson, line 24:
```suggestion
    when: '2020-06-30T16:00Z', // 9am PDT (-7)
```
nit, in case it's important.

PST is -8, PDT is -7.

We're in PDT now and we'll still be in PDT when the event happens. PDT ends Nov 1. I double checked the timestamp and 16:00 UTC == 09:00 am PDT 👍

src/site/_data/event.js
```diff
@@ -129,9 +137,13 @@ const eventData = [
   },
   {
     title: 'Day 3',
-    when: '2020-07-02T16:00Z',
-    duration: 8,
+    when: '2020-07-02T07:30Z', // 1pm IST (+5:30)
```
robdodson, line 140: double checked and india does not do daylight savings to IST is correct

PR #2933, opened 2020-05-20
subscribe copy just for live page

### LINKED ISSUE #2901: web.dev/live subscribe component has incorrect copy
As we pull in the generic subscribe partial, it still says "Developer Newsletter".

### LINKED ISSUE #2903: web.dev/live subscribe link doesn't offset
When you click on "Get Updates", the title of the subscribe area is cut off because of our header. We should pad it out by ~64px.

samthor:
Fixes #2903

Fixes #2901

I don't believe we use the `id="subscribe"` on `<web-subscribe>` directly for any reason, so I've repurposed it for the copy page.

There's still an issue; the subscribe element repositions itself because the event table and its tabs size is unknown until the JS loads. I'll create this issue when this PR is merged.

### DIFF
--- a/src/site/content/en/live/index.njk
+++ b/src/site/content/en/live/index.njk
@@ -55,7 +55,7 @@ draft: true
 
   <hr />
 
-  <section class="w-event-section">
+  <section id="schedule" class="w-event-section">
 
     <div class="w-event-section__column-left">
       <h3>What to expect</h3>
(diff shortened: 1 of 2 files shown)


src/site/content/en/live/index.njk
```diff
@@ -55,7 +55,7 @@ draft: true
 
   <hr />
 
-  <section class="w-event-section">
+  <section id="schedule" class="w-event-section">
```
mathiasbynens, line 58: Browsing to https://web.dev/live/#schedule in a new tab results in:



Could we make it behave more like this?



(Asking because I'm using this for a demo and have to currently manually fix the scroll position)

samthor:
Yup, we can do that. Can you file a bug?

We're just doing the trick of making the element bigger on top to deal with
the header.

mathiasbynens:
@samthor Done: #2990

PR #2941, opened 2020-05-20
chore(ci): run lighthouse CI on master

patrickhulce:
Sorry for the double PR today but just updated the LHCI server and noticed there weren't any datapoints from master in the past 2 months. Double checked the config and looks like the move to GitHub actions disabled running on master. It's best to run on master so that individual PRs *delta* can be surfaced in LHCI.

- Runs LHCI on master commits as well to establish a baseline


robdodson:
rad thank you!

PR #2943, opened 2020-05-21
deals with fouc for web-tabs and web-event-time

### LINKED ISSUE #2910: web-tabs has noticeable FOUC
**Describe the bug**
The schedule tabs have a pretty big FOUC/CLS as their styles lazy load. We may need to give them some placeholder styles. It reminds me of [this twitter thread](https://twitter.com/rob_dodson/status/1259263207635595264).

samthor:
Fixes #2910.

This adds specialized unresolved CSS to deal with FOUC, and only shows one page of event contents by default.


samthor:
As an aside, I think this would be "nicer" if web-tabs was a pure CSS component which did the `<label>` + `<input type="radio">` hack, since we could move this all to light DOM and JS wouldn't be required to tab through them.

But I'm splitting hairs and our userbase without CE/SD support is so low (even once we remove polyfills).

samthor:
I don't know enough about our CSS to know whether this is the right place to put the unresolved CSS.

robdodson:
> As an aside, I think this would be "nicer" if web-tabs was a pure CSS component which did the <label> + <input type="radio"> hack, since we could move this all to light DOM and JS wouldn't be required to tab through them.

If I recall, that trick can have keyboard access issues. I believe it can be made to work but it may require some tabindexing.

PR #2945, opened 2020-05-21
Calendar options

### LINKED ISSUE #2902: web.dev/live Add To Calendar
The "Add To Calendar" button doesn't work yet. Additionally, the design calls for individual per-day buttons, which aren't yet on the page.

I'm actually not an expert on this. I have a hunch that we need two buttons: one to add to Gcal, and one to download an ICS file. I don't think there's a single-button solution which works for both.

samthor:
Fixes #2902 with UX and calendar changes.

This points the user to actually subscribe to a calendar, one that's owned by a number of Googlers. It uses its public ICS link to subscribe for iCal/Outlook and also to Google Calendar. (The previous approach by @devnook used a calendar as a _template_, so if we change it, users won't see updates.)

Confusingly, Google Calendar recognizes that it's just a real Google Calendar, and subscribes to it directly—ignoring the ICS file itself. The primary downside to this is that users can see individual events are created by me, see:

Screen Shot 2020-05-21 at 13 28 30

(This is on a test account that has a random test name).

I personally don't care about this, but the workaround here is literally just to rehost the ICS file somewhere else. This could even happen at build (we sync/fetch the real ICS). Thoughts?


devnook:
The previous calendar linked to a calendar (and event) dedicated to WebDay created to Harleen. I think it's a good idea to have her as a calendar owner, just in case? It is better to keep ical file on calendar.google.com, as it is easy to refresh it without code changes then.

samthor:
I shared the new calendar with Harleen too.

src/site/content/en/live/index.njk
```diff
@@ -42,10 +42,23 @@ draft: true
...
-          <a class="w-button w-button--with-icon w-button--secondary gc-analytics-event" data-category="web.dev" data-label="live, add to calendar" data-action="click" data-icon="calendar_today" target="_blank"
-            href="https://calendar.google.com/event?action=TEMPLATE&tmeid=MzVwaDlxYWllMWZuZjlicjhpcGc4Y3NjOWMgZ29vZ2xlLmNvbV9rMGM1a2tzbGg1ZjlrdDU3bTU5bGhqYjRsZ0Bn&tmsrc=google.com_k0c5kkslh5f9kt57m59lhjb4lg%40group.calendar.google.com">
-            Add to calendar
-          </a>
+          <div class="w-event-calendar">
+            <div class="w-event-calendar__note">
+              <span>Add to calendar</span>
+            </div>
+            <a class="w-button w-button--secondary w-button--small gc-analytics-event" data-category="web.dev" data-label="live, add to Google Calendar" data-action="click" target="_blank"
+              href="https://calendar.google.com/calendar/r?cid=webcal://calendar.google.com/calendar/ical/c_ta9vmmo7jqm7j2bqgg3bsn47oc%40group.calendar.google.com/public/basic.ics">
```
devnook, line 50:
```suggestion
              href="https://calendar.google.com/calendar/r?cid=webcal://calendar.google.com/calendar/ical/google.com_k0c5kkslh5f9kt57m59lhjb4lg%40group.calendar.google.com/public/basic.ics">
```

src/site/content/en/live/index.njk
```diff
@@ -42,10 +42,23 @@ draft: true
...
+              Google
+            </a>
+            <a class="w-button w-button--secondary w-button--small gc-analytics-event" data-category="web.dev" data-label="live, add to iCal" data-action="click" target="_blank"
+              href="webcal://calendar.google.com/calendar/ical/c_ta9vmmo7jqm7j2bqgg3bsn47oc%40group.calendar.google.com/public/basic.ics">
```
devnook, line 54:
```suggestion
              href="webcal://calendar.google.com/calendar/ical/google.com_k0c5kkslh5f9kt57m59lhjb4lg%40group.calendar.google.com/public/basic.ics">
```

src/site/content/en/live/index.njk
```diff
@@ -42,10 +42,23 @@ draft: true
...
+              iCal
+            </a>
+            <a class="w-button w-button--secondary w-button--small gc-analytics-event" data-category="web.dev" data-label="live, add to Outlook" data-action="click" target="_blank"
+              href="webcal://calendar.google.com/calendar/ical/c_ta9vmmo7jqm7j2bqgg3bsn47oc%40group.calendar.google.com/public/basic.ics">
```
devnook, line 58:
```suggestion
              href="webcal://calendar.google.com/calendar/ical/google.com_k0c5kkslh5f9kt57m59lhjb4lg%40group.calendar.google.com/public/basic.ics">
```

samthor:
We're going to leave as-is for the calendar I added.

PR #2951, opened 2020-05-21
Remove Francois talk

devnook:


samthor:
FWIW Percy hasn't caught up with the Community Guidelines addition yet, hence why it thinks there's changes.

PR #2965, opened 2020-05-21
fix(authors): authors without posts will use their twitter profile instead of their author page used as their link, fixes #2904

### LINKED ISSUE #2904: web.dev/live speakers without articles go to 404
We're adding more speakers (see #), but if those authors haven't written articles, their individual pages (e.g. https://web.dev/author/sebabenz) aren't generated and they 404.

We can either solve this by always generating the page (easier), or perhaps disabling the link if they've not written anything (harder).

MichaelSolati:
Fixes #2904

- Set the `authors` collection to include all contributors, however if they have no posts their `href` field is set to their Twitter profile.
- Created a new collection called `authorsWithPosts` which does what the old `authors` collection did, returns an array of all authors with their posts (only if they have posts). I also moved the filter authors for percy to here too.
- Added `jsonfig.json` file so that JSDoc `@typedef`s are recognized across the app.
- Added `src/@typedefs.js` to hold some custom data types for the site, such as `Author` objects and `Paginated` objects. I also updated the typing of functions to use these new types, rather than the more verbose object typing.
- Refactored the `AuthorsDate` component/shortcode to use the `authors` collection if it exists to get the contributor's data. Also instead for the link to the author's page we use the author object's `href` field (which will link to Twitter if the author has no posts).
- Ensure that the `authors` collection is passed into the `EventTable`, which then passes it to `AuthorsDate`.
- Update the `authors/feed.njk` to use the `authorsWithPosts` collection.


robdodson:
Discussed a bit in the team chat. I feel like AuthorDate is doing a bit too much work right now by reaching into the global scope and that's made it so we need to do a pretty big change to override the href that it uses.

Probably a nicer option would be to refactor AuthorDate to be a bit "dumber" and only render the href, image, and text it gets handed.

Because we need this for Tuesday's launch I think we should move ahead with this PR but might be good to do a follow up refactor to see if we can pair back some of the complexity.

PR #2967, opened 2020-05-21
system-ui article thumbnail patch

argyleink:
fixes text content being cutoff barely

image


robdodson:
looks better! thank you!

PR #2973, opened 2020-05-22
refactor(collections): remove collections and move pagination into `before` callback of 11ty pagination

MichaelSolati:
Instead of having a collection for all the variations of paginated authors, tags, etc... This will ensure we only have one collection for each category, then in the templated that use the paginated collections I've implemented a `before` hook in 11ty's pagination to paginate the data for the template.

It (possibly) speeds up the build by a fraction of a second

Before:
Screen Shot 2020-05-21 at 4 56 16 PM

After:
Screen Shot 2020-05-21 at 4 55 32 PM

One other thing done though is I set the `ELEVENTY_ENV` to `dev` when we run `npm run dev` that way we can ignore the author and tag rss feeds with the `before` hook, which definitely speeds up the build time.

Screen Shot 2020-05-21 at 5 01 11 PM


MichaelSolati:
PTAL @robdodson

src/site/_utils/_hooks/authors.js
```diff
@@ -0,0 +1,85 @@
...
+const addPagination = require('../add-pagination');
+
+/**
+ * @param {Author[]} authors
+ * @return {Author[]}
+ */
+const authorsFeed = (authors) => {
+  let filteredAuthors = [];
+
+  if (process.env.ELEVENTY_ENV !== 'dev') {
```
robdodson, line 26: Whenever possible I think doing an early return can be better than doing a not condition. It's a bit less cognitive overhead.  So another way to write this would be:

```js
let filteredAuthors = [];
if (process.env.ELEVENTY_ENV === dev) {
  return filteredAuthors;
}

filteredAuthors = authors.reduce(...);
return filteredAuthors.
```

src/site/_utils/_hooks/authors.js
```diff
@@ -0,0 +1,85 @@
...
+    filteredAuthors = authors.reduce((accumulator, author) => {
```
robdodson, line 27: I might be missing something but could this be written as a loop instead?

```js
for (author of authors) {
  if (author.elements.length > 0) {
    filteredAuthors.push(author);
  }
}
```

Might be easier to grok than a reduce.

src/site/_utils/_hooks/authors.js
```diff
@@ -0,0 +1,85 @@
...
+  ];
+  const authorsWithPosts = authors.reduce((accumulator, author) => {
+    if (process.env.PERCY && testAuthors.includes(author.key)) {
+      accumulator.push(author);
+    } else if (!process.env.PERCY && author.elements.length > 0) {
+      accumulator.push(author);
+    }
+
+    return accumulator;
+  }, []);
```
robdodson, line 60: Rather than combine the logic in one reduce() here, it might be easier to split it into separate branches depending on if we're in percy or not.

```
let authorsWithPosts;
if (process.env.Percy) {
  // map() or for/of
} else {
  // map() or for/of
}
```

src/site/_utils/_hooks/tags.js
```diff
@@ -0,0 +1,85 @@
...
+
+  return filteredTags;
+};
+
+/**
+ * @param {any[]} tags
+ * @return {Paginated[]}
+ */
+const tagsIndex = (tags) => {
+  const testTags = [
```
robdodson, line 44: with the exception of the `testTags` and `testAuthors` arrays, are tags.js and authors.js mostly the same code? If so, perhaps we can combine them?

PR #2974, opened 2020-05-22
Add missing body argument to banner.

robdodson:

- Adds missing body argument to shortcode. This just makes the font and padding match other stuff in the body. I'm making some changes to the base banner class to be fixed position but banner--body will still be statically positioned. Wanted to add this argument so we don't break this article.


kaycebasques:
Oh I thought I had changed that to an aside. Did you change it back to a banner @argyleink?

argyleink:
we had 2 of the same kind of asides next to each other, i did change this one to be more of a mention yeah. here's what it was before:

image

i'm not too partial what it uses, just felt 2 blue star asides next to each other made neither of them important

PR #2994, opened 2020-05-25
Split ShareAction into a separate component for reusability

### LINKED ISSUE #2899: <web-actions> uses global styles and isn't extensible
Our `<web-actions>` is currently only used in the bottom-right of post pages (via `position: fixed`). It also uses global styles (`.w-action` and friends, which aren't used by anything else) and it seems like it just wasn't ported when we migrated all the styles to the elements themselves.

(As an aside, `<web-header>` also uses global styles, but it doesn't render itself and just enhances DOM already on the page.)

This issue/report/proposal replaces #2888 and the (already closed) #2881.

## Current

The element is currently default `display: inline`, but is effectively always `position: fixed` as its first child is `<div class="w-actions">` which is.

It provides a Share (to Twitter, or Web Share where available) button, which shares the current page with its current `document.title` as well as a link to Subscribe (you can choose which of these to support _already_ through an attribute `actions`). Finally, it allows an optional list of authors to share along with the title.

Only the "post" page currently allows Share (it also uses Subscribe), but various other pages allow just Subscribe.

The element is also responsive (showing Subscribe and Share _labels_ on wide screens). This makes it difficult to immediately embed it as a WC elsewhere where we don't know the size it's intended to be displayed at.

## Proposal

Having a single CTA element is quite useful and I'd argue for it to remain as a whole. While it's not common I can see a need for having both share and subscribe shown at once.

* Bring its styles inline and remove `position: fixed`
* Make the pages that currently contain `<web-actions>` wrap it in something like `<div class="w-footer-cta">` which places it in the bottom right
* Add some boolean attributes:
  * "small" or similar, which disables responsive design and always displays as icons without text
  * "extendedOptions" or similar, which adds Facebook/Instagram to the Share options (where Web Share isn't supported)

The element can then be used in both places: inline on web.dev/LIVE, as well as a global element on other pages.

cc @robdodson and @MichaelSolati for discussion

devnook:
Fixes #2899

It:
* Creates a new ShareAction component that is css independent
* Keeps w-actions in global styles, as a regular css component
* Subscribe action is effectively a regular link with no behaviour, so I jus moved it to a .njk template for easier reuse.

It didn't seem that there is a reason to keep Actions or Subscribe button as individual Web Components, but if you think otherwise I can be convinced too.


src/site/_includes/partials/post.njk
```diff
@@ -11,7 +11,18 @@
   {% endif %}
 
   {# Make sure Actions are first in the tab order. #}
-  <web-actions actions="share|subscribe" authors="{{ authors | expandContributors | join('|') }}"></web-actions>
+  <div class="w-actions">
+    <share-action authors="{{ authors | expandContributors | join('|') }}"
+      class="gc-analytics-event w-actions__fab w-actions__fab--share"
+      data-category="web.dev"
+      data-label="live, share"
```
samthor, line 18: tiny nit: this should have a different label ("live" was for the live page)

Actually, previously we had separate Analytics events for Web Share vs Twitter share. I think we can just have a default and let the WC modify its own property.

I'll push a PR to this branch to add this.

src/lib/components/ShareAction/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+  get shareUrl() {
+    return window.location.href;
+  }
+
+  get shareText() {
+    let authorText = '';
+    if (this.authors && this.authors.length) {
+      const authors = this.authors
+        .split(',')
```
samthor, line 54: this was previously split by |, so I don't think this splits properly now.

samthor:
@googlebot I consent.

PR #2995, opened 2020-05-25
Add Live link in the header

devnook:


samthor:
Screen Shot 2020-05-26 at 10 07 54

I think we end up going under the search box here since there's too much stuff—we need to either have an overflow, or more aggressively show the sidebar instead.

You can merge this on Tuesday your time as the launch time is then :)

PR #2996, opened 2020-05-26
ensure CSS padding at top for section

### LINKED ISSUE #2990: Improve behavior of hotlinks to web.dev LIVE schedule
Copied from https://github.com/GoogleChrome/web.dev/pull/2933#pullrequestreview-417585837:

Browsing to https://web.dev/live/#schedule in a new tab results in:

Could we make it behave more like this?

(Asking because I'm using this for a demo and have to currently manually fix the scroll position)

samthor:
Fixes #2990 and ensures that every # on the page links to correctly and puts its heading below the header with a sane amount of padding.

### DIFF
--- a/src/styles/pages/_live.scss
+++ b/src/styles/pages/_live.scss
@@ -86,14 +86,30 @@
   display: flex;
   flex-flow: row wrap;
   justify-content: center;
-  margin: 64px auto;
+  margin: (64px - $WEB_HEADER_HEIGHT - 48px) auto 64px;
   max-width: 960px;
+  padding-top: ($WEB_HEADER_HEIGHT + 48px);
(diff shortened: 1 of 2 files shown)


src/styles/pages/_live.scss
```diff
@@ -86,15 +86,31 @@
   display: flex;
   flex-flow: row wrap;
   justify-content: center;
-  margin: 64px auto;
+  margin: (64px - $WEB_HEADER_HEIGHT - 48px) auto 64px;
```
devnook, line 89: Where does this 48 come from?

samthor: It's just to push the content 48px past the header. I confess it's a magic number.

PR #3005, opened 2020-05-26
fix: remove bug notice from lighthouse offscreen images

patrickhulce:

- Removes the bug notice from the Lighthouse document on offscreen images audit now that it has been fixed.

Alternative: change the text to mention that this only applies to v5 and earlier.


kaycebasques:
@robdodson the Netlify preview barfed

robdodson:
I restarted the build. I think I can safely swap you in for sam on our netlify account since you'll likely be interacting with it more often.

PR #3013, opened 2020-05-26
fix(analytics): prevent false triggering of `subscribe` event

MichaelSolati:


src/lib/components/Subscribe/index.js
```diff
@@ -105,17 +105,19 @@ class Subscribe extends BaseElement {
...
     this.subscribeError.textContent = '';
     this.subscribeMessage.textContent = "Thank you! You're all signed up.";
     this.form.removeEventListener('submit', this.onSubmit);
     this.form.parentElement.removeChild(this.form);
-    trackEvent({
-      category: 'web.dev',
-      action: 'submit',
-      label: 'subscribe, newsletter',
-    });
+    if (isRobot === false) {
```
robdodson, line 114: Suggest redoing this with an early return.

```
if (isRobot) {
  return;
}

trackEvent({...})
```

PR #3026, opened 2020-05-27
Images transform

devnook:
An example of string replacement transform.

DO NOT MERGE


robdodson:
@devnook if it's ok i'm going to close this as I think we decided regex was too unsafe to use. thank you for putting together this experiment!

PR #3056, opened 2020-05-28
Add a language switch element

devnook:
I also simplify styles for the subscribe select - please let me know if it's ok (regarding the design)


src/lib/actions.js
```diff
@@ -182,3 +183,41 @@ export const setUserAcceptsCookies = store.action(() => {
...
+  },
+);
+
+export const setLocale = store.action((state, preferredLocale) => {
+  if (!isValidLocale(preferredLocale)) {
+    console.warn('Invalid locale: ', preferredLocale);
+    return;
+  }
+  const options = {
+    expires: 365,
```
robdodson, line 213: I'm not a cookie expert, but would we want this to be some date far in the future, so it essentially never expires?

src/lib/actions.js
```diff
@@ -182,3 +183,41 @@ export const setUserAcceptsCookies = store.action(() => {
...
+    samesite: 'strict',
+  };
+  cookies.set('preferred_lang', preferredLocale, options);
+  if (preferredLocale !== state.userPreferredLocale) {
+    location.pathname = preferredLocale;
```
robdodson, line 218: I think this will redirect the user back to `web.dev/${preferredLocale}`

Would you want to factor in their current location so it instead refreshes the article that they're on?

devnook:
PTAL

devnook:
now really PTAL

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @fileoverview A select for choosing language option.
+ */
+
+import {html} from 'lit-element';
+import {BaseStateElement} from '../BaseStateElement';
+import {store} from '../../store';
+import {setLanguage, checkUserPreferredLanguage} from '../../actions';
+
+const languageNames = {
```
robdodson, line 26: Maybe both this and the array of locales used by isValidLang should be defined in a constants file that both actions.js and LanguageSelect can pull from? Then if we add another language we only need to update in one place.

src/lib/actions.js
```diff
@@ -182,3 +183,42 @@ export const setUserAcceptsCookies = store.action(() => {
     // overwrite it.
   };
 });
+
+// TODO(devnook): Extract isSupportedLocale to a common format.
```
robdodson, line 187: In a follow up PR we should break up this actions.js file because it's getting quite large. I think it'd be better to have an actions folder and have individual files for the related actions.

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,85 @@
...
+class LanguageSelect extends BaseStateElement {
+  static get properties() {
+    return {
+      userPreferredLanguage: {type: String},
+    };
+  }
+
+  constructor() {
+    super();
+    this.onStateChanged = this.onStateChanged.bind(this);
```
robdodson, line 40: I think BaseStateElement will set this up for you so you can delete this line.

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+  connectedCallback() {
+    super.connectedCallback();
+    checkUserPreferredLanguage();
+  }
+
+  onStateChanged() {
+    const state = store.getState();
+    this.userPreferredLanguage = state.userPreferredLanguage;
+  }
```
robdodson, line 51:
```suggestion
  onStateChanged({userPreferredLanguage}) {
    this.userPreferredLanguage = userPreferredLanguage;
  }
```

unistore will pass the state object to onStateChanged so you can destructure it.

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,85 @@
...
+        `;
+  }
+
+  render() {
+    return html`
+      <div class="w-display-flex">
+        <label class="w-visually-hidden" for="preferred-language">
+          Choose language
+        </label>
+        <select id="preferred-language" @change=${this.onSelect}>
```
robdodson, line 77: nit: would you mind renaming onSelect to onChange? At a glance I thought it was trying to use the `select` event which is only for text elements.

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,85 @@
...
+          ${['en', 'pl'].map((language) => this.renderOption(language))}
```
robdodson, line 78: If we go with a language constants file then we should use it for this array as well.

src/lib/actions.js
devnook: Filed an issue: https://github.com/GoogleChrome/web.dev/issues/3103

devnook:
PTAL

devnook:
PTAL

src/lib/actions.js
```diff
@@ -182,3 +184,36 @@ export const setUserAcceptsCookies = store.action(() => {
     // overwrite it.
   };
 });
+
+function getCanonicalPath(path) {
+  const parts = path.split('/');
+  if (lang.isValidLanguage(parts[1])) {
```
robdodson, line 190: probably good to check that parts[1] exists just in case.

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+  onChange(e) {
+    setLanguage(e.target.value);
+  }
+
+  renderOption(language) {
+    return this.userPreferredLanguage === language
+      ? html`
+          <option value="${language}" selected>
+            ${lang.languageNames[language]} (${language})
```
robdodson, line 50: A couple thoughts:
- Should we check that `lang.languageNames[language]` exists?
- I think the design calls for `lang.languageNames[language]` to be in all uppercase

src/lib/utils/language.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * @fileoverview Tools for language selection and validation.
+ * TODO(devnook): Extract isSupportedLocale to a common format.
```
robdodson, line 3: Would you mind creating a bug to track this and linking the bug in this comment?

robdodson:
Just a few small remaining comments.

I was wondering... should we delete the language select element from the page templates? Right now it doesn't actually do anything (aside from setting the cookie). I wouldn't want folks to start filing bugs saying that our translation support was broken.

devnook:
I'd like to keep it to make testing in dev possible, but I commented out other languages than English.

PTAL

src/lib/components/LanguageSelect/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+  renderOption(language) {
+    const languageName = lang.languageNames[language];
+    if (!languageName) {
+      return '';
+    }
+    return this.userPreferredLanguage === language
+      ? html`
+          <option value="${language}" selected>
+            ${languageName.toUpperCase()} (${language})
```
robdodson, line 54: could you do:


```
let languageName = lang.languageNames[language];
if (!languageName) { return ''; }
languageName = languageName.toUpperCase();
```

PR #3080, opened 2020-05-29
Fix 404ing redirect with query strings.

### LINKED ISSUE #3079: Links from devtools with query strings are 404'ing.
**Describe the bug**
Attempt to visit: https://web.dev/largest-contentful-paint/?utm_source=devtools
I think this is breaking because our redirect handler is failing.

robdodson:
Fixes #3079

The redirect handler was trying to match the url's pathname against the pathnames in the redirects.yaml. However, because we were using `req.url` it was including the query string in the match and not properly finding the redirect.

- Use `req.path` instead of `req.url`. `req.path` is the pathname and will not include query strings.

I'll do a follow up PR to add the query params back to the redirected string. But I'd like to first fix the 404s.

### DIFF
--- a/redirect-handler.js
+++ b/redirect-handler.js
@@ -68,7 +68,7 @@ module.exports = function buildRedirectHandler(filename, code = 301) {
   groupMatcher = new RegExp(`^(${escaped.join('|')})`);
 
   return (req, res, next) => {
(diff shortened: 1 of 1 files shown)


samthor:
LGTM. Thanks.

hashseed:
Thanks for the quick fix!

PR #3054, opened 2020-05-28
Migrate a11y article

rachelandrew:
This is my migration of the content from [Accessible Styles](https://developers.google.com/web/fundamentals/accessibility/accessible-styles), into four new posts. Three are updated content from that article. I have also added an article about content re-ordering.

Images have been compressed with tinypng.

Related to #3054

PR #3084, opened 2020-05-30
Update event.js

### LINKED ISSUE #3040: web.dev/live Talk title change
Please change this the talk title 
FROM: Building for Modern Browsers and Progressively Enhancing Like It's 2003 
TO: Progressively Enhancing Like It's 2003

@tomayac FYI

Thanks.

digitalPlayer1125:
Fixes #3040

"Please change this the talk title
FROM: Building for Modern Browsers and Progressively Enhancing Like It's 2003
TO: Progressively Enhancing Like It's 2003"

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -3,6 +3,6 @@
   "//2": "Use git push --no-verify to bypass the pre-push hook",
   "hooks": {
     "pre-commit": "npm run precommit",
-    "pre-push": "npm run lint -- --fix"
+    "pre-push": "npm run lint"
   }
 }
--- a/.huskyrc
+++ b/.huskyrc
@@ -3,6 +3,6 @@
   "//2": "Use git push --no-verify to bypass the pre-push hook",
   "hooks": {
     "pre-commit": "npm run precommit",
-    "pre-push": "npm run lint"
+    "pre-push": "npm run lint -- --fix"
   }
 }
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -31,7 +31,7 @@ jobs:
         run: npm ci
 
       - name: Lint
-        run: npm run lint
+        run: npm run lint -- --fix
 
       - name: Test
         run: npm run test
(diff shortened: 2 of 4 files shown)


digitalPlayer1125:
@googlebot I signed it!

digitalPlayer1125:
As suggested "1 error and 0 warnings potentially fixable with the `--fix` option." , the command 
'npm run lint -- --fix' will do the job.

robdodson:
Hey @digitalPlayer1125 

Would you mind running the lint fix command and committing the changes to this pull request?

digitalPlayer1125:
Hey @robdodson 

Well, --fix option didn't work out as expected, but after reading the .eslintrc.js, I exactly knew what to do. The rule "indent: ['error', 2, " was the cause of the error, and removing a newline against the title of thomassteiner in the event.js file did the work!

.huskyrc
```diff
@@ -3,6 +3,6 @@
   "//2": "Use git push --no-verify to bypass the pre-push hook",
   "hooks": {
     "pre-commit": "npm run precommit",
-    "pre-push": "npm run lint"
+    "pre-push": "npm run lint -- --fix"
```
robdodson: We don't want to use fix here because it won't have any effect. The flow of events will look like this:

```
git commit
git push
pre-push hook runs and modifies a file
push executes
modified file remains, needing to be added to a separate commit.
```

robdodson:
@digitalPlayer1125 

Take a look at this comment from someone hitting a similar issue:
https://github.com/GoogleChrome/web.dev/pull/3054#discussion_r433374819

If you're using VS Code you can install the ESLint extension and set it to fixOnSave which will automatically fix any files that you change.

I went ahead and updated the text for this talk in a separate PR because I wanted to get it out quickly. But thank you for sending in a PR!

digitalPlayer1125:
Hey @robdodson ,
Definitely you are right about not having the --fix in .huskyrc and that's why I discarded all of my earlier commits and just removed the newline in the required file as been done in #3093 . The issue was because of the rule that is defined in the .eslintrc.js.

I might not be familiar with rules that might make this PR for not being eligible to merge, but really happy to be a part of the identifying the solution!

.huskyrc
digitalPlayer1125: Oh, right. 

I removed it in my next commit of .huskyrc though.
I will be more careful while I make another PR :D

PR #3085, opened 2020-05-30
Fix broken gulp image output.

robdodson:
I think as part of the i18n work we started outputting html/json into language directories, but we didn't update the gulpfile to move their image assets to the right location. So we were outputting:

```
dist/
  foo/
    a.jpg
  en/
    foo/
      index.html
      index.json
```

This didn't break the site because our express server will look in the top dir, and then the en dir for assets. But we really want the images to live with their html files.

- Update the gulpfile so images in dist/ end up in the same directory as their html file.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -69,7 +69,7 @@ gulp.task('copy-content-assets', () => {
           const parts = assetPath.dirname.split('/');
           // Landing pages should keep their assets.
           // e.g. en/vitals, en/about
-          if (parts.length === 2) {
+          if (parts.length <= 2) {
             return;
           }
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-const path = require('path');
 const gulp = require('gulp');
 const mozjpeg = require('imagemin-mozjpeg');
 const pngquant = require('imagemin-pngquant');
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -67,13 +66,30 @@ gulp.task('copy-content-assets', () => {
...
+          // rest of the site and they're allowed to nest.
+          const subdir = parts[1];
           const imagePassThroughs = ['handbook', 'newsletter'];
-          if (imagePassThroughs.includes(leftPart)) {
+          if (imagePassThroughs.includes(subdir)) {
             return;
           }
-          return (assetPath.dirname = path.basename(assetPath.dirname));
+
+          // Some assets are nested under directories which aren't part of
```
samthor, line 85: This feel fragile, but I suppose the entire way we do permanent URLs for posts is (lots of hardcoded 'handbook' and 'newsletter'). Just my 2c.

samthor: Actually, it's not hardcoded that much, but I still think it might not be immediately obvious.

gulpfile.js
```diff
@@ -67,13 +66,30 @@ gulp.task('copy-content-assets', () => {
       // they belong to.
       .pipe(
         rename((assetPath) => {
-          const leftPart = assetPath.dirname.split('/')[1];
-          // Let certain path's images to pass through.
+          const parts = assetPath.dirname.split('/');
+          // Landing pages should keep their assets.
+          // e.g. en/vitals, en/about
+          if (parts.length === 2) {
```
samthor, line 72: can you make this <= 2? I don't think any images will ever appear at the actual top-level but the rest of the method mostly works by chance (`subdir` will be `undefined`, `splice` will do nothing).

gulpfile.js
robdodson: 100% agree it's not good. In fact the fragility is evident because this used to work but then we broke it without really noticing.

I think we should flatten the directory structure so the output matches the src. Ewa's also working towards this for i18n reasons. However, I'm concerned about making that big of a change during the run up to web.dev/live because it'll mean needing to rebase all inflight content PRs and update what folders they target.

I'll put this on the agenda for our next eng sync for us to discuss. In the meantime I'd like to land this PR to at least fix what's currently broken.

PR #3090, opened 2020-06-01
pass-through query string on redirect

### LINKED ISSUE #3082: Redirects strip query strings.
Related to https://github.com/GoogleChrome/web.dev/pull/3080.
The fix for this will probably need to go inside the redirect handler (https://github.com/GoogleChrome/web.dev/blob/39afd8c1a30576ba06d1b438e196fb921252791a/redirect-handler.js#L70)

**To Reproduce**
1. Go to https://web.dev/largest-contentful-paint/?utm_source=foo
2. The query string is stripped after the redirect.

**Expected behavior**
The query string is preserved after the redirect.

**Screenshots**

samthor:
Fixes #3082.

### DIFF
--- a/redirect-handler.js
+++ b/redirect-handler.js
@@ -18,6 +18,9 @@ const yaml = require('js-yaml');
 const fs = require('fs');
 const escapeStringRegexp = require('escape-string-regexp');
 
+const baseUrlPrefix = 'https://web.dev/';
+const baseUrl = new URL(baseUrlPrefix);
+
 /**
  * Normalizes the passed URL to ensure that it ends with a simple trailing
  * slash. Removes "index.html" if found.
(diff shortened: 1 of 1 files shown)


redirect-handler.js
```diff
@@ -18,6 +18,16 @@ const yaml = require('js-yaml');
...
 
+const baseUrlPrefix = 'https://web.dev/';
+const baseUrl = new URL(baseUrlPrefix);
+
+if (!baseUrlPrefix.endsWith('/')) {
+  // Prevents abuse by having a domain "https://web.dev.another.owner.com".
+  throw new Error(
+    `warning: baseUrlPrefix must end with /, was: ${baseUrlPrefix}`,
+  );
+}
```
robdodson, line 29: do we need this if we defined the value for `baseUrlPrefix` ourselves up above?

redirect-handler.js
```diff
@@ -68,18 +78,36 @@ module.exports = function buildRedirectHandler(filename, code = 301) {
...
+    const requestParams = new URLSearchParams(req.url.substr(req.path.length));
+    for (const [key, value] of requestParams) {
+      target.searchParams.append(key, value);
     }
 
-    next();
+    // If the result URL starts with "https://web.dev/" (with trailing slash),
+    // then strip it and treat as local. This allows redirects to work in dev,
+    // where the domain is localhost or a staging URL.
+    let s = target.toString();
```
robdodson, line 107: Instead of doing this and hard coding the host on line 21, what if you did something like:

```
var url = require('url');

function baseUrl(req) {
  return url.format({
    protocol: req.protocol,
    host: req.get('host'),
  });
}
```

I think that should give you either localhost:8080 or web.dev.

robdodson: although I don't think that includes the port... so might need to grab that from app settings or something.

robdodson:
Because this code is so important I'd like to write some tests before landing this.

redirect-handler.js
samthor: I guess I didn't trust Express to give us the host properly. So the flip side is we don't do this at all, because I'm literally just trying to save a few bytes (rather than _always_ redirecting to https://web.dev/—I'm worrying too much).

redirect-handler.js
samthor: No, but it felt nice to ensure as an invariant in the code. I can flip it to a comment or assert it in a test.

redirect-handler.js
robdodson: > I guess I didn't trust Express to give us the host properly.

Why's that? Not trying to be contrarian; genuinely curious.

> So the flip side is we don't do this at all, because I'm literally just trying to save a few bytes (rather than always redirecting to https://web.dev/—I'm worrying too much).

I could see it being useful to do. I've definitely ended up in situations where I thought I was developing locally and clicked on a link to went to the live site and didn't realize it and wondered why my changes weren't taking affect.

samthor: I'm not worried about having a hard-coded "web.dev" string in the server, as we do it in a few places.

PR #3098, opened 2020-06-02
feat(log-errors): log errors from subscribe form

MichaelSolati:


src/lib/components/Subscribe/index.js
```diff
@@ -65,15 +65,23 @@ class Subscribe extends BaseElement {
     }).then((r) => r.json());
   }
 
-  onError(errors) {
+  /**
+   *
+   * @param {Error} [error]
+   * @param {boolean} [useDefault]
```
robdodson, line 71:
```suggestion
   * @param {boolean} [useDefault=false]
```

PR #3114, opened 2020-06-03
Fix broken link in LCP article

csabapalfi:

- Fix usage example link in LCP article


robdodson:
thank you!

PR #3126, opened 2020-06-04
Fix inconsistency in manifest file name

ericvera:
It seems the document has manifest.webmanifest everywhere but one reference to manifest.json.

Fixes #SOME_ISSUE_NUMBER

- Update file name to what seems to be the new file name to be used.


ericvera:
@googlebot I signed it!

PR #3127, opened 2020-06-04
feat(subscribe-analytics): add tracking for clicking subscribe/newsletter buttons, fixes #3106

### LINKED ISSUE #3106: Add GA event when someone clicks the subscribe button
We want to track how often folks click the Subscribe button in order to see how many subscriptions it's driving, versus folks linking directly to subscribe.

Let's add a GA event back to the subscribe button. The label should be different from 'subscribe'.

MichaelSolati:
Fixes #3106

@robdodson PTAL

### DIFF
--- a/src/site/content/en/live/index.njk
+++ b/src/site/content/en/live/index.njk
@@ -39,7 +39,7 @@ date: 2020-06-31
         </p>
 
         <div class="w-event-buttons">
-          <a class="w-button w-button--primary" href="#subscribe">
+          <a class="w-button w-button--primary gc-analytics-event" data-category="web.dev" data-label="view-subscribe" data-action="click" href="#subscribe">
             Get notified
           </a>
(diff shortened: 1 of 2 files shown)


src/site/content/en/live/index.njk
```diff
@@ -39,7 +39,7 @@ date: 2020-06-31
         </p>
 
         <div class="w-event-buttons">
-          <a class="w-button w-button--primary" href="#subscribe">
+          <a class="w-button w-button--primary gc-analytics-event" data-category="web.dev" data-label="view-subscribe" data-action="click" href="#subscribe">
```
robdodson, line 42: Let's go with a different name for this label to disambiguate it from the FAB. It's possible to sort by page type in analytics but I think a different label name will be quicker. Might call it 'scroll-to-subscribe' or something?

Issue #2720, opened 2020-04-28
Add typescript comments and enforce with tsc compiler.


robdodson:
I think if we were to use the tsc compiler with the `noEmit` and `allowJs` option, then we could have typescript check our code without needing to use actual .ts files. Paul Lewis shot a video demonstrating how to set it up: https://www.youtube.com/watch?v=YHvqbeh_n9U

@MichaelSolati and @devnook I think you've both mentioned being interested in exploring TypeScript this so I'm tentatively assigning this to y'all. I marked it as p3 (nice to have) but if y'all think it's important or start to see promising results we can always change the priority to a higher one.

robdodson:
keep alive

PR #3130, opened 2020-06-04
refactor(lint:js): use `gts` for JS linting

MichaelSolati:
This will be the first step to implement more TS based features (like `d.ts` files). This easily allows us to implement Google's TypeScript style guide. It also caught A LOT of weirdness in the code, so I fixed those up and just have questions on 2-3 bits. (I'll add those q's in a diff comment)

Part of #2720


src/lib/utils/generate-salt.js
```diff
@@ -1,15 +1,15 @@
...
  * @return {string} Id salt.
  */
 export const generateIdSalt = (idPrefix) => {
-  const salt = Math.random()
-    .toString(36)
-    .substr(2, 9);
+  const salt = Math.random().toString(36).substr(2, 9);
   return document.getElementById(idPrefix + salt)
-    ? BaseElement.generateIdSalt(idPrefix)
+    ? BaseElement.generateIdSalt(idPrefix) // eslint-disable-line
```
MichaelSolati, line 13: @robdodson @samthor Where does BaseElement come from?

MichaelSolati:
@robdodson @samthor PTAL

Also curious about the following comments...

src/lib/utils/router.js
```diff
@@ -99,7 +96,7 @@ export function listen(handler) {
   if (!handler) {
     throw new Error('need handler');
   }
-  listen = () => {
+  listen = () => { // eslint-disable-line
```
MichaelSolati, line 99: @robdodson @samthor How is listen being reassigned as a function?

src/site/_includes/components/Assessment.js
```diff
@@ -125,8 +123,8 @@ function responseTemplate(response) {
 
   if (
     response.columns &&
-    response.columns != true &&
-    response.columns != false
+    response.columns !== true &&
```
MichaelSolati, line 126: @robdodson @samthor `typeof response.columns !== 'boolean'` instead of checking if not true and not false?

src/lib/components/LighthouseScoresMeta/index.js
```diff
@@ -37,7 +37,7 @@ class LighthouseScoresMeta extends BaseElement {
           this.auditedOn,
         );
       } catch (err) {
-        auditedOnText = auditedOn.toLocaleString();
+        auditedOnText = this.auditedOn.toLocaleString();
```
MichaelSolati, line 40: @robdodson @samthor I'm assuming this is the correct change?

src/lib/utils/generate-salt.js
robdodson: src/lib/components/BaseElement but it looks like they didn't import it here. Perhaps we're getting lucky that it's already in the bundle so this isn't throwing an error.

src/lib/utils/router.js
robdodson: Sam can correct me. I think this might be some hecka clever JS. Basically it's reassigning the parent function to itself, then running to completion so its side effects still happen. After that if you call listen() ever again it will throw.

src/site/_includes/components/Assessment.js
robdodson: hm sure. I think this code was reviewed in a rush when Michael F. was leaving. Should have caught this.

src/lib/components/LighthouseScoresMeta/index.js
robdodson: seems like it.

src/lib/utils/generate-salt.js
MichaelSolati: Hrm, ok, but it doesn't seem like `generateIdSalt` is a method of BaseElement or LitElement.

.eleventy.js
```diff
@@ -84,7 +84,7 @@ const {
   serviceWorkerPartials,
 } = require(`./${transformsDir}/service-worker-partials`);
 
-module.exports = function(config) {
+module.exports = function (config) {
```
robdodson, line 87: the space in between function and parens will kill me... but i will learn to live with it.

src/lib/bootstrap.js
```diff
@@ -32,12 +36,13 @@ ga('send', 'pageview');
 // site code. This includes Shadow DOM.
 const browserSupport = 'noModule' in HTMLScriptElement.prototype;
 if (browserSupport) {
-  function prepare() {
+  const prepare = () => {
```
robdodson, line 39: just curious, is this gst style?

src/lib/components/BaseStateElement/index.js
```diff
@@ -22,9 +22,4 @@ export class BaseStateElement extends BaseElement {
     super.disconnectedCallback();
     store.unsubscribe(this.onStateChanged);
   }
-
-  /**
-   * @param {!Object<string, *>} state
-   */
-  onStateChanged(state) {}
```
robdodson, line 29: removing this seems like it would break line 12?

MichaelSolati: Missed that, I'll add it back in.

.eleventy.js
MichaelSolati: Lol, I'm sorry! It seemed inconsistent across the code so I didn't think anyone had a preference.

src/lib/bootstrap.js
MichaelSolati: It would want a function declared at the top level, rather than in an `if` statement.

src/lib/components/BaseStateElement/index.js
robdodson: it might be worth adding back the state argument and the comment. In case folks are referencing this class to see what it does.

src/lib/utils/router.js
robdodson: @MichaelSolati I think we may want to remove the TODO on line 90 and replace it with an explanation of how this thing works.

robdodson: Also Sam should double check that my assessment is correct :)

src/site/_includes/components/Author.js
```diff
@@ -23,14 +23,14 @@ module.exports = ({id, showSocialMedia = false, small = false}) => {
   if (!author) {
     throw new Error(
       `Can't create Author component for "${id}" without contributor ` +
-        `information. Please check '_data/contributors.js' and make sure the ` +
-        `author you provide is a key in this object.`,
+        "information. Please check '_data/contributors.js' and make sure the " +
+        'author you provide is a key in this object.',
     );
```
robdodson, line 28: this seems odd? it's using both double and single quotes

src/site/_includes/components/Author.js
```diff
@@ -23,14 +23,14 @@ module.exports = ({id, showSocialMedia = false, small = false}) => {
...
   }
 
   if (!author.name) {
     throw new Error(
-      `Can't create Author with missing author.name. author object: ` +
+      "Can't create Author with missing author.name. author object: " +
```
robdodson, line 33: why's it using double quotes here?

src/site/_includes/components/AuthorInfo.js
```diff
@@ -20,7 +20,7 @@ const {html} = require('common-tags');
 
 module.exports = ({author, id, showSocialMedia = false}) => {
   if (!author) {
-    throw new Error(`Can't generate AuthorInfo without author object`);
+    throw new Error("Can't generate AuthorInfo without author object");
```
robdodson, line 23: double quotes instead of single quotes

src/site/_includes/components/AuthorsDate.js
```diff
@@ -102,15 +102,15 @@ const renderAuthorsDate = (
     if (!info) {
       throw new Error(
         `Can't create Author component for "${id}" without contributor ` +
-          `information. Please check '_data/contributors.js' and make sure the ` +
-          `author you provide is a key in this object.`,
+          "information. Please check '_data/contributors.js' and make sure the " +
+          'author you provide is a key in this object.',
```
robdodson, line 106: same weirdness as above, it's using both double and single quotes.

src/site/_includes/components/AuthorsDate.js
```diff
@@ -102,15 +102,15 @@ const renderAuthorsDate = (
...
       );
     }
 
     if (!info.title) {
       throw new Error(
-        `Can't create Author with missing 'title'. author object: ` +
-          JSON.stringify(author),
+        "Can't create Author with missing 'title'. author object: " +
```
robdodson, line 112: double quotes

src/site/_includes/components/Blockquote.js
```diff
@@ -21,7 +21,7 @@ module.exports = (content, source) => {
   if (!source) {
     /* eslint-disable max-len */
     throw new Error(
-      `Can't create Blockquote component without a source. Did you forget to pass the source as a string?`,
+      "Can't create Blockquote component without a source. Did you forget to pass the source as a string?",
```
robdodson, line 24: double quote

src/site/_includes/components/Compare.js
```diff
@@ -22,7 +22,7 @@ module.exports = (content, type, labelOverride) => {
   if (!type) {
     /* eslint-disable max-len */
     throw new Error(
-      `Can't create Compare component without a type. Did you forget to pass the type as a string?`,
+      "Can't create Compare component without a type. Did you forget to pass the type as a string?",
```
robdodson, line 25: double quote

src/site/_includes/components/DetailsSummary.js
```diff
@@ -30,17 +30,15 @@ function renderPreview(contentArr) {
 
   const preview = contentArr.join('\n');
 
-  return html`
-    <p class="w-details__preview">${md.renderInline(preview)}</p>
-  `;
+  return html` <p class="w-details__preview">${md.renderInline(preview)}</p> `;
```
robdodson, line 33: is there supposed to be a space after html\`

src/site/_includes/components/Author.js
MichaelSolati: I believe this is happening because the string has single quotes that are not escaped. I can update the strings to use single quotes and escape the inner single quotes

src/lib/utils/generate-salt.js
robdodson: oh i see what you're saying.

yeah I don't know what's up with that... @devnook did you work on the generateIdSalt code?

src/site/_includes/components/Author.js
robdodson: ah I see.

Can we tell the linter to ignore this entire block and leave it with the backtick strings? We are doing string interpolation on the first line, so it seems like overkill to have to escape.

src/site/_includes/components/AuthorInfo.js
robdodson: oh i see why :P

src/site/_includes/components/Instruction.js
```diff
@@ -101,21 +101,15 @@ module.exports = (type, listStyle = 'ul') => {
       break;
 
     case 'disable-cache':
-      instruction = html`
-        ${bullet}Select the **Disable cache** checkbox.
-      `;
+      instruction = html` ${bullet}Select the **Disable cache** checkbox. `;
```
robdodson, line 104: probably not a huge deal but maybe we should delete the whitespace at the start and end of these strings.

src/site/_includes/components/Label.js
```diff
@@ -18,7 +18,5 @@ const {html} = require('common-tags');
 const md = require('markdown-it')();
 
 module.exports = (content) => {
-  return html`
-    <p class="w-label">${md.renderInline(content)}</p>
-  `;
+  return html` <p class="w-label">${md.renderInline(content)}</p> `;
```
robdodson, line 21: remove extra whitespace?

src/site/_includes/components/tags/Image.js
```diff
@@ -87,10 +87,10 @@ function renderImage({src, alt, maxWidth}, ctx) {
  * @return {function}
```
robdodson, line 87: we can delete this /tags/ directory. We ended up not using these.

test/integration/index.js
```diff
@@ -1,4 +1,4 @@
-describe('Integration tests', function() {
+describe('Integration tests', () => {
```
robdodson, line 1: everything in /test/ will need to use `function() {}` instead of arrow functions.
https://mochajs.org/#arrow-functions

tools/linting/bad-urls.js
```diff
@@ -34,11 +34,11 @@ function checkURL(tree, file) {
     if (!nodeUrl) {
       return;
     }
-    const parsed = url.parse(nodeUrl);
+    const parsed = new url.URL(nodeUrl, 'https://web.dev');
```
robdodson, line 37: ?

src/lib/utils/generate-salt.js
devnook: Nope, but here's the relevant commit: https://github.com/GoogleChrome/web.dev/pull/2229/commits/f714e3ea2b6bc9f1e0ac1117640d13f690d8a96a

MichaelSolati: Thank you @devnook I'm modifying that now

tools/linting/bad-urls.js
MichaelSolati: [`url.parse` was deprecated since Node v11.0.0](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost). (expand the history)

This effectively does the same thing, though since internal URLs don't have a base URL I'm adding one (it only get's used if the `nodeUrl` doesn't have a base URL).

src/lib/components/BaseStateElement/index.js
```diff
@@ -24,7 +24,7 @@ export class BaseStateElement extends BaseElement {
   }
 
   /**
-   * @param {!Object<string, *>} state
+   * Required for `constructor`.
    */
-  onStateChanged(state) {}
+  onStateChanged() {}
```
robdodson, line 29: I was wondering if we can add all of it back?

```
* @param {!Object<string, *>} state
*/
onStateChanged(state) {}
```

I've seen folks using this class not know that the state object gets passed to this function, so they end up calling `this.store.getState()` when they don't need to.

MichaelSolati: It doesn't do anything though. I'm not sure what it's supposed to look like or do 😬 (linter doesnt like unused arguments)

src/site/_data/contributors.js
```diff
@@ -585,7 +585,7 @@ const contributors = {
     github: 'jpmedley',
     twitter: 'medleyjp',
     description: {
-      en: "If an API's not documented it doesn't exist",
+      en: `If an API's not documented it doesn't exist`,
```
robdodson, line 588: I think using double quotes for strings that contain a single quote would be ok... does the linter grump about that? is that why they all needed to be switched to backticks?

src/site/_includes/components/Author.js
```diff
@@ -24,22 +24,23 @@ module.exports = ({id, showSocialMedia = false, small = false}) => {
     throw new Error(
       `Can't create Author component for "${id}" without contributor ` +
         `information. Please check '_data/contributors.js' and make sure the ` +
-        `author you provide is a key in this object.`,
+        'author you provide is a key in this object.',
```
robdodson, line 27: is it possible to use backticks here as well? seems like it might be good to be consistent with the two lines above in case someone goes in to change this block in the future

src/site/_includes/components/AuthorsDate.js
```diff
@@ -103,14 +103,15 @@ const renderAuthorsDate = (
       throw new Error(
         `Can't create Author component for "${id}" without contributor ` +
           `information. Please check '_data/contributors.js' and make sure the ` +
-          `author you provide is a key in this object.`,
+          'author you provide is a key in this object.',
```
robdodson, line 106: similar note as above, might be good to use backticks for all three lines

tools/linting/bad-urls.js
robdodson: is it possible to delete the require for `url` on line 18 and change this to:

```js
const parsed = new URL(nodeUrl, site.url);
```

I think you'll need to add a require for _date/site.js

src/lib/components/BaseStateElement/index.js
robdodson: responded in chat

PR #3135, opened 2020-06-04
Simplify ArticleNavigation

devnook:
I remove the "next" link on ArticleNavigation, as it is scarcely used & complicated the code quite a lot.

### DIFF
--- a/src/site/_includes/codelab.njk
+++ b/src/site/_includes/codelab.njk
@@ -36,14 +36,12 @@ show_banner: false
       {% if related_post %}
         {% ArticleNavigation {
           back: '/' + related_post,
-          backLabel: 'Return to article',
-          lang: lang
+          backLabel: 'Return to article'
         } %}
       {% else %}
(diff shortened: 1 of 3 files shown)


src/site/_includes/codelab.njk
```diff
@@ -36,14 +36,12 @@ show_banner: false
       {% if related_post %}
         {% ArticleNavigation {
           back: '/' + related_post,
-          backLabel: 'Return to article',
-          lang: lang
+          backLabel: 'Return to article'
         } %}
       {% else %}
         {% ArticleNavigation {
           back: '/' + path.slug,
```
robdodson, line 43: are you still thinking of getting rid of `path` in a future PR?

devnook: yes, but it still requires some untangling (e.g. postToCollections collection relies on it) and bigger files reshuffling, and you said you wanted small PRs:)

PR #3140, opened 2020-06-04
Quick and dirty player and chat embed.

### LINKED ISSUE #3073: [During event] Add side-by-side video and chat embeds.
image

We can use this placeholder premiere to test with:
https://www.youtube.com/watch?v=ukVbZUPkxx0&feature=youtu.be

robdodson:
Fixes #3073

Quick and dirty attempt to make sure these embeds still work as expected. Just used inline styles to get things looking approximately correct.

- Adds side by side video and chat embeds.

### DIFF
--- a/src/lib/components/LivestreamContainer/_styles.scss
+++ b/src/lib/components/LivestreamContainer/_styles.scss
@@ -1,0 +1,21 @@
+@import '../../../styles/settings/colors';
+
+web-livestream-container {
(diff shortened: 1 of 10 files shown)


robdodson:
note the chat will only work on localhost because that's what i set the embed_domain to.

samthor:
note: we should update this to reload itself if the sign-in state changes

PR #3147, opened 2020-06-05
Update 'why https'

dinakajoy:
I hope my pull request gets accepted. 💯

Just thought the need for website owners to be alerted that users can leave their websites simply because it's not secured needs to be well emphasized
- 
- 
-


dinakajoy:
@googlebot I signed it!

kaycebasques:
Thanks @dinakajoy let me discuss the change with the Chrome security team

kaycebasques:
A person on the Chrome security team is going to review the document in mid-July. They're very precise about how they word their documentation so we should leave the updates to them. Thank you for the contribution and I do think that your idea will get incorporated into the updated document.

Issue #2687, opened 2020-04-27
Refactor site structure to make /en/ directory canonical.


robdodson:
We currently put content in language directories, with `src/site/content/en/` being the home of all of our content. This matches how things were done on devsite, but makes things a little annoying with eleventy because it wants to add `/en/` onto the beginning of our urls and output all content to `/dist/en/`. To remedy this we've resorted to using hacks like [the `stripLanguage()` filter](https://github.com/GoogleChrome/web.dev/blob/master/src/site/_filters/strip-language.js).

A better approach may be to remove the `/en/` directory altogether and just put content in `src/site/content`. Then if we need to _translate_ content, we can move that content to a language subdirectory, `src/site/content/pl`. This works because we _want_ the urls for translated content to be prefixed with the language dir.

Assigning this to Ewa and myself to work on as it'll be a big refactor and probably need to take place over a few steps.

PR #3155, opened 2020-06-05
Refactoring path 11tydata files into _data

devnook:
Main goal: Make data in posts less dependent on directory structure. This is in preparation to flattening of the dir structure as per #2687 

This PR:
- Adds get-paths filter to dynamically find paths the post belongs to.
- Removes all occurrences of pathItem tag.
- Moves all path definitions to /src/_data.
- Adds postToPaths map, available in _data.
- Refactors postToCollections to use postToPaths.

It is a bit unfortunate that all these changes need to be made in one go (sorry for a big PR), but the build won't work if we approach it gradually (all pieces must fit).


src/site/_data/postToPaths.js
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const paths = require('./paths/all.js');
+
+const postToPathsMap = {};
+Object.keys(paths).forEach((path) => {
+  (paths[path].topics || []).forEach((topic) => {
+    (topic.pathItems || []).forEach((slug) => {
```
robdodson, line 22: Note we also support subtopics, which the progressive-web-apps path uses.
https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/progressive-web-apps/progressive-web-apps.11tydata.js#L35

src/site/_collections/post-to-collections.js
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
```
robdodson, line 14: I think this collection is only used by the SignPosts shortcode. Could it be removed and could the Signposts shortcode use the post-to-path filter to look up the slug it was given?

robdodson: Thinking about this a bit more, it might be good if we could give the SignPosts shortcode the title and href of the paths it needs to link to, instead of having it do the look up logic itself. My thinking is this might make it much more testable and reusable.

src/site/_includes/path.njk
```diff
@@ -2,7 +2,9 @@
 layout: layout
 ---
 
-{% set topics = path.topics | removeDrafts(lang) %}
+{% set path = paths[pathName] %}
+{% set topics = path.topics  %}
+{# {% set topics = path.topics | removeDrafts(lang) %} #}
```
robdodson, line 7: Should this comment be removed?
Also, I think the behavior is that it used to remove whole topic sections if every item in the section was a draft. Does this new code preserve that behavior?

src/site/_includes/path.njk
```diff
@@ -2,7 +2,9 @@
 layout: layout
 ---
 
-{% set topics = path.topics | removeDrafts(lang) %}
+{% set path = paths[pathName] %}
```
robdodson, line 5: Since all of the path data lives in _data/paths I think we could use [Eleventy's pagination](https://www.11ty.dev/docs/pagination/) to generate the path pages (/accessible/, /secure/, etc). Basically the same thing we do to generate the authors and the tags pages.

path.njk would set the pagination size to 1, and the data to _data/paths/all.
@MichaelSolati does that sound right to you?

MichaelSolati: Possibly, I don't think pagination works well with data files, but we could probably "hack" it and bypass the use of a collection with an 11ty.data.js file.

robdodson:
@devnook were you planning to add tests to this PR for the collections/filters that were changed/added?

src/site/_includes/path.njk
robdodson: hm that seems weird. It just can't recognize the data in scope or something?

robdodson: seems to be supported? https://www.11ty.dev/docs/pagination/#paginate-a-global-or-local-data-file

robdodson: threw together a quick example
https://glitch.com/edit/#!/pale-safe-chance

src/site/_collections/post-to-collections.js
devnook: Indeed, I did the same for ArticleNavigation lately. I think it is better to have separate "logic" shortcoder/macros and "UI" shortcodes just for visual output.

.eleventy.js
```diff
@@ -70,12 +70,12 @@ const gitlocalizeLink = require(`./${filtersDir}/gitlocalize-link`);
 const htmlDateString = require(`./${filtersDir}/html-date-string`);
 const md = require(`./${filtersDir}/md`);
 const pagedNavigation = require(`./${filtersDir}/paged-navigation`);
-const postsLighthouseJson = require(`./${filtersDir}/posts-lighthouse-json`);
```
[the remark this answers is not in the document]
devnook: Oh wow. It was in display: none, so I thought it is not used anymore!

src/site/_filters/find-by-url.js
```diff
@@ -68,12 +68,7 @@ const findByUrl = (url) => {
     throw new Error(`No collection has been memoized yet.`);
   }
 
-  const found = memo[url];
-  if (!found) {
```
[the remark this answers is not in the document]
devnook: Path definitions are now common to all languages. If a given post is not translated yet, the file will be missing in that language, which is fine, so I think we can be less strict about errors in case of missing files.

src/site/_data/postHost.js
```diff
@@ -1,47 +0,0 @@
-/*
```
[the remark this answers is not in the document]
devnook: Reverted

src/site/_filters/find-by-url.js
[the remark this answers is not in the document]
robdodson: ok. I'm curious if this might lead to other files requesting a post by url, getting back null, and trying to act on it. But maybe we can see if that happens and address it then.

src/site/_includes/path.njk
devnook: I don't really see what would be a benefit of using pagination instead of the data file directly?

devnook:
PTAL

I might add tests when I get internet back, but don't want to drag too long due to merge conflicts.

src/site/_includes/components/SignPosts.js
```diff
@@ -15,28 +15,28 @@
...
+module.exports = (slug, x) => {
+  const postPaths = postToPaths[slug];
+  if (!postPaths) {
+    return '';
+  }
+  const aTags = postPaths.map((pathName) => {
     return html`
-      <a class="w-post-signpost__link" href="${collection.href}"
-        >${collection.title}</a
+      <a class="w-post-signpost__link" href="${paths[pathName].href}"
```
robdodson, line 28: this doesn't seem to be working. try the signpost element on this page:
https://deploy-preview-3155--web-dev-staging.netlify.app/document-title/

both links go back to the page itself, not the paths

src/site/_includes/path.njk
robdodson: You would still use the data files, but you could delete files like: `content/en/accessible/index.jk` and instead have a single file that outputs all of our path files. It also means if we change the path template we only need to do it in one place, instead of having to update `content/en/accessible/index.jk`, `content/en/secure/index.jk`, etc.

robdodson:
Regarding this comment: https://github.com/GoogleChrome/web.dev/pull/3155#discussion_r437631550

I think it's fine if we want to defer that to another PR. I understand the urgency to land this because it causes conflicts with other files.

src/site/_includes/path.njk
devnook: I'd like to keep the .njk files to be honest. They do not include any markup, only frontmatter anyways. They will be useful to keep translated content for each path in language folders (e.g. a translated title and description).

src/site/_includes/components/SignPosts.js
devnook: Forgot to update property name (s/href/slug)

devnook:
PTAL

devnook:
PRs that depend on 11tydata file structure:
https://github.com/GoogleChrome/web.dev/pull/3176/files
https://github.com/GoogleChrome/web.dev/pull/3190
https://github.com/GoogleChrome/web.dev/pull/3162/files

I'm happy to help rebase them if needed.

rachelandrew:
The lazy-loading one is mine, based on the fact that I seem to manage to destroy git branches on a weekly basis some help rebasing it would be good!

PR #3171, opened 2020-06-09
When sw updates, don't refresh the page.

### LINKED ISSUE #3089: Service worker is causing double loads in Lighthouse
**Describe the bug**
Our LH and PSI perf scores seem much worse than they should be. Locally I get around 87 in LH and 61(!) in PSI.

image

---

Looking at the trace I noticed that the page seems to load twice, even if the url contains a trailing slash. Here's an example from `localhost:8080/`. Note the stats like LCP get logged twice:

image

After poking around a bit I think I've narrowed the issue down to our service worker. If I just comment out [the lines that register the SW](https://github.com/GoogleChrome/web.dev/blob/64bcffdf3e82ab1a2d24022e6815f279d0741fae/src/lib/app.js#L62-L66) my local LH score jumps to 100. I also noticed that [we don't register the sw on the staging site](https://github.com/GoogleChrome/web.dev/blob/64bcffdf3e82ab1a2d24022e6815f279d0741fae/src/lib/app.js#L70) and sure enough its PSI score is ~30 points higher:

image

@samthor I know you've got a lot going on with /live/ so if you'd like I could have the rest of the team dig into this to isolate and fix the issue.

robdodson:
Fixes #3089

- Feature detect and register sw after load event.
- Consolidate removeServiceWorker calls.
- Don't refresh the page if the service worker has been updated.

### DIFF
--- a/src/lib/app.js
+++ b/src/lib/app.js
@@ -58,12 +58,20 @@ function onGlobalStateChanged({isSignedIn, isPageLoading}) {
 store.subscribe(onGlobalStateChanged);
 onGlobalStateChanged(store.getState());
 
-// Ensure/update the Service Worker, or remove it if unsupported (this should
-// never happen here unless the valid domains change, but left in for safety).
-if (serviceWorkerIsSupported(window.location.hostname)) {
-  ensureServiceWorker();
+const featureDetectServiceWorker = () => {
+  // Ensure/update the Service Worker, or remove it if unsupported (this should
+  // never happen here unless the valid domains change, but left in for safety).
+  if (serviceWorkerIsSupported(window.location.hostname)) {
+    registerOrEnsureServiceWorker();
+  } else {
+    removeServiceWorkers();
+  }
+};
+
+if (document.readyState === 'complete') {
+  featureDetectServiceWorker();
 } else {
-  removeServiceWorkers();
+  window.addEventListener('load', featureDetectServiceWorker);
 }
 
 function serviceWorkerIsSupported(hostname) {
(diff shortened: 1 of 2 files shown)


src/lib/bootstrap.js
```diff
@@ -41,8 +40,4 @@ if (browserSupport) {
   s.type = 'module';
   s.src = '/' + entrypoint;
   document.head.append(s);
-} else {
-  // If we've transitioned into becoming an unsupported browser, then any
```
samthor, line 45: We still need this, for the reason listed in the comment.

src/lib/app.js
```diff
@@ -102,11 +110,9 @@ function ensureServiceWorker() {
     // This isn't the first install, but ensure some partials are up-to-date.
     ensurePartialCache();
 
-    // We claim active clients if the Service Worker's architecture rev changes. We can't
-    // reliably force a reload via the Client interface as it's unsupported in Safari.
-    navigator.serviceWorker.addEventListener('controllerchange', () => {
-      window.location.reload();
-    });
+    // TODO(robdodson): Offer a page reload to users if they want to get the
```
samthor, line 113: :'(

src/lib/bootstrap.js
robdodson: Are there browsers that support service workers but not modules? I figure older versions of the evergreen browsers did but how many versions back are they?

samthor:
I think you're correct.

But this conditional is here so we can easily add other checks, e.g., for
Shadow DOM or WCs. If we're removing those polyfills then we should be
adding the check for their native existence to this line.

I don't want to remove it, because basically we'll be adding it back again
real soon now, and we might otherwise forget :)

robdodson:
I noticed this line as well https://github.com/GoogleChrome/web.dev/blob/master/src/lib/sw.js#L57

I'll do some research to see if folks to clients.claim when they're showing the "A new version of this app is available" toast.

src/lib/bootstrap.js
samthor: no, but I think this comment and code path should still be included since as we up the requirement level (e.g. Shadow DOM), it's obvious that old browser should transition away from having a SW.

PR #3172, opened 2020-06-09
Force cookie banner off.

robdodson:
Forcing the cookie banner off to see if it affects our PSI paint scores.


robdodson:
doesn't seem to have a noticeable affect

Issue #3170, opened 2020-06-09
Workbox error: isSWEnv is not a function


robdodson:
**Describe the bug**
I noticed recently that if you build the site locally and open localhost:8080 you'll see this issue (ignore the Redux devtools warning):

![image](https://user-images.githubusercontent.com/1066253/84097512-b0e40980-a9b9-11ea-864b-6d3a95e87c47.png)

![image](https://user-images.githubusercontent.com/1066253/84097508-ade91900-a9b9-11ea-903c-980be0e7ade0.png)

@jeffposnick any ideas where we messed up? 😅

jeffposnick:
Workbox 5.1.3 removed that code: https://github.com/GoogleChrome/workbox/releases/tag/v5.1.3

Your module dependencies are a mix of 5.1.2 and 5.1.3—that's causing the issue. I can file a PR to update everything to 5.1.3.

PR #3173, opened 2020-06-09
Move all of Workbox to 5.1.3

jeffposnick:
Fixes https://github.com/GoogleChrome/web.dev/issues/3170

This brings all of the Workbox dependencies up to the same release. (They all depend on `workbox-core`,  so having `workbox-core` mismatched is particularly problematic.)


robdodson:
thank you!

PR #3184, opened 2020-06-10
[Update] Mention `minsize` value and point to full config in granular chunking article

housseindjirdeh:

- Mention `minsize` value and point to full config in granular chunking article

CC @spanicker


kaycebasques:
Not sure if the CC on @spanicker is just an FYI or a request for a review so I'll just wait a couple days

PR #3196, opened 2020-06-11
schedule modal

samthor:
Creates a modal for schedule content. Positions this on the correct row. Works reasonably well on mobile even without CSS changes.

Screen Shot 2020-06-13 at 13 08 24
Screen Shot 2020-06-13 at 13 08 27

I've made changes based on feedback. You can now click whole row to select and I've addressed some accessibility feedback.

This is ready for review _but_ I've not imported everyone's abstracts. The original design calls for a single line inline with each session, but since most abstracts are at most one line _anyway_, I'm less inclined to show it there.

If you're happy I'll import everyone's abstract before submit.

### DIFF
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview A schedule manager which opens and closes modals based on
+ * the current href. Looks in children for content.
+ *
+ * This does not inherit from BaseStateElement as it is not a LitElement.
+ */
+
+class EventSchedule extends HTMLElement {
+  constructor() {
+    super();
+
+    this._positionElement = document.createElement('div');
+    this._positionElement.className = 'w-event-modal-position';
+    this._positionElement.textContent = 'Hello';
+    this._positionElement.tabIndex = -1;
+
+    this._positionElement.addEventListener(
+      'blur',
+      () => {
+        console.info('active element', document.activeElement);
+
+        window.setTimeout(() => {
+          console.debug(
+            'after a frame, hash=',
+            window.location.hash,
+            'session=',
+            this._currentSession,
+            'active=',
+            this.activeElement,
+          );
+          if (!this._currentSession) {
+            return;
+          }
+
+          const url = window.location.pathname + window.location.search;
+          window.history.pushState(null, null, url);
+          this._updateHash();
+        }, 0);
+      },
+      {passive: true, capture: true},
+    );
+
+    this._currentSession = null;
+    this._updateHash = this._updateHash.bind(this);
+    this._updatePosition = this._updatePosition.bind(this);
+  }
+
+  _updateHash() {
+    const id = this.isConnected ? window.location.hash.substr(1) : '';
+    const session =
+      (id && this.querySelector(`[data-session-id="${id}"]`)) || null;
+
+    if (session === this._currentSession) {
+      return;
+    }
+
+    this._currentSession = session;
+    if (!session) {
+      this._positionElement.remove();
+      return;
+    }
+
+    this._updatePosition();
+    this._positionElement.focus();
+    this._positionElement.scrollIntoView(true);
+  }
+
+  _updatePosition() {
+    const bounds = this._currentSession.getBoundingClientRect();
+    this._positionElement.style.top = window.scrollY + bounds.top + 'px';
+    this._positionElement.style.left = window.scrollX + bounds.left + 'px';
+    this._positionElement.style.width = bounds.width + 'px';
+    this.append(this._positionElement);
+  }
+
+  connectedCallback() {
+    window.addEventListener('hashchange', this._updateHash);
+    window.addEventListener('resize', this._updatePosition);
+    this._updateHash();
+  }
+
+  disconnectedCallback() {
+    window.removeEventListener('hashchange', this._updateHash);
+    window.removeEventListener('resize', this._updatePosition);
+    this._updateHash();
+  }
+}
+
+customElements.define('web-event-schedule', EventSchedule);
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -27,9 +27,6 @@ class EventSchedule extends HTMLElement {
   constructor() {
     super();
 
-    this._positionElement = document.createElement('div');
-    this._positionElement.className = 'w-event-modal-position';
-
     this._modalElement = document.createElement('web-event-schedule-modal');
     this._modalElement.className = 'web-modal';
     this._modalElement.open = false;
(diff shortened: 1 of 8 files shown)


src/site/_includes/components/EventTable.js
```diff
@@ -41,17 +42,34 @@ module.exports = (event, authorsCollection) => {
...
   const renderSession = ({speaker, title}) => {
     // Always pass an Array of author IDs.
     const authors = typeof speaker === 'string' ? [speaker] : speaker;
 
+    const id = slugForTitle(title);
+
     return html`
-      <tr>
-        <td class="w-event-schedule__speaker">
+      <div class="w-event-schedule__row" data-session-id=${id}>
```
robdodson, line 65: I think we may want the entire row to be clickable to open the modal.

src/site/_includes/components/EventTable.js
```diff
@@ -41,17 +42,34 @@ module.exports = (event, authorsCollection) => {
...
+        <div class="w-event-schedule__cell w-event-schedule__speaker">
           ${AuthorsDate({authors}, authorsCollection)}
-        </td>
-        <td class="w-event-schedule__session">${title}</td>
-      </tr>
+        </div>
+        <div class="w-event-schedule__cell w-event-schedule__session">
+          <a class="w-event-schedule__open" href="#${id}">${title}</a>
```
robdodson, line 70: I dont' think we want the default blue styling for these links. I think the links should be black and hovering anywhere in the row should trigger the hover effect.

src/site/_includes/components/EventTable.js
samthor: The challenge with this is that the author's heads and links are also link targets.

The other option is that we only make them clickable from within the modal. WDYT?

src/lib/components/EventSchedule/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+    this._modalElement = document.createElement('web-event-schedule-modal');
+    this._modalElement.className = 'web-modal';
+    this._modalElement.open = false;
+    this._modalElement.addEventListener('close-modal', () => {
+      if (!window.location.hash.substr(1)) {
+        return; // we closed ourselves, so got an event that can be ignored
+      }
+      const url = window.location.pathname + window.location.search;
+      window.history.pushState(null, null, url);
```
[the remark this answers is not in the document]
samthor: Changed to `replaceState` and added a click handler for the #-links which do the same.

src/lib/components/EventSchedule/index.js
```diff
@@ -0,0 +1,101 @@
...
+    this._positionElement.style.width = bounds.width + 'px';
+
+    document.body.append(this._positionElement);
+
+    // TODO(samthor): If we resize, we might have to scroll the modal into view
+    // again.
+  }
+
+  connectedCallback() {
+    window.addEventListener('hashchange', this._updateHash);
```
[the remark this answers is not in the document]
samthor: Fixed

samthor:
Ping, ready for review again.

src/lib/components/EventSchedule/index.js
```diff
@@ -0,0 +1,134 @@
...
+
+    this._modalElement = document.createElement('web-event-schedule-modal');
+    this._modalElement.className = 'web-modal';
+    this._modalElement.open = false;
+    this._modalElement.addEventListener('close-modal', () => {
+      if (!window.location.hash.substr(1)) {
+        return; // we closed ourselves, so got an event that can be ignored
+      }
+      const url = window.location.pathname + window.location.search;
+      window.history.replaceState(null, null, url);
```
robdodson, line 38: It seems like pressing the back button now doesn't do anything which feels a bit confusing.
I thought it might be possible to discard the hash changes entirely so pressing the back button took you to the previous page. If that's not possible then we should probably revert to the behavior you had before where the back button reopened the modals.

src/lib/components/EventSchedule/index.js
```diff
@@ -0,0 +1,134 @@
...
+    if (!session) {
+      this._modalElement.open = false;
+      this._modalElement.remove();
+      return;
+    }
+
+    // Clone the session node and pass it to our session. This is kinda gross but basically we use
+    // it as the canonical source of truth for the modal. We also have to remove all tabindex
+    // attributes as they may have been added by the inert polyfill.
+    const clone = session.cloneNode(true);
```
robdodson, line 76: I think this implementation is totally fine, but as I was working on the livestream player I realized I was hitting a similar problem. Basically we don't have an API to fetch data, and we're not bootstrapping the page with data. So we're doing workarounds.

For the measure page we serialize a data object and add it to the page. This is how the measure tool knows how to link an audit to a specific guide. I was curious if we should consider using that approach more often in the future.

src/lib/components/EventSchedule/index.js
```diff
@@ -0,0 +1,134 @@
...
+      ? this._elementForHash(window.location.hash)
+      : null;
+    if (session === this._currentSession) {
+      return;
+    }
+
+    this._currentSession = session;
+    if (!session) {
+      this._modalElement.open = false;
+      this._modalElement.remove();
```
robdodson, line 69: Pressing the close button on the modal causes us to go down this code path. Normally the modal [waits for animationEnd to restore focus to the triggering element](https://github.com/GoogleChrome/web.dev/blob/master/src/lib/components/BaseModalElement/index.js#L134). Since we remove it right away, the focus blurs and goes back to `body`. It's not a great experience for a screen reader because they'll have to navigate the page all over again.

Do we need to remove the modal element or could we just let it handle its own close behavior?

src/lib/components/EventScheduleModal/_styles.scss
```diff
@@ -0,0 +1,39 @@
...
+    color: $GREY_600;
+    font: inherit;
+    font-size: 14px;
+    margin: 8px 0;
+
+    &:first-of-type {
+      margin-top: 16px;
+    }
+  }
+}
```
robdodson, line 39: i thought the sass linter grumped if files didn't end in a new line

src/lib/components/EventScheduleModal/index.js
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+class EventScheduleModal extends BaseModalElement {
+  static get properties() {
+    return {
+      sessionRow: Element,
+      _sessionName: String,
+      _authorsPart: Element,
+      _abstractPart: Element,
+      _titleId: String,
```
robdodson, line 33: Should these be written as?:

```
_sessionName: { type: String }
```

https://lit-element.polymer-project.org/guide/properties#conversion-type

Does lit support the Element type? I thought it could only do Object
https://lit-element.polymer-project.org/guide/properties#conversion-type

src/styles/pages/_live.scss
```diff
@@ -207,6 +230,36 @@ table.w-event-schedule {
...
+      top: 0;
+      left: 0;
+      right: 0;
+      bottom: 0;
+      display: block;
+      position: absolute;
+      z-index: 0;
+    }
+
+    // When the link is hovered, highlight the whole row, even though the
```
robdodson, line 257: I think this is creating an effect where once you hover the author, the row hover goes away which creates a kind of flicker that we probably don't want.

Our blog post card elements retain their hover state even when you're hovering over sub items. I'm going to see if I can figure out how to do similar css with these rows and offer that as a suggestion.

robdodson: I pushed a CSS tweak

src/lib/components/EventScheduleModal/index.js
```diff
@@ -0,0 +1,92 @@
...
+    // Remove the link that was used to open us.
+    const link = row.querySelector('.w-event-schedule__open');
+    if (link) {
+      this._sessionName = link.textContent;
+    }
+    return true;
+  }
+
+  render() {
+    return html`
```
robdodson, line 73: Should have "About this session" in the top left corner.

image

src/site/_includes/components/EventTable.js
```diff
@@ -41,17 +42,48 @@ module.exports = (event, authorsCollection) => {
...
-        </td>
-        <td class="w-event-schedule__session">${title}</td>
-      </tr>
+        </div>
+        <div class="w-event-schedule__cell w-event-schedule__session">
+          <a class="w-event-schedule__open" href="#${id}">
+            <span>${title}</span>
+          </a>
+          <div class="w-event-schedule__abstract" hidden>
+            ${abstract.map((part) => html`<p>${part}</p>`)}
```
robdodson, line 83: I was curious if we should pull in markdown-it and run the abstract text through. That way folks could include links (Phil and Paul Lewis's abstract contains a link).

robdodson: On second thought, if folks want to add a link we can just have them include the html in their abstract. I'll update phil and paul's talk to do so.

robdodson:
added talk abstracts

robdodson:
playing around with the site on my phone. kinda feels like we could use a bit more top/bottom padding on rows otherwise the tap targets are pretty close together

src/lib/components/EventScheduleModal/index.js
samthor: Apparently I forgot how to write lit. Using `Element` looks pretty but I'm pretty sure triggers default behavior. Turns out the type only helps you when you're converting from property to attribute, and we're not doing this here, so...

I've fixed this up and changed everything to `Object`.

src/lib/components/EventScheduleModal/_styles.scss
samthor: It does, but there's no extra line? Newlines are hard.

src/lib/components/EventSchedule/index.js
samthor: I agree but also... I think at this point I'm happy for bespoke things for this event.

src/lib/components/EventSchedule/index.js
samthor: This was a bug. I think the router was stealing the behavior (I wasn't calling `stopPropagation`). It should be fixed now.

src/lib/components/EventSchedule/index.js
samthor: I fixed this. Basically we need to listen to `animationend` as you mentioned. I'm worried it's a bit flaky but eh.

src/lib/components/EventSchedule/index.js
samthor: Not a router issue—I just messed up some logic. Fixed now!

PR #3199, opened 2020-06-11
Updates codeowners for PWA 11tydata.js to add petele

petele:
Changes to the [`pwa-11tydata.js`](https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/progressive-web-apps/progressive-web-apps.11tydata.js) file requires approval from Rob or Sam (because it is a `js` file). 

If we're adding new content, it have to be reviewed and approved by Kayce. But it also needs approval from either Rob/Sam - because it touches a `.js` file. This extra approval feels unnecessary.

This PR adds @kaycebasques and me (the area/content owner) to the approval for the PWA `11tydata.js` file, so that either of us can approve any new content added to the PWA section.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -27,3 +27,5 @@ content/en/angular/ @mgechev @kaycebasques
 # Exceptions
 src/site/_data/contributors.js @kaycebasques @robdodson
 src/site/_data/postTags.js @kaycebasques @robdodson
+src/site/_data/postTags.js @kaycebasques @robdodson
+src/site/content/en/progressive-web-apps/progressive-web-apps.11tydata.js @petele @kaycebasques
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -27,3 +27,5 @@ content/en/angular/ @mgechev @kaycebasques
 # Exceptions
 src/site/_data/contributors.js @kaycebasques @robdodson
 src/site/_data/postTags.js @kaycebasques @robdodson
+src/site/_data/postTags.js @kaycebasques @robdodson
```
robdodson, line 30: it looks like postTags was duplicated

robdodson:
> This extra approval feels unnecessary.

Happy to have more folks help with review. Be mindful that these files do execute during our build so keep a watchful eye for third party contributions that look sketchy. @MichaelSolati is working on moving these kinds of config files to .json so we can be a bit less paranoid.

.github/CODEOWNERS
petele: Well, that's embarrassing. Fixed.

petele:
> moving these kinds of config files to .json so we can be a bit less paranoid.

Happy to remove this special case once that lands.

PR #3200, opened 2020-06-11
Replace tag "fast" with "performance"

devnook:
Replaces tag "fast" with "performance" to be compatible with postTags tags list.


src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -10,7 +10,9 @@ description: |
   techniques reduces payload sizes and in turn improves page load times.
 codelabs:
   - codelab-text-compression
-  - codelab-text-compression-brotli
+	- codelab-text-compression-brotli
```
robdodson, line 13: hm, the indentation seems a little off here?

robdodson: ah it looks like it's also causing the tests to fail

PR #3202, opened 2020-06-11
Reorders topics on home & learn page

petele:
@kaycebasques - we talked about moving PWA higher up in the topic listing page, it fell off my plate and I am only finally getting around to it now.

- Reorders the Topics on the Home and Learn page to put our more critical content higher up. Vitals and PWA are now the top two items, followed by the other content.

### DIFF
--- a/src/site/content/en/en.11tydata.js
+++ b/src/site/content/en/en.11tydata.js
@@ -11,12 +11,12 @@ const allPaths = require('../../_data/paths');
 
 module.exports = function () {
   const paths = [
-    allPaths['fast'],
+    allPaths['vitals'],
+    allPaths['progressive-web-apps'],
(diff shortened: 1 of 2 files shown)


kaycebasques:
For the record, this was originally my idea because our PWA content is so popular (not taking credit, but rather providing rationale).

@robdodson it seems like this interferes with Ewa's refactoring (#3155), though?

robdodson:
> @robdodson it seems like this interferes with Ewa's refactoring (#3155), though?

Hm yep. If it's ok I think I'd like to land Ewa's PR and rebase this one. I can land her PR right now if that's ok?

robdodson:
rebased. Should be able to land once the tests finish.

petele:
Thank you!

PR #3213, opened 2020-06-12
Fix for "How to define your install strategy" article. 

demianrenzulli:
In the current version of the article, in the "Using JavaScript APIs" section, instead of showing how these APIs can be used, it shows how to obtain these signals via client-hints, which is the topic that comes immediately after that.

This change aims to fix this.

cc // @kaycebasques @petele


kaycebasques:
@petele PTAL

PR #3223, opened 2020-06-13
performance-scoring: tweaks for LH v6

paulirish:
some updates.. primary goals:

* address that desktop scoring was introduced in v6
* explain that scores of 100 are not expected

also made changes:

* newer desmos for scorecurve (with p10)
* updated scorecalc link
* slow/moderate/fast -> poor/needsimprovement/good


kaycebasques:
@paulirish LGTM, please merge yourself if this is ready (the `DO NOT MERGE` label is applied so I don't know if you're waiting on other reviews)

PR #3227, opened 2020-06-15
Update WebOTP blog to clarify that "host" should be used

majido:
Per [spec](https://wicg.github.io/sms-one-time-codes/#parsing) the item after the '@' is the **host** part of the URL.

Origin includes scheme, and port as well which we don't want authors to include in the SMS as
there are automatically set to 'https' and 'null' by user agent to produce an origin given the host.


agektmr:
Good call. Thanks for the pull request!

PR #3236, opened 2020-06-15
Add overflowing thumbnail carousel.

### LINKED ISSUE #3075: [During event] Add video embed for previous day to schedule.
image

We will not be embedding youtube players for these. Instead we'll reuse the YT thumbnail (possibly adding the play button if it's needed) and these will just be links that go to the YT watch page. They should use target=_blank

robdodson:
Fixes #3075

- Adds `.w-event-carousel` for the overflow scrolling video thumbnails thinger

### DIFF
--- a/src/site/content/en/live/index.njk
+++ b/src/site/content/en/live/index.njk
@@ -84,7 +84,24 @@ date: 2020-06-31
 
     <div class="w-event-section__column-right w-event-section__column-right__schedule">
       {% if event.isDuringEvent or event.isPostEvent %}
-        {# Include carousel #}
+        <div class="w-event-carousel">
+          <a href="#" class="w-event-carousel__day">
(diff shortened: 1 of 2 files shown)


src/site/content/en/live/index.njk
```diff
@@ -83,6 +83,26 @@ date: 2020-06-31
     </div>
 
     <div class="w-event-section__column-right w-event-section__column-right__schedule">
+
+      <div class="w-event-carousel">
+        <a href="#" class="w-event-carousel__day">
+          <div class="w-event-carousel__thumbnail">
+            <img src="https://i1.ytimg.com/vi/T0fAznO1wA8/mqdefault.jpg" alt="" width="178" height="110">
```
samthor, line 90: Yes but obviously we can't launch with this image here :)

PR #3251, opened 2020-06-17
Add blurbs to schedule. Improve styling of live page.

robdodson:
- Loads of small changes to try and match the mocks a bit more on the live page and tweak the schedule layout so we can fit more text content on mobile.
- Adds talk `blurbs` to schedule. I did my best to wordsmith these based on the abstracts folks submitted.
- Small tweak to AuthorCard images to space them out a bit more.


src/styles/pages/_live.scss
```diff
@@ -207,7 +240,9 @@ web-tabs.w-event-tabs {
   }
 
   .w-event-schedule__speaker {
-    width: #{$WIDTH_SPEAKER};
+    // width: #{$WIDTH_SPEAKER};
```
samthor, line 243: remove?

samthor: We can also remove `$WIDTH_SPEAKER` generally since both speaker and text is now full-width.

robdodson: oh ok

PR #3255, opened 2020-06-17
Consolidate tags

### LINKED ISSUE #3254: Clean up unused tags
We use tags that are inconsistent across the site. We should clean them up and make sure they are compatible with the https://github.com/GoogleChrome/web.dev/blob/master/src/site/_data/postTags.js file.

devnook:
1. Remove tags that are not compatible with https://github.com/GoogleChrome/web.dev/blob/master/src/site/_data/postTags.js file.
2. Move valid tags from 11tydata files to posts' frontmatter.
3. Rename tag "secure" to "security" to make it valid.

Fixes #3254


src/site/_includes/layout.njk
```diff
@@ -1,5 +1,6 @@
 ---
 show_banner: true
+tags: []
```
robdodson, line 3: This removes prior tag inheritance, correct?

Do you need it if we're getting rid of directory data files?

robdodson:
@kaycebasques @jpmedley @mihajlija @rachelandrew 

Wanted to give y'all a heads up regarding this change.

Previously we relied on eleventy's data cascade to propagate `tags` down the directory structure. This is why putting a post inside of the `angular/` directory automatically gives it the `angular` tag.

This PR will remove that behavior because we're planning to get rid of those directories (they make localization really tricky because ultimately we remove the directories from our URLs and flatten everything).

This means for new content, you'll want to ensure the post itself has the correct tags. So if someone is writing a new angular post, and they add it to the /angular/ directory, they'll need to also include the `tags: - angular` bit in their post's YAML frontmatter.

src/site/_includes/layout.njk
devnook: Inheritance still works, just through frontmatter and not through data files. Layout.njk is a top-level template, so this is just a fallback tags value for posts that do not define any tags in their frontmatter.

kaycebasques:
I think @jpmedley is going to have a problem with step 1 because he has previously argued that it's important for tags to support synonyms. I personally have no issue with the proposed changes.

jpmedley:
@kaycebasques is correct. 

Pretend for a moment that a new page isn't generated for every single tag. An index has to get more granular as the information set it describes gets bigger. That means it will tend to grow slighter faster than what it describes. So if I have ten pages and only one of them describes widgets, I only need one tag. If I have a 1000 pages and 30 describe widgets, I need enough tags to save someone from having to review 30 pages to find what they're looking for. Full text search doesn't solve this if the word someone is looking for is overloaded. Nor does it solve it if the word someone is looking for isn't in the actual text. 

Unless you decided to consider another tag page design, there's a case to be made for keeping things under control. But please be careful about it and don't eliminate tags just to eliminate tags.

src/site/_includes/layout.njk
robdodson: Does removing this break anything? I'm not sure if we actually need a tags fallback (the site originally didn't have one) but I added one myself because I needed to reset inheritance at the directory data file level.

robdodson:
> Pretend for a moment that a new page isn't generated for every single tag.

I'm not sure if we're talking about the same thing so I want to make sure we're on the same page.

A new page is generated only for the tags that appear in [postTags.js](https://github.com/GoogleChrome/web.dev/blob/master/src/site/_data/postTags.js). If a tag does not exist in that file then:
- We never show it in the UI to the user
- We never generate a page for it.

So the tags that Ewa is proposing removing are presently completely ignored by the site.

But looking at the changed files in the PR, it looks like Ewa did not remove any tags. This PR only adds tags to posts that are missing them. So maybe we can remove Step 1 from the commit message.

> Unless you decided to consider another tag page design, there's a case to be made for keeping things under control. But please be careful about it and don't eliminate tags just to eliminate tags.

I don't suspect that we'll change the design of the tags implementation anytime soon. From my perspective it might be good to remove tags that don't appear in `postTags.js` because they give authors the false sense that those tags actually do something. [The linter in this PR](https://github.com/GoogleChrome/web.dev/pull/2758) does validate that the tags used on a post appear in postTags.js but we've put reviewing it on hold until after web.dev LIVE.

robdodson:
If it's ok with the group I'd like to move ahead with merging this PR. We need it for our i18n/l10n work and as noted above, the PR does not actually remove any tags.

PR #3256, opened 2020-06-17
Hashes

### LINKED ISSUE #3243: Live page is broken because of cached CSS in service worker
**Describe the bug**
Tried deploying #3196 but the service worker is serving cached CSS so the page renders broken if you've previously visited the site:

Screen Shot 2020-06-16 at 8 50 31 AM

samthor:
Fixes #3243. Although I suspect it has a bunch of issues so I'm marking it DO NOT MERGE :)

This does a couple of things.

1. Adds hashes to the CSS/JS files (as "?v=hash") based on Eleventy computed data
2. Reorders the build process a bit, and makes the App Shell template fetched and cached by the Service Worker (previously we inlined it as a variable)
3. Adds a "resource version", which is the hash of CSS/JS/template, and includes it in both:
  - all partials
  - the payload which includes the App Shell template
4. If we load a new page, which involves fetching the partial and hydrating the template, we check if there's a version mismatch: if so, _immediately_ refetch the template and fetch its associated assets

To be clear, I couldn't just do 1 and 2 on their own. The build process now looks like:

a. Rollup and SASS (can happen independently)
b. Eleventy (as it needs generated hashes)

Rollup depended on knowing about the partial, as it used Workbox's `getManifest` helper. So instead, we actually generate a manifest inside Eleventy, and include this in what is effectively the App Shell template (it's now included in a JSON file).

An important implication of this design is that we _do not_ include general resources and hashes in the SW (to pass to e.g., `precacheAndRoute`). My goal here is actually to make it so that CSS, JS and template changes, and misc 'top level assets', do **not** cause a change in the Service Worker. This means we'll have very few SW changes (which currently causes a site refresh) and more consistency about what's being served, because an updated partial forces an update.

Some TODOs:

- We only force a template update when the SW tries to build a page (on opening a full page), _not_ on internal navigation
  - This only effects a user who has the site open, we do a deploy, and then they navigate—we need to force a refresh here but it's less important
  - ... I still need to build this
  - ... This is as broken as it is now

PTAL @jeffposnick. I've left some comments for you, mostly around whether inserting directly into the cache makes sense, e.g. sw.js line 309.

### DIFF
--- a/src/site/_transforms/service-worker-partials/index.js
+++ b/src/site/_transforms/service-worker-partials/index.js
@@ -25,6 +25,15 @@ const fs = require('fs').promises;
 const path = require('path');
 const cheerio = require('cheerio');
 
+const computedData = require('../../_data/eleventyComputed');
+
+const revision = {
+  // Refers to the version of the site the partial was built at, and the build timestamp. We use
+  // this inside the Service Worker to determine whether any cached layout HTML is out-of-date.
+  resourcesVersion: computedData.resourcesVersion,
+  builtAt: computedData.builtAt,
+};
+
 const writePartial = async (to, raw) => {
   await fs.mkdir(path.dirname(to), {recursive: true});
   await fs.writeFile(to, JSON.stringify(raw));
@@ -38,13 +47,30 @@ const getPartial = (content) => {
     title: $('title').text(),
     rss: $('link[type="application/atom+xml"]').attr('href'),
     offline: Boolean($('meta[name="offline"]').attr('content')) || false,
+    ...revision,
   };
   return partial;
+};
+
+const writeTemplate = async (to, raw) => {
+  const o = {
+    template: raw,
+    ...revision,
+  };
+  await fs.writeFile(to, JSON.stringify(o));
 };
 
 const serviceWorkerPartials = async (content, outputPath) => {
   // Page has permalink set to false and will not be rendered.
   if (!outputPath) {
+    return content;
+  }
+
+  // Special-case the output partial. We need to include it as well as its revision information in
+  // a single JSON file.
+  if (outputPath === 'dist/sw-partial-layout.partial') {
+    // nb. This does not end with ".json" as we don't want it to be confused with an actual partial.
+    await writeTemplate('dist/template-json', content);
     return content;
   }
 
--- a/src/site/_transforms/service-worker-partials/index.js
+++ b/src/site/_transforms/service-worker-partials/index.js
@@ -26,6 +26,8 @@ const path = require('path');
 const cheerio = require('cheerio');
 
 const computedData = require('../../_data/eleventyComputed');
+const {getManifest} = require('workbox-build');
+const isProd = process.env.ELEVENTY_ENV === 'prod';
 
 const revision = {
   // Refers to the version of the site the partial was built at, and the build timestamp. We use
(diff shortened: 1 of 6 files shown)


jeffposnick:
Hey @samthor!

I am still digesting some of these changes. From checking out a local copy, building, and running it I feel like we might be entering "what benefit is this service worker offering" territory, where just using proper `Cache-Control` headers for subresources would end up offering a similar experience with less complexity. I get the impression that you're now fighting against the service worker lifecycle, instead of using it to your advantage when it comes to versioning.

The offline experience, even for pages that you've previously visited, seems unreliable in this implementation, since CSS and JS subresources aren't guaranteed to be present in the precache.

Do you all want to take a step back and chat about the bigger service worker picture to ensure that the benefits that they could offer—network-independent navigations, functional offline experiences, and atomic cache expiration—are achievable with your HTML architecture?

samthor:
> aren't guaranteed to be present in the precache.

That's not the case? I'm just updating dependent resources like CSS/JS
independently of the install event: we have a "mini install" when a new
version of the site's resources are seen. I am trying to work out if
Workbox has a helper which optionally fetches potentially updated resources
in this way (as precache() _sets up_ a fetch on 'install', and if that's
already happened..).

I honestly think the Workbox standard model doesn't fit well with SPAs
here. If we do a proper navigation and load a whole new page then we have a
simpler problem: the hashes in our CSS and JS deps deal with that. But I
don't see how we avoid this general problem of "partial is from newer site
build" without deviating from the model which y'all encourage.

I was chatting to Rob yesterday and told him what I really want is a way to
force the Service Worker to update inline before a request is serviced. The
point being that we're happy for a delay here if we're just going to dump
an invalid page out, and then the user sees a refresh when the Service
Worker reinstalls _anyway_. This PR solves that problem. If we don't fit
Workbox then that's fine but I don't understand what else I can do here.

jeffposnick:
> That's not the case? I'm just updating dependent resources like CSS/JS
> independently of the install event: we have a "mini install" when a new
> version of the site's resources are seen. I am trying to work out if
> Workbox has a helper which optionally fetches potentially updated resources
> in this way (as precache() _sets up_ a fetch on 'install', and if that's
> already happened..).

Based on doing an `npm run build && npm run start`, I'm not seeing caching of JS or CSS anymore. I guess it's the code at https://github.com/GoogleChrome/web.dev/pull/3256/files#diff-c42a7f340a989aa2a3926dce2cbc90efR295-R309 that is supposed to handle caching that, but if you start from a fresh Incognito window and visit `http://localhost:8080`, that code doesn't seem to be run. 

> I honestly think the Workbox standard model doesn't fit well with SPAs
> here. If we do a proper navigation and load a whole new page then we have a
> simpler problem: the hashes in our CSS and JS deps deal with that. But I
> don't see how we avoid this general problem of "partial is from newer site
> build" without deviating from the model which y'all encourage.

The [App Shell model](https://developers.google.com/web/fundamentals/architecture/app-shell) works well with aggressively cached  HTML that then gets populated via client-side code with the content. In that model, the shell HTML/JS/CSS are always in sync if they're all precached, but the downside is that they might be potentially stale when responding to a navigation, and it's possible to create new content that's incompatible with an old shell.

> I was chatting to Rob yesterday and told him what I really want is a way to
> force the Service Worker to update inline before a request is serviced. The
> point being that we're happy for a delay here if we're just going to dump
> an invalid page out, and then the user sees a refresh when the Service
> Worker reinstalls _anyway_. This PR solves that problem. If we don't fit
> Workbox then that's fine but I don't understand what else I can do here.

So my larger point is that if you need those level of guarantees around freshness, maybe using a service worker to respond to your navigations isn't the right choice. Assuming the code in this PR does what you intend, what is the ultimate benefit that you'll end up getting from using this service worker vs. just properly configuring HTTP `Cache-Control` headers and versioning your subresources?

The biggest performance benefit of using a service worker derives from responding to navigation requests without blocking on the network, with at least some shell content if not a full page. Accomplishing that with a content-driven, non-app site is tough, and when I've seen it [happen successfully](https://developers.google.com/web/updates/2018/05/beyond-spa), it comes from making deliberate choices in HTML architecture to be "service worker friendly" from the start. That's why it might be worthwhile to take a step back and have a larger discussion about the goals for the service worker.

src/site/_transforms/service-worker-partials/index.js
```diff
@@ -38,16 +49,72 @@ const getPartial = (content) => {
...
+    if (toplevelManifest.warnings.length) {
+      throw new Error(`toplevel manifest: ${toplevelManifest.warnings}`);
+    }
+    manifest.push(...toplevelManifest.manifestEntries);
+
+    // We need this manifest to be separate as we pretend it's rooted at the
+    // top-level, even though it comes from "dist/en".
+    const contentManifest = await getManifest({
+      globDirectory: 'dist/en',
+      globPatterns: ['offline/index.json'],
```
robdodson, line 88: is `offline/index.json` guaranteed to exist at this point?

samthor: No, I think it's probably unknown since we're relying on Eleventy's output order.

PR #3261, opened 2020-06-17
Remove stripLanguage filter

devnook:
It replaces the reference to page.url with page.fileSlug, to avoid the need to strip language from the url.


src/site/content/en/tags/feed.njk
```diff
@@ -32,9 +32,10 @@ pagination:
   <logo>{{ renderData.metadata.logo }}</logo>
   <subtitle>{{ renderData.metadata.subtitle }}</subtitle>
   {% for post in paged.elements -%}
-  {% set absolutePostUrl -%}{{ post.url | stripLanguage | url | absoluteUrl(renderData.metadata.url) }}{%- endset %}
+  {% set postUrl = '/' + post.fileSlug + '/' %}
+  {% set absolutePostUrl -%}{{ postUrl | url | absoluteUrl(renderData.metadata.url) }}{%- endset %}
   <entry>
-    <title>{{ post.data.title }}</title>
+    <title>xxx{{ post.data.title }}</title>
```
samthor, line 38: xxx left for testing?

src/site/content/en/feed.njk
```diff
@@ -29,7 +29,8 @@ permalink: "/{{lang}}/feed.xml"
   <icon>{{ metadata.icon }}</icon>
   <logo>{{ metadata.logo }}</logo>
   {% for post in collections.blogPosts -%}
-  {% set absolutePostUrl -%}{{ post.url | stripLanguage | url | absoluteUrl(metadata.url) }}{%- endset %}
+  {% set postUrl = '/' + post.fileSlug + '/' %}
+  {% set absolutePostUrl -%}{{ postUrl | url | absoluteUrl(metadata.url) }}{%- endset %}
```
samthor, line 33: it's not important but why doesn't this one reference `metadata`?

src/site/_collections/algolia-posts.js
```diff
@@ -72,7 +71,7 @@ module.exports = (collection) => {
       objectID: data.page.url + '#' + lang,
       lang,
       title: data.title,
-      url: stripLanguage(data.page.url),
+      url: `/${data.page.fileSlug}/`,
```
robdodson, line 74: omg `fileSlug`! so useful!

.eleventy.js
```diff
@@ -76,7 +76,6 @@ const prettyDate = require(`./${filtersDir}/pretty-date`);
 const removeDrafts = require(`./${filtersDir}/remove-drafts`);
 const strip = require(`./${filtersDir}/strip`);
 const stripBlog = require(`./${filtersDir}/strip-blog`);
-const stripLanguage = require(`./${filtersDir}/strip-language`);
```
robdodson, line 79: should the strip-language.js file be deleted as well?

src/site/content/en/authors/feed.njk
```diff
@@ -30,7 +30,8 @@ pagination:
   <logo>{{ renderData.metadata.logo }}</logo>
   <subtitle>{{ renderData.metadata.subtitle }}</subtitle>
   {% for post in paged.elements -%}
-  {% set absolutePostUrl -%}{{ post.url | stripLanguage | url | absoluteUrl(renderData.metadata.url) }}{%- endset %}
+  {% set postUrl = '/' + post.fileSlug + '/' %}
+  {% set absolutePostUrl -%}{{ postUrl | url | absoluteUrl(renderData.metadata.url) }}{%- endset %}
```
robdodson, line 34: Do we need [the url filter](https://www.11ty.dev/docs/filters/url/) here? It seems like it might be a no-op because you turned fileSlug into an absolute url in the step before.

Btw would you mind include a comment above that links to the absoluteUrl docs
https://www.11ty.dev/docs/plugins/rss/#supplies-the-following-nunjucks-filters

Just so folks know where that filter comes from.

src/site/content/en/feed.njk
robdodson: I think it may come from this: https://www.11ty.dev/docs/plugins/rss/#supplies-the-following-nunjucks-filters

nunjucks syntax is super gross, but I think to pass two arguments to a filter you have to call it as:

```
foo | bar(baz) // == bar(baz, foo)
```

robdodson: @devnook see note above about adding a comment (maybe at the top of this file?) to https://www.11ty.dev/docs/plugins/rss/#supplies-the-following-nunjucks-filters so folks know where these filters come from

devnook:
PTAL

PR #3267, opened 2020-06-18
feat(subscribe): add warning when javascript is disabled

MichaelSolati:
Some people may bet having a weird experience when JavaScript is disabled and they try to sign up for the newsletter/subscribe. This will give them a friendly warning that the experience may not be as expected.

### DIFF
--- a/src/site/_includes/partials/subscribe.njk
+++ b/src/site/_includes/partials/subscribe.njk
@@ -74,7 +74,7 @@
       </div>
     </form>
     <div class="w-subscribe__error w-text--center w-mt--sm">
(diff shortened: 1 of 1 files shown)


src/site/_includes/partials/subscribe.njk
```diff
@@ -73,6 +73,8 @@
         </button>
       </div>
     </form>
-    <div class="w-subscribe__error w-text--center"></div>
+    <div class="w-subscribe__error w-text--center w-mt--sm">
+      <noscript>It seems like you have JavaScript disabled. On form submission you will redirect you to a page that may not look too familiar, don't worry, just click back when you're done.</noscript>
```
robdodson, line 77: playing around with wording a bit

```suggestion
      <noscript>It seems like you have JavaScript disabled or it failed to load. After you submit the form, press the back button to return to the site.</noscript>
```

PR #3268, opened 2020-06-18
web-event-store for time management

samthor:
Adds `<web-event-store>` and a few things to help manage the 'active day' of the event.

- you can pass e.g. "?_now=wdl-day2" to demo the event at a specific time
- the schedule is made available in unistore (with a boolean `.complete` as to whether the day is over)
- the active day is available in unistore—it prefers the upcoming day, but will retain the previous day until an hour before the next day (in case the user is still using the embed)
- I made it so deep-linking to a session ensures that tab is open when you load the page
- but otherwise it focuses the correct tab on load (this is _also_ done by the server, but maybe it's out of whack or we don't deploy properly)

### DIFF
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -95,9 +95,9 @@ class EventSchedule extends HTMLElement {
 
     // If the user opens this page from externally on a specific session, make sure we're showing
     // the correct day of tab.
-    const index = this._tabsElement.indexOfTabParent(session);
+    const index = this._tabsElement.indexOfTabByChild(session);
     if (index !== -1) {
-      this._tabsElement.focusTab(index);
+      this._tabsElement.activeTab = index;
     }
 
     this._modalElement.sessionRow = clone;
@@ -187,7 +187,7 @@ class EventSchedule extends HTMLElement {
     // tabs, which is pretty safe, since it comes from the same source.
     // Don't change the tab for the user if a modal is already open.
     if (!this._modalElement.open && activeEventDay) {
-      this._tabsElement.focusTab(activeEventDay.index);
+      this._tabsElement.activeTab = activeEventDay.index;
     }
   }
 }
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -162,7 +162,7 @@ class EventSchedule extends HTMLElement {
       }
 
       store.subscribe(this.onStateChanged);
-      this.onStateChanged(store.getState());
+      this.onStateChanged(store.getState()); // subscribe doesn't trigger listener
       this.onHashChange();
     });
   }
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -15,6 +15,7 @@
  */
 
 import '../EventScheduleModal';
+import '../Tabs';
 
 import {store} from '../../store';
 
@@ -29,31 +30,35 @@ class EventSchedule extends HTMLElement {
   constructor() {
     super();
     this.onStateChanged = this.onStateChanged.bind(this);
+    this.onCloseModal = this.onCloseModal.bind(this);
+    this.onHashChange = this.onHashChange.bind(this);
+    this.onClick = this.onClick.bind(this);
+    this.onModalAnimationEnd = this.onModalAnimationEnd.bind(this);
 
     this._activeEventDay = null;
-    this._tabsElement = this.querySelector('web-tabs');
+    this._currentSession = null;
+    this._tabsElement = null;
 
+    // This just creates an element, we're not yet making it part of the DOM, so it's allowed here
+    // in the constructor.
     this._modalElement = document.createElement('web-event-schedule-modal');
     this._modalElement.className = 'web-modal';
     this._modalElement.open = false;
-    this._modalElement.addEventListener('close-modal', () => {
-      if (!window.location.hash.substr(1)) {
-        return; // we closed ourselves, so got an event that can be ignored
-      }
-      const url = window.location.pathname + window.location.search;
-      window.history.replaceState(null, null, url);
-      this.onHashChange();
-    });
+    this._modalElement.addEventListener('close-modal', this.onCloseModal);
 
-    this._currentSession = null;
-    this.onHashChange = this.onHashChange.bind(this);
-    this.onClick = this.onClick.bind(this);
-
-    this.onModalAnimationEnd = this.onModalAnimationEnd.bind(this);
     this._modalElement.addEventListener(
       'animationend',
       this.onModalAnimationEnd,
     );
+  }
+
+  onCloseModal() {
+    if (!window.location.hash.substr(1)) {
+      return; // we closed ourselves, so got an event that can be ignored
+    }
+    const url = window.location.pathname + window.location.search;
+    window.history.replaceState(null, null, url);
+    this.onHashChange();
   }
 
   _elementForHash(hash = window.location.hash) {
@@ -143,13 +148,17 @@ class EventSchedule extends HTMLElement {
   }
 
   connectedCallback() {
+    this._tabsElement = this.querySelector('web-tabs');
+    if (!this._tabsElement) {
+      throw new Error(`web-event-schedule expects web-tabs child element`);
+    }
+
     window.addEventListener('hashchange', this.onHashChange);
     this.addEventListener('click', this.onClick);
 
-    // nb. Gross, but we need to wait for <web-tabs> to be ready.
-    window.requestAnimationFrame(() => {
+    customElements.whenDefined('web-tabs').then(() => {
       if (!this.isConnected) {
-        return; // disconnected while we waited for rAF
+        return; // disconnected while we waited for web-tabs
       }
 
       store.subscribe(this.onStateChanged);
@@ -164,6 +173,8 @@ class EventSchedule extends HTMLElement {
     window.removeEventListener('hashchange', this.onHashChange);
     this.removeEventListener('click', this.onClick);
     this.onHashChange();
+
+    this._tabsElement = null;
   }
 
   onStateChanged({activeEventDay}) {
--- a/src/lib/components/EventSchedule/index.js
+++ b/src/lib/components/EventSchedule/index.js
@@ -16,6 +16,8 @@
 
 import '../EventScheduleModal';
 
+import {store} from '../../store';
+
 /**
  * @fileoverview A schedule manager which opens and closes modals based on
  * the current href. Looks in children for content.
@@ -26,6 +28,10 @@ import '../EventScheduleModal';
 class EventSchedule extends HTMLElement {
   constructor() {
     super();
+    this.onStateChanged = this.onStateChanged.bind(this);
+
+    this._activeEventDay = null;
+    this._tabsElement = this.querySelector('web-tabs');
 
     this._modalElement = document.createElement('web-event-schedule-modal');
     this._modalElement.className = 'web-modal';
@@ -82,6 +88,13 @@ class EventSchedule extends HTMLElement {
       el.removeAttribute('tabindex');
     });
 
+    // If the user opens this page from externally on a specific session, make sure we're showing
+    // the correct day of tab.
+    const index = this._tabsElement.indexOfTabParent(session);
+    if (index !== -1) {
+      this._tabsElement.focusTab(index);
+    }
+
     this._modalElement.sessionRow = clone;
     this._modalElement.open = true;
     document.body.append(this._modalElement);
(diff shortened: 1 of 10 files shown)


src/site/content/en/live/index.njk
```diff
@@ -149,6 +149,4 @@ date: 2020-06-31
 
 </div>
 
-<code id="event-data" style="display: none;">
-  {{ event.days | dump | safe }}
-</code>
+<web-event-store style="display: none;">{{ event.days | dump | safe }}</web-event-store>
```
robdodson, line 152: It looks like Ewa's maps widget need data form event.json as well. I think it might be good to merge that entire object into the store and not just event.days.

src/lib/components/EventSchedule/index.js
```diff
@@ -26,6 +28,10 @@ import '../EventScheduleModal';
 class EventSchedule extends HTMLElement {
   constructor() {
     super();
+    this.onStateChanged = this.onStateChanged.bind(this);
+
+    this._activeEventDay = null;
+    this._tabsElement = this.querySelector('web-tabs');
```
robdodson, line 34: You probably want to do this (and really all DOM operations) in the connectedCallback.

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+    for (const day of this._data) {
+      const timeOffsetBy = (hours) => {
+        const d = new Date(day.when);
+        d.setHours(d.getHours() + hours);
+        return +d;
+      };
+
+      const bufferStart = timeOffsetBy(-bufferHours);
+      const bufferEnd = timeOffsetBy(day.duration + bufferHours);
```
robdodson, line 57: Maybe another way of doing this would be to make bufferEnd = (utcTimeOfNextDay - 1hr). If we do that then you might be able to remove the nextPending day stuff?

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+      // Are we past the completion of this day? This allows the YT link to show up.
+      const isComplete = now >= bufferEnd;
+      if (day.isComplete !== isComplete) {
+        day.isComplete = isComplete;
+        change = true;
+      }
+      if (!isComplete && nextPendingDay === null) {
```
robdodson, line 65: so this sets nextPendingDay to the current active day, correct?

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // If there was no previously active day, then choose the upcoming day.
+    if (this._activeDay === null) {
+      this._activeDay = nextPendingDay;
+      change = true;
+    }
+
+    // Gate this with a boolean as to not trigger recursive updates.
+    if (change) {
+      console.warn('active day', this._activeDay);
```
robdodson, line 85: Do we want to leave this in for production?

src/lib/store.js
```diff
@@ -4,6 +4,27 @@ import getMeta from './utils/meta';
 import {localStorage} from './utils/storage';
 import {isProd} from 'webdev_config';
 
+const initialParams = new URLSearchParams(window.location.search);
+
+let timeOffset = 0;
+if (initialParams.has('_now')) {
+  const overrides = {
```
robdodson, line 11: What do you think about something like:

```
?day1=2020-06-30T16:02&day2=2020-06-30T16:03
```

to override the days and make them 1 minute apart so we can watch the state transitions happen without needing hardcode timestamps into the js itself

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+    this._data = [];
+  }
+
+  /**
+   * @param {boolean=} change this should force a change
+   */
+  _update(change = false) {
+    const now = +new Date() + this._timeOffset;
+
+    // If there was a previously active day (because the user has their tab open for a long time),
```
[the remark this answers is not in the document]
robdodson: Blah sorry, I wasn't thinking through all of the states. Maybe we should add a comment at the top of the file to explain all of the states. Something like...

```
// If you hit the page 1hr before the day, you'll see that day
// If you refresh 1hr after the day, you'll see the next day
// If you leave the tab open it will continue to show the day until you refresh OR it's 1hr before the next day's start time
```

src/lib/components/Tabs/index.js
```diff
@@ -232,10 +234,9 @@ class Tabs extends BaseElement {
   focusTab(index) {
     const tabs = this.querySelectorAll('.web-tabs__tab');
 
-    if (!tabs[index]) {
-      throw new RangeError('There is no tab at the specified index.');
+    if (tabs[index]) {
```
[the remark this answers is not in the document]
samthor: Just dealing with race conditions. I know this might also be solved by waiting for `whenDefined` but it seemed particularly harsh to explode here.

src/lib/components/Tabs/index.js
```diff
@@ -267,6 +268,19 @@ class Tabs extends BaseElement {
 
     this.activeTab = tabs.length - 1;
   }
+
+  /**
+   * @param {!Node} node to check
+   * @return {number} the index of the tab containing this node
+   */
+  indexOfTabParent(node) {
```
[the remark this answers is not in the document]
samthor: I don't agree since I'm not passing the panel, I'm passing.. some element within the panel. Happy to have a better name though.

src/lib/components/EventStore/index.js
samthor: No, because we have several ranges.

a. Within day 1
b. Ambiguous time between days
c. Within day 2
d. Ambiguous time between days
e. Within day 3

So the end is around each individual day.

src/lib/components/EventStore/index.js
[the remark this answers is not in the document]
samthor: Thanks. I've added this and the timer.

src/lib/components/EventSchedule/index.js
```diff
@@ -132,14 +145,40 @@ class EventSchedule extends HTMLElement {
   connectedCallback() {
     window.addEventListener('hashchange', this.onHashChange);
     this.addEventListener('click', this.onClick);
-    this.onHashChange();
+
+    // nb. Gross, but we need to wait for <web-tabs> to be ready.
```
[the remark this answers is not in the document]
samthor: Neat. Done.

src/lib/components/Tabs/index.js
```diff
@@ -103,8 +103,10 @@ class Tabs extends BaseElement {
   }
 
   panelTemplate(i, child) {
+    const index = i - 1; // i is 1-indexed
```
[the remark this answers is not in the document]
samthor: I think the goal was just to have nicely formatted panels. I don't have an opinion, the comment is just worth explaining it.

src/lib/store.js
samthor: Um, I don't know. It's a lot of extra code for this. Having said that, testing with "_now" the way it is is also a pain because of our implicit day chooser: it's hard to see the transition.

src/lib/components/Tabs/index.js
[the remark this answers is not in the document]
robdodson: yeah sorry, not a dig at your stuff. mostly a frown that i let tabs get built with an odd index thinger

src/lib/components/Tabs/index.js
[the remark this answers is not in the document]
robdodson: hm, `indexOfTabByChild` ?

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,141 @@
...
+ * limitations under the License.
+ */
+
+import '../EventScheduleModal';
+
+import {store} from '../../store';
+
+// Pretend the date is active within this buffer. This is useful so we can
+// reveal the player for a bit more time.
+const bufferHours = 1;
```
robdodson, line 23: I think we might need two buffers, one for video and one for chat. We want to disable the chat maybe about 10 minutes or so after the video is done so our moderators can sign off.

src/lib/components/EventSchedule/index.js
```diff
@@ -82,6 +93,13 @@ class EventSchedule extends HTMLElement {
       el.removeAttribute('tabindex');
     });
 
+    // If the user opens this page from externally on a specific session, make sure we're showing
+    // the correct day of tab.
+    const index = this._tabsElement.indexOfTabParent(session);
+    if (index !== -1) {
+      this._tabsElement.focusTab(index);
```
robdodson, line 100: I think you may not want this because it's literally calling focus() on the tab when the page loads (note the blue ring). So if someone opens the page with a screen reader it'll immediately announce "Day 1, tab"

image

robdodson: oh it also scrolls the page down so we definitely don't want that.

src/lib/components/Tabs/index.js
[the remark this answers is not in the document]
samthor: SGTM

src/lib/components/EventSchedule/index.js
samthor: Good catch. I now just set `.activeTab`.

robdodson:
I think the PR introduced a behavior where clicking on a schedule item scrolls the page. Doesn't seem to be happening on the prod version.

samthor:
I think that behavior is from the older version (since Netlify didn't deploy due to a random error); it previously focused on the tab which scrolled.

src/lib/components/EventStore/index.js
samthor: Since we don't know how long the event is going to play for yet (runtimes are still fluid), how about we make this the exact event start/end?

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+      store.setState({
+        eventDays: this._data,
+        activeEventDay: this._activeDay,
+      });
+    }
+  }
+
+  connectedCallback() {
+    store.subscribe(this.onStateChanged);
+    this.onStateChanged(store.getState());
```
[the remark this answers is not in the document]
samthor: Yes, I added a comment. Subscribe doesn't trigger the listener.

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,141 @@
...
+
+    const raw = JSON.parse(this.textContent.trim());
+
+    for (let i = 0; i < raw.days.length; ++i) {
+      const d = raw.days[i];
+      d.index = i;
+      d.complete = false;
```
robdodson, line 115: this needs to be changed to isComplete

src/lib/components/EventStore/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+    this._timeOffset = 0;
+    this._activeDay = null;
+    this._data = [];
+  }
+
+  /**
+   * @param {boolean=} change this should force a change
+   */
+  _update(change = false) {
```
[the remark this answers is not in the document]
samthor: There's no negative behavior if it runs early.

src/lib/store.js
```diff
@@ -4,6 +4,27 @@ import getMeta from './utils/meta';
 import {localStorage} from './utils/storage';
 import {isProd} from 'webdev_config';
 
+const initialParams = new URLSearchParams(window.location.search);
```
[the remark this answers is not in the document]
samthor: I imagine that we might want to use this same logic in other places, plus, this way it runs on every load (not just for /live/, which we might only navigate to later). I've moved some of it out to a helper utils script. Let me know if you insist on this only being part of `EventStore`.

src/lib/components/EventStore/index.js
samthor: I've added `activeChatDay`.

src/lib/components/EventStore/index.js
```diff
@@ -80,12 +83,19 @@ class EventStore extends HTMLElement {
...
-      const isActive = now >= bufferStart && now <= bufferEnd;
+      // Is this day active (within the buffer time range)?
+      const isActive = now >= bufferStart && now < bufferEnd;
       if (isActive && this._activeDay !== day) {
         this._activeDay = day;
         change = true;
       }
+
+      // Is this the active day for chat (within the actual time range)?
+      const isActiveChat = now >= actualStart && now < actualEnd;
```
robdodson, line 94: my hunch is we'll want to buffer this by some amount (10 minutes?) as well to let folks wrap up their conversations. But we can go with this for now and ask Harleen what she wants.

src/lib/components/EventStore/index.js
```diff
@@ -80,12 +83,19 @@ class EventStore extends HTMLElement {
...
+      if (isActiveChat && this._activeChatDay !== day) {
+        this._activeChatDay = day;
```
robdodson, line 96: Thinking through this...

I don't think I need to know the activeChatDay, I think I only need to know if chat is enabled or not. So if we are within the bounds of the actual day then chat is enabled, and if not it's disabled.

src/lib/components/EventStore/index.js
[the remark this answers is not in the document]
robdodson: I think you added a comment to EventSchedule not EventStore. Was that what you intended?

The reason I was asking is because I think this line is going to cause onStateChanged() to run, which may execute _update() if timeOffset is different but at that point none of the data will have been pulled in from the raw JSON (this._days will be null still) so update() will be executing with mostly null values. That was also why I added the comment about doing an early return because this._days won't be defined at that point. 
To your point, it's true that _update doesn't have any negative side effects at the moment, but I was concerned that if we kept adding to that function and we have the possibility of running it with null values that we may inadvertently add a negative side effect down the line. Would you be ok adding the early return to that function if this._days is undefined just as a safety check?

src/lib/components/EventStore/index.js
samthor: I've made this ten minutes, and updated all the duration stuff to be in minutes to correspond with that.

samthor:
Not really, but it's still a relevant comment in both places. Note that
`BaseStateElement`, which we don't use since it inherits from `LitElement`,
actually also calls the onStateChanged method immediately. In general we
can't guarantee when the subscriber triggers so relying on it not being
called immediately is problematic.

I've made the for loop iterate through `this._days || []`. Does that work?

robdodson:
> Note that
`BaseStateElement`, which we don't use since it inherits from `LitElement`,
actually also calls the onStateChanged method immediately. 

Yep I know.

> I've made the for loop iterate through `this._days || []`. Does that work?

Oh I wasn't worried about the loop. I think starting on line 102 it does stuff that's outside of that loop, then it calls setState. It was more so that area that I was concerned about. My read of the function was that we didn't really want it to execute if this._days is null so I'm not sure what the downside is to adding an early return but I guess it's not worth debating.

PR #3286, opened 2020-06-19
Fix wording

ekafyi:
Four factors are listed but the text says "three factors". This PR replaces "three" with "four".

### DIFF
--- a/src/site/content/en/metrics/lcp/index.md
+++ b/src/site/content/en/metrics/lcp/index.md
@@ -357,7 +357,7 @@ the article on [custom metrics](/custom-metrics/#element-timing-api).
 
 ## How to improve LCP
 
-LCP is primarily affected by four factors:
+LCP is primarily affected by the following factors:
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/lcp/index.md
```diff
@@ -357,7 +357,7 @@ the article on [custom metrics](/custom-metrics/#element-timing-api).
 
 ## How to improve LCP
 
-LCP is primarily affected by three factors:
+LCP is primarily affected by four factors:
```
kaycebasques, line 360:
```suggestion
LCP is primarily affected by the following factors:
```
How about we avoid the problem in the future by avoiding any hardcoded reference to a specific number of factors?

ekafyi: Absolutely, great idea.

kaycebasques:
I think this was just fixed in a different PR.

PR #3294, opened 2020-06-19
Make EventStore a bit more stateless. Get LiveStream working with unistore.

### LINKED ISSUE #3074: [During event] Add disabled chat embed
Related to #3073 

At the end of each day we'll need to disable chat and include a link to the youtube page.

image

### LINKED ISSUE #3143: [During event] Controller that enables/disables video player based on timestamp
If a user has the web.dev/live page open around our event, they should see the page change to bring in the player and remove it once each day's done, without needing to reload.

We'll need to update the schedule stuff with a bunch of JS timestamps that we push to some element which controls the visibility of the player (and the player/chat only makes sense with JS anyway, so it's all enhanced interactive foo).

robdodson:
Fixes #3143, Fixes #3074

I tested the following scenarios:

- [x] Before day 1, video shows day 1, chat is disabled
- [x] Before day 1, chat transitions from disabled -> active
- [x] Refresh during day 1, video and chat are active
- [x] After day 1 video buffer, video is marked isComplete
- [x] User leaves their tab open:
  - [x] After day 1 chat buffer, chat is disabled
  - [x] After day 1 video buffer, video transitions to day 2 (not what we intended, but also totally fine).
  - [x] After day 2 video buffer start, video transitions to day 2
  - [x] After day 2 chat buffer start, chat transitions to day 2
- [x] Refresh during day 2, day 2 video and chat are active
- [x] After day 3, chat is disabled, video stays on day 3 as a fallback.

The `isChatActive` code wasn't working because day 3 in the loop was setting `this._isChatActive` to false which then updated it in the store. So I changed some of the code in EventStore to make it always call setState and to make it a bit more stateless. Genuinely not trying to step on toes or anything 😅  I found it made it easier for me to reason through the code if it always called setState and if it left the decision of whether or not to render up to the components themselves. I don't think lit triggers a render if you set the same value over and over so I didn't need to add any guards to the elements. I watched the iframe src's in the elements panel and they stayed the same every tick.

- Set the update interval to 1 minute instead of 5. Just in case folks have the tab open before the event and are wondering why it hasn't started yet.
- Adds disabled chat state to livestream
- Connects LivestreamContainer to unistore

### DIFF
--- a/src/lib/components/LivestreamContainer/index.js
+++ b/src/lib/components/LivestreamContainer/index.js
@@ -7,14 +7,14 @@ class LivestreamContainer extends BaseStateElement {
   static get properties() {
     return {
       videoId: {type: String},
-      isChatEnabled: {type: Boolean},
+      isChatActive: {type: Boolean},
     };
   }
 
   constructor() {
     super();
     this.videoId = null;
-    this.isChatEnabled = true;
+    this.isChatActive = true;
   }
 
   render() {
@@ -23,6 +23,7 @@ class LivestreamContainer extends BaseStateElement {
     }
 
     // prettier-ignore
+    /* eslint-disable indent */
     return html`
       <div class="web-livestream-container__col-yt">
         <div class="w-youtube">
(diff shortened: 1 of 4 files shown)


robdodson:
Regarding this comment:

> After day 1 video buffer, video transitions to day 2 (not what we intended, but also totally fine).

I'm going to double check that this was the behavior I saw but I believe this is the case because after day 1 is marked isComplete then day 2 becomes the nextPendingDay on the next call to update() which would make it the fallback value for activeDay.

I'd like to keep this behavior if that's ok because I did a bunch of testing on EventStore and don't want to repeat it if I can avoid it. FWIW I asked Jake what they did for CDS and it sounded like they just revved the player the second the day was over so I think we're already going above and beyond with the 1hr margin.

src/lib/components/LivestreamContainer/index.js
```diff
@@ -37,24 +38,45 @@ class LivestreamContainer extends BaseStateElement {
...
    */
-  onStateChanged() {
-    // Example implementation depending on what we end up putting in our
-    // state object:
-    // const {videoId, isChatEnabled} = state;
-    // this.videoId = videoId;
-    // this.isChatEnabled = isChatEnabled;
+  onStateChanged({activeEventDay}) {
+    if (!activeEventDay) {
+      return;
```
samthor, line 74: Should this reset the state of this element? If for some reason our logic is bad and `activeEventDay` transitions back to `null`, then we won't clear state.

But I think `EventStore` will only ever go from `null` to a value, not the other way around.

robdodson: > Should this reset the state of this element? If for some reason our logic is bad and activeEventDay transitions back to null, then we won't clear state.

If it transitions to null then I think we'd want to keep the previous video on screen. Otherwise the player would show a blank rectangle.

PR #3317, opened 2020-06-22
Fix error where we check for PERCY mode

devnook:


robdodson:
I was having a bit of a tough time following the original conditional so instead of using ! I switched to early returns. Hope this is ok with folks!

PR #3320, opened 2020-06-22
Signedin reload

### LINKED ISSUE #3296: web.dev/live: reload page on signin
We need to reload the live page on signin as otherwise the YouTube chat/etc embeds won't authorize properly. We could do this only if chat is showing.

samthor:
Fixes #3296. I did this yesterday but didn't PR it.


src/lib/components/LivestreamContainer/index.js
```diff
@@ -15,11 +15,12 @@ class LivestreamContainer extends BaseStateElement {
     super();
     this.videoId = null;
     this.isChatActive = true;
+    this._isSignedIn = undefined;
```
robdodson, line 18:
```suggestion
    this.isSignedIn = undefined;
```

I guess to be consistent with the other properties.

src/lib/components/LivestreamContainer/index.js
```diff
@@ -69,14 +70,29 @@ class LivestreamContainer extends BaseStateElement {
...
-      return;
-    }
+  onStateChanged({activeEventDay, isSignedIn}) {
+    const {videoId, isChatActive} = activeEventDay || {
+      videoId: null,
+      isChatActive: false,
+    };
 
-    const {videoId, isChatActive} = activeEventDay;
+    const updateFrame = this.videoId !== videoId;
```
robdodson, line 79:
```suggestion
    const shouldUpdateIFrames = this.videoId !== videoId;
```

Bikeshedding (sorry) but I like using `is`, `should`, or `has` to quickly signify if something is going to be a boolean vs a function.

test/unit/src/lib/components/LivestreamContainer/index.js
```diff
@@ -0,0 +1,105 @@
...
+      },
+      isSignedIn: false,
+    });
+    await container.updateComplete;
+
+    const ytIframe = container.renderRoot.querySelector('iframe');
+    assert(ytIframe !== null, 'iframe should render');
+
+    const iframes = container.renderRoot.querySelectorAll('iframe');
+    assert(iframes.length === 1, 'should be single iframe');
```
robdodson, line 56: Might also add a test to verify that the src on the iframe matches the videoId

test/unit/src/lib/components/LivestreamContainer/index.js
```diff
@@ -0,0 +1,105 @@
...
+        isChatActive: false,
+      },
+      isSignedIn: false,
+    });
+    await container.updateComplete;
+
+    const timeout = new Promise((r) => setTimeout(r, 2000)).then(() => {
+      throw new Error("unload event didn't arrive for forced reload");
+    });
+    await Promise.race([p, timeout]);
```
robdodson, line 103: instead of racing is it possible to await `p` and assert that it resolves? Just in case, for whatever reason, the timeout ends up being flaky.

I think mocha has a built in 5s timeout so if the assert doesn't get triggered within that time then it would produce a similar behavior to the race you have now.

src/lib/components/LivestreamContainer/index.js
samthor: Moot since I've removed this boolean

PR #3330, opened 2020-06-23
explode on missing author posts/social

samthor:
Currently we just log loudly that someone has no posts and no social information.

Instead we should throw. Or possibly not. But the warning is annoying and doesn't help anyone.

### DIFF
--- a/src/site/_collections/authors.js
+++ b/src/site/_collections/authors.js
@@ -38,26 +38,51 @@ const findAuthorsPosts = (posts) => {
 };
 
 /**
+ * @param {!Author} author to update
+ * @param {!Array<*>} allAuthorPosts posts including drafts
+ * @return {boolean} whether this author is allowed here
+ */
+const maybeUpdateAuthorHref = (author, allAuthorPosts) => {
+  if (author.elements.length !== 0) {
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


robdodson:
@samthor It used to throw but ewa and I removed it yesterday and switched to a warning. The current person it's logging is in the process of adding a post that they're coauthoring so they snuck in a quick PR to add themselves to the contributors.

robdodson:
Here's the PR we did. https://github.com/GoogleChrome/web.dev/pull/3317

I think there was a bug in the previous code that wasn't properly blowing up when they did the PR to add themselves but started blowing up afterwards.

Not saying we have to keep this behavior, just giving some extra context.

samthor:
hm, okay. I accept that exploding here might not always be useful especially in intermediate states.

I'm happy to remove the `throw` but I think this PR does generally clean up this code. We can just log once and list all "bad" authors. WDYT?

robdodson:
> I'm happy to remove the throw but I think this PR does generally clean up this code. We can just log once and list all "bad" authors. WDYT?

I think that sounds good. Might want to wait on @MichaelSolati to land his [contributors.json PR](https://github.com/GoogleChrome/web.dev/pull/3339) before landing this and then rebase.

PR #3338, opened 2020-06-23
feat(web-event-q-and-a): add Q&A element, fixes #3121

### LINKED ISSUE #3121: [Post-event] Add Q&A section
image

- Should have a categories dropdown
- We'll probably implement this using summary/details elements

MichaelSolati:
Fixes #3121

Add rendering of Q&As as well as element to filter them

### DIFF
--- a/src/lib/components/EventQAndA/_styles.scss
+++ b/src/lib/components/EventQAndA/_styles.scss
@@ -40,6 +40,7 @@ web-event-q-and-a {
       list-style: none;
 
       &::-webkit-details-marker {
+        cursor: pointer;
         display: none;
       }
     }
--- a/src/lib/components/EventQAndA/_styles.scss
+++ b/src/lib/components/EventQAndA/_styles.scss
@@ -1,52 +1,12 @@
 @import '../../../styles/settings/colors';
 @import '../../../styles/settings/type';
 
-$DETAILS_TB_PADDING: 16px;
-$DETAILS_LR_PADDING: 8px;
-
 web-event-q-and-a {
   display: block;
 
   select {
     color: $GREY_700;
-    font-family: Google Sans;
+    font-family: $HEADLINE_FONT;
     width: auto;
-  }
-
-  details {
-    padding: $DETAILS_TB_PADDING $DETAILS_LR_PADDING;
-
-    &[open] {
-      border-bottom: 1px solid $GREY_800;
-      > * {
-        color: $GREY_800;
-      }
-    }
-
-    p {
-      margin-bottom: 0 !important;
-    }
-
-    span {
-      font-family: Google Sans;
-      font-size: 18px;
-      margin-right: 13px;
-    }
-
-    summary {
-      color: $GREY_600;
-      font-family: Google Sans;
-      font-size: 18px;
-      list-style: none;
-
-      &::-webkit-details-marker {
-        cursor: pointer;
-        display: none;
-      }
-    }
-  }
-
-  .w-event-question-element {
-    border-bottom: 1px solid $GREY_600;
   }
 }
--- a/src/lib/components/EventQAndA/_styles.scss
+++ b/src/lib/components/EventQAndA/_styles.scss
@@ -1,0 +1,51 @@
+@import '../../../styles/settings/colors';
+@import '../../../styles/settings/type';
+
+$DETAILS_TB_PADDING: 16px;
+$DETAILS_LR_PADDING: 8px;
+
+web-event-q-and-a {
+  display: block;
+
+  select {
+    color: $GREY_700;
+    font-family: Google Sans;
+    width: auto;
+  }
+
+  details {
+    padding: $DETAILS_TB_PADDING $DETAILS_LR_PADDING;
+
+    &[open] {
+      border-bottom: 1px solid $GREY_800;
+      > * {
+        color: $GREY_800;
+      }
+    }
+
+    p {
+      margin-bottom: 0 !important;
+    }
+
+    span {
+      font-family: Google Sans;
+      font-size: 18px;
+      margin-right: 13px;
+    }
+
+    summary {
+      color: $GREY_600;
+      font-family: Google Sans;
+      font-size: 18px;
+      list-style: none;
+
+      &::-webkit-details-marker {
+        display: none;
+      }
+    }
+  }
+
+  .w-event-question-element {
+    border-bottom: 1px solid $GREY_600;
+  }
+}
--- a/src/lib/components/EventQAndA/index.js
+++ b/src/lib/components/EventQAndA/index.js
@@ -21,7 +21,7 @@ class EventQAndA extends BaseElement {
   connectedCallback() {
     super.connectedCallback();
 
-    this.childElements = Array.from(this.querySelectorAll('[data-category]'));
+    this.childElements = Array.from(this.children);
 
     const categories = new Set();
     this.childElements.forEach((element) =>
@@ -47,7 +47,9 @@ class EventQAndA extends BaseElement {
   }
 
   closeDetail($event) {
-    const category = $event.target.closest('[data-category]');
+    const category = $event.target.hasAttribute('[data-category]')
+      ? $event.target
+      : $event.target.closest('[data-category]');
     if (!category) {
       return;
     }
--- a/src/lib/components/EventQAndA/index.js
+++ b/src/lib/components/EventQAndA/index.js
@@ -1,0 +1,88 @@
+import {BaseElement} from '../BaseElement';
+import './_styles.scss';
+
+/**
+ * Filters out Q&A's based on `data-category` property.
+ *
+ * @extends {BaseElement}
+ * @final
+ */
+class EventQAndA extends BaseElement {
+  constructor() {
+    super();
+    this.selectCategory = this.selectCategory.bind(this);
+
+    this.catageories = new Set();
+    this.categorySelectElement = document.createElement('select');
+    this.categorySelectElement.classList.add('w-select--borderless');
+    this.categorySelectElement.addEventListener('change', this.selectCategory);
+  }
+  connectedCallback() {
+    super.connectedCallback();
+    this.addDropDown();
+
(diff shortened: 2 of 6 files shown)


src/lib/components/EventQAndA/_styles.scss
```diff
@@ -0,0 +1,51 @@
...
+      margin-right: 13px;
+    }
+
+    summary {
+      color: $GREY_600;
+      font-family: Google Sans;
+      font-size: 18px;
+      list-style: none;
+
+      &::-webkit-details-marker {
```
samthor, line 42: If we're going to disable this, can we make the cursor a pointer?

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @extends {BaseElement}
+ * @final
+ */
+class EventQAndA extends BaseElement {
+  constructor() {
+    super();
+    this.selectCategory = this.selectCategory.bind(this);
+
+    this.catageories = new Set();
```
samthor, line 15: typo: `categories` ?

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+    this.categorySelectElement = document.createElement('select');
+    this.categorySelectElement.classList.add('w-select--borderless');
+    this.categorySelectElement.addEventListener('change', this.selectCategory);
+  }
+  connectedCallback() {
```
samthor, line 20: nit: add newline above

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+  }
+
+  disconnectedCallback() {
+    this.categorySelectElement.removeEventListener(
+      'change',
+      this.selectCategory,
+    );
+
+    this.querySelectorAll('[data-category]').forEach((element) => {
```
samthor, line 35: I'm confused, doesn't `<summary>` and `<details>` handle opening and closing on its own? This looks like we're adding extra JS.

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+    super.connectedCallback();
+    this.addDropDown();
+
+    this.querySelectorAll('[data-category]').forEach((element) => {
+      element.addEventListener('click', this.closeDetail);
+    });
+  }
+
+  disconnectedCallback() {
+    this.categorySelectElement.removeEventListener(
```
samthor, line 30: you add this handler in the constructor but then remove in disconnected.

We manage the subordinate element's lifecycle, you don't need this remove here.

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+    this.querySelectorAll('[data-category]').forEach((element) =>
+      this.catageories.add(element.getAttribute('data-category')),
+    );
+
+    const divContainer = document.createElement('div');
+    divContainer.classList.add('w-event-question-element');
+    divContainer.append(this.categorySelectElement);
+    const allOption = document.createElement('option');
+    allOption.setAttribute('value', '');
+    allOption.innerText = 'All categories';
```
samthor, line 51: `innerText` considered harmful (it's an old IE property). Use `textContent`.

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,88 @@
...
+      .forEach((element) => {
+        if (element !== this) {
+          element.removeAttribute('open');
+        }
+      });
+  }
+
+  selectCategory($event) {
+    this.querySelectorAll('[data-category]').forEach((element) => {
+      if (
```
samthor, line 76: minor nit: could be neater like

```js
const show = !$event.target.value || element.getAttribute('data-category') === $event.target.value;
element.classList.toggle('hidden', !show);
```

... or similar. Up to you.

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,80 @@
...
+      option.setAttribute('value', category);
+      option.textContent = category;
+      this.categorySelectElement.append(option);
+    });
+
+    this.prepend(divContainer);
+  }
+
+  closeDetail() {
+    this.parentElement
```
samthor, line 61: I think you want `parentElement` because `closeDetail` isn't bound properly to the instance of the this class.

samthor: Actually, if you're going to `querySelectorAll` all the time, maybe add a getter?

```js
  get categoryElements() {
    return this.querySelectorAll(...);
  }

  // later
  closeDetail() {
    this.categoryElements.forEach(...);
  }

File: src/lib/components/EventQAndA/index.js [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+    this.categorySelectElement = document.createElement('select');
+    this.categorySelectElement.classList.add('w-select--borderless');
+    this.categorySelectElement.addEventListener('change', this.selectCategory);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.addDropDown();
+
+    this.querySelectorAll('[data-category]').forEach((element) => {
```
samthor, line 25: I don't want to bikeshed but I think doing something like:

```js
  this.addEventListener('click', this.closeDetail);

// later
  closeDetail(e) {
    const category = e.target.closest('[data-category]');
    if (!category) {
      return;
    }
    this.categoryElements.forEach((cand) => {
      if (cand !== this) { ... }
    });
  }
```

is a bunch neater as we don't have to try to maintain a bunch of click listeners.

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  disconnectedCallback() {
+    this.querySelectorAll('[data-category]').forEach((element) => {
+      element.removeEventListener('click', this.closeDetail);
+    });
+    super.disconnectedCallback();
+  }
+
+  addDropDown() {
```
samthor, line 37: this will run every time the element is connected. Which, to be fair, is only once right now. Maybe we can have a boolean or see if `this.categories.size` is truthy?

src/lib/components/EventQAndA/index.js
MichaelSolati: @samthor not binding this also give's me the element that triggers the event as my `this`. I found this helpful, though I can go through this again with the feedback and see how I can refactor it.

src/site/content/en/live/index.njk
```diff
@@ -122,8 +122,33 @@ date: 2020-06-31
...
+        <p>
+          During web.dev LIVE, we were happy to address some very insightful and interesting questions from the community across the globe. Here are some of the top questions asked.
+        </p>
+        <a href="/live#content" class="w-event-link">Back to top</a>
+      </div>
 
+      <div class="w-event-section__column-right">
+        <web-event-q-and-a>
+        {%- for entry in event.qAndAs  %}
+          <details class="w-event-question-element" data-category="{{entry.category}}">
```
robdodson, line 140: I think if we're trying to be BEM-like then this class name would be:

```
web-event-q-and-a__details
```

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,80 @@
...
+ * @extends {BaseElement}
+ * @final
+ */
+class EventQAndA extends BaseElement {
+  constructor() {
+    super();
+    this.selectCategory = this.selectCategory.bind(this);
+
+    this.categories = new Set();
+    this.categorySelectElement = document.createElement('select');
```
robdodson, line 16: Would it make sense to define this in the same chunk of code where you're defining the div container element?

Also, since we know the prerendered children won't change, you might be able to use a `render()` function to cut down on code.

```js
render() {
  if (!this.prerenderedChildren_) {
    this.prerenderedChildren_ = [];
    for (const child of this.children) {
      this.prerenderedChildren_.push(child);
    }
  }

  return html`
    <div class="w-event-question-element">
      <select>
        <option value="">All options</option>
        ${this.categories.map(...)}
      </select>
    </div>
    ${this.prerenderedChildren_}
  `;
}
```

src/lib/components/EventQAndA/index.js
MichaelSolati: Because of the padding on the elements I took this in a different direction, but have also moved away from all the click listeners (kinda), so it should be much neater.

src/styles/pages/_live.scss
```diff
@@ -482,3 +481,41 @@ web-tabs.w-event-tabs {
...
+    }
+  }
+
+  p {
+    margin-bottom: 0 !important;
+  }
+
+  span {
+    font-family: $HEADLINE_FONT;
+    font-size: 18px;
```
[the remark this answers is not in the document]
MichaelSolati: The design for the Q and the A (font size and font type) are the same I though?

src/lib/components/EventQAndA/index.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  renderDetails() {
+    return this.childElements.map(
+      (e) => html`<div @click="${this.closeDetail}">${e}</div>`,
+    );
+  }
+
+  closeDetail($event) {
+    const category = $event.target.hasAttribute('[data-category]')
```
samthor, line 53: Closest includes itself. :)

robdodson:
@MichaelSolati when you've reached a stopping point let me know. I have [another PR in progress](https://github.com/GoogleChrome/web.dev/pull/3340) that touches a bunch of files and will conflict with this one. Because the changes in your PR are much more scoped, I'd like to land my PR and then I can manually migrate your changes and resolve the conflicts for you.

src/styles/pages/_live.scss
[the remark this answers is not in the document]
robdodson: I think if you remove the styles here and move them up to .w-event-question-element they'll cascade down so the span will get the same styles. The `p` is actually supposed to be 16px Roboto it seems. But everything else is 18px Google Sans.

PR #3339, opened 2020-06-23
refactor(authors): turn `_data/contributors.js` to `_data/authorsData.js`

### LINKED ISSUE #3055: Make authors and tags files into JSON. Move behavior into other files.
We'd like to give the tech writers the ability to edit the authors and tags files without needing an eng review. But since these files are `.js` we always require an eng review because we don't want folks sneaking malicious js into our build.

I think we should move the behavioral bits out of these files and into their own .js files that import the authors/tags file and then export the final product. Eng would be owners of these new files and we could have writers own the .json files.

Convert postTags.js and contributors.js to json, move behavior into separate files.

MichaelSolati:
Turn `_data/contributors.js` to `_data/authorsData.js`, then use only `_collections/authors.js` for author data, and finally move away from the term `contributor`.

Part one of #3055.


src/site/_collections/authors.js
```diff
@@ -71,56 +71,68 @@ module.exports = (collections) => {
...
+    author.title = parts.join(' ');
+    author.href = `/authors/${key}/`;
+
+    author.description =
+      author.description && author.description.en
+        ? author.description.en
+        : `Our latest news, updates, and stories by ${author.title}.`;
+
+    // This updates the shared contributors object with meta information and is safe to be called multiple times.
+    author.url = path.join('/en', author.href);
```
robdodson, line 92: What's author.url being used for? I'm just cautious about hardcoding any language paths given @devnook's localization work.

src/site/_collections/authors.js
```diff
@@ -71,56 +71,68 @@ module.exports = (collections) => {
...
+        }
+      } else {
+        author.href = `https://twitter.com/${author.twitter}`;
       }
+    }
 
-      if (process.env.PERCY) {
-        author.elements = author.elements.slice(-6);
-      }
+    const authorsImage = findAuthorsImage(key);
```
robdodson, line 124: I wonder how slow this function is since it needs to do 3 disk reads per author... I don't think we have any 3x photos at the moment and we encourage everyone to add a 2x photo. If there are any existing cases without a 2x photo maybe we should just duplicate their 1x, rename it 2x, and always use the 2x version for all of our authors?

src/site/_collections/authors.js
```diff
@@ -71,56 +71,68 @@ module.exports = (collections) => {
...
-          if (posts.length) {
-            console.warn(
-              `author ${
-                author.title
-              } has no posts and no social: ${JSON.stringify(author)}\n`,
-            );
-          }
-        } else {
-          author.href = `https://twitter.com/${author.twitter}`;
+  Object.keys(contributors).forEach((key) => {
```
robdodson, line 74: I think this seems pretty chill and we can go with this approach instead of the interstitial file i suggested

src/site/_collections/authors.js
MichaelSolati: I believe it's being used by the BaseCard for [the url](https://github.com/GoogleChrome/web.dev/blob/master/src/site/_includes/components/BaseCard.js#L42).

MichaelSolati: This can and would be removed when the cards are refactored.

src/site/_collections/authors.js
```diff
@@ -38,89 +37,83 @@ const findAuthorsPosts = (posts) => {
...
-            console.warn(
-              `author ${
-                author.title
-              } has no posts and no social: ${JSON.stringify(author)}\n`,
-            );
-          }
-        } else {
-          author.href = `https://twitter.com/${author.twitter}`;
+    // If the author doesn't have any posts, use their Twitter profile.
+    if (author.elements.length === 0) {
```
robdodson, line 86: Would it be possible to pull @samthor's changes from https://github.com/GoogleChrome/web.dev/pull/3330 into this PR?

src/site/_collections/authors.js
robdodson: ah i see. sneaky old stripLanguage..

src/site/_collections/authors.js
robdodson: Or we could land this in do a separate PR if you'd like to reduce the number of changes. Up to you.

src/site/_collections/hooks/utils.js
```diff
@@ -56,6 +56,8 @@ const index = (items, href, testItems) => {
     itemsWithPosts = items.filter((item) => item.elements.length > 0);
   }
 
+  itemsWithPosts.sort((a, b) => a.title.localeCompare(b.title));
```
robdodson, line 59: Is this alphabetizing the author's posts?

robdodson: if so, we may want to leave them in publish order

MichaelSolati: This doesn't sort the posts, rather the tags and the authors themselves.

src/site/_collections/authors.js
MichaelSolati: @samthor would you mind if I merged your PR into mine?

MichaelSolati:
@googlebot I fixed it.

src/site/_collections/authors.js
samthor: Go ahead.

PR #3340, opened 2020-06-24
Update styles for isPreEvent, isDuringEvent, and isPostEvent phases

### LINKED ISSUE #3115: [During event] Add Share Feedback button
image

Tentatively assuming this links to an external feedback form.

### LINKED ISSUE #3116: [During event] Remove section heading. Move logo into schedule section.
image

Currently the page says "What to expect". But we'll be removing that text and moving the header into this area for the 'day of' page.

robdodson:
Fixes #3115, Fixes #3116

A bunch of CSS changes to make the pre event and during event pages look correct. We could land these now if we want. I still have some more changes for the during and post event pages but it might be good to break those apart. My hope is this will be the bigger of the two PRs because it contained a lot of cleanup.

- Adds a second carousel that displays only on mobile and hides the carousel in the right column.
- A lot of CSS :\
- Adds a share feedback button
- Adds the correct header for the during event phase.
- Adds a `.w-live-flow` class which can be used for consistent vertical rhythm.

### DIFF
--- a/src/site/_data/event.js
+++ b/src/site/_data/event.js
@@ -404,6 +404,7 @@ const communityEvents = {
 };
 
 module.exports = {
+  isPreEvent: true,
   isDuringEvent: false,
   isPostEvent: false,
   days,
--- a/src/site/_data/event.js
+++ b/src/site/_data/event.js
@@ -405,8 +405,8 @@ const communityEvents = {
 
 module.exports = {
   isPreEvent: false,
-  isDuringEvent: true,
-  isPostEvent: false,
+  isDuringEvent: false,
+  isPostEvent: true,
   days,
   communityEvents,
 };
--- a/src/site/_data/event.js
+++ b/src/site/_data/event.js
@@ -23,7 +23,7 @@ const days = [
     title: 'Day 1',
     when: '2020-06-30T16:00Z', // 9am PDT (-7)
     duration: 3 * 60, // minutes
-    videoId: null,
+    videoId: 'oHHSSJDJ4oo',
     sessions: [
       {
         speaker: 'dalmaer',
@@ -124,7 +124,7 @@ const days = [
     title: 'Day 2',
     when: '2020-07-01T12:00Z', // 12pm GMT/UTC (+0), note UK time will be 1pm
     duration: 3 * 60, // minutes
-    videoId: null,
+    videoId: 'k6TWO-ESC6A',
     sessions: [
       {
         speaker: 'dalmaer',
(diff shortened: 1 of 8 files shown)


src/site/content/en/live/index.njk
```diff
@@ -22,78 +22,127 @@ date: 2020-06-31
...
+          questions real-time, in your active hours. We encourage you to join us
+          on all three days but if that's not possible, you can watch the videos
+          on demand and leave your questions in the Comments section on YouTube
+          and we'll try our best to respond over the three days of the event.
+        </p>
+        <a href="https://docs.google.com/forms/d/e/1FAIpQLSdkWh94xheuR3Y_p0z6GML0AkdWLCa767KhFXBUPM3VI9-jCA/viewform" class="w-feedback-button w-button w-button--primary">
+          <svg width="24" height="24" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <path fill-rule="evenodd" clip-rule="evenodd" d="M4 2h16c1.1 0 2 .9 2 2v12c0 1.1-.9 2-2 2H6l-4 4 .01-18c0-1.1.89-2 1.99-2zm0 14h16V4H4v12zm6.5-2H18v-2h-5.5l-2 2zm3.86-6.58c.2.2.2.51 0 .71L8.47 14H6v-2.47l5.88-5.88c.2-.2.51-.2.71 0l1.77 1.77z" fill="#fff"/>
+          </svg>
+          Share Feedback
```
samthor, line 138: For some reason I imagine this as opening in `target="_blank"`. But whatever.

src/site/_data/event.js
```diff
@@ -23,7 +23,7 @@ const days = [
     title: 'Day 1',
     when: '2020-06-30T16:00Z', // 9am PDT (-7)
     duration: 3 * 60, // minutes
-    videoId: null,
+    videoId: 'oHHSSJDJ4oo',
```
samthor, line 26: remove before submit (unless this is real?) - I see a video unrelated to us in the Netlify preview.

samthor:
The double-carousel is not amazing, but playing with the CSS for a bit... unless we actively switch to say, grid layout, I can't really see a practical way to solve it by only having one element right now.

Can you add a comment adjacent to the two carousels calling out explicitly why we have two?

src/styles/pages/_live.scss
```diff
@@ -191,69 +217,89 @@
...
   }
 
   .w-event-buttons {
     margin: 16px 0;
   }
+
+  .w-feedback-button {
+    margin-top: 49px;
+
+    :first-child {
```
samthor, line 266: tiny nit: adding `svg` here might make it more obvious that this is just moving the image around.

robdodson:
> The double-carousel is not amazing, but playing with the CSS for a bit... unless we actively switch to say, grid layout, I can't really see a practical way to solve it by only having one element right now.

yeah it's by design 🤷‍♂️

image

I think the main difference is I added the explicit scrollbar. My hunch is depending on the device, if day2 lines right up with the edge then people won't know to scroll it.

PR #3312, opened 2020-06-22
Add Rob Kochman to contributors.

jpmedley:
@rkochman Please do two things:
* Edit the [links to profiles](https://github.com/GoogleChrome/web.dev/pull/3312/files#diff-979786d25a45cd5b7cc5f60fcf97a2a7R1310) as you see fit.
* (Optional) Add profile images as [described in the handbook](98776ae5ac5467e9b15749a88dffe1cb944f4077). Not everyone has images. Placeholders are substituted.

PR #3351, opened 2020-06-24
Add author images for rkochman

rkochman:


samthor:
The other PR (#3312) lists your key as "robko", so your images need to match that, or we need to change the key there.

PR #3354, opened 2020-06-24
fix(sitemap): use `canonicalUrl` for pages with a `url`, fixes #3352

### LINKED ISSUE #3352: sitemap.xml contains /en/ urls
**Describe the bug**
The sitemap.xml contains urls like

```xml
<url>
    <loc>https://web.dev/en/read-files/</loc>
    <lastmod>2020-05-08</lastmod>
</url>
```

**To Reproduce**
1. Do a prod build, `ELEVENTY_ENV=prod npm run build`
2. Look at the `sitemap.xml` in the `dist` directory.

**Expected behavior**
URLs should not contain /en/

MichaelSolati:
Fixes #3352

This way we can do away with the language, as it's not in the file slug.

### DIFF
--- a/src/site/content/sitemap.xml.njk
+++ b/src/site/content/sitemap.xml.njk
@@ -7,7 +7,7 @@ permalink: "/sitemap.xml"
 {%- for page in collections.all %}
 {%- if not page.data.noindex and not page.data.draft %}
   {%- if page.url %}
-  {% set absoluteUrl %}{{ page.url | absoluteUrl(site.url) }}{% endset %}
+  {% set absoluteUrl %}{{ page.fileSlug | absoluteUrl(site.url) }}{% endset %}
   <url>
     <loc>{{ absoluteUrl }}</loc>
     {%- if page.data.updated %}
--- a/src/site/content/sitemap.xml.njk
+++ b/src/site/content/sitemap.xml.njk
@@ -6,6 +6,9 @@ permalink: "/sitemap.xml"
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
 {%- for page in collections.all %}
 {%- if not page.data.noindex and not page.data.draft %}
+  {# Verify that the page has a permalink so we don't pull in things like #}
+  {# the service worker partial. #}
+  {# But use the page's fileSlug so we can strip out the language directory. #}
   {%- if page.url %}
   {% set absoluteUrl %}{{ page.fileSlug | absoluteUrl(site.url) }}{% endset %}
   <url>
--- a/src/site/content/sitemap.xml.njk
+++ b/src/site/content/sitemap.xml.njk
@@ -10,7 +10,7 @@ permalink: "/sitemap.xml"
   {# the service worker partial. #}
   {# But use the page's fileSlug so we can strip out the language directory. #}
   {%- if page.url %}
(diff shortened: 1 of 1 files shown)


samthor:
Isn't the issue is that we _also_ have "/en/"? This looks like it just changes the output.

robdodson:
> Isn't the issue is that we also have "/en/"? This looks like it just changes the output.

I think we do technically have /en/ but I don't think we want folks to use those URLs. In some instances they even break (https://web.dev/en/measure/ for example). So it seemed like a good idea to only use canonical urls in the sitemap?

robdodson:
btw what initiated this was kayce started a thread showing that search console was erroring on one of these /en/ pages

image

image

my thinking was that since we don't want /en/ pages to be canonical (I think we have a link rel canonical tag but still...) it seemed good for the sitemap to use the canonical url as well. Let me know if I'm missing something?

src/site/content/sitemap.xml.njk
```diff
@@ -6,8 +6,11 @@ permalink: "/sitemap.xml"
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
 {%- for page in collections.all %}
 {%- if not page.data.noindex and not page.data.draft %}
+  {# Verify that the page has a permalink so we don't pull in things like #}
+  {# the service worker partial. #}
+  {# But use the page's fileSlug so we can strip out the language directory. #}
   {%- if page.url %}
-  {% set absoluteUrl %}{{ page.url | absoluteUrl(site.url) }}{% endset %}
+  {% set absoluteUrl %}{{ page.fileSlug | absoluteUrl(site.url) }}{% endset %}
```
devnook, line 13: FileSlug is not reliable. E.g. for the file /newsletters/archive/2020/03, slug is just "03", which does not give you a correct url for this resource. in PR #3261 I introduced canonicalUrl to address that. Maybe use canonicalUrl instead?

src/site/content/sitemap.xml.njk
```diff
@@ -6,8 +6,11 @@ permalink: "/sitemap.xml"
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
 {%- for page in collections.all %}
 {%- if not page.data.noindex and not page.data.draft %}
+  {# Verify that the page has a permalink so we don't pull in things like #}
+  {# the service worker partial. #}
+  {# But use the page's fileSlug so we can strip out the language directory. #}
   {%- if page.url %}
-  {% set absoluteUrl %}{{ page.url | absoluteUrl(site.url) }}{% endset %}
+  {% set absoluteUrl %}{{ page.data.canonicalUrl }}{% endset %}
```
devnook, line 13: I'm not entirely clear why would we want to strip language from url in a sitemap? We want e.g. google bot to see both /en/blah and /pl/blah, right?

MichaelSolati: I think we'd only want to remove english, other languages should have their prefix. @robdodson?

robdodson: I think Michael is correct, though I'm not an SEO expert.

But I think the idea is we'd want google bot to index:

/web-vitals/
/pl/web-vitals/
/ja/web-vitals/

PR #3358, opened 2020-06-25
hashes JS/CSS source

samthor:
This creates makes our JS and CSS files generate with a hash (e.g., "bootstrap.js" is now "bootstrap-HASH.js").

It also creates files inside Eleventy's data folder as part of their respective builds "resourcesJS.json" and "resourcesCSS.json" which contain the hashed entrypoints of our JS and CSS, so they can be included in the layout. Importantly:
- without these files, we just get files built with no CSS or JS (literally the blank string)
- if these files don't exist then get created, `eleventy --watch` will notice and rebuild
- if these files change, `eleventy --watch` will notice and rebuild

This PR also:
- moves Eleventy to be the second last step in the build
- adds a Service Worker build task which is the last step (as it needs to know Eleventy hashes for `precacheAndRoute`)
- (implicitly) the Service Worker now caches the new files, hash and all

It does not yet:
- remove the generation of the Live partial (but that will still work exactly the same way)

Honestly I think this is PR is _more_ complicated than I'd like because we now have a separate SW build task. We'd be able to remove this if Eleventy generated the "cache manifest" as I do in the other PR.


src/build/hash.js
```diff
@@ -0,0 +1,45 @@
...
+const hashLength = 8;
+
+/**
+ * Hashes the passed path. If the file doesn't exist, return a random hash. If
+ * we're in prod, then crash.
+ *
+ * @param {string} files to hash
+ * @return {string}
+ */
+function hashFor(...files) {
```
robdodson, line 13: Is this function used anywhere?

samthor: No, it was leftover from the "resources" version in the other PR. I'll remove it for now.

.eleventy.js
```diff
@@ -237,6 +241,41 @@ module.exports = function (config) {
...
+    // of their build scripts, so make sure they exist in prod. In watch mode,
+    // we skip this check as builds will occur at all sorts of random times.
+    const checkJSONDataPath = (name) => {
+      const f = `src/site/_data/${name}.json`;
+      try {
+        const raw = JSON.parse(fs.readFileSync(f), 'utf-8');
+        if (!raw['path']) {
+          throw new Error('could not find path');
+        }
+        const check = 'dist' + raw['path'];
```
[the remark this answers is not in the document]
samthor: Yup, that's better. I was just being lazy. 👍

PR #3362, opened 2020-06-25
versioned partials

samthor:
Supersedes #3256.

* Adds a common "resource version" known by the Service Worker and a few other places
* Uses this to force an update of the template and manifest _before_ serving a hydrated partial
* Similarly, forces a real page load (which'll go through to the SW) inside "loader.js", in the client

Again, importantly, we don't include general resources and hashes in the SW. It should not change on a regular basis. Most of the code added to the SW in PR exists because Workbox doesn't have a general way to cache/serve its generated manifest—it _only_ works in a `precacheAndRoute()` call, which has strict lifecycle requirements (you _must_ update the SW to update the list, as it has to happen before events fire).

(There might be a few holes, I wrote this then stopped for the evening.)


robdodson:
Hey @samthor, is this something you'd like to continue working on?

samthor:
Yes. I will bring it up to HEAD today and make the description clearer.

PR #2857, opened 2020-05-13
Adding origin trial info & prettier reformat

jakearchibald:
This adds details of the [Portals origin trial](https://www.chromium.org/blink/origin-trials/portals) to the Portals article.

This is draft & "DO NOT MERGE" since the origin trial hasn't launched yet.

@uskay since you're the author of the original, can you take a look and let me know if you're ok with the changes?

PR #3375, opened 2020-06-25
Use standard idoms and language for flags and origin trials.

jpmedley:


jpmedley:
@kaycebasques FYI, this is against #2857

PR #3378, opened 2020-06-25
add cleanup versions script and workflow

### LINKED ISSUE #2696: Write a script that cleans up old app engine versions on prod and staging.
App engine has a hard limit of around 210 versions before it stops letting you deploy. Unfortunately there's no automated way (that I know of) to tell app engine to clean up old unused versions. We need to write a script that can be run as part of a cron (probably a scheduled GitHub Action) that deletes old app engine versions.

samthor:
Fixes #2696.

I haven't tested the workflows—I do think that GitHub will pass through the environment variables (as our other scripts which run `gcloud` work fine, and the internal call to `gcloud` in the script by default also passes through `process.env`), but basically I'd like to commit and fix if there's some problem.

### DIFF
--- a/.github/workflows/cleanup-versions-workflow.yml
+++ b/.github/workflows/cleanup-versions-workflow.yml
@@ -1,0 +1,23 @@
+name: Cleanup old versions
+on:
+  schedule:
+    # At 18:00 UTC / 7:00 PST every day.
+    - cron: '0 18 * * *'
+
+jobs:
+  ci:
+    name: Run gcloud to snapshot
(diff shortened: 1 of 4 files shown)


robdodson:
Looks good but adding the DO NOT MERGE flag until after the /live/ event is over. I know the pantheon GUI won't let you delete a version that's serving traffic but not sure if the cli has that same boundary. I know the code is checking for it, but out of caution I'd like to wait until after day 3 is over before enabling the script.

PR #3381, opened 2020-06-26
support no-hash in dev for css, better source map support, cleanup 'dev' use

samthor:
- Uses app.css in dev for faster dev cycles
- Fixes up source map and serves /src/ in dev so CSS is correctly attributed/displayed during development
- Removes use of `ELEVENTY_ENV === 'dev'`, updates README to provide guidance (we should set to 'prod' or leave blank)
- Deploys staging as `ELEVENTY_ENV='prod'`


.eleventy.js
```diff
@@ -90,7 +90,6 @@ const {
 
 module.exports = function (config) {
   const isProd = process.env.ELEVENTY_ENV === 'prod';
-  const isWatch = process.argv.includes('--watch');
```
MichaelSolati, line 93: Is this analogous for `dev`?

samthor:
Yeah, the goal is that dev is only when we're watching stuff locally or
otherwise are much more flexible about build order.

PR #3392, opened 2020-06-27
fallback to loading other css/js

samthor:
If anyone is seeing this problem I'm concerned because it means that Workbox is not always correctly precaching assets. And the real challenge there is they'll never upgrade their SW since the registration code is inside our JS.

To be fair, the lifetime of cached assets is not guaranteed inside `Cache`.

The simpler way to solve this is to go back to "?v=...." search params.

### DIFF
--- a/server.js
+++ b/server.js
@@ -16,6 +16,7 @@
 
 const isProd = Boolean(process.env.GAE_APPLICATION);
 
+const fs = require('fs');
 const compression = require('compression');
 const express = require('express');
 const cookieParser = require('cookie-parser');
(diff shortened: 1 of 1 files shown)


robdodson:
> If anyone is seeing this problem I'm concerned because it means that Workbox is not always correctly precaching assets.

I'm not sure if we know if this is a workbox issue just yet, or if maybe something in our sequence of events is off. Perhaps our sw is getting tripped up somehow when it segues to new versions. It would be good if we could repro somehow and see the response that the sw returned. If the html contained out of date hashes then we'd know it had to do with the way the sw builds the partials

robdodson:
Adding more info to this ticket.

I was able to temporarily repro the issue using an incognito window. It looked like the html was requesting the previous css hash. This leads me to believe that perhaps something is caching our html or the server is serving a stale version.

robdodson:
Going to merge this to get the site back in working order

PR #3394, opened 2020-06-27
Add hero banner.

robdodson:
Instructions from marketing are that this should go live later this week. Adding the DO NOT MERGE flag till then.

- Add a custom hero banner component.


robdodson:
Turns out the original design was a misconfigured thing from one of the other sites so going with just the single banner.

Issue #3363, opened 2020-06-25
Cache-busting app.css?v=... isn't cached by old SW


samthor:
**Describe the bug**
At some point, we added a request to "app.css?v=..." to help cache-bust our CSS loading. Unfortunately, this was never cached or served by the Service Worker. (This changed in #3358, which removed app.css in favor of a hashed version.)

**To Reproduce**
1. Load and install the site before #3358
2. either: go offline (and see no CSS), or move past that PR (and see a flash of missing CSS)

**Expected behavior**
Styles!

**Screenshots**
<img width="757" alt="Screen Shot 2020-06-26 at 00 12 16" src="https://user-images.githubusercontent.com/119184/85736284-b4a9b700-b741-11ea-81bd-3dc3db91ec1c.png">
(see e.g. "bootstrap.js" coming from SW, even though it no longer exists as a real file).

samthor:
This issue is "fixed" but I'm filing it to provide more context in code.

PR #3395, opened 2020-06-29
use bootstrap.js/app.css again

samthor:
We're seeing odd caching problems in the Real World(™) whereby older HTML files are being served, which request hashed CSS/JS files that don't exist. This was mitigated by #3392 but this PR cleans up the behavior further.

- instead of generating "app-hashed.css" and serving it at all "app.css", "app-whatever.css", this PR now generates _only_ "app.css" and also catches "app-whatever.css".
- we include a cache-busting query param in our layout (e.g., "app.css?v=1234")
- this same exact URL, including cache-busting, is cached by our SW (avoiding #3363)

So if we see odd caching issues, "app.css" and "bootstrap.js" will _still be served_, because the server is ignoring the query string.

This will effect the "bigger refactor" in #3362 but that's fine.

### DIFF
--- a/src/site/_filters/strip-query-params-dev.js
+++ b/src/site/_filters/strip-query-params-dev.js
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const isProd = process.env.ELEVENTY_ENV === 'prod';
+const noop = (url) => url;
+const strip = (url) => url.split('?')[0];
+
+// Strip query params from URLs in dev only. Used for lazy cache-busting
+// e.g. in non-prod, app.css?blah => app.css
+module.exports = isProd ? noop : strip;
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -79,6 +79,7 @@ const removeDrafts = require(`./${filtersDir}/remove-drafts`);
 const strip = require(`./${filtersDir}/strip`);
 const stripBlog = require(`./${filtersDir}/strip-blog`);
 const stripLanguage = require(`./${filtersDir}/strip-language`);
+const stripQueryParamsDev = require(`./${filtersDir}/strip-query-params-dev`);
 const getPaths = require(`./${filtersDir}/get-paths`);
 
 const transformsDir = 'src/site/_transforms';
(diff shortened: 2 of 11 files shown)


robdodson:
looks like the test failed trying to do a prod build

samthor:
Tests fixed. PTAL. This can wait until after WDL, because the functionality is the same, it's just a bit less fragile (we don't need to look up _what_ "app-hashedname.css" we currently have).

robdodson:
> We're seeing odd caching problems in the Real World(™) whereby older HTML files are being served, which request hashed CSS/JS files that don't exist.

Before moving forward maybe we should solve this problem. Looking at the MDN docs on cache-control, it looks like setting something to public;max-age=0 (which is what express.static() gives us today) will not prevent intermediaries from caching it. Instead we should set our html to `no-store` https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Preventing_caching

My concern is if we make it so query params always fallback to the current CSS then we could end up with a different kind of mismatching (cache'd HTML is trying to reference some CSS selector that we deleted in the newest CSS).

samthor:
Right but we don't deploy the old CSS. That scenario you posit will happen anyway since those browsers don't have access to the older CSS.

I'm not worried about a user seeing a more up-to-date version of our CSS or JS, even if they have slightly older HTML, although we should try to avoid that too.

src/site/_filters/strip-query-params-dev.js
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+const isProd = process.env.ELEVENTY_ENV === 'prod';
+const noop = (url) => url;
+const strip = (url) => url.split('?')[0];
+
+// Strip query params from URLs in dev only. Used for lazy cache-busting
+// e.g. in non-prod, app.css?blah => app.css
+module.exports = isProd ? noop : strip;
```
MichaelSolati, line 23: Ok, so not gonna hold this PR up, but simple question... Could this be a generic filter (not dev specific), where it takes a second argument which would tell the filter if it should strip the params?

```njk
<link rel="stylesheet" href="{{ resourceCSS.path | stripQueryParams(ifDev) }}" />
```

and 

```JS
module.exports =(url, condition) => condition ? noop(url) : strip(url);
```

samthor: Yup. I don't think we'd ever set the boolean to anything but `isDev` though. Happy to change this if we ever need to.

robdodson:
> Right but we don't deploy the old CSS. That scenario you posit will happen anyway since those browsers don't have access to the older CSS.

I'm confused. Why would they request old CSS if we tell the intermediaries to never cache stale html? Wouldn't the html always be current then?

> I'm not worried about a user seeing a more up-to-date version of our CSS or JS, even if they have slightly older HTML, although we should try to avoid that too.

It seems like this could also cause breakages (we remove something in the new CSS or new JS is trying to reference a thing that doesn't exist on the page). Maybe #3362 will resolve it

samthor:
We saw stale HTML in practice.

I'm happy to revisit the partial hashes stuff next week, we can chat about it before merging anything.

robdodson:
> We saw stale HTML in practice.

oh yeah, I think that may be because we weren't using `no-store`. I'm curious if we used `no-store` if we would hit that problem again

PR #3434, opened 2020-07-03
Ship post event flag. Add note to Q&A

### LINKED ISSUE #3428: [Post event] Add Q&A module
Hey folks,

Here's the first set of questions across the Web Vitals and Performance categories. As a reminder, would love to have a message on top of the Q&A module where we can let folks know that we'll have more Q&A coming soon.

Thanks!

robdodson:
Got a bit confused as to who was supposed to ship this so I did a quick PR this morning.

Fixes #3428

- Ships post event phase 
- Enables Q&A
- Adds note to Q&A section

### DIFF
--- a/src/site/_data/event.js
+++ b/src/site/_data/event.js
@@ -597,8 +597,8 @@ const qAndAs = [
 
 module.exports = {
   isPreEvent: false,
-  isDuringEvent: true,
-  isPostEvent: false,
+  isDuringEvent: false,
+  isPostEvent: true,
   days,
   communityEvents,
   qAndAs,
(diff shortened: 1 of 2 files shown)


robdodson:
The percy snapshot for the map looks off (the map doesn't load) but I suspect that's because the maps API call isn't working for whatever reason when percy is running it.

harleenkbatra08:
Hey @samthor in the note for Q&A, the line "Keep an eye on this space as we add more questions in the coming days." has been repeated twice. Could you please remove the one that's hidden in the first para so that it stands out and isn't repeated? Thanks!

PR #3388, opened 2020-06-26
Update getInstalledRelatedApps() for Windows support.

jpmedley:
@petele I've used the `scheduled` tag so that we don't need to remember to push this at the end of next month.

PR #3454, opened 2020-07-08
Revert "Update getInstalledRelatedApps() for Windows support."

kaycebasques:
Reverts GoogleChrome/web.dev#3388

### DIFF
--- a/src/site/content/en/blog/get-installed-related-apps/index.md
+++ b/src/site/content/en/blog/get-installed-related-apps/index.md
@@ -1,17 +1,11 @@
 ---
 title: "Is your app installed? getInstalledRelatedApps() will tell you!"
-subhead: |
-  "The `getInstalledRelatedApps()` method allows your web app to check whether
-  your native app or PWA is installed on a user's device."
+subhead: "The `getInstalledRelatedApps()` method allows your web app to check whether your native app or PWA is installed on a user's device."
 authors:
   - petelepage
-description: |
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/get-installed-related-apps/index.md
```diff
@@ -1,17 +1,11 @@
...
 authors:
   - petelepage
-description: |
-  The getInstalledRelatedApps() API is a new web platform API that allows your
-  web app to check whether your native app or PWA is installed on the user's
-  device.
+description: The getInstalledRelatedApps() API is a new web platform API that allows your web app to check whether your native app or PWA is installed on the user's device.
 date: 2018-12-20
-updated: 2020-07-22
-scheduled: true
```
kaycebasques, line 14: Hey @jpmedley & @petele after I merged #3388 I realized that I think the intended use of `scheduled` here was to specify when the updates to the post should go live. I'm not sure if `scheduled` is that sophisticated so I'm reverting #3388 in case you don't want these updates live yet.

cc @robdodson for awareness about a use of the `scheduled` flag that perhaps you weren't anticipating

robdodson: correct, the scheduled feature only works with new files. I think if you were to put it onto an existing doc it would actually unpublish it (because it converts it back to being `draft: true` essentially)

PR #3474, opened 2020-07-10
Nested lists look ugly

agektmr:
Simple list in web.dev looks great, but once it's nested, it looks ugly.

Screen Shot 2020-07-10 at 17 04 01

The above image is current state with simple nested list. Notice there's unnecessary space before and after nested lists. These are because `ul` and `ol` tags have extra margin at the top and the bottom.

Screen Shot 2020-07-10 at 17 02 56

With this change, the nested lists look much better with minimized margin for top and bottom.

This change will make nested lists like [this](https://web.dev/chrome-ux-report-api/#what's-next) look better. (And I have a pull request to make it look better.)

### DIFF
--- a/src/styles/generic/_lists.scss
+++ b/src/styles/generic/_lists.scss
@@ -88,6 +88,11 @@ ol:not([class]) > li > p:last-child {
   margin-bottom: 0;
 }
 
+li > ul:not([class]),
+li > ol:not([class]) {
+  margin: 8px 0;
+}
+
 .w-unstyled-list {
   list-style: none;
 }


jpmedley:
Nested lists are considered an anti-pattern in formal writing. Improving the styling should be done regardless, but employing one of the tricks for avoiding them is always the better options. 

Can you share the content where you ran into this?

agektmr:
Two examples:
https://web.dev/install-criteria/
https://web.dev/chrome-ux-report-api/#what's-next
And one from my pull request
https://deploy-preview-3472--web-dev-staging.netlify.app/registering-a-web-based-payment-app/#configuring-a-payment-app-with-a-web-app-manifest

PR #3476, opened 2020-07-10
Cleanup of /en/ directory

devnook:


src/site/content/en/index.njk
```diff
@@ -6,6 +6,8 @@ description: |
   guidance and analysis from web.dev.
 show_banner: false
 date: 2018-11-05
+eleventyComputed:
+  canonicalUrl: "/"
```
samthor, line 10: I'm confused about this line. This makes the English homepage canonical?

devnook: We added canonicalUrl property to each page (e.g. see #3354 )  to be able to find "canonical version of a post" (e.g. for posts at urls:  
/blah 
/en/blah
/pl/blah 
canonicalUrl is /blah, because it contains the original (untranslated) version of the post.

robdodson:
@devnook would you mind writing a description for the PR?

src/site/content/en/index.njk
robdodson: We should merge #3354 before this PR lands then, correct?

robdodson: just merged #3354

devnook: No, this is just an example of usage of canonicalUrl (PRs don't depend on each other)

PR #3486, opened 2020-07-13
Allow to build eleventy for only 1 target language at a time

devnook:
Right now, our eleventy build process assumes we always rebuild the whole site (even in watchers). As partial builds are not available in eleventy, I propose to allow for building only part of the site (in this case, files for 1 language, like "en") to improve the speed of builds in some development workflows that do not need the entire site to be present (e.g. authoring a particular post or updating css).  

Depends on #3476 https://github.com/GoogleChrome/web.dev/pull/3476


robdodson:
@devnook would you mind writing a description for the PR?

devnook:
PTAL

.eleventy.js
```diff
@@ -264,12 +264,13 @@ module.exports = function (config) {
   config.setDataDeepMerge(true);
 
   // https://www.11ty.io/docs/config/#configuration-options
+  const targetLang = process.env.ELEVENTY_LANG || '';
   return {
     dir: {
-      input: 'src/site/content',
+      input: 'src/site/content/' + targetLang,
```
MichaelSolati, line 270: `path.join()`?

robdodson: +1

devnook:
PTAL

.eleventy.js
devnook: ok

robdodson:
One issue I noticed is if you do `ELEVENTY_LANG=pl npm run dev` it'll throw an error because many of the authors don't have posts. I think we'll need to change that behavior, otherwise we won't be able to translate some of the docs to pl but let the rest fall back to en. Joe hit this in #3537

It seems like if someone is in the authorsData.json but doesn't have content we should just ignore it because folks find the warning and the error annoying.

PR #3503, opened 2020-07-15
Add language routing to internal navigation in SPA

devnook:
This PR implements the following language recognition flow:

In the client (SPA, during internal navigations, e.g. click on /foo):
=> 1. Read language from the current url (eg "pl" in /pl/bar). If not found:
=> 2. Read language from a cookie ("preferred_lang"). If not found:
=> 3. Read language from browser settings (navigator.language). If not found:
=> 4. Use defaultLanguage ("en").
=> 5. Update location.href & load partial accordingly.
Note: If localised content in PL missing, an EN version will be shown, but the url will stay PL (as the "frame" template stays in PL and only the partial is in EN. Also so that it can be used for future navigations which may not be missing).

In server (first render, direct visit to a url, e.g. visit to /pl/foo):
=> 1. Read language from the current url (eg "pl" in /pl/bar).  If not found:
=> 2. Read language from a cookie ("preferred_lang"). If not found:
=> 3. Read language from 'Accept-Language' header.
=> 4. If not found, or same as default language, continue with /en/ content.
=> 5. If found, and resource is present, serve localised content.
=> 6. If resource is not present, redirect to canonical content (without language in url). 
Note: If localised content in PL missing, an EN version will be shown, and the url will convert to a canonical url (e.g. /foo).

Please note that in order to be able to test it, you need to have line 13 uncommented:
https://github.com/GoogleChrome/web.dev/compare/master...devnook-spa?expand=1#diff-c4a591823cbbc82a83f39c11e9f0fb7fR13

Is should be commented out before merge, until we have polish content ready to serve.


samthor:
This PR will make it so:

1. I load /foo, I end up on English /foo
2. I navigate to /bar, this actually loads /en/bar/

Is that correct?

devnook:
FYI Currently available urls in "pl", for testing, are:
https://deploy-preview-3503--web-dev-staging.netlify.app/pl/accessible/ and 
https://deploy-preview-3503--web-dev-staging.netlify.app/pl/control-focus-with-tabindex/

src/lib/actions.js
```diff
@@ -195,10 +195,17 @@ function getCanonicalPath(path) {
...
   ({userPreferredLanguage}) => {
-    if (!userPreferredLanguage) {
-      return {
-        userPreferredLanguage: cookies.get('preferred_lang'),
-      };
+    userPreferredLanguage =
+      // Use currently set language.
+      userPreferredLanguage ||
+      // Or check in the url.
+      lang.getLanguageFromPath(location.pathname) ||
```
samthor, line 202: If I switch to PL using the dropdown but then type "web.dev/en/blah", I get the English page then further links switch back to Polish.

devnook: This happens currently because in the repo we have now an incomplete pl directory. English page is a fallback for Polish urls that might be missing in /pl/ directory. If you go to a page that has currently a polish version - e.g. /accessible/ and then switch language, you should see the correct behavior.

src/site/_includes/layout.njk
```diff
@@ -6,7 +6,6 @@ eleventyComputed:
 ---
 
 <!DOCTYPE html>
-{% set lang = "en" %}
 {% set locale="en_US" %}
 <html lang="{{lang}}">
```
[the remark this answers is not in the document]
devnook: Yup.

src/site/content/pl/pl.11tydata.js
```diff
@@ -7,11 +10,23 @@
...
+  ].filter(livePaths);
+
   const lang = 'pl';
-  const translation = 'none'; // Default translation status.
 
   return {
     lang,
-    translation,
+    translation: 'none',
+    test: 'im pl',
```
samthor, line 27: What is this?

src/lib/utils/router.js
```diff
@@ -107,6 +109,12 @@ export function listen(handler) {
     if (!isNavigation) {
       url = getUrl();
       hash = window.location.hash;
+    } else {
+      // If user has a preferred language, show it in the url.
```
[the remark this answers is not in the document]
samthor: Can you still update this comment?

src/lib/actions.js
devnook: I pushed the full language content to staging: https://web-dev-staging.uc.r.appspot.com/ so you can safely test there.

devnook:
PTAL

I pushed the full language content to staging: https://web-dev-staging.uc.r.appspot.com/ so it is best to test there.

src/site/_includes/layout.njk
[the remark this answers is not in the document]
robdodson: Should we also set locale in 11tydata.js?

src/site/_includes/layout.njk
```diff
@@ -171,6 +170,7 @@ eleventyComputed:
         {% if show_banner %}
           {% include 'partials/banner.njk' %}
         {% endif %}
+        {% include 'partials/translation-disclaimer.njk' %}
```
robdodson, line 173: I was only able to trigger this banner by going to localhost:8080/pl/

Here's the code for the partial:
```
{% set gitlocalizeUrl =  page.inputPath | gitlocalizeLink(lang) %}
{% if translation %}
{% Banner 'neutral' %}
    {% if translation === 'none' %}
    This page is not translated yet.
    {% elif translation === 'machine' %}
    This page is automatically translated using the Google Translation
    API.
    {% elif translation === 'manual' %}
    This article was translated from
    [the original post](/en{{ page.data.canonicalUrl }}).
    {% endif %}
    {% if gitlocalizeUrl %}
    Consider improving it by [contributing your translation]({{ gitlocalizeUrl }}).
    {% endif %}
{% endBanner %}
{% endif %}
```

- Since we redirect articles with missing translations would they ever show the `'none'` condition?
- Why does the homepage show the `'none'` condition but articles don't? Should the homepage do a similar redirect?

devnook: This is because we don't have a pl translation folder fully checked in to the repo. Please use stage (https://web-dev-staging.uc.r.appspot.com/pl/measure/) instead of local dev env and you should see banner everywhere in /pl/

src/lib/utils/language.js
```diff
@@ -10,9 +10,15 @@
  */
 const languageNames = {
   en: 'English',
-  // pl: 'Polish', TODO: Uncomment when PL language is prod-ready.
+  pl: 'Polish', // TODO: Uncomment when PL language is prod-ready.
```
robdodson, line 13: Will you comment this back out before landing the PR?

src/site/_includes/layout.njk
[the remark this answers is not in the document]
devnook: Yes, we can.

src/site/_includes/layout.njk
robdodson: ah I see

src/lib/utils/language.js
devnook: Yes, that's the plan.

PR #3509, opened 2020-07-15
chore: convert learning paths to json. update codeowners.

### LINKED ISSUE #3506: Convert _data/paths to json.
This will make them safer to contribute to. I don't think these files need to be .js so it may be as simple as changing their extensions and fixing quote marks.

robdodson:
Fixes #3506

- Add two new teams, @GoogleChrome/web-dev-eng and @GoogleChrome/web-dev-content 
- Add GH teams for eng and writers to CODEOWNERS. This will make it easier for the maintainers to give access control. @kaycebasques is the maintainer for web-dev-content.
- Convert the files in _data/paths over to json so they're safer to edit and allow codeownership.


kaycebasques:
Nitpick: `@GoogleChrome/web-dev-content` instead of `@GoogleChrome/web-dev-writers` but no big deal if this is a pain to change and you want to keep it as `@GoogleChrome/web-dev-writers`

The reason being is that I've sometimes referred to my team as the writers and then other people in Web DevRel were confused because they've written lots of content for the site

robdodson:
@kaycebasques np. I updated the team to @GoogleChrome/web-dev-content

robdodson:
PR is failing on an unrelated test. Let me see if I can fix that.

src/site/_data/paths/discoverable.json
```diff
@@ -0,0 +1,14 @@
+{
```
samthor, line 1: super nit: not formatted correctly

.github/CODEOWNERS
```diff
@@ -3,34 +3,35 @@
 # Read more @ https://help.github.com/articles/about-code-owners/
 
 # Default owners for everything
-* @robdodson @samthor
+* @GoogleChrome/web-dev-eng
```
samthor, line 6: actually, where are these groups defined?

robdodson: https://github.com/orgs/GoogleChrome/teams

robdodson:
> As a total curveball, I wonder if it's ever worth switching to say, JSON5 so we can have comments but still be fundamentally non-code data.

I guess if it doesn't make node or eleventy grump then i'd be cool with exploring that

PR #3510, opened 2020-07-16
Fix typo

deadlyhifi:
Missing “you” in “If you don't use an image CDN need to save…”


deadlyhifi:
@googlebot I signed it!

PR #3514, opened 2020-07-16
Filter blog collection by language

devnook:
On a localized blog home page (/pl/blog) we want to show blog posts in particular language (e.g. pl).


src/site/content/en/lang.11tydata.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
robdodson, line 1: Does this file work? Eleventy's docs say:

image

https://www.11ty.dev/docs/data-template-dir/

robdodson: I opened an issue about it a while back https://github.com/11ty/eleventy/issues/492

robdodson:
Percy failure can be ignored. It's just the top nav underline being flaky
image

src/site/content/en/lang.11tydata.js
MichaelSolati: I don't think it'll trickle to other files, but seeing how it's being required in the blog it should be fine.

devnook: Actually it does not need to be an 11tydata file, normal .js would do too.

robdodson:
@MichaelSolati @devnook Circling back to this. I was wondering... why have a `lang.js` file when we could put that info into the `en.11tydata.js` file and have other files require `en.11tydata.js` if they need to access it?

PR #3515, opened 2020-07-16
Create separate performance section on /learn, update captions

khempenius:
Given that /vitals, /metrics, and /fast are highly coupled, I think grouping them together makes it easier to locate content - for both the "performance" and "realize what the web can be" sections.

Before
current layout

After
Screen Shot 2020-07-16 at 5 36 13 PM

cc @philipwalton

### DIFF
--- a/src/site/_data/paths/fast.json
+++ b/src/site/_data/paths/fast.json
@@ -3,7 +3,7 @@
   "cover": "/images/collections/fast.svg",
   "title": "Fast load times",
   "updated": "July 17, 2020",
-  "description": "Guarantee your site loads quickly to avoid user drop off.",
+  "description": "Techniques for improving site performance.",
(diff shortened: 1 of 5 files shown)


robdodson:
@khempenius this looks cool and I'm not opposed to it but I was wondering if you think we could also retitle the section that currently say "Realize what the web can be".

I guess "Fundamentals" is not allowed 😅  so we need something else.

Also note that we just changed files like fast.js over to being .json to make it safer for folks to edit (less worry of folks sneaking in executable code) so you have a few merge conflicts.

kaycebasques:
I like it but this is a bit tricky because web.dev/pwa is a major driver of traffic to the site.

We can rephrase `Realize what the web can be` to `Build excellent websites`.

khempenius:
Fixed merge conflict and switched "Realize what the web can be" to "Build excellent websites".

robdodson:
> By the way let me note for the record that I've been lax about homepage changes in the past but I think this should ultimately be my judgment call because it pretty squarely falls under the category of content changes. In this case we're re-organizing the presentation of content. LMK if you see differently @robdodson and want to discuss 1:1.

that sounds fine to me 👍

khempenius:
Fixed 👍

kaycebasques:
Percy changes are expected

PR #3522, opened 2020-07-17
Restore reverted #3503 and fix #3519 (lang bug)

### LINKED ISSUE #3519: The site is not working
**Describe the bug**
The site is not working, im clicking in every place and nothing is refreshed, nothing is loaded, the site doesn't work. 

**To Reproduce**
1. Go to 'anything'
2. Click on 'anything'

**Expected behavior**
Site works , go to blog , test site , etc , nothing works

**Screenshots**

devnook:
Fixes #3519 

The fix is in the second commit: https://github.com/GoogleChrome/web.dev/commit/f8df17086e273a0cf6253d6cb9e958c6e06528df

The rest of the changes was reviewed previously in #3503


src/lib/utils/router.js
```diff
@@ -113,7 +114,7 @@ export function listen(handler) {
       // If user has a preferred language, use it in the request.
       const lang = store.getState().userPreferredLanguage;
       if (lang && lang !== language.defaultLanguage) {
-        url = '/' + lang + url;
+        url = path.join('/', lang, url);
```
robdodson, line 117: ah I overlooked the fact that `path` isn't available in the browser! 🤦‍♂️

I think `path-browserify` adds 53kb to the bundle when it's unpacked so it's probably not worth the size. I think we can just use what you had before: `url = '/' + lang + url;`

devnook: ok!

devnook:
PTAL

PR #3528, opened 2020-07-17
Filter by lang tags and authors collections

devnook:
Similar filtering as in /blog, used to filter authors, tags and newsletters pages.


robdodson:
@MichaelSolati would you mind adding descriptions to the functions in `hooks/utils.js`? I don't think I understand what the `individual` function is doing.

robdodson:
I think it's ok to push those comments to this PR

MichaelSolati:
I'll add some docs to the hooks

PR #3534, opened 2020-07-20
Disable updated hook.

### LINKED ISSUE #3368: Automatic update of updated field might be more trouble than it's worth
**Describe the bug**

When our contributors update their branches, a bunch of unrelated files sometimes get updated:

image

The PR says that there's 74 modified files while this PR has actually only touched 5-20.

This just happened to @samdutton and I believe it happened to @rachelandrew also.

This can be risky and burdensome to fix because it seems like it requires rewriting git history. So there's the possibility of contributors losing their draft content if they do it wrong. I've been telling people to just copy their content over into a fresh branch.

**To Reproduce**

I don't really know. Our contributors are using some git workflow that doesn't jive well with the auto-update behavior. Rather than training all of our contributors to use a specific git workflow I think it's easier to just disable the auto-update logic and use other means for ensuring that the `updated` field stays up-to-date.

robdodson:
Fixes #3368

- Don't run `update-updated` as a pre-commit hook.
- Rename npm script from `precommit` to `updated`. If folks want to use the script they can still do `npm run updated`.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -2,7 +2,6 @@
   "//1": "Use git commit --no-verify to bypass the pre-commit hook",
   "//2": "Use git push --no-verify to bypass the pre-push hook",
   "hooks": {
-    "pre-commit": "npm run precommit",
     "pre-push": "npm run lint"
   }
 }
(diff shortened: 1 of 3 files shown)


kaycebasques:
I'm going to merge this because the auto-updated just munged another PR

PR #3535, opened 2020-07-21
Fix searchbox

robdodson:
While recording a screencast for Una I noticed a few issues with the search box.

If you were using a screen reader it would close every time you pressed the up arrow key. I think this is because the cursor was moving to the beginning of the text field and briefly losing focus. Adding preventDefault for the relevant keys seemed to fix this.

I also took it as an opportunity to replace some of the existing material icons with actual SVG.

- Fix broken screen reader behavior on search box
- Switch to svg icons
- Remove generic/_buttons.scss. This file was setting a min-width that we copied from devsite and it was messing with my icon buttons. I don't think it's really useful anywhere else.


src/lib/components/Search/_styles.scss
```diff
@@ -62,16 +62,6 @@ web-search {
     outline: none;
   }
 
-  .web-search__open-btn::before {
-    @include font-material-icon();
-    content: 'search';
-  }
-
-  .web-search__close-btn::before {
-    @include font-material-icon();
```
samthor, line 71: very yes 👍

PR #3540, opened 2020-07-21
feat(q-and-a): add questions from day 2, fixes #3511

### LINKED ISSUE #3511: Adding more questions in the Q&A section on web.dev/live
Hi there,

I've finally managed to curate top questions from day 2 - PWA & Tools that we can add to the Q&A section. Could you please help add them? They're available here: https://docs.google.com/document/d/1_-bo8thRmGMeIL1uNtLpXzgi6sok9atGNqHcqK52u38/edit#bookmark=id.r8tpxd2tfccp

MichaelSolati:
Fixes #3511

Just added some more questions. However the list is getting crazy long, maybe a bit unruly... Maybe we can add a starred questions so we can limit the list on load? So instead of seeing all qs at first, you see starred qs (which would be much shorter).

@harleenkbatra08 thoughts?

### DIFF
--- a/src/lib/components/EventQandA/index.js
+++ b/src/lib/components/EventQandA/index.js
@@ -27,6 +27,12 @@ import './_styles.scss';
  * @final
  */
 class EventQAndA extends BaseElement {
+  static get properties() {
+    return {
+      selected: {type: String},
+    };
+  }
+
   constructor() {
     super();
     this.closeDetail = this.closeDetail.bind(this);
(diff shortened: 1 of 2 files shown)


robdodson:
@harleenkbatra08 you can see what Michael is referring to here: https://deploy-preview-3540--web-dev-staging.netlify.app/live/#q-and-a

what if we get rid of the All Categories option and just default it to the first category?

harleenkbatra08:
Yeah, agree with you Rob. Michael's suggestion also sounds neat but feels
like more work than it's worth. So happy to go with your solution, Rob :)

Thanks both!

MichaelSolati:
First category? I mean we don't really organize them... So I think it'll start with web vitals. Should I keep all, but move it so it's not the default?

harleenkbatra08:
Yeah I think that works. There's more questions but I'm going to look at
some data before we bother adding more. Not sure if we have a ton of
traffic on the page any more.

MichaelSolati:
@harleenkbatra08 Ok, so now it'll display the first category ("Web Vitals") and allow users to toggle between the options. "All categories" is now listed as the last option in the drop down, if a user wants to see all the questions.

@robdodson PTAL

PR #3541, opened 2020-07-21
refactor(tags): turn tags data into a json file

MichaelSolati:
Simple refactor of moving the logic that cooks the data in `_data/postTags.js` to `_collections/tags.js`. Use `_data/tagsData.json` for all tag data pre-cooking (also used the name to be consistent with the authors data json). And update places that use `_data/postTags.js` to use `_collections/tags.js`.

This should close out #3055


.github/workflows/lighthouse-ci-workflow.yml
```diff
@@ -4,14 +4,14 @@ name: Lighthouse CI
...
-    branches: 
+    branches:
       - master
   pull_request:
     paths:
       - '**.js'
       - '!src/site/_data/contributors.js'
       - '!src/site/_data/countries.js'
-      - '!src/site/_data/postTags.js'
+      - '!src/site/_data/tagsData.json'
```
robdodson, line 14: I think you can remove this line entirely. The line at the top is telling it to include only .js files. Since this is now a .json file it should be ignored.

.github/workflows/percy-workflow.yml
```diff
@@ -10,7 +10,7 @@ on:
       - '**.js'
       - '!src/site/_data/contributors.js'
       - '!src/site/_data/countries.js'
-      - '!src/site/_data/postTags.js'
+      - '!src/site/_data/tagsData.json'
```
robdodson, line 13: can remove this

.github/workflows/percy-workflow.yml
```diff
@@ -19,7 +19,7 @@ on:
...
```
robdodson, line 22: can remove this

robdodson:
LGTM, let's merge it in the morning just to be on the safe side :)

PR #3550, opened 2020-07-22
Remove required check action.

robdodson:
The required check workflow fails a build if it has the DO NOT MERGE label. However, this Action runs anytime anyone makes any change to a PR. It's currently running _a lot_ and it's very expensive to spin up an entire linux VM to do this check. Recently we've had issues with our Actions getting queued up and it seems like this is probably not helping.

If folks don't want a PR to be merged they should use [the Drafts feature](https://github.blog/2019-02-14-introducing-draft-pull-requests/). If they would like a PR to be reviewed, and approved, but not merged then they can still use the DO NOT MERGE label, but we'll no longer enforce a failing build with this Action. If folks feel like they still need the failing build then we can look into implementing this as a GitHub app using [Probot](https://github.com/probot/probot) which would be much cheaper as it's only a webhook.

- Remove required check workflow


robdodson:
fyi @jpmedley @kaycebasques

jpmedley:
Is there no low-resource way to just read the labels. GitHub's draft feature is clearly designed for an individual working in a pre-review stage.

robdodson:
> Is there no low-resource way to just read the labels.

We'd need to write a Github App to do it. We could probably do that using Probot (we used to use that on this repo). Just takes a bit of time to set up.

PR #3554, opened 2020-07-23
[WebHID article] Replace colon entity with just :

tomayac:

- The entity `&#58;` appears like that in search. We can just use a colon instead.

### DIFF
--- a/src/site/content/en/blog/hid-examples/index.md
+++ b/src/site/content/en/blog/hid-examples/index.md
@@ -1,5 +1,5 @@
 ---
-title: Human interface devices on the web&#58; a few quick examples
+title: Human interface devices on the web: a few quick examples
 subhead: Connecting to uncommon devices from your app.
 authors:
   - joemedley


jpmedley:
No because it breaks eleventy.

```
5:34:34 AM: > incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 2, column 42:
5:34:34 AM:      ... man interface devices on the web: a few quick examples
```

Also, that seems like a problem that search shouldn't have and will probably be fixed sooner rather than later.

tomayac:
> No because it breaks eleventy.
> 
> ```
> 5:34:34 AM: > incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 2, column 42:
> ```
> 
> Also, that seems like a problem that search shouldn't have and will probably be fixed sooner rather than later.

Yuck. Would an em-dash work instead of the colon in the title?

jpmedley:
I believe so.

src/site/content/en/blog/hid-examples/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Human interface devices on the web&#58; a few quick examples
+title: Human interface devices on the web: a few quick examples
```
tomayac, line 2:
```suggestion
title: Human interface devices on the web—a few quick examples
```

tomayac:
Alright, now using an em-dash. PTAL.

mathiasbynens:
> No because it breaks eleventy.
> 
> ```
> 5:34:34 AM: > incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 2, column 42:
> ```

FWIW, this is a YAML syntax issue, the fix for which is:

```diff
-title: Human interface devices on the web: a few quick examples
+title: 'Human interface devices on the web: a few quick examples'
```

tomayac:
Wow, so simple. Thanks, @mathiasbynens, for sharing the actual fix!

PR #3578, opened 2020-07-27
Moves PWA content into PWA folder

petele:
Updates to the PWA info architecture 
- Re-orders top line topics to move install closer to the top since it's a key component of PWAs
- Moves `badging`, `app shortcuts`, and `share target` into the PWA folder
  - These were originally posted as blog posts, but have been moved into PWA area since they're only available to installed PWAs.
- Creates new advanced topics section where deeper dive content will go
  - Moves PWA multi-origin from blog into advanced topics

See https://deploy-preview-3578--web-dev-staging.netlify.app/progressive-web-apps/


kaycebasques:
We should get @robdodson's input on this, because I believe that they are moving all content into a flat directory structure

petele:
@kaycebasques & @robdodson - sounds good. Even if everything goes flat, I assume we'll still have the landing pages and they'll look as they do now?

kaycebasques:
> Even if everything goes flat, I assume we'll still have the landing pages and they'll look as they do now?

@petele yes, the published site will look the same, this is just an implementation change

robdodson:
yeah we're planning to flatten things as it will make i18n _much_ easier. Currently we have these interstitial directories which aren't reflected in the urls and it makes things a bit confusing.

The appearance and function of the site will remain the same, it'll just be an internal implementation change. I think @devnook  is close to being at a point where she can flatten the structure, but she's currently OOO on vacation so this probably won't happen for another few weeks.

src/site/_data/paths/progressive-web-apps.json
```diff
@@ -48,6 +41,29 @@
...
+    },
+    {
+      "title": "Capable",
+      "pathItems": [
+        "app-like-pwas",
+        "progressively-enhance-your-pwa"
+      ],
+      "subtopics": [
+        {
+          "title": "Improve your experience with PWA exclusive features",
```
jpmedley, line 53:
```suggestion
          "title": "Improve your experience with PWA-exclusive features",
```

petele:
@kaycebasques - is there any harm in moving it?

robdodson:
> @kaycebasques - is there any harm in moving it?

I believe you'll need to add the `blog` tag to the items if you move them out of the blog directory. Otherwise they will stop appearing as blog posts. You should be able to test locally to confirm.

petele:
@robdodson - When I moved them out, and added the `blog` tag, I was getting an error when trying to view any of the landing pages. If I removing the `blog` tag from the post seems to fix it, and things render. The error said: 
`Error: ENOENT: no such file or directory, stat '/Users/petele/Sites/web.dev/dist/en/404/index.html`

This happens whether the files live in the `/blog/` directory or the `/pwa/` directory.

robdodson:
I'll take a look. That usually happens if you're refreshing the page while eleventy is still building.

petele:
> I'll take a look. That usually happens if you're refreshing the page while eleventy is still building.

Yah, I've seen that before, but I've tried waiting until I see the "Watching..." notice, to make sure the build is done.

robdodson:
@petele the watching notice is for gulp I think, not eleventy. You have to wait until it says "Wrote 681 files in 11.66 seconds"
annoyingly we have two processes printing to stdout at the same time.

image

I did a test of adding the `blog` tag to the app shortcuts article and it worked as expected (started appearing in the blog again).

petele:
I could have sworn I wait extra long just to make sure, but yep - that did it, thanks @robdodson 

I'm going to replace this PR with #3583, which does not move the content, but simply updates the tags and the landing page data.

PR #3585, opened 2020-07-28
Fix the HID examples title

jpmedley:

PR #3586, opened 2020-07-28
Disable flaky test.

robdodson:

- Disable a flaky test. #3585 was getting tripped up by the test and I've seen it happen to other PRs.


robdodson:
Going to self merge this to unblock folks. If we want to keep this test we could maybe look at increasing the timeout rate?

PR #3608, opened 2020-07-31
Fix code indentation for Trust Token examples

baz1:


baz1:
@googlebot I signed it!

PR #3613, opened 2020-07-31
[Shape Detection article] Move Barcode Detection up

tomayac:

- Barcode is supported w/o flag, so move it first.

### DIFF
--- a/src/site/content/en/blog/shape-detection/index.md
+++ b/src/site/content/en/blog/shape-detection/index.md
@@ -5,7 +5,7 @@ authors:
   - thomassteiner
 description: The Shape Detection API detects faces, barcodes, and text in images.
 date: 2019-01-07
-updated: 2020-07-31
+updated: 2020-08-10
 tags:
   - blog
   - capabilities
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/shape-detection/index.md
```diff
@@ -224,6 +201,29 @@ try {
 }
 ```
 
+### Working with the `FaceDetector` {: #facedetector}
+
+The `FaceDetector` always returns the bounding boxes of faces it detects in
+the `ImageBitmapSource`. Depending on the platform, more information
+regarding face landmarks like eyes, nose, or mouth may be available.
```
jpmedley, line 208: Is this API more about finding where the faces are than about identifying people? Maybe we should say something about that.

tomayac: There's an important difference in literature between face detection (there's *a* face) and face identification (there's a face, and it's *Joe's*). This API is only about detection.

src/site/content/en/blog/shape-detection/index.md
```diff
@@ -5,7 +5,7 @@ authors:
   - thomassteiner
 description: The Shape Detection API detects faces, barcodes, and text in images.
 date: 2019-01-07
-updated: 2020-06-04
+updated: 2020-07-31
```
tomayac, line 8:
```suggestion
updated: 2020-08-10
```

src/site/content/en/blog/shape-detection/index.md
tomayac, line 208:
```suggestion
regarding face landmarks like eyes, nose, or mouth may be available.
It is important to note that this API only detects faces.
It does not identify who a face belongs to.
```

PR #3620, opened 2020-08-01
feat(iframe): create iframe shortcode to enforce lazy loading of iframes, fixes #3154

### LINKED ISSUE #3154: Waiting on document load event to load js delays table of contents.
On pages with a lot of iframes, e.g., https://web.dev/web-animations/, it seems like maybe the iframes are pushing out the load event so it takes quite a long time for our custom elements to show up. This isn't a great situation because the table of contents needs to be a custom element but we're not loading it until many seconds after the page paints.

cc @samthor @MichaelSolati for thoughts.

MichaelSolati:


robdodson:
This is looking good but I think we should also add a Glitch shortcode (maybe it can use the IFrame shortcode in its implementation?)

```
{% Glitch 'fav-kitties' %}
```

It'll need to support three arguments:
- id (the id of the glitch to display)
- path (this is the file the glitch will display first)
- previewSize (this controls how much of the app preview is visible)

PR #3621, opened 2020-08-01
Update index.md

SimplGy:
Suggest rm this section because it's misleading.

This doesn't seem to be the case any more. While I see LCP on my timeline, I cannot find any such marks or measures in `Performance.entries()`.

I do see entries of type `paint` (not `largest-contentful-paint`), but not the largest-contentful-paint.

I would update with the new/correct way to get this measure, but I can't figure out how to see it, maybe it's no longer gettable?

image

Fixes #SOME_ISSUE_NUMBER

- 
- 
-

### DIFF
--- a/src/site/content/en/metrics/lcp/index.md
+++ b/src/site/content/en/metrics/lcp/index.md
@@ -148,14 +148,6 @@ considered.
 Web pages often load in stages, and as a result, it's possible that the largest
 element on the page might change.
 
-To handle this potential for change, the browser dispatches a
-[`PerformanceEntry`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry)
-of type `largest-contentful-paint` identifying the largest contentful element
-as soon as the browser has painted the first frame. But then, after rendering
-subsequent frames, it will dispatch another
-[`PerformanceEntry`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceEntry)
-any time the largest contentful element changes.
-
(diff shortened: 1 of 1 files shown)


SimplGy:
image

image

kaycebasques:
@philipwalton PTAL

SimplGy:
Hi Philip -- learned from a colleague that you actually can get LCP out programmatically, but it's not ever *stored* in `Performance.getEntriesByType('largest-contentful-paint')`, which is how I read the doc.

From MDN, I can get the value by observing mutations in real time:

```
      const po = new PerformanceObserver((entryList) => {
        const entries = entryList.getEntries();
        const lastEntry: any = entries[entries.length - 1];
        entry = lastEntry;
        console.log('lcp', lastEntry);
      });
      po.observe({type, buffered: true});
```

I can update the docs if this is the WAI way of getting this. Not sure what's happening though -- maybe the browser deletes the stored mark/measurement as soon as it's recorded because the design is for it to get updated frequently? I'm not sure I understand why this one is so different from the other paint marks and measures.

Would love to be able to add more color to the explanation of how it's being measured and the best way to get the most recent value.

Thanks!

philipwalton:
@SimplGy you are correct that you cannot use `performance.getEntriesByType('largest-contentful-paint')` to access `largest-contentful-paint` entries, and that's by design. Most new performance APIs are only accessible via `PerformanceObserver`, and that's because there isn't any "known" point in time where you could call `performance.getEntriesByType('largest-contentful-paint')` and be sure it would be available—which is why you need to register a callback to observe it.

I'm going to close this PR as I don't think there is anything misleading in the post. I understand why you were confused (the APIs are inconsistent), but the post is pretty clear on how to [measure LCP in JavaScript](https://web.dev/lcp/#measure-lcp-in-javascript), so I don't think any changes need to be made.

PR #3624, opened 2020-08-03
feat(Glitch): add shortcode for glitch, fixes #3337

### LINKED ISSUE #3337: A Glitch shortcode
It's easy for authors to accidentally use the wrong url for their glitch embed if they're writing the iframe code directly.

image

The top image using /edit/ is bad (it'll try to iframe the literal glitch site). The second image is correct as it uses /embed/.

We should make this easier for folks by providing a Glitch shortcode so they just tell us the id of their project and we do the right thing for them.

MichaelSolati:
Fixes #3337

### DIFF
--- a/src/site/_includes/components/Glitch.js
+++ b/src/site/_includes/components/Glitch.js
@@ -15,39 +15,45 @@
  */
 
 const {html} = require('common-tags');
+const iframe = require('./IFrame');
 
 /**
  *
- * @param {string | {src: string; allow?: string; style?: string; title?: string;}} param
+ * @param {string | { id: string; path?: string; previewSize?: number; style?: string; }} param
  * @return string
  */
 module.exports = (param) => {
-  let iframeProps = {
-    allow: '',
-    src: null,
-    style: 'height: 100%; width: 100%; border: 0;',
-    title: 'IFrame content',
+  const allow = 'geolocation; microphone; camera; midi; encrypted-media';
+  let glitchProps = {
+    id: null,
+    path: '',
+    previewSize: 0,
+    style: 'height: 420px; width: 100%;',
   };
 
   if (typeof param === 'string') {
-    iframeProps.src = param;
+    glitchProps.id = param;
   } else if (param.constructor === {}.constructor) {
-    iframeProps = {...iframeProps, ...param};
+    glitchProps = {...glitchProps, ...param};
   }
 
-  const {allow, src, style, title} = iframeProps;
+  const {id, path, previewSize, style} = glitchProps;
 
-  if (!src) {
+  if (!id) {
     return;
   }
 
+  let src = `https://glitch.com/embed/#!/embed/${id}?`;
+  if (path) {
+    src += `path=${path}&`;
+  }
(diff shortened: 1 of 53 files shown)


src/site/_includes/components/Glitch.js
```diff
@@ -0,0 +1,59 @@
...
+    glitchProps = {...glitchProps, ...param};
+  }
+
+  const {id, path, previewSize, style} = glitchProps;
+
+  if (!id) {
+    return;
+  }
+
+  let src = `https://glitch.com/embed/#!/embed/${id}?`;
```
samthor, line 46: I'd be happier if we used the URL constructor here for escaping reasons. But also `${id}` would need to be escaped manually anyway. Can we just escape `path` and `id`?

MichaelSolati: URL constructor doesn't seem to work here as it'll strip the necessary `/#!` however I can use `querystring.escape` on the id and `querystring.stringify` for an object of the query params (which includes the path).

src/site/content/en/handbook/web-dev-components/index.md
```diff
@@ -663,30 +663,28 @@ at.
...
 
-Query parameters for modifying how the embed is presented:
+Shortcode object fields allow for modifying how the embed is presented:
 
-* `previewSize=100` displays the app, rather than the source code.
-* `path=index.html` lets you specify which source code file to show.
-* `sidebarCollapsed=true` collapses the file navigator sidebar.
+* {`string`} `id` ID of Glitch project.
+* {`string`} `path?` Lets you specify which source code file to show.
+* {`number`} `previewSize?` Displays the app, rather than the source code.
```
robdodson, line 686: I'd add to this that previewSize 0 will just display the code and previewSize 100 will just display the app

src/site/_includes/components/Glitch.js
```diff
@@ -0,0 +1,67 @@
...
+ * @param {string | { id: string; path?: string; previewSize?: number; style?: string; }} param
+ * @return string
+ */
+module.exports = (param) => {
+  const allow =
+    'camera; clipboard; clipboard-read; clipboard-write; geolocation; encrypted-media; microphone; midi; vr;';
+  let glitchProps = {
+    id: null,
+    path: '',
+    previewSize: null,
```
robdodson, line 32: kind of wondering if this should default to 100 as that looks like what the majority of embeds use?

robdodson:
@kaycebasques @jpmedley wanted to let y'all know that this + a new IFrame shortcode have landed (https://github.com/GoogleChrome/web.dev/pull/3620).

src/site/content/en/handbook/web-dev-components/index.md
MichaelSolati: I'll throw together another quick PR tomorrow the two bits of feedback.

MichaelSolati: @robdodson should I just do a boolean instead of numebrs? If we're just using 0 (false) and 100 (true)?

PR #3628, opened 2020-08-04
Add manual build trigger and updated scheduled build cron to be hourly.

robdodson:
Moves the site to an hourly deploy cadence (Mon-Fri), instead of deploying every time we merge something into master. Also adds a 'Run Workflow' button so folks can manually deploy the site on-demand.

**Why do this?**
We've recently been seeing issues with App Engine failing our builds. The App Engine team is still debugging why this is happening, but one likely culprit is when we queue multiple builds to run at the same time. This is pretty easy to do if we merge two PRs close to one another. It's also wasteful to enqueue so many builds. This change will essentially bucket our builds, so if we merge 10 PRs in an hour, it'll only do 1 deploy. The addition of the manual trigger also means folks can deploy the site immediately if they need to.

- Merging to master will no longer deploy the site. This is because we often see folks going through and merging several PRs at once, which queues up a number of builds that begin to race one another and seem to be breaking App Engine.
- Adds [a manual trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) to the Action. This will add a button to the Actions tab so folks on the team can easily manually deploy the site.
- Updates the schedule so the site does an automated build every hour, Mon-Fri.
- Renames 'Continuous Integration' task to 'Deploy' so it's easier for folks to grok at a glance.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -4,11 +4,9 @@ on:
   # manually deploy the site.
   workflow_dispatch:
   
-  # Automatically build and deploy the site hourly,
-  # between the hours of 5am and 5pm
-  # every day-of-week from Monday through Thursday.
+  # Automatically build and deploy the site hourly, Mon-Fri
   schedule:
-    - cron: '0 5-17 * * 1-4'
+    - cron: '0 * * * 1-5'
 
 jobs:
   ci:
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -1,11 +1,14 @@
-name: Continuous integration
+name: Deploy
 on:
-  push:
-    branches:
-      - master
+  # Adds a 'Run Workflow' button to the GitHub Actions UI so the team can
+  # manually deploy the site.
+  workflow_dispatch:
+  
+  # Automatically build and deploy the site hourly,
+  # between the hours of 5am and 5pm
+  # every day-of-week from Monday through Thursday.
   schedule:
-    # At 15:00 UTC / 7:00 PST on every day-of-week from Monday through Friday.
-    - cron: '0 15 * * 1-5'
+    - cron: '0 5-17 * * 1-4'
 
 jobs:
   ci:


robdodson:
cc @kaycebasques @jpmedley

kaycebasques:
Might be good to document this under https://web.dev/handbook/#content-infrastructure

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -1,11 +1,14 @@
...
 on:
-  push:
-    branches:
-      - master
+  # Adds a 'Run Workflow' button to the GitHub Actions UI so the team can
+  # manually deploy the site.
+  workflow_dispatch:
+  
+  # Automatically build and deploy the site hourly,
+  # between the hours of 5am and 5pm
```
mathiasbynens, line 8: Why only during these hours? web.dev contributors are spread across all kinds of timezones. Let's just deploy once per hour?

jpmedley:
Looks awesome. 

Can you confirm my new procedure. I usually have to merge three or four articles or updates on the day before beta. It seems like my procedure on that day should be:

1. Merge articles as needed throughout the day.
2. Click deploy late in the day.

Step 2 brings up another point. Can I get deploy access?

.github/workflows/continuous-integration-workflow.yml
robdodson: Yeah I've gone back and forth on this. Originally I had it hourly Mon-Fri but was worried about something bad happening when the web.dev team was asleep (maybe a deploy that takes down the site or something).

But since the web.dev team is spread out between zurich-sf-sydney, we have quite a lot of overlap and could probably roll back easily enough.

What do you think about us not doing automatic deploys on Friday? I've gone back and forth on that one as well. I'd like to avoid us having to debug on the weekends if possible.

robdodson:
@jpmedley 

> Merge articles as needed throughout the day.

Those articles would go live 1hr after you merged them, so that doesn't seem like what you'd want. The above changes make it so the site deploys (essentially) hourly Mon-Thur. Check out the thread above with Mathias.

> Step 2 brings up another point. Can I get deploy access?

I _think_ everyone who has write access to the repo (you and the rest of devrel) will be able to click the deploy button but I haven't confirmed that yet. GitHub only has a few tiers of repo access (read, write, admin) and it's not clear to me how those roles tie in to Actions permissions. I guess we'll find out when this lands :)

robdodson:
> Those articles would go live 1hr after you merged them, so that doesn't seem like what you'd want. The above changes make it so the site deploys (essentially) hourly Mon-Thur. Check out the thread above with Mathias.

Actually maybe it is what you want. If you're ok with the articles going live shortly after you merge them then this should work fine and you'd only need to do a manual deploy if you wanted something to go live instantly.

.github/workflows/continuous-integration-workflow.yml
mathiasbynens: My gut feeling would be to start with hourly deploys Mon-Fri, and then seeing what happens. It should be an _extremely_ rare event that a successful build (needed to trigger a deployment) results in the site being down, so I wouldn't optimize for it unless this actually occurs. Just my $0.02, it's totally your call!

Re: my original suggestion, making the deploy hourly 24 times a day would be more inclusive, timezone-wise.

jpmedley:
Being live an hour after isn't a problem. I'm only picky about the date because it's stuck between two immovable parts of the schedule. Authors need as much time to create as possible, yet articles must absolutely be up before the beta post goes live. Making things live the day before beta adds padding in case something goes wrong.

robdodson:
Per the discussion with @mathiasbynens above, I decided to switch this to hourly, Mon-Fri. We'll monitor this and adjust if we want to deploy less often or if we hit issues with deploying on Fridays.

PR #3640, opened 2020-08-05
Replace hero image on Badging API article

petele:
Previous hero image implied badging would work on iOS & mobile, when it doesn't, so providing a more generic hero image.


jpmedley:
My favorite hero images real-world images that capture dead-on something about the digital world.

Issue #3605, opened 2020-07-30
Clean up CLI output so authors know when to refresh the page.


robdodson:
When you run `npm run dev` we're actually running a few watchers in parallel, including eleventy. Unfortunately eleventy takes a while to build but what folks see in the console is this:
![image](https://user-images.githubusercontent.com/1066253/88971171-6750c600-d268-11ea-8655-d8ad116a891b.png)

However, if you wait an additional 12 seconds you'll see this:
![image](https://user-images.githubusercontent.com/1066253/88971211-76377880-d268-11ea-881b-22e055e5b3b9.png)

It's only at this point that eleventy is done building and you can actually refresh the page.

We should see if we can make the output in the terminal more clear so folks know that eleventy is still building. Maybe by suppressing all of the other log messages?

robdodson:
cc @petele @kaycebasques @jpmedley

petele:
Personally I like the time stamp that gulp spits out, makes it easy to verify it built after I hit save. Beyond that, for me it's helpful to know it's currently building, and when it's ready. I don't *think* I care about the build steps that Gulp spits out.

robdodson:
Gulp has a nice feature called [`ignoreInitial`](https://gulpjs.com/docs/en/api/watch/#options) which you can pass to `watch()` which tells it to not run the task the first time you call `gulp watch` but to only react to file changes.

I'm kind of curious if we could move _all_ of our tasks into gulp and use it as our file watcher, instead of depending on the rollup's file watcher + eleventy's file watcher, etc.

Eleventy's file watcher already has some issues in that it won't rerun eleventy when you modify `.eleventy.js` or any of the shortcodes ([issue](https://github.com/11ty/eleventy/issues/1052)).

robdodson:
If we did this, we could rewrite the `npm run dev` command to be:

```
"dev": "npm-run-all clean gulp --parallel --race gulp:watch server",
```

If we're using gulp to trigger eleventy, then this will ensure eleventy has built before the task says "Watching...".

The alternative is to do this:

```
"dev": "npm-run-all clean gulp rollup eleventy --parallel --race gulp:watch rollup:watch eleventy:watch server",
```

Unfortunately this will run eleventy twice because calling `eleventy --watch` (which is what `eleventy:watch does) will trigger an eleventy build.

robdodson:
Submitted an issue for this to the eleventy repo.
https://github.com/11ty/eleventy/issues/1336

[...]

PR #3654, opened 2020-08-07
Clean up CLI output a bit.

robdodson:
This improves #3605 a little bit but doesn't fully fix the problem.

I think after [Eleventy's `afterBuild` events ship in 1.0](https://www.11ty.dev/docs/events/) we should be able to use https://www.npmjs.com/package/cli-spinner to show when a build is in progress or finished.


src/site/_filters/find-by-url.js
```diff
@@ -33,11 +30,6 @@ let memo;
  * @see {@link https://github.com/11ty/eleventy/issues/399}
  */
 const memoize = (collection) => {
```
robdodson, line 32: The warning that was previously here was essentially noise (it logged every time you edited a file in dev mode) so I just removed it.

.eleventy.js
```diff
@@ -91,6 +92,7 @@ const {
 } = require(`./${transformsDir}/service-worker-partials`);
 
 module.exports = function (config) {
+  console.log(`${chalk.black.bgGreen('Eleventy is building, please wait…')}`);
```
samthor, line 95: This only happens once, right?

Also, super nit, the `${` doesn't really help here, we could just remove the wrapping.

samthor: Also, not sure if you're a fan, but we use `fancy-log` in a few place which includes timestamp on the left. We could use it here too.

robdodson: Yeah, that's why it doesn't help a ton. It happens when you first run `npm run dev` but if you subsequently edit a file it won't log it again.

> Also, super nit, the ${ doesn't really help here, we could just remove the wrapping.
oh right, I was combining strings earlier and then put it all into the same message. I'll fix it.

PR #3656, opened 2020-08-07
Add YAML front matter reference for Handbook

kaycebasques:

- Adds https://web.dev/handbook/yaml-front-matter

FYI @jpmedley


robdodson:
I think this change broke the redirects file (I guess the hash isn't supported). I'm going to push another PR to fix it. Be sure to let the eng team approve stuff if they're tagged in as reviewers. I've tried to make it so we're only tagged in on behavior related files now.

PR #3659, opened 2020-08-07
fix(authors): add fallback images in `prod` if authors don't have them, fixes #3318

### LINKED ISSUE #3318: Add fallback images if authors don't have them.
If we don't fine an image for an author we currently show a broken img.

image

We should use a fallback img.

MichaelSolati:
Fixes #3318

When in `prod` mode the authors collection will check if the author has a profile picture. If they do not it'll log a warning in the console and swap in `no-photo@2x.jpg`.

### DIFF
--- a/src/site/_collections/authors.js
+++ b/src/site/_collections/authors.js
@@ -125,7 +125,7 @@ module.exports = (collections) => {
         console.warn(
           `No 2x image was found for ${author.title} (${author.key}), replacing with placeholder.`,
         );
-        authorsImage = path.join('/images', 'authors', 'no-photo@2x.jpg');
+        authorsImage = path.join('/images', 'authors', 'no-photo.svg');
       }
     }
     author.data.hero = authorsImage;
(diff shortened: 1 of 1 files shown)


src/site/_collections/authors.js
```diff
@@ -117,7 +118,16 @@ module.exports = (collections) => {
...
 
-    const authorsImage = path.join('/images', 'authors', `${key}@2x.jpg`);
+    let authorsImage = path.join('/images', 'authors', `${key}@2x.jpg`);
+    if (process.env.ELEVENTY_ENV === 'prod') {
+      const authorsImageExists = fs.existsSync(path.join('src', authorsImage));
+      if (!authorsImageExists) {
+        console.warn(
+          `No 2x image was found for ${author.title} (${author.key}), replacing with placeholder.`,
+        );
+        authorsImage = path.join('/images', 'authors', 'no-photo@2x.jpg');
```
samthor, line 128: total nit: can we use the SVG here? It's smaller and nicer when the image is actually geometric.

MichaelSolati: Yes we can! And I just updated it to do that.

PR #3660, opened 2020-08-07
feat(glitch): expose `allow` property

MichaelSolati:


jpmedley:
LGTM

src/site/_includes/components/Glitch.js
```diff
@@ -57,6 +67,12 @@ module.exports = (param) => {
     queryParams.previewSize = previewSize;
   }
 
+  const allow = Array.from(
+    typeof userAllow === 'string'
+      ? new Set([...defaultAllow, ...userAllow.toLowerCase().split('; ')])
```
samthor, line 72: can you add a helper, something like:

```js
function expandAllowSource(s) {
  if (typeof s === 'string') {
    s = s.split(/;\s*/g);
  }
  return s.map((a) => a.toLowerCase());
}
```

Then it's easier:

```js
const allow = Array.from(new Set([...defaultAllow, ...expandAllowSource(userAllow)]));
```

MichaelSolati: Done, thanks for the advice.

PR #3670, opened 2020-08-10
Make translation link work with explicit language in path

devnook:
Currently. we use only canonical urls (/foo) internally in web.dev. The only exception was where from a /lang/ page we link to canonical one (e.g. from /pl/foo to /en/foo, to enforce language version of the target despite user preferences). This Pr forces this link to open in a new page, to trigger a server-side redirect.

At the same time, I removed a server-side redirection for partials (from /en/foo/index.json to /foo/index.json) as it was unnecessary.


src/site/_includes/partials/translation-disclaimer.njk
```diff
@@ -8,7 +8,7 @@
     API.
     {% elif translation === 'manual' %}
     This article was translated from
-    [the original post](/en{{ page.data.canonicalUrl }}).
+    <a href="/en{{ canonicalUrl }}" target="_blank">the original post</a>.
```
robdodson, line 11: @samthor could we add a data attribute to links to signal to the router to ignore them? My concern with relying on target="_blank" here is that someone might one day remove it and not realize it has implications with the server.

devnook: This comment made me realise that we do skip absolute urls in the router, so I changed an internal /en/ link to an absolute one.

robdodson: ah clever!

Would you mind also adding a comment to njk to explain why youre doing the full url

robdodson:
@devnook this looks good to merge. Just remember to add a comment to the html explaining why it's using a full path.

PR #3675, opened 2020-08-10
fix(authors): only check once if author has an image

MichaelSolati:


robdodson:
Chatted with Michael about this. I think we're going to switch to having the path to the author image added to authorData.json. Then we'll use that path in our templates instead of attempting to read from disk. Should save time over the long run.

samthor:
Sure. Can we also remove the @2x files and just serve them through imgix or perhaps generate the 1/2 size images on build?

PR #3678, opened 2020-08-11
[Idle Detection article] Reflect permission changes

tomayac:

- The API now requires `IdleDetector.requestPermission()` rather than `Notificaiton.requestPermission()`.

### DIFF
--- a/src/site/content/en/blog/idle-detection/index.md
+++ b/src/site/content/en/blog/idle-detection/index.md
@@ -8,7 +8,7 @@ description: |
   interaction with the keyboard, mouse, screen, activation of a screensaver, locking of the screen,
   or moving to a different screen. A developer-defined threshold triggers the notification.
 date: 2020-05-18
-updated: 2020-08-05
+updated: 2020-08-11
 tags:
   - blog
   - idle-detection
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/idle-detection/index.md
```diff
@@ -107,19 +107,27 @@ this definition is left to the user agent.
...
   // Need to request permission first.
-  return console.log('Notifications permission not granted.');
+  return console.log('Idle detection permission not granted.');
 }
 ```
 
+{% Aside %}
+  Initially, idle detection was gated behind the
+  notifications permission.
+  While many, however, not all, use cases of this API
```
jpmedley, line 126:
```suggestion
  While many, but not all, use cases of this API
```

src/site/content/en/blog/idle-detection/index.md
```diff
@@ -107,19 +107,27 @@ this definition is left to the user agent.
...
+  involve notifications, we have decided to gate it
```
jpmedley, line 127: Who is 'we'? Spec authors?

tomayac, line 127:
```suggestion
  involve notifications, the Idle Detection spec editors have decided to gate it
```

tomayac: Yes, changed accordingly.

PR #3682, opened 2020-08-11
Don't fetch hero image if small screens

beaufortfrancois:
This PR make sure the hero image is not fetched on small screen devices. It is done by simply adding `loading=lazy` to the hero image element.

Without this fix, the hero image is always downloaded but never shown in the page on small screen devices.


robdodson:
nice one, thank you!

PR #3686, opened 2020-08-12
Feat/inclusive spellcheck lint

### LINKED ISSUE #3604: Add linting/testing for spelling and language
I'd love to see a spell check and test for non-inclusive words included in the linting process.

There are a few tools available that could help, for example [Alex](https://alexjs.com/), or [textlint](https://github.com/textlint/textlint). textlint seems more flexible, and includes rules for Alex, as well as spelling, common misspellings, etc. 

We had something like this on WebFu, but it was homegrown and would occasionally run into issues. 

Using a tool like this should only generate warnings, since in some cases, there will be false positives, but it's easier to catch them early then miss them.

PS: There's even an [Alex GitHub Action](https://github.com/marketplace/actions/alex-recommends)

robdodson:
Fixes #3604

- Updates the workflow for linting and testing to be a bit more intelligent and only run certain steps if files relevant to those steps were changed. 
- Adds [alex-recommends](https://github.com/brown-ccv/alex-recommends) GH Action to catch insensitive language.


src/site/content/en/accessible/control-focus-with-tabindex/index.md
```diff
@@ -10,6 +10,16 @@ description: |
   tabindex to ensure that they're keyboard accessible.
 ---
 
+isnt
```
robdodson, line 13: These changes are just to test this PR. They should be removed before merging.

PR #3690, opened 2020-08-12
Always link to /feature instead of /features

jpmedley:
URLs with /features links to a list view and the identified item is supposed to expand and scroll into view. This is broken in the current iteration of Chrome Status.

@tomayac FYI


tomayac:
Thanks for letting me know, and thanks for the fix!

samthor:
This is also broken in "src/site/_data/event.js", if you wanted to fix that (for a Q&A entry for web.dev/live).

kaycebasques:
@jpmedley I can put together an audit to catch this in the review extension

kaycebasques:
Audit should be live in v0.0.8, LMK if it doesn't work as expected

image

PR #3697, opened 2020-08-13
Test alex action.

robdodson:
Fixes #SOME_ISSUE_NUMBER

- 
- 
-


robdodson:
Looks like Alex will flags "slave" because it's part of retext-profanities, but not "master".
However, it will flag them if they're used together, i.e. "master and slave", because that pattern is in retext-equality.

I'm chatting with the maintainer to figure out where to add "master".
https://github.com/get-alex/alex/issues/298#issuecomment-673490607

PR #3707, opened 2020-08-13
Add link to setting up a Thumbor configuration file.

khempenius:


jpmedley:
@robdodson This is the second dorny error I've seen today. The log does not tell me how to fix it.

robdodson:
Pushing up a fix, but I'll also merge this PR to unblock folks.

PR #3715, opened 2020-08-14
chokidar --silent fix

samthor:
Rebuilds at least for me weren't working as: `-c` took an argument, but `--silent` did not. We had them the wrong way around.


robdodson:
hm weird, percy exploded with:

```
Error: Cannot find module 'puppeteer'
```

robdodson:
i'm rerunning it

PR #3718, opened 2020-08-14
Fix flaky test

devnook:
Integration tests seemed to be bottlenecking lately, so this should help.

### DIFF
--- a/test/integration/src/site/_collections/index.js
+++ b/test/integration/src/site/_collections/index.js
@@ -10,7 +10,7 @@ describe('_collections', function () {
     });
 
     after(function () {
-      fs.rmdirSync(path.join('.', '.tmp'), {recursive: true});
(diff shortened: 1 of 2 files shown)


test/integration/src/site/_collections/index.js
```diff
@@ -10,7 +10,7 @@ describe('_collections', function () {
     });
 
     after(function () {
-      fs.rmdirSync(path.join('.', '.tmp'), {recursive: true});
+      fs.rmdirSync(path.join(__dirname, '.tmp'), {recursive: true});
```
samthor, line 13: This will be the path of this file—are we creating `.tmp` under the test/integration/src/site/_collections path?

devnook: yes

PR #3720, opened 2020-08-14
updates SW to cache payload contents

samthor:
This does a couple of things to improve site loading and updates. We no longer have to reload the SW on site updates: we just update assets as required (diffing, adding and removing only what we need).

Some specific notes:

* It adds Eleventy's lifecycle methods early (monkeypatches for now)
* We generate "/sw-payload", which contains the page template, plus a list of assets to cache, as well as build info (hash of assets and build time)—this happens basically at the start of Eleventy's build (we block all other files until we see "/sw-payload" and generate the build info to use)
* When a page is fetched, we check that the partial's `resourceVersion` matches: if not, we reload "/sw-payload", our cache is probably out-of-date

This is a bit different than #3362, which I closed. (The issue with it was that we didn't hash anything but the top-level JS or CSS. Now, we hash exactly the same list of things that we cache when we render a page.)


samthor:
So hilariously the Percy diffs here are I think because we're including the banner in both partials and template HTML.

src/site/content/sw-payload.njk
```diff
@@ -2,8 +2,7 @@
 lang: ''
 layout: layout
 nometa: true
-permalink: /sw-partial-layout.partial
-show_banner: false
+permalink: /sw-payload
```
robdodson, line 5: I think you need to add `show_banner: false` back, otherwise it displays twice on any page with a banner (see the percy tests)

samthor:
Percy is now failing because of just one image failing to load—so it's effectively passing. PTAL!

robdodson:
Thanks for doing the work on this.

I am a bit concerned that it feels like we're going even deeper down the service worker rabbit hole. The more I think about it, the more I feel that we should consider going back to the approach of force refreshing if there's a pending service worker, and then moving to a model where we do the prompt thing. It's not a perfect UX _but_ it's a very well worn path and the rest of the team can easily grok it and help maintain it. Happy to discuss in a hangout.

samthor:
> to the approach of force refreshing if there's a pending service worker

We'll still do that. It's quite useful. But this approach means we won't refresh literally whenever we make a tiny CSS or JS change.

I really do have strong feelings about this. I don't think the well-worn path has friendly UX in a world where we're doing network-first partial loads (if we're cache-first, then the well-worn path is _better_, because you don't want to do _any_ network traffic).

robdodson:
> I really do have strong feelings about this. I don't think the well-worn path has friendly UX in a world where we're doing network-first partial loads (if we're cache-first, then the well-worn path is better, because you don't want to do any network traffic).

Right it's dangerous if we're network-first for partials because we could change the html in a partial but not have the CSS for it in the cache so we get a broken state.

Personally I am less concerned about the UX but I do feel very strongly about how complicated the set up has become. My main concern is my ability (and the rest of the team's ability) to help maintain it. I'd really like to free you up from having to work on this stuff.

Since we have a fairly standard content site, I'm wondering if we could take a step back and aim for a simpler setup (no-cache the html, hash the js and css assets, maybe only use a sw as a slightly more durable runtime cache). I'm concerned that dipping our toe into precaching has created a poison pill where we're now building pretty complex machinery to invalidate stuff. It feels like a model that works well for apps but less well for a content site.

robdodson:
@samthor ok for me to close this?

samthor:
Yes please

Issue #2717, opened 2020-04-28
Refactor cards component to be more flexible.


robdodson:
The card component has been expanded to support different pieces of content (authors, tags) and it doesn't feel very extensible. We're kind of repurposing the PathCard to act as a base for all other cards. It might be worth seeing if we can make the card more composable  (perhaps broken into smaller shortcodes or utility functions) that folks could compose together, as opposed to inheriting from one big card class. It would be nice if the CSS were more flexible as well.

robdodson:
Also wanted to note that there is a slight redesign we could work with. It ditches the blue line and moves the word 'collection' to the bottom, and that word can be replaced with 'guide'.

![image](https://user-images.githubusercontent.com/1066253/81005547-f9306a80-8e02-11ea-9cc8-2e4d6c605afb.png)

robdodson:
I think switching to utility CSS and card macros will help with this a lot.

robdodson:
Basically I think we should emulate what was done on d.c.c.

PR #3738, opened 2020-08-18
refactor(collections): save processed 11ty collections, fixes #3532

### LINKED ISSUE #3532: Post cards author links going to twitter not author's page
Post cards author links going to twitter not author's page

MichaelSolati:
Fixes #3532

The first time the collections get called is by 11ty when it passes the 11ty collection object into it. Every time we call it after, elsewhere, we are unable to pass in the 11ty collection object, which means that if we call it to process authors posts it'll think there are no authors posts and provide a twitter link instead of a link to the authors page.

By storing the result of the authors and tags, we can return the collection object with the posts and the correct links. (And we don't need to process it again)

Memoization FTW!(?)

### DIFF
--- a/src/site/_collections/authors.js
+++ b/src/site/_collections/authors.js
@@ -19,6 +19,8 @@ const authorsData = require('../_data/authorsData');
 const {livePosts} = require('../_filters/live-posts');
 const setdefault = require('../_utils/setdefault');
 
+let processedCollection;
+
 /**
  * Generate map the posts by author's username/key
  *
@@ -68,6 +70,10 @@ const maybeUpdateAuthorHref = (author, allAuthorPosts) => {
  * @return {Object.<string, Author>}
  */
 module.exports = (collections) => {
+  if (processedCollection) {
+    return processedCollection;
+  }
+
   let allPosts = [];
 
   if (collections) {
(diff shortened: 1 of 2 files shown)


samthor:
I presume it's impossible for this to be called _early_.

I'm also a bit worried about—in dev mode, will this prevent changes on the disk from showing up?

MichaelSolati:
It is impossible to be called early, but yes the data could be not changed. I will add even if it gets called early it's fine cause then the processed data wouldnt exist and it would continue.

What I could do is have it only return the processed data if it does exist and there is no collection? It wong provide any speedups, but will allow for us to use it in other places and still prevent the defaulting to twitter.

robdodson:
> Every time we call it after, elsewhere, we are unable to pass in the 11ty collection object, which means that if we call it to process authors posts it'll think there are no authors posts and provide a twitter link instead of a link to the authors page.

Where else do we call it?

robdodson:
Ah I think this may be the cause:
https://github.com/GoogleChrome/web.dev/blob/master/src/site/_includes/components/AuthorsDate.js#L100-L102

Rather than having the author's info passed into the card as an attribute or property, it's reaching outside of itself to require and then invoke the author's collection. I think we want to reverse this behavior. I think this is the ticket that was previously tracking it but it got closed (https://github.com/GoogleChrome/web.dev/issues/2717).

I realize that's a bigger investment than quick-fixing this bug, so if this PR resolves the issue I think we can land it, but we really do need to redo these cards. We're sort of adding technical debt to handle them now.

PR #3745, opened 2020-08-19
 Update pl directory with current en content

devnook:
This effectively enables Polish as an official language on we.dev. This will alow to send content to gitlocalize on broader scale.
It's staged at https://web-dev-staging.uc.r.appspot.com/pl/measure/


robdodson:
Hey @devnook can you take a look at the Percy failures? It seems like it's making some articles disappear and some articles get duplicated on the home page.

devnook:
@robdodson it is addressed in this PR: https://github.com/GoogleChrome/web.dev/pull/3752

devnook:
I'm adding TO NOT MERGE label until PR #3752 gets resolved.

devnook:
PTAL

src/site/content/pl/pl.11tydata.js
```diff
@@ -1,3 +1,7 @@
+const livePaths = require('../../_filters/live-paths');
```
robdodson, line 1: It seems like it'd be good for us to remove the paths data from this file in the future so we don't have to update every language if we change the paths.

src/site/content/pl/offline.njk
```diff
@@ -0,0 +1,17 @@
+---
+layout: layout
+title: Offline
+description: |
+  Network Offline
+noindex: true
+offline: true
+---
+
+<div class="w-layout-container--large">
```
robdodson, line 10: Should this change to an include that uses the `i18n` frontmatter approach to replace the strings?

src/site/content/pl/feed.njk
```diff
@@ -0,0 +1,48 @@
+---
```
robdodson, line 1: Should this change to using an include for the html?

robdodson:
I'm not sure why the Update translated date action is failing. @devnook you should be able to merge this when you're ready if you think it'll be safe to do so.

PR #3749, opened 2020-08-20
trick eleventy into watching generated files

samthor:
We had a bug, if we compiled CSS or JS during dev, Eleventy wouldn't rebuild. It was ignoring changes in the top-level `.gitignore`, but some of these files change during dev, and point to the latest JS/CSS.

Work around this by tricking Eleventy. The alternative is we create a whole `.eleventyignore` file which mostly mirrors `.gitignore`.


robdodson:
I'd prefer to use .eleventyignore since that's really what it's there for.

robdodson:
There's a lot we could remove from our .gitignore as well. i think it may have been a template I copied from devsite or something.

samthor:
I'm happy to use Eleventy's ignore file. I'll swap over.

PR #3750, opened 2020-08-20
move large shared image so we don't cache it

samthor:
Not sure how to help enforce this in future. Currently we are caching "rss-banner.png" because it globs as part of the SW manifest build, but it's quite large.


robdodson:
We could have a /images/rss/ folder and put it in there and ignore it?

robdodson:
I think the images in /images/shared/ are (in theory) used by our includes so it may be worth caching them? 🤷‍♂️ 
I guess I don't feel very strongly about it.

samthor:
The images in /images/shared/, right now, are used where assets are shared between posts. I don't think it makes sense to cache them.

PR #3752, opened 2020-08-20
Remove recent posts collection

devnook:
recentBlogPosts collection was used only in 1 place (on the homepage to display latest posts). It was also somewhat redundant to the already existing blogPosts collection. So instead of making it language-aware, I decided to remove it and use blogPosts directly. Let me know wdyt.


src/site/content/en/index.njk
```diff
@@ -128,11 +128,19 @@ eleventyComputed:
     </div>
     <div class="w-grid w-pb--std">
       <div class="w-grid__columns w-grid__columns--featured" role="list">
-        {% for post in collections.recentBlogPosts %}
-          {% if loop.first %}
-            {% PostCard post, true %}
-          {% else %}
-            {% PostCard post %}
+        {% set counter = 0 %}
+        {% for post in collections.blogPosts %}
```
MichaelSolati, line 132: Not saying that it's computationally intensive to iterate over all of the objects here, I doubt it is, but I think this could be made cleaner and easier to read...

The logic could be moved into an 11ty data file like so...

```JavaScript
module.exports = {
    eleventyComputed: {
        recentPosts: (data) => {
            const recent = [];
            for (const item of data.collections.blogPosts) {
                if (item.data.hero || item.data.thumbnail && item.data.lang === data.lang) {
                    recent.push(item);
                    if (recent.length === 3) { break; }
                }
            }
            return recent;
        }
      }
};
```

And then in the njk file the for loop would just be:

```njk
{% for post in recentPosts %}
    {% PostCard post, loop.first %}
{% endfor %}
```

devnook:
@robdodson language-aware means that we need to calculate the collections in eleventy separately for each language, or to filter them by language before pagination happens. E.g. a page /pl/blog should paginate only through Polish blog posts while /en/blog only through English ones etc. If we calculate all blog posts together in one collection, we either have duplicate entries (for canonical blog post and for the translation), or if we later filter by language we break pagination counts (filtering needs to happen before pagination). Let me know if this makes sense.

The logic itself does not need to be duplicated, it can live in the shared location (I'm not sure if _collections//hooks is the best location as it's not exactly a hook. On the other hand it does similar thing to hooks so why not).

@MichaelSolati /pl/ dir is already heavily outdated so I don't want to check in individual changes there right now. It will get bulk=updated in this PR: #3745

src/site/content/pl/index.njk
```diff
@@ -74,8 +74,20 @@ date: 2018-11-05
     </div>
     <div class="w-grid w-pb--std">
       <div class="w-grid__columns w-grid__columns--gapless w-grid__columns--three" role="list">
-        {% for post in collections.recentBlogPosts %}
-          {% PostCard post %}
+        {% set counter = 0 %}
```
robdodson, line 77: Should this look like the en/index.njk example? It seems like you'd add a similar /pl/index.11tydata.js file.

src/site/content/en/index.njk
```diff
@@ -128,7 +128,7 @@ eleventyComputed:
     </div>
     <div class="w-grid w-pb--std">
```
robdodson, line 129: I think we should turn this file into a layout in the _includes directory, instead of copying it. My concern is keeping it in sync over time. If we make a change to the HTML we'll want that to show up in all languages but it could be easy to forget to copy it over.

For the strings of text that appear in this file I think there are a few options. We could either move all of that text into the YAML front matter or the file and just translate those strings. Or we could create a file of translated strings and use eleventy data to access that file. For example, I think something like this might be possible:

`/_includes/homepage.njk`
```njk
<a class="w-button" href="/blog">
  {% helpers.translate('See all articles', lang) %}
</a>
```

test/unit/src/site/_collections/hooks/recent-blog-posts.js
```diff
@@ -0,0 +1,75 @@
+const assert = require('assert');
```
robdodson, line 1: 🙌  thank you for adding tests!

src/site/content/en/index.njk
robdodson: We can do this in a separate PR. I don't want to block this change.

devnook: Filed: https://github.com/GoogleChrome/web.dev/issues/3769

src/site/content/pl/index.njk
devnook: Updated

PR #3755, opened 2020-08-21
Switch from Google Fonts to self hosted fonts.

### LINKED ISSUE #826: Self-host Material Icons font.
To prevent a flash of unstyled text, we could consider self hosting the Google Fonts version of Material Icons and using link rel preload to load it earlier.

https://www.youtube.com/watch?v=Mv-l3-tJgGk&feature=youtu.be&t=24m58s

robdodson:
Fixes #826

- Switches from using Roboto to system fonts
- Adds preload only for Material Icons
- Updates SW to cache fonts
- Adds a clean up step to the SW to remove unused caches

Big savings!

**Before:**
image

**After:**
image


src/lib/sw.js
```diff
@@ -43,6 +47,19 @@ self.addEventListener('install', (event) => {
   event.waitUntil(self.skipWaiting());
 });
 
+self.addEventListener('activate', () => {
+  // Define a list of allowed caches.
+  // If a cache does not appear in the list then it will be deleted.
+  const allowedCaches = new Set(Object.values(cacheNames));
+  caches.keys().then((cacheNames) => {
```
samthor, line 54: can I suggest

```js
const p = Promise.resolve().then(async () => {
  const allowedCaches = new Set(Object.values(cacheNames));
  const cacheNames = await caches.keys();
  for (const cacheName of cacheNames) {
    if (!allowedCaches.has(cacheName)) {
      await caches.delete(cacheName);
    }
  }
});
event.waitUntil(p);
```

src/lib/sw.js
```diff
@@ -89,26 +106,12 @@ const assetExpirationPlugin = new ExpirationPlugin({
...
-  new workboxStrategies.StaleWhileRevalidate({
-    cacheName: 'google-fonts-stylesheets',
-    plugins: [externalExpirationPlugin],
-  }),
-);
-
 // Cache the underlying font files with a cache-first strategy for 1 year.
 workboxRouting.registerRoute(
-  /^https:\/\/fonts\.gstatic\.com/,
+  ({request}) => request.destination === 'font',
```
samthor, line 111: nice, it's the future!

There's a small set of old browsers which don't support this (but which otherwise will get our SW—like Chrome 61-64). I don't think it's worth worrying about them.

src/styles/settings/_type.scss
```diff
@@ -21,9 +21,8 @@ $FONT_SMOOTHING_RENDERING: optimizeSpeed;
 
 // Fonts.
 $HEADLINE_FONT: 'Google Sans', sans-serif;
-$PRIMARY_FONT: 'Roboto', sans-serif;
-$CODE_FONT: 'Roboto Mono', monospace;
-$ASIDE_FONT: 'Roboto Condensed', sans-serif;
+$PRIMARY_FONT: Segoe UI, system-ui, -apple-system, sans-serif;
```
samthor, line 24: 👍

src/lib/sw.js
robdodson: updated!

src/styles/settings/_type.scss
robdodson: 😁

src/lib/sw.js
robdodson: 👍

robdodson:
> I like it. I spent some time looking at visual differences now that we use the system font, and it's honestly just fine. Not really better or worse.

yeah it's a very subtle visual change and I think some things, like Asides, are actually more legible.

robdodson:
Before:
image

After:
image

PR #3760, opened 2020-08-22
fix SW registration code to allow Netlify staging

samthor:
I don't know if this used to be on netlify.com, or whether it never actually worked.


robdodson:
> I don't know if this used to be on netlify.com, or whether it never actually worked.

yeah i think it used to be netlify.com but they recently changed everything to .app

PR #3763, opened 2020-08-23
Update Notification Triggers status

christianliebel:

- In the Fugu status list, move Notification Triggers from origin trial back to "behind a flag"
- More prominently state that Notification Triggers are currently on hold


jpmedley:
@petele @tomayac The entry on the status page needs to be moved. Can you please look at what he suggests for the article. Is this how the message should be stated?

tomayac (approved):
LGTM, thanks!

This is indeed the current status, I hope we continue our work on this.

PR #3764, opened 2020-08-23
Add Idle Detection to the Fugu list

christianliebel:

- Add Idle Detection API to the list of Fugu APIs
- Consolidate dates

### DIFF
--- a/src/site/content/en/blog/fugu-status/index.md
+++ b/src/site/content/en/blog/fugu-status/index.md
@@ -95,6 +95,19 @@ trials in the [Origin Trials Guide for Web Developers][ot-guide].
       </tr>
       <tr>
         <td>
+          <a href="/idle-detection/">Idle Detection API</a>
+        </td>
+        <td>
+          The Idle Detection API notifies developers when a user is idle,
+          indicating such things as lack of interaction with the keyboard,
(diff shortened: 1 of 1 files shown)


jpmedley:
@petele @tomayac How does this text look to you all.

christianliebel:
> @petele @tomayac How does this text look to you all.

Note: The text is taken from the Chrome Platform Status page: https://www.chromestatus.com/feature/4590256452009984

tomayac:
LGTM, thanks!

(We should really automate generating this. It’s essentially a collection in Eleventy terms of everything tagged “capabilities”. @robdodson, is this doable?)

robdodson:
@tomayac it may be possible but I'm not sure if the eng team will have cycles to build it out as they're heads down working on developers.chrome.com stuff.

If you wanted to take a shot at a PR to add it we could take a look. I'm not sure if this page would be able to use the standard blog post template anymore though, I think it would need to use a new nunjucks page so you could access the collection and iterate over it.

tomayac:
Sounds good. It's likewise *best effort* for me, but FWIW, I've just self-assigned the newly created https://github.com/GoogleChrome/web.dev/issues/3807.

PR #3765, opened 2020-08-23
feat(algolia): index author, newsletter, and tag pages

MichaelSolati:


types/eleventy/collection-item.d.ts
```diff
@@ -0,0 +1,116 @@
+/*
```
robdodson, line 1: Is eleventy on definitely typed? Just curious if there's a place where this stuff is maintained so we don't have to stay in sync with eleventy when it revs.

src/site/_collections/algolia.js
```diff
@@ -58,10 +64,12 @@ module.exports = (collection) => {
   // For now, hard-code language to English.
   const lang = 'en';
 
-  const authorsCollection = authorsCollectionFn();
+  const authorsCollection = authorsCollectionFn(collection);
+  const newslettersCollection = newslettersCollectionFn(collection);
+  const tagsCollection = tagsCollectionFn(collection);
```
robdodson, line 69: I'm wondering if we can avoid having to recompute these collections.

You can define a javascript function in the _data directory and pass collections to it as an argument. So in algolia.njk you could do something like:

```
{{ helpers.algolia(collections.all, collections.authors, collections.newsletters, collections.tags) }}
```

Where helpers.algolia is defined as:

```
// helpers.js

const algolia = (all, authors, newsletter, tags) => {
  ...
  // return some json
}
module.exports = {algolia};
```

types/eleventy/collection-item.d.ts
MichaelSolati: Unfortunately not... It wasn't super difficult to generate this though!

src/site/_collections/algolia.js
MichaelSolati: So other than the newsletter, author and tags does not get recomputed. In earlier feature I added the authors and tags gets stored and returned instead of it being we recomputed.

I could do the same for the newsletter, but let me know what you think.

robdodson: I think my preference is for these dependencies to be passed in as arguments. It feels a bit cleaner and more testable since we could mock them. (side note, let's add some tests as well).

Memoizing the collections works, but it feels like we're working around the system. I keep thinking that eleventy collections should be as simple as possible because they have such a funky lifecycle, and if we can pass them to helper functions that live in _data then we might end up with simpler, more testable code.

types/eleventy/collection-item.d.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare global {
+  export interface EleventyCollectionItem {
```
samthor, line 18: I'm happy with this contribution but is there like, definitelytyped types for Eleventy already?

MichaelSolati: Haha, Rob actually asked the same thing. Unfortunately there is not, at least not one that I found (I checked for `@types/11ty` and `@types/eleventy`).

src/site/_collections/algolia.js
MichaelSolati: I set the "helper" as a filter in `_filter`, it made more sense to me there, though the logic should behave the same in any folder. It just didn't make sense in the data folder as those should be just, well, data objects.

src/site/content/algolia.11tydata.js
```diff
@@ -0,0 +1,141 @@
...
+        const lang = 'en';
+        const validTags = ['post'];
+        /** @type EleventyCollectionItem[] */
+        const eleventyPosts = collections.all
+          .filter((item) => {
+            // nb. There's no easy 'getFilteredByMultipleTag' method in Eleventy.
+            if (!Array.isArray(item.data.tags)) {
+              return false;
+            }
+            return item.data.tags.some((tag) => validTags.includes(tag));
```
robdodson, line 60: Since there's only one tag could you instead start with `collections.post` instead of `collections.all`? Eleventy will have already filtered this collection for us.

robdodson (approved):

PR #3766, opened 2020-08-24
a bunch of misc tsc fixes

samthor:
Fixes a bunch of small TSC typing issues.

We still have 78 errors, according to my local TSC. This is mostly in three categories:
* assuming a `Node` or `Element` can be used as a `HTMLElement` (just needs lots of casting), especially in component code
* we have untyped virtual imports for config
* no included types for SW code (the default DOM code doesn't seem to know about e.g., `ServiceWorkerGlobalScope`)


robdodson:
percy is just flaking on images still so you can ignore that failure

PR #3771, opened 2020-08-24
A proposal how to handle i18n messages

devnook:
Using md files instead of njk in content directory has multiple advantages:
+ more readable content
+ structural html stays in _includes folder
+ it helps with code reuse and keeping many languages in sync
+ frontmatter in .md can be parsed directly by gitlocalize


src/site/content/en/index.md
```diff
@@ -0,0 +1,32 @@
+---
+layout: homepage
+description: |
+  Get the web's modern capabilities on your own sites and apps with useful
+  guidance and analysis from web.dev.
+date: 2018-11-05
+i18n:
+  banner_blm: |
```
samthor, line 8: If we think we're going to include HTML in a tag, can we suffix it with `_html`, so we know that it should be passed through `| safe`?

samthor (approved):

src/site/content/en/index.md
devnook: +1

samthor:
It's not a requirement but another way this sort of string stuff is done is to provide the English text alongside the i18n key. This allows this to be the default fallback in case we're missing a string. It would look like:

```html
<div>{{ i18n_string i18n.blah_key "Blah string" }}</div>
```

(I forget how to call functions in Nunjucks, but i18n_string is the function here which is basically trying the i18n map first, then the string fallback.)

Notably this actually means that the English version is just showing entirely fallback strings.

robdodson:
I think in nunjucks you can put a function into the _data folder and do something like:

```
// _data/i18n.js
const translate = (key, string) => {...}
module.exports = {translate};
```

```
<div>{{ i18n.translate(blah_key, "Blah string") }}</div>
```

Lighthouse uses this approach (https://github.com/GoogleChrome/lighthouse/blob/41175fec205f735d3303bff615b89c6d3d03f9da/lighthouse-core/lib/i18n/README.md)

I'm curious which approach is easier for GitLocalize to translate?

robdodson:
It's worth noting that we also have some strings sprinkled around in various JS files so having a function we can call might be really useful for those cases as well.

robdodson:
Ewa and I chatted about this and it sounds like the solution this PR will work well with the way that we're currently using GitLocalize.

Essentially the YAML frontmatter is content, and we're already using Git Localize to translate the content (including frontmatter) for things like blog posts. For example, it'll translate the `title` and `description` for us.

Ewa says that by default Git Localize will translate every field unless we specifically exclude it. That's nice because it means we don't have to write a giant include list.

I also like that this frontmatter is co-located with the page itself.

Ewa mentioned that she's going to open an issue so we can discuss how we want to localize the handful of strings in our js components and shortcodes. It might be that we use the string file + function only for those widgets.

devnook:
Issue filed: https://github.com/GoogleChrome/web.dev/issues/3775

PR #3792, opened 2020-08-26
Track Table of Contents (ToC) opens and closes

kaycebasques:

- Tracks ToC interactions via Google Analytics Events

### DIFF
--- a/src/lib/components/TableOfContentsButton/index.js
+++ b/src/lib/components/TableOfContentsButton/index.js
@@ -17,6 +17,7 @@
 import {html} from 'lit-element';
 import {BaseStateElement} from '../BaseStateElement';
 import {openToC} from '../../actions';
+import {trackEvent} from '../../analytics';
 
 /**
  * Element that renders table of contents open button.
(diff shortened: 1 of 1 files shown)


src/lib/components/TableOfContentsButton/index.js
```diff
@@ -47,6 +48,12 @@ class TableOfContentsButton extends BaseStateElement {
 
   onStateChanged({isTocOpened}) {
     this.opened = isTocOpened;
+    trackEvent({
+      category: 'Site-Wide Custom Events',
+      action: 'click',
+      label: 'ToC',
+      value: (isTocOpened ? 1 : 0),
```
kaycebasques, line 55: The category and label are the same whether the user opens or closes the ToC. We would send a value of 1 when opening and a value of 0 when closing. We would therefore expect an average event value of 1 if people open the ToC and never close it. If we see a value of 0.5 it would mean that people are always closing the ToC after opening it (first event sends 1, second event sends 0). If we agree on this implementation maybe I should add a code comment explaining this?

Issue #3783, opened 2020-08-26
Site has horizontal scroll on mobile.


robdodson:
I think this is caused by the ToC. It seems to be pushing the social buttons offscreen on mobile.

robdodson:
Some additional information.

It seems like there might be two bugs, both relating to the ToC button and the introduction of self-hosted fonts.
If you check out 237cd82cd and then 65e230401 you can compare the difference. Adding Google Fonts seems to have caused some overflow, though it's not clear why.

If you add `overflow: hidden` to `.post-content` it _almost_ fixes the issue, but you also need to remove the ToC button.

Just removing the ToC button on its own doesn't seem to fix the issue EXCEPT on this one post:
/aria-poison-or-antidote/

It's not clear why that one post behaves differently from the others but I did notice that it doesn't have any code blocks or images. However, removing the code blocks an images from other posts didn't fix the overflow.

robdodson:
Copying over from chat:

when we load web components we lazy load their styles. I was looking at a page with a few glitches that cause it load our js late and the overflow doesn't happen until our custom elements upgrade and we inject the style elements into the head

testing on this page: http://localhost:8080/local-fonts/
if you delete the second to last style element it seems to fix the issue

```
.web-copy-code__button {
  color: #202124;
  height: 48px;
  margin: 8px;
  opacity: 0;
  position: absolute;
  right: 0;
  width: 48px; }
  .web-copy-code__button:focus, .web-copy-code__button:hover {
    background-color: rgba(63, 196, 255, 0.2);
    opacity: 1; }
  .web-copy-code__button:active {
    background-color: rgba(63, 196, 255, 0.4);
    opacity: 1; }

web-copy-code {
  display: block;
  position: relative; }
  web-copy-code:active .web-copy-code__button, web-copy-code:focus .web-copy-code__button, web-copy-code:hover .web-copy-code__button {
    opacity: 1; }
  web-copy-code pre {
    min-height: 64px; }
```

I think this is why /aria-poison-or-antidote/ might be "fixed" by only removing the toc-button because it doesn't have any code blocks (so it doesn't have any copy code buttons)

and I think there's a bug about the copy code button not showing up (probably because my PR to switch to self hosted fonts broke it)

so i bet there some interplay between the broken icon and the styles above which is causing an overflow

PR #3798, opened 2020-08-27
Fix copy button, hide ToC for now.

robdodson:
Partial fix for #3783 

- Fixes the broken font on the copy code button.
- Hides the ToC for now as the `web-table-of-contents` seems to cause a bit of horizontal overflow on mobile.


robdodson:
The percy errors are just percy being flaky. I need to ping them to figure out why it keeps showing images a 404'ing

robdodson:
I pinged Percy support to try and figure out why the images are flaking but we can ignore it for this PR

PR #3804, opened 2020-08-28
test: Update `@percy/script` to avoid Percy SDK timeouts

Robdel12:
## What is this?

TL;DR: This should fix intermittent missing assets: https://percy.io/web-dev/site/builds/6678051

When there is a `loading` attribute on an iframe, it causes request interception to hang in Puppeteer, which caused Percy's asset discovery process to timeout after 30 seconds (causing missing assets).

This PR updates the SDK to fix the asset discovery timeout (fixed upstream in Percy's SDKs). It also contains a few small updates to the `snapshots.js` script to make the tests run faster. When debugging, it made my feedback loop more manageable (2 mins --> 40 second test runs). With the small refactor to check for the `active` state (rather than a hard coded 5 second timeout), it shook out a small bug. It seems in tests the "about" page never gets the active state applied. It doesn't fail the script, just logs it out.

By setting the viewport to a desktop size, I believe this can be removed: https://github.com/GoogleChrome/web.dev/blob/1b18c4500f3691c8aeb84f1b0c0774d81ec1db05/src/lib/components/Codelab/index.js#L86-L93

Let me know if there's anything else I can do :)

cc: @robdodson

### DIFF
--- a/tools/percy/snapshots.js
+++ b/tools/percy/snapshots.js
@@ -8,6 +8,7 @@ const pagesToTest = [
   {
     url: '/learn/',
     title: 'Learn page',
+    navPage: true,
   },
   {
     url: '/accessible/',
@@ -16,10 +17,12 @@ const pagesToTest = [
   {
     url: '/measure/',
     title: 'Measure page',
+    navPage: true,
   },
   {
     url: '/blog/',
     title: 'Blog page',
+    navPage: true,
   },
   {
     url: '/test-post/',
@@ -28,6 +31,7 @@ const pagesToTest = [
   {
     url: '/about/',
     title: 'About page',
+    navPage: true,
   },
   {
     url: '/codelab-avoid-invisible-text/',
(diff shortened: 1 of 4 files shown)


samthor:
Thanks @Robdel12. We do need the CLA signed by all contributors. If that's not possible then we can recreate the change based on your guidance, which we do very much appreciate.

Re:

> By setting the viewport to a desktop size, I believe this can be removed:

Am I correct in understanding that this isn't solving the underlying problem—i.e., when the desktop size gets resized to mobile, we'd just be hiding the Codelab with CSS, and none of the resize handlers will run.

tools/percy/snapshots.js
```diff
@@ -60,18 +64,39 @@ const pagesToTest = [
...
+  } catch (err) {
+    console.log(`Couldn't find active link for ${page.url}`);
+  }
+}
+
 // A script to navigate our app and take snapshots with Percy.
 PercyScript.run(
   async (browser, percySnapshot) => {
+    // set the viewport to a desktop size
+    await browser.setViewport({height: 1024, width: 1280});
```
robdodson, line 92: @Robdel12 we're also using a .percy.conf.yml which sets dimensions
https://github.com/GoogleChrome/web.dev/blob/master/tools/percy/.percy.conf.yml

Should we remove this line so the yml file can set mobile and tablet sizes?

Robdel12: @robdodson That's a good question! There's a bit of a disconnect between those two settings. And I admit its a little confusing.

The snapshot options are passed along to our browsers in our infrastructure. The widths being passed there are to tell our browsers to resize to that and then capture the screenshot.

This line in the test script is changing the width of the _test_ browser, which can impact how the DOM renders (at this moment Percy is capturing a DOM snapshot & not a screenshot). If the applications DOM changes between screen sizes, you'll have to make some changes to the script for accurate screenshots. At the moment Percy makes the assumption that the DOM snapshot is the same across widths and CSS media queries is what changes the pages look.

robdodson: ah I see

robdodson:
thanks for the patch @Robdel12!

robdodson:
@Robdel12 I think we're still seeing issues. Here's a recent test run that started after this PR was merged:
https://percy.io/web-dev/site/builds/6722391

Robdel12:
Hey @robdodson wanted to let you know I've seen this and plan on getting to it, just a little backed up at the moment :)

PR #3813, opened 2020-09-02
feat(web-feedback): check if form was likely filled by bot, fixes #3812

### LINKED ISSUE #3812: Improve confidence in "Was this page helpful?" data

After a few months of having the "Was this page helpful?" (WTPH) component running on the site, a suspicious pattern has emerged. A lot of pages have the exact same number of responses (**Total Events**):

image

There's many more like this.

This seems to suggest that bots are prowling around our pages.

Add some kind of mechanism that prevents non-humans from interacting with our feedback component.

reCAPTCHA provides the most confidence, but from what I remember, the size of the library is prohibitively large.

MichaelSolati:


robdodson:
I think this is slightly changing the appearance of the form, see the comparison below:

image

You might check to see if you can explicitly remove the underline from this input

robdodson:
@MichaelSolati I don't think that last commit fixed the issue, i'm still seeing:
image

PR #3817, opened 2020-09-02
Allow smart quotes as they are necessary in some languages

devnook:


robdodson:
I think we should delete this file and remove it from .remarkrc

MichaelSolati (approved):

PR #3825, opened 2020-09-02
Refactor gulpfile. Switch to canonical sass for everything.

### LINKED ISSUE #3823: Refactor gulpfile into tasks
d.c.c. has all of its gulp tasks factored into single files which is quite nice. We should do the same for web.dev.

robdodson:
Fixes #3823 

- Refactors the gulpfile so it imports tasks instead of having them defined in one big file.
- Uses chokidar to watch eleventy, instead of relying on `eleventy --watch`. The reason we do this is because `eleventy --watch` will always attempt to build eleventy first, which makes it hard to build eleventy _and then watch eleventy_ in an npm script. This resolves that issue and guarantees eleventy will always have finished running by the time the server starts when we do `npm run dev`.
- Replaces compile-css.js with sass.js gulp task
- Renamed all.scss to app.scss because gulp.dest() only works on directories so it was outputting "all.css" but our other scripts except the file to be named "app.css".
- I ran `npm update` to fix an issue in autoprefixer which revved a few packages like eslint and markdown-it-anchor. Had to fix a few issues like a broken eslint rule and duplicate anchor ids in articles as a result. We should probably increase the cadence that we run dependabot...


gulpfile.js
```diff
@@ -14,112 +14,35 @@
...
-/* eslint-disable max-len */
-const assetTypes =
-  'jpg,jpeg,png,svg,gif,webp,webm,mp4,mov,ogg,wav,mp3,txt,yaml';
-/* eslint-enable max-len */
-
-const isProd = process.env.ELEVENTY_ENV === 'prod';
+const {parallel, watch} = require('gulp');
+
+// Pull in each task
+const fonts = require('./src/gulpt-tasks/fonts');
```
samthor, line 20: gulpt => gulp?

src/gulpt-tasks/assets.js
```diff
@@ -0,0 +1,77 @@
+const {dest, src} = require('gulp');
+const mozjpeg = require('imagemin-mozjpeg');
+const pngquant = require('imagemin-pngquant');
+const imagemin = require('gulp-imagemin');
+const through2 = require('through2');
+const rename = require('gulp-rename');
+
+const isProduction = process.env.ELEVENTY_ENV === 'prod';
+
+/* eslint-disable max-len */
```
samthor, line 10: do we still need this disable?

src/gulpt-tasks/sass.js
```diff
@@ -0,0 +1,45 @@
...
+
+// We want to be using canonical Sass, rather than node-sass
+sassProcessor.compiler = require('sass');
+
+// Flags whether we compress the output, generate sourcemaps, etc
+const isProduction = process.env.ELEVENTY_ENV === 'prod';
+
+/**
+ *
+ * @param {Buffer} vinly.contents
```
samthor, line 17: vinly => vinyl

src/gulpt-tasks/sass.js
```diff
@@ -3,19 +3,43 @@ const cleanCSS = require('gulp-clean-css');
...
 // Flags whether we compress the output, generate sourcemaps, etc
 const isProduction = process.env.ELEVENTY_ENV === 'prod';
 
+/**
+ *
+ * @param {Buffer} vinly.contents
+ * @param {string} vinly.basename
+ */
+const hashOutput = ({contents, basename}) => {
+  // gulp.dest() requires that we return a path to an output directory.
```
samthor, line 21: Can you put this in the comment of the function (inside `/** */`?) I understand now but I imagined the internal comment to be a comment on _how_ this function worked, not what it did.

Issue #3836, opened 2020-09-03
web.dev is "downloading" pages because of content indexing API


robdodson:
**Describe the bug**
If a user has the experimental web platform features flag enabled then web.dev will show a download indicator as they navigate. This is caused by a bug in the current version of the content indexing API.

I'm sending in a PR to temporarily disable our use of this API until after the bug is resolved.

samthor:
Assigning to @jeffposnick

jeffposnick:
...there's not much to be done at the moment, but once the engineering team responsible for the Content Indexing API resolves the bug in desktop Chrome while running with "experimental web platform features," we can hopefully turn this back on.

robdodson:
@jeffposnick just checking in as i'm doing a big of bug triage. Do you know if the underlying Chrome issue was resolved?

jeffposnick:
I'm not able to reproduce it anymore on https://so-pwa.firebaseapp.com/ (which still uses the Content Indexing API) in macOS Chrome 90.0.4396.0 with the experimental web platform features flag on.

So... I can't say for certain about what would happen if it were re-enabled on https://web.dev/, but it does appear to be resolved elsewhere.

robdodson:
I'm going to close this since we removed the content indexing API. We can look into if we want to add this back when we redo our service worker in the future.

jeffposnick:
(Just for posterity, I can confirm that the Chrome engineering team did fix the underlying issue a month or so ago.)

PR #3837, opened 2020-09-03
Temporarily disable content indexing.

robdodson:
Related to #3836 

- Temporarily disable content indexing API. We should turn this back on when the underlying bug is fixed in the API itself.

### DIFF
--- a/src/lib/app.js
+++ b/src/lib/app.js
@@ -8,14 +8,16 @@
 
 import './webcomponents-config'; // must go before -loader below
 import '@webcomponents/webcomponentsjs/webcomponents-loader.js';
-import {trackError} from './analytics'; // side effects & named export
 import {swapContent, getHTML} from './loader';
 import * as router from './utils/router';
 import {checkUserPreferredLanguage} from './actions';
 import {store} from './store';
 import {localStorage} from './utils/storage';
 import removeServiceWorkers from './utils/sw-remove';
-import {syncContentIndex} from './content-indexing';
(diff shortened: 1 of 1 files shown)


MichaelSolati (approved):

src/lib/app.js
```diff
@@ -8,14 +8,16 @@
...
 import {swapContent, getHTML} from './loader';
 import * as router from './utils/router';
 import {checkUserPreferredLanguage} from './actions';
 import {store} from './store';
 import {localStorage} from './utils/storage';
 import removeServiceWorkers from './utils/sw-remove';
-import {syncContentIndex} from './content-indexing';
+// TODO: Enable this when #3836 is fixed.
+// // https://github.com/GoogleChrome/web.dev/issues/3836
+// import {trackError} from './analytics'; // side effects & named export
```
jeffposnick, line 19: I think you want to keep this? Or at least double-check what the side-effects of the import is. I remember that the `./analytics` import was around before my PR.

jeffposnick (changes requested):

src/lib/app.js
robdodson: It looks like it's ok being commented out because your snippet on line 70 was the only one using it.

robdodson: OH WAIT you're right

robdodson: didn't realize that module had its own side effects. Good catch.

PR #3840, opened 2020-09-04
Initial german localization

eyecatchup:

- Add initial localization for German


eyecatchup:
@googlebot I fixed it.

PR #3847, opened 2020-09-08
feat(feedback): allow conditional feedback questions to be enabled

MichaelSolati:
This will allow us to have conditional questions added to our feedback element, and they can be enabled by the front matter of posts.

For example, here we are only enabling a question about if this API met a user's needs:

```md
---
title: Feedback Page
feedback:
  - api
---
```


src/site/_includes/partials/post.njk
```diff
@@ -113,8 +113,7 @@
     {% if codelabs %}
       {% CodelabsCallout codelabs, lang %}
     {% endif %}
-
-    <web-feedback></web-feedback>
+    <web-feedback conditionals="{{ feedback }}"></web-feedback>
```
robdodson, line 116: Suggest renaming `conditionals` to `additional-questions`

types/lib/components/Feedback.d.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+declare global {
+  export namespace WebFeedback {
```
robdodson, line 18: Should we be namespacing these typings with `Web` or would it make sense to instead call them `Feedback`?

src/lib/components/Feedback/index.js
```diff
@@ -6,18 +6,51 @@ import {BaseElement} from '../BaseElement';
...
+      conditionals: {type: String},
+    };
+  }
+
   constructor() {
     super();
     this.robotName = 'is-it-just-me-or-was-this-form-filled-out-by-a-robot';
     this.submitted = false;
     this.submit = this.submit.bind(this);
+    this.conditionals = '';
```
robdodson, line 53: If you use the converter suggestion above then you can set this to `this.conditionals = []`

src/lib/components/Feedback/index.js
```diff
@@ -6,18 +6,51 @@ import {BaseElement} from '../BaseElement';
...
 /**
  * Element that handles feedback on posts to submit to Analytics.
  *
  * @extends {BaseElement}
  * @final
  */
 class Feedback extends BaseElement {
+  static get properties() {
+    return {
+      conditionals: {type: String},
```
robdodson, line 44: It seems like you want to treat this as Array-like data. You can use a [LitElement converter](https://lit-element.polymer-project.org/guide/properties#conversion) to do that. I think that'll be tidier.

```js
return {
  conditionals: {
    converter: (value) => value.split(','),
  },
};
```

src/site/content/en/handbook/web-dev-components/index.md
```diff
@@ -5,6 +5,8 @@ date: 2019-06-26
 updated: 2020-07-17
 description: |
   Learn how to use web.dev's UI and content components.
+feedback:
```
robdodson, line 8: At first glance I thought this was going to enable all of the default questions, but I think it's actually enabling all of the _conditional_ feedback questions.

You might rename it to something like `additional_feedback_questions`.

src/lib/components/Feedback/index.js
```diff
@@ -6,18 +6,51 @@ import {BaseElement} from '../BaseElement';
...
+  {
+    text: "Was this page's information accurate?",
+    label: 'accuracy',
+    name: 'Accuracy',
+  },
+  {text: 'Was this page easy to read?', label: 'readable', name: 'Readability'},
+];
+
+/** @types {WebFeedback.Question[]} */
+const conditionalQuestions = [
```
robdodson, line 31: Maybe rename to `additionalQuestions` and add a comment that these questions are optional and enabled by the `additional_feedback_questions` key?

src/lib/components/Feedback/index.js
```diff
@@ -6,18 +6,51 @@ import {BaseElement} from '../BaseElement';
 import {trackEvent} from '../../analytics';
 import './_styles.scss';
 
+/** @types {WebFeedback.Question[]} */
+const questions = [
```
robdodson, line 10: You might add a comment explaining that these are default questions that will _always_ be displayed. You could also consider renaming this to `defaultQuestions`.

robdodson (changes requested):

src/site/content/en/handbook/web-dev-components/index.md
kaycebasques: In the spirit of YAGNI I think we can get rid of the option of enabling all conditional questions. When we need to enable conditional questions, we'll just do it one-by-one. If we add other conditional questions, I expect that they'll be for dramatically different purposes. E.g. some will be related to APIs, others may be along the lines of "Are you planning on implementing this feature?", other ones might be about browser compatibility. So I don't think it's likely that I would want to enable them all.

kaycebasques: Also, I think we can remove this test page because I just pushed a commit that adds the feature to a real page (the change password URL post)

src/lib/components/Feedback/index.js
```diff
@@ -6,18 +6,51 @@ import {BaseElement} from '../BaseElement';
...
+    text: 'Did this page have the information you needed?',
+    label: 'complete',
+    name: 'Completeness',
+  },
+  {
+    text: "Was this page's information accurate?",
+    label: 'accuracy',
+    name: 'Accuracy',
+  },
+  {text: 'Was this page easy to read?', label: 'readable', name: 'Readability'},
```
kaycebasques, line 27:
```suggestion
```

I'm worried about the feedback widget getting ridiculously long. I'm OK with dropping these questions. Getting signals about whether the page was helpful and whether it helped them complete their goals are probably the most important feedback we can get. Perhaps we should do it in a separate PR though, because it's not related to this feature? In the future I may change my mind and want to add questions back.

types/lib/components/Feedback.d.ts
MichaelSolati: I chose this because `Web` is the prefix of our custom elements, I'm not sure how having the namespace of `Web` declared in different files would work though...

src/site/content/en/handbook/web-dev-components/index.md
MichaelSolati: Ok, I removed the `all` flag and the `feedback` front matter from the handbook, but @robdodson @kaycebasques should we rename the property from `feedback` to `additional_feedback_questions`?

samthor (approved):

kaycebasques (approved):

src/lib/components/Feedback/index.js
```diff
@@ -85,112 +143,8 @@ class Feedback extends BaseElement {
...
-              <label>
-                <input
-                  type="radio"
-                  aria-labelledby="readable-label no-label"
-                  name="Readability"
-                  value="0"
-                />
-              </label>
-            </div>
+            ${this.renderQuestions(additionalQuestionsToDisplay)}
```
robdodson, line 146: nit, can probably rename to `additionalQuestion` to match `defaultQuestions`

robdodson (approved):

src/lib/components/Feedback/index.js
MichaelSolati: `additionalQuestions` is used already on line 20 though 😬

Imma merge it in for now, and if we come up with a different naming we can always change it hopefully quickly.

PR #3849, opened 2020-09-09
Get latest version of alex recommends

robdodson:

- alex-recommends wasn't using the latest version of alex, which was failing to detect singular uses of 'master' and 'slave'. Previously it would only flag these if they were used in the same phrase, i.e. "master and slave".


robdodson:
@kaycebasques as an fyi since I saw you were working on some inclusive language updates

devnook (approved):

PR #3865, opened 2020-09-12
fix(toc): set overflow of TOC button to be hidden, fixes #3783

### LINKED ISSUE #3783: Site has horizontal scroll on mobile.

MichaelSolati:


src/lib/components/TableOfContents/index.js
```diff
@@ -98,11 +98,14 @@ class TableOfContents extends BaseStateElement {
     for (const heading of headings) {
       const href = `#${heading.target.getAttribute('id')}`;
       const link = links.find((l) => l.getAttribute('href') === href);
```
samthor, line 100: Unrelated but is it worth converting `links` into an object mapping from href to link, so we can remove this find?

samthor (approved):

src/lib/components/TableOfContents/index.js
MichaelSolati: Done, well, I used a Map instead of just making an object as it's easier to iterate through if needed, but it saves the need to use the `find` method.

PR #3874, opened 2020-09-15
Remove "native" from the codebase

kaycebasques:
- [ ] check if any payment articles need to be redirected (I think "native" was only mentioned in the titles, not the URLs)


src/site/_data/paths/fast.json
```diff
@@ -47,7 +47,7 @@
         "lazy-loading",
         "lazy-loading-images",
         "lazy-loading-video",
-        "native-lazy-loading",
+        "built-in-image-lazy-loading",
```
robdodson, line 50: should this be browser-level-image-lazy-loading?

PR #3883, opened 2020-09-15
Inline sw manifest to avoid weirdness with etags.

### LINKED ISSUE #3879: Site is taking extremely long to load on repeat load
**Describe the bug**
Visit web.dev (assuming you have a preexisting service worker). The site takes a very long time to load and the loading spinner never stops.

Now try opening an incognito window and visiting the site, notice that the site loads almost immediately.

robdodson:
Fixes #3879

- Inline the sw manifest to avoid etags getting out of sync.

### DIFF
--- a/sw-manifest.js
+++ b/sw-manifest.js
@@ -61,6 +61,9 @@ async function buildCacheManifest() {
   }
 
   const toplevelManifest = await injectManifest(config);
+  log(
+    `Building the Service Worker manifest, caching: ${toplevelManifest.count} files`,
+  );
   if (toplevelManifest.warnings.length) {
     if (isProd) {
       throw new Error(`toplevel manifest: ${toplevelManifest.warnings}`);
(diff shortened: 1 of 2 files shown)


sw-manifest.js
```diff
@@ -59,20 +60,15 @@ async function buildCacheManifest() {
     config.globPatterns.push('bootstrap.js', 'app.css');
   }
 
-  const toplevelManifest = await getManifest(config);
+  const toplevelManifest = await injectManifest(config);
```
jeffposnick, line 63: You might want to log the `toplevelManifest.count` value somewhere, which is the number of items precached. That was previously being logged, but you remove that message.

The properties available in the return value is documented at https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.injectManifest

jeffposnick (approved):
The usage seems fine, and when I tested things via `export ELEVENTY_ENV=prod && npm run build`, the correct output was created.

sw-manifest.js
robdodson: added, thanks!

PR #3904, opened 2020-09-18
Fix code example

krutoo:
Fixes code example in article "How to use CSS clipping"

P.S. I just use "Improve this article" link. Maybe i doing something wrong...


krutoo:
@googlebot I signed it!

PR #3912, opened 2020-09-21
Force express to no-store our html.

### LINKED ISSUE #3911: JavaScript is failing to load.
**Describe the bug**
Note the failing JS in devtools.
I was able to reproduce this in Firefox as well without a service worker. I think the issue is our html is using weak etags (because it comes from express.static()) and so it's sending the user a 304 even after we've pushed up new js. This means the user's html is trying to load a fingerprinted js file that no longer exists on the server.

Screen Shot 2020-09-21 at 10 32 09 AM

robdodson:
Fixes #3911 

- Add express handler that forces html to not be cached at all. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Directives for more info.

@jeffposnick mentioned that we might be able to get away with disabling etags in express.static() and relying on Last-Modified:

```
app.use(express.static('dist', {
  etag: false,
  lastModified: true,  // Just being explicit about the default.
}));
```

I'd like to land this no-store change as a quick fix to see if it unbreaks the site. Then we can see if using the `etag:false` approach works.

### DIFF
--- a/server.js
+++ b/server.js
@@ -119,12 +119,22 @@ const safetyHandler = (req, res, next) => {
   return next();
 };
 
+// Express' static middleware uses weak etags which will sometimes serve
+// stale html (which then requests the wrong versioned js resources).
+// This handler forces express to not allow any caching of the html
+// whatsoever.
+const staticHandler = (res, path) => {
+  if (path.endsWith('.html')) {
+    res.setHeader('Cache-Control', 'no-store');
+  }
+};
+
 const handlers = [
   safetyHandler,
   buildSafetyAssetHandler(),
   localeHandler,
-  express.static('dist'),
-  express.static('dist/en'),
+  express.static('dist', {setHeaders: staticHandler}),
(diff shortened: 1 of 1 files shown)


devnook (approved):
lgtm

jeffposnick:
Based on the fact that I see `Last-modified: Tue, 01 Jan 1980 00:00:01 GMT` on all of the static responses, and that each file's `mtime` goes into forming the Etag, I don't think App Engine is correctly preserving timestamps either.

[Here's the code](https://github.com/jshttp/etag/blob/4664b6e53c85a56521076f9c5004dd9626ae10c8/index.js#L126-L131) that Express uses under the hood to create the weak Etag, which depends on a unique combination of file size and `mtime`.

(Setting `etag: false` [apparently does not work](https://stackoverflow.com/a/29961486/385997) for anything served by `express-static`.)

PR #3920, opened 2020-09-22
Add isolate feature

robdodson:
This adds a (very experimental) feature to improve the quality of life for folks who are developing or authoring content on web.dev.

`npm run isolate` can be used to move all files out of `src/site/content` and into an `_exile` directory which will be ignored for builds. `isolate` takes an optional glob (or space separated list of globs) of files that should be preserved.

`npm run integrate` will restore all of the files in `_exile` to their previous location. It will also delete the `_exile` directory.

Example usage:

```bash
# Move everything except the style-focus directory to _exile
npm run isolate -- src/site/content/en/accessible/style-focus/**
```

(Do some work on style-focus/index.md)

```bash
# Restore everything from _exile back to src/site/content
npm run integrate
```

## Benchmark

Without isolate:
- Initial Eleventy build: **10.48 seconds**
- Eleventy rebuild: **8.27 seconds**

With isolate:
- Initial Eleventy build: **1.06 seconds**
- Eleventy rebuild: **0.53 seconds**

## Changes

- Adds `isolate` and `integrate` npm scripts
- Adds git commit hook to prevent committing while the `_exile` directory is active
- Prevents authors collection from throwing if there are authors without posts


robdodson:
Percy failure can be ignored. It's flaking on an image (related to #3804).

src/site/_collections/authors.js
```diff
@@ -150,14 +150,6 @@ module.exports = (collections) => {
     authors[key] = author;
   });
 
-  // Only complain that authors are invalid if we've got any posts *at all*
```
devnook, line 153: is it ok to remove this checkaltogether, or should it be disabled only in "isolate" mode?

tools/isolate/index.js
```diff
@@ -0,0 +1,73 @@
...
+ * Items matching these globs will not be moved to _exile.
+ * @param {Array<string>} ignore An array of glob patterns to be ignored.
+ */
+async function isolate(ignore = []) {
+  // Eleventy's rssLastUpdatedDate filter will blow up if we pass it an empty
+  // collection. To avoid this we also move all RSS feeds into exile.
+  const matches = await glob('src/site/content/**/{feed.njk,*.md}', {ignore});
+  for (const oldPath of matches) {
+    let newPath = oldPath.split(path.sep);
+    newPath.splice(2, 1, '_exile');
```
devnook, line 23: consider moving '_exile' to a const

tools/isolate/index.js
```diff
@@ -0,0 +1,73 @@
...
+const {existsSync} = require('fs');
+
+/**
+ * Move all markdown files and RSS feeds in src/site/content to the _exile
+ * directory.
+ * Optionally pass in an array of globs to be ignored.
+ * Items matching these globs will not be moved to _exile.
+ * @param {Array<string>} ignore An array of glob patterns to be ignored.
+ */
+async function isolate(ignore = []) {
```
devnook, line 17: I'd run integrate or restore here before isolate. Otherwise, if you run isolate twice, e.g.:

```
npm run isolate -- src/site/content/en/fast/**/*
```

and sometime later:
```
npm run isolate -- src/site/content/en/accessible/**/*
```

you'd expect that you isolated 'accessible' dir, but in reality that dir is empty. So at least a restored() warning should be in place before re-running isolate command.

devnook (commented):
Left some comments. Also - maybe consider adding a test for it?

tools/isolate/index.js
robdodson: ah good call. Instead of running integrate I think I may check if the `_exile` directory already exists and if it does then `isolate` will halt.

src/site/_collections/authors.js
robdodson: @MichaelSolati wdyt? Is there any danger to removing this check?

nhoizey:
Wouldn't it be enough to restrict the number of items in every collection when the environment is "dev"?

robdodson:
> Wouldn't it be enough to restrict the number of items in every collection when the environment is "dev"?

Maybe, though I think that would require editing the code for all of our collections to use a common filter and we'd need to ensure we use it for new collections as well. It's not a bad idea though and may be safer than physically moving the files (though that seemed to work surprisingly well). I'd also need a way to specify to the filter that it should allow certain posts to remain in the collection(s) (basically whichever post or directory of posts I'm working on).

nhoizey:
I agree it might not be easy, and you probably don't want to remove items from a navigation collection, for example… 😅

src/site/_collections/authors.js
MichaelSolati: I don't think there is any danger in removing it, bit I think it would be safer to conditionally ignore it if you want. Are there any authors that this brakes for?

PR #3921, opened 2020-09-22
refactor(types): add more typings to codebase

MichaelSolati:


robdodson (approved):

src/lib/components/BaseResponseElement/index.js
```diff
@@ -187,7 +189,9 @@ export class BaseResponseElement extends BaseElement {
         } else if (isSelected && !isCorrect) {
           option.setAttribute('data-submitted', '');
           this.disableOption(option);
-          this.deselectOption(option);
+          if (typeof this['deselectOption'] === 'function') {
```
samthor, line 192: Where is this coming from? Can we set `this.deselectOption = /** @type {function(Node)} */ null` in the ctor to solve this?

src/lib/components/Codelab/index.js
```diff
@@ -48,16 +48,20 @@ class Codelab extends BaseElement {
...
-    // Normally LitElement will remove any light DOM children that are not
-    // slotted when we call render().
-    // Because we don't use slots, and we _do_ want to preserve this element's
-    // light DOM children (they hold the codelab instructions) we create a new
-    // renderRoot for LitElement.
-    // https://lit-element.polymer-project.org/guide/templates#renderroot
-    // This will render the glitch element as a sibling to the existing light
-    // DOM children.
-    const container = document.createElement('div');
+    const container = /** @type this */ (
```
samthor, line 62: Can you explain why `this` is needed?

src/lib/components/EventSchedule/index.js
```diff
@@ -61,9 +65,13 @@ class EventSchedule extends HTMLElement {
...
   }
 
+  /**
+   * @param {string} hash
+   * @returns {HTMLElement|null}
+   */
   _elementForHash(hash = window.location.hash) {
     const id = hash.substr(1);
-    return (id && this.querySelector(`[data-session-id="${id}"]`)) || null;
+    return id ? this.querySelector(`[data-session-id="${id}"]`) : null;
```
samthor, line 74: I suppose `querySelector` also returns null if it can't find a value.

src/lib/components/Tabs/index.js
```diff
@@ -271,10 +275,11 @@ export class Tabs extends BaseElement {
   }
 
   /**
-   * @param {!Node} node to check
+   * @param {HTMLElement} node to check
    * @return {number} the index of the tab containing this node, or -1 for none
    */
   indexOfTabByChild(node) {
+    /** @type HTMLDivElement */
```
samthor, line 282: I don't think we care that this is a `div`, explicitly. Could just be `HTMLElement`?

src/site/_filters/slugify.js
```diff
@@ -19,12 +19,12 @@ const slugify = require('slugify');
 /**
  * Slugifies string.
  *
- * @param {string | String} str
+ * @param {string|StringConstructor} str
```
samthor, line 22: Does this actually happen in practice? I've always been scared of the upper-case `String` type.

samthor (approved):

src/lib/components/Codelab/index.js
MichaelSolati: I think because of the fact we're extending from the `BaseElement` we end up getting this weird error:

```
Property 'createRenderRoot' in type 'Codelab' is not assignable to the same property in base type 'BaseElement'.
  Type '() => HTMLDivElement' is not assignable to type '() => this'.
```

So the casting was necessary in order to prevent errors.

src/site/_filters/slugify.js
MichaelSolati: I'm unsure if we ever use String objects instead of primitives, but if we were this would be the correct typing (from what I could tell).

PR #3922, opened 2020-09-22
Add the Goibibo case study

### LINKED ISSUE #3882: content: goibibo [2020-09-18]
* [draft](https://docs.google.com/document/d/1p00e7vtyo3llv7msm5EgwHtRlSBV2Ua3p-lSQqQ4qJQ/edit)
* [thread](https://groups.google.com/a/google.com/g/webdev-content/c/6TldI0GRPY8)

kaycebasques:
Fixes #3882


src/site/content/_redirects.yaml
```diff
@@ -215,3 +215,8 @@ redirects:
 
 - from: /image-support-for-async-clipboard
   to: /async-clipboard
+
+# Vanity URLs
+
+- from: /goibibo
+  to: /goibibo-case-study
```
robdodson, line 222: do you think we need a vanity url for this? Are we planning to put something else at `/goibibo` ?

robdodson:
I added the correct hero image and also added an option which visually hides the h1 (so it's still accessible) but only displays the logo.

btw @kaycebasques I think you have a markdown linter error. I keep having to push using the `--no-verify` flag.

```
src/site/content/en/blog/goibibo-case-study/index.md
  57:1-57:26  warning  Don’t pad `strong` with inner spaces  no-inline-padding  remark-lint
```

kaycebasques:
@robdodson it looks like the image that you added for the tags page is missing an `alt`?

robdodson:
@kaycebasques I set it to alt="" which is an explicitly empty alt. That instructs screen readers to ignore the image and treat it as decorative. The h1 is still in the DOM (but visually hidden) so a screen reader will interact with that instead.

The result is that it will announce something like this:

```
"Scale on Web Case Studies, heading level 1"
```

If we give the image an alt then it will announce something like this:
```
"Scale on Web logo", image
"Scale on Web Case Studies, heading level 1"
```

I figured removing the image from the screen reader flow might make the experience a bit nicer but we could add it back if you prefer?

kaycebasques:
I trust your judgment, just wanted to make sure it wasn't an accident

robdodson:
Thanks for checking. I did wonder about it and sorta took a gamble that this approach might be more efficient for the reader. Perhaps if someone wants to use another image in the future we may revisit the decision based on the image they're using.

src/site/content/en/blog/goibibo/index.md
```diff
@@ -0,0 +1,147 @@
...
+    <p class="w-stat__figure">60<sub class="w-stat__sub">%</sub></p>
+    <p class="w-stat__desc">Increase in conversions</p>
+  </div>
+  <div class="w-stat">
+    <p class="w-stat__figure">20<sub class="w-stat__sub">%</sub></p>
+    <p class="w-stat__desc">Increase in logged-in users</p>
+  </div>
+</div>
+
+<br>
```
robdodson, line 38: This br seems to add a fair amount of white space which looks a bit odd on mobile. I removed it with devtools and thought it looked a bit better.

src/site/content/_redirects.yaml
kaycebasques: I realized it was unlikely we would need `web.dev/goibibo` for anything else so I changed it to that and removed the redirect

src/site/content/en/blog/goibibo/index.md
kaycebasques: SG; IIRC those `<br>` elements were leftover from my earlier tinkerings

PR #3940, opened 2020-09-24
Remove self-referencing tag from /tags/* page

imrobinized:

As title described, I noticed that there are empty card/post on certain tags page (`/tags/*`). Including:
- https://web.dev/tags/metrics/
- https://web.dev/tags/notifications/
- https://web.dev/tags/payments/
- https://web.dev/tags/progressive-web-apps/

It looks like this:
Screen Shot 2020-09-25 at 1 28 46 AM

I think it's because there are `tags` been defined which is exactly the same as `pathName`. And sort of created a self-referencing that resulted in the page been curated under itself.

Removing them should fix this.

### DIFF
--- a/src/site/content/en/metrics/index.njk
+++ b/src/site/content/en/metrics/index.njk
@@ -3,6 +3,4 @@ layout: path
 override:tags: []
 date: 2019-11-07
 pathName: metrics
-tags:
-  - metrics
 ---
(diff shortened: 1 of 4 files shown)


robdodson:
oh good catch!

cc @kaycebasques @jpmedley it looks like if we're making learning path landing pages they shouldn't use `tags`

robdodson (approved):

kaycebasques:
Thank you for fixing this @imrobinized

jpmedley:
So if someone clicks on a media tag, they can't see the media landing page? Seems like that's the first thing they should see.

PR #3943, opened 2020-09-24
Respect screenshot image widths from markup

developit:
Web.dev's CSS uses `*{box-sizing:border-box}`. However, for `.w-screenshot` images/videos that supply a `width=".."` attribute, this means that images are rendered 10 pixels smaller than the markup's specified width. With this PR, the width attribute value for `.w-screenshot` images/videos is taken to be the _content_ width, so that images are not slightly scaled. This only affects `.w-screenshot` because it applies a border and padding.

You can see the blurry effect in the before & after images below:

## Before: (current)
Screen Shot 2020-09-24 at 7 09 33 PM

## After: (with this change)
Screen Shot 2020-09-24 at 7 09 39 PM

Here's a live example image you can use to compare the deploy preview's sizing VS what's on stable:

https://web.dev/text-fragments/#mixing-element-and-text-fragments
https://deploy-preview-3943--web-dev-staging.netlify.app/text-fragments/#mixing-element-and-text-fragments

### DIFF
--- a/src/styles/components/_images.scss
+++ b/src/styles/components/_images.scss
@@ -51,7 +51,12 @@
 .w-figure--screenshot > img {
   border: 1px solid $GREY_300;
   padding: 4px;
+  
+  // This ensures images are scaled to the exact size specified by the
+  // width attribute, preventing HiDPI screenshots from being blurry:
+  //  <img src="800x500.png" width="400" class="w-screenshot">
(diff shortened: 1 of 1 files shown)


robdodson (approved):

robdodson:
good catch, thank you!

PR #3946, opened 2020-09-25
Fix undefined codelab paths

### LINKED ISSUE #3939: "Return to articles" button in codelabs not working
The "Return to all articles" button on some of the codelabs is not working, it links to [https://web.dev/undefined/](https://web.dev/undefined/). 

Here are the affected pages that I've found:

[https://web.dev/codelab-two-ways-to-prefetch/](https://web.dev/codelab-two-ways-to-prefetch/)
[https://web.dev/codelab-preload-critical-assets/](https://web.dev/codelab-preload-critical-assets/)
[https://web.dev/codelab-preload-web-fonts/](https://web.dev/codelab-preload-web-fonts/)

This is related to #3928 but since these links are generated I'm not sure what's the problem.

cc @robdodson

samthor:
Fixes #3939.

Having said that, I wonder if we should just simplify the `getPaths` function: we only use it literally for one reason, to get the default path for "back" URLs.

(We use `postToPathsMap` for a bunch of things, but that's more on the _discovery_ angle, things we explicitly want to list in certain places.)

### DIFF
--- a/src/site/_filters/get-paths.js
+++ b/src/site/_filters/get-paths.js
@@ -14,7 +14,24 @@
  * limitations under the License.
  */
 const postToPathsMap = require('../_data/postToPaths.js');
+const paths = require('../_data/paths');
 
 module.exports = (post) => {
-  return postToPathsMap[post.fileSlug];
+  const out = postToPathsMap[post.fileSlug];
+  if (out) {
+    return out;
+  }
+
+  // Not all posts are correctly attributed to a path (they're not linked to in the path JSON).
+  // However, guess based on their filename. This is only needed for some "return" links.
+  const parts = post.filePathStem.split('/');
+
+  if (parts.length > 2) {
+    const pathFolder = parts[2];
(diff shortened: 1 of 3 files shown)


src/site/_filters/get-paths.js
```diff
@@ -14,7 +14,24 @@
...
 module.exports = (post) => {
-  return postToPathsMap[post.fileSlug];
+  const out = postToPathsMap[post.fileSlug];
+  if (out) {
+    return out;
+  }
+
+  // Not all posts are correctly attributed to a path (they're not linked to in the path JSON).
+  // However, guess based on their filename. This is only needed for some "return" links.
+  const parts = post.filePathStem.split('/');
```
robdodson, line 27: This seems good but worth mentioning that as part of the i18n work we were thinking of getting rid of directories like /fast/, /accessible/, etc. Though it's unclear when we would actually do that.

robdodson (approved):

Issue #3864, opened 2020-09-12
CSS animations learning path


kaycebasques:
- [ ] Create `src/site/content/en/animations/animations-overview.md` (@rachelandrew)
  * Goal: conceptually explain why some animations are slow and how to create fast ones
  * Draw from [DOM to pixels](https://www.html5rocks.com/en/tutorials/speed/high-performance-animations/#toc-dom-to-pixels) and [Animations and performance](https://developers.google.com/web/fundamentals/design-and-ux/animations/animations-and-performance#tldr)
  * Non-goal: explaining in-depth how to actually create high-performance animations
- [ ] Create `src/site/content/en/animations/animations-guide.md` (@rachelandrew)
  * Goal: high-level general guidance of how to create high-performance animations (i.e. position, scale, rotation, opacity) and which properties to avoid (i.e. go to csstriggers.com and avoid anything that affects layout or paint)
- [ ] Create `src/site/_data/paths/animations.json` (@rachelandrew)
  * We'll fill out it later but for now we just need a prototype
- [ ] Create `src/site/content/en/animations/animations-examples.md` (@rachelandrew)
  * Embed a collection of Glitches (and possibly codepens?) that demonstrate common animation use cases
  * OK to draw from community (i.e. embed other people's examples onto the page)
    * Need to be [minimal](https://stackoverflow.com/help/minimal-reproducible-example) and high-quality however
- [ ] Create `src/site/content/en/animations/animations-tutorial.md` (@rachelandrew)
  * Goal: hands-on experience building the most common animation use case (we'll just have to guess at which one is most common… slide-out menu?) and using DevTools (probably Chrome *and* Firefox)
  * This one will probably require brainstorming between Rachel & Kayce
- [ ] Add support on index pages for external links (@kaycebasques)

rachelandrew:
Just adding a note that I've been referencing:

- https://developers.google.com/web/fundamentals/performance/rendering/stick-to-compositor-only-properties-and-manage-layer-count
- https://developers.google.com/web/fundamentals/performance/rendering/simplify-paint-complexity-and-reduce-paint-areas

The first two articles will cover this content, so it may be we can also redirect these.

PR #3948, opened 2020-09-25
animations overview

rachelandrew:
Part of #3864 

- Creates new folder for animations
- Adds animations.json
- First draft of "Why are some animations slow?"


src/site/content/en/animations/index.njk
```diff
@@ -0,0 +1,8 @@
+---
+layout: path
+override:tags: []
+date: 2020-09-25
+pathName: animations
+tags:
+  - animations
```
robdodson, line 7: ah we should remove these. They were causing a bug with our /tags/ landing page.

kaycebasques:
Thanks, Rachel.

My first impression is that this overview is blurring the lines between explaining 1) why some animations are slow and 2) how to create high-performance animations.

[Divio's Documentation System](https://documentation.divio.com/) does a good job explaining why we want this "separation of concerns" as it were.

I would like the overview to focus on conceptual explanation of why animations are slow. In the beginning and the end of the overview we can say, "if you don't care about the theory, check out the guide"

And then we have a guide that focuses on how to create high-performance animations, with less explanation on why they're fast. This will optimize for the type of web developer who doesn't really care why, they just want to get it done. In the beginning or end we can say "if you want to understand the theory behind high-performance animations, check out the overview"

Logistical note: feel free to do all of the work related to the animations learning path within this PR (i.e. you don't need to create a separate PR for the guide)

rachelandrew:
@kaycebasques I'm not sure about having a conceptual overview that doesn't tie this stuff back to what people know about. I don't think I'm over explaining the how, I'm giving one example (the very simple top left one) in order to tie some theory to something a reader might reasonably understand. As mostly the key thing to understand is where in the steps things happen, the top left example is slower because of the place in the process it touches.

I could remove the example, personally I'd rather keep it.

The forcing layers creation stuff could move.

I'll look at adding in your links.

I'll add to this PR the next article (the guide to creating animations) as perhaps it will be easier to figure out what goes where in the light of both.

kaycebasques:
Work is continuing in #4019

PR #3951, opened 2020-09-25
Add "memory" tag

kaycebasques:

- Adds a `memory` tag
- Applies the tag to relevant pages

### DIFF
--- a/src/site/_data/tagsData.json
+++ b/src/site/_data/tagsData.json
@@ -131,6 +131,10 @@
   "media-queries": {
     "title": "Media Queries"
   },
+  "memory": {
+    "title": "Memory",
+    "description": "News and guidance related to memory management on the web."
+  },
   "metrics": {
     "title": "Metrics"
   },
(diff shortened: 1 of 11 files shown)


kaycebasques:
@jpmedley PTAL

src/site/_data/tagsData.json
```diff
@@ -131,6 +131,10 @@
   "media-queries": {
     "title": "Media Queries"
   },
+  "memory": {
+    "title": "Memory",
+    "description": "News and guidance related to memory management on the web."
```
jpmedley, line 136:
```suggestion
    "description": "Memory management on the web."
```

jpmedley: If we wrote "News and guidance related to" on everything It would be something readers would have to wade through instead of being helpful.

jpmedley (changes requested):

developit (approved):

PR #3984, opened 2020-10-01
HTTP/3-related changes to CDN article

### LINKED ISSUE #3924: content: HTTP/3 explanation in CDN article might be too high-level

https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/fast/content-delivery-networks by @khempenius 

The page currently touches upon HTTP/3, but does so from a relatively high level. This leads to a few areas where the text/images isn't necessarily wrong, but in my opinion also not entirely accurate/nuanced. As these are new and highly complex technologies which have a high potential to be mis-interpreted and mis-used (similar to the problems we see with HTTP/2 today), I feel it's important to make these things a bit clearer from the start. 

One example is the statement that H3 uses UDP (https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/fast/content-delivery-networks/index.md#http3). This is technically correct, but not the whole story of course, as it uses QUIC as the reliable channel. By abstracting this away, it might give some readers the impression that HTTP/3 isn't reliable and/or doesn't use congestion control (enforced by the last statement of that paragraph), a mistake I've seen many people make over the past months. I can appreciate the intent behind abstracting out QUIC here for simplicity, but feel this might be a bit too much. 

A more confusing (and even potentially misleading) example though is Figure 5, showing resumed connections (https://github.com/GoogleChrome/web.dev/raw/master/src/site/content/en/fast/content-delivery-networks/cdn5.png). Here, it seems as if TLS 1.3 always enables resumed 0-RTT, which is not technically correct. Over TCP, it can only do so with TCP fast open, which the article (rightly) doesn't mention and which the left TLS 1.2 side of the image also doesn't use for fair comparison. Simply changing the headers to H2 and H3 instead of TLS also wouldn't be entirely correct, since H2 can also use TLS 1.3's early data. My suggestion would be to change the right side and show the TCP handshake there as well, showing that it just saves 1 RTT, not 2, in the TCP case. You could potentially add a third image on the right showing the QUIC/H3-specific case to drive home the point.  

A final minor nit is the statement that most major browsers support H3 since April 2020, while Safari only properly added support since September 2020. 

Thanks for considering these suggestions.
Robin

khempenius:
Fixes #3924


src/site/content/en/fast/content-delivery-networks/index.md
```diff
@@ -252,12 +252,12 @@ Although switching your CDN instance to HTTP/2 is largely a matter of flipping a
...
+    HTTP/2 introduced multiplexing, a feature that allows a single connection to be used to transmit multiple streams of data simultaneously. However, with HTTP/2, a single dropped packet blocks all streams on a connection (a phenomena known as a head-of-line blocking). With HTTP/3, a dropped packet only blocks a single stream. This improvement is largely the result of HTTP/3 using [UDP](https://en.wikipedia.org/wiki/User_Datagram_Protocol) (HTTP/3 uses UDP via QUIC) rather than [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol). This makes HTTP/3 particularly useful for data transfer over congested or lossy networks.
```
rmarx, line 260: Maybe it's interesting to include a link to the QUIC wikipedia page here as well:

```
via [QUIC](https://en.wikipedia.org/wiki/QUIC)
```

Other than that, I'd agree more detail is probably not necessary for most readers.

src/site/content/en/fast/content-delivery-networks/index.md
```diff
@@ -268,7 +268,7 @@ Although switching your CDN instance to HTTP/2 is largely a matter of flipping a
     HTTP/3 uses TLS 1.3 and therefore shares its performance benefits: establishing a new connection only requires a single round-trip and resuming an existing connection does not require any roundtrips.
 
 <figure class="w-figure">
-  <img src="./cdn5.png" alt="Comparison of connection resumption between TLS 1.2 and TLS 1.3" class="w-screenshot">
+  <img src="./cdn5.png" alt="Comparison of connection resumption between TLS 1.2, TLS 1.3, TLS 1.3 0-RTT, and HTTP/3" class="w-screenshot">
```
rmarx, line 271: I can't seem to comment on the new image itself via GitHub UI, so I'll try it here.

Main comment: the asterisk next to 1 RTT should be the double asterisk to match the legend on the bottom

Other than that, I think this approach is technically correct and better than the previous image. However, I wonder if the bottom right 0 RTT will feel strange to some readers, as it seems no network traffic is being exchanged. In similar images, I personally have taken to including the first HTTP exchange in all sequences as well to (hopefully) prevent this confusion. I'm not sure if that's needed here, but maybe something to consider.

rmarx (commented):
Thank you for the changes! Besides the double asterisk change, I've left some suggestions in the review, but feel free to ignore them.

rmarx (approved):

jpmedley (approved):

PR #3987, opened 2020-10-01
Update guidance around unload event

### LINKED ISSUE #3983: Unload event example in an article example
Just wanted to point out that these pages send conflicting signals:

https://developers.google.com/web/updates/2018/07/page-lifecycle-api#the-unload-event
https://web.dev/disallow-synchronous-xhr/#sendbeacon()

Here's a quote from developers.google.com:

> **Warning**: Using the unload event is never recommended because it's unreliable and can hurt performance in some cases. See the legacy APIs section for more details. 

* https://web.dev/disallow-synchronous-xhr
* https://github.com/GoogleChrome/web.dev/blob/master/src/site/content/en/blog/disallow-synchronous-xhr/index.md

The two pages send slightly different signals. 

Even though "`XMLHttpRequest` vs `fetch`" is the focus here, it seems a bit like the article is encouraging `unload` usage and it's possible that at least some readers aren't aware about the recommendation to avoid the unload event.

cc'ing the author: @jpmedley

jpmedley:
Fixes #3983

### DIFF
--- a/src/site/content/en/blog/disallow-synchronous-xhr/index.md
+++ b/src/site/content/en/blog/disallow-synchronous-xhr/index.md
@@ -50,15 +50,15 @@ use the  `AllowSyncXHRInPageDismissal` policy flag, which ends at the same time.
 
 Regardless of how you send data back to the server, it's best to avoid waiting
 until page unload to send all the data at once. Aside from creating a bad user
-experience, you risk data loss if something goes wrong.  Unload events [often
-don't fire on mobile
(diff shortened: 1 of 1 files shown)


samthor (approved):

kaycebasques:
Issue filer LGTM'd the update: https://github.com/GoogleChrome/web.dev/issues/3983#issuecomment-707529684

kaycebasques (approved):

PR #4007, opened 2020-10-05
"Handling range requests in a service worker" post

### LINKED ISSUE #3974: content: Range request header in a Service Worker [2020-10-14]
* [thread](https://groups.google.com/a/google.com/g/webdev-content/c/D176Jqwlwjk)
* [draft](https://docs.google.com/document/d/1FxqIQlC0Ft0WT0tvAHRRMGPEsdyrvXECUcXzoKkJWuI/edit#)

jeffposnick:
Fixes #3974

@jpmedley has already provided an editorial review.


src/site/content/en/blog/sw-range-requests/index.md
```diff
@@ -0,0 +1,95 @@
...
+
+```javascript
+self.addEventListener('fetch', (event) => {
+  // The Range: header will pass through in browsers
+  // that behave correctly.
+  event.respondWith(fetch(event.request));
+});
+```
+
+The server now gets a chance to properly handle the range request and return a partial response with a <code>[206 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206)</code>.
```
jpmedley, line 77: What's the reasoning for making `206 status code` code-formatted? Seems like this should be regular text.

jpmedley (changes requested):
This looks pretty good. My failsafe check turned up two items. Please take a look.

src/site/content/en/blog/sw-range-requests/index.md
jeffposnick: That was an artifact produced by the Google Doc to Markdown conversion script. I'll fix here and elsewhere.

jeffposnick:
Please take another look.

jpmedley (approved):

PR #4016, opened 2020-10-06
Update translated workflow

devnook:
This PR changes the translation workflow Github Action to update the translation date in a separate json file instead of front matter of the translated post.

### DIFF
--- a/.github/workflows/tranaslation-status-workflow.yml
+++ b/.github/workflows/tranaslation-status-workflow.yml
@@ -1,6 +1,6 @@
 name: Update translated date
 on:
-  push:
+  pull_request:
     branches:
       - i18n-gitlocalize
 
@@ -10,6 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Setup node
         uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/tranaslation-status-workflow.yml
```diff
@@ -32,12 +34,11 @@ jobs:
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action: Update translated date"
+          git add --all
```
robdodson, line 37: I thought `git commit -am` did an add all for you?

robdodson (approved):
I think this seems fine but I'm curious why you wanted to change to using json files instead of the file's frontmatter?

robdodson:
oh one thing to note. I think since the .json file has the same name as the post (index) then 11ty may interpret it as [a template data file](https://www.11ty.dev/docs/data-template-dir/). To avoid that you may want to give the json file a different name.

devnook:
I am using a separate file, because the frontmatter field was a bit problematic. If a field in frontmatter is present in the translated file <lang>/foo/index.md, but not in the original file en/foo/index.md, gitlocalize will always mark this file as conflicted. Moving this field to a separate file makes it transparent for gitlocalize, as it is concerned only with .md files.

At the same time, we want to have "translated" field available in the template, and this is why i keep the same name "index" - so that eleventy can see this file and include translated field in the build process.

.github/workflows/tranaslation-status-workflow.yml
devnook: it will add changed files but not newly created (untracked in git).

PR #4031, opened 2020-10-09
Add SXG article

### LINKED ISSUE #3167: content: Signed Exchanges (SXG) [Date: TBD]
* draft: none provided yet
* [thread](https://groups.google.com/a/google.com/g/webdev-content/c/n3zpXLo1nU0)
* next step: katie to draft content and ping the thread when ready for review

khempenius:
Fixes #3167


kaycebasques:
@khempenius in the future please remember to break lines at 100 characters. In the interest of time I just pushed up a commit and did it myself.

Here are two tools that make it easy:

* [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) for VS Code

* [Docs To X](https://gsuite.google.com/u/0/marketplace/app/docs_to_x/243920501110?hl=en&pann=docs_addon_widget) for Google Docs

src/site/content/en/fast/signed-exchanges/index.md
```diff
@@ -0,0 +1,328 @@
+---
+layout: post
+title: Signed Exchanges (SXGs) overview and guide
```
kaycebasques, line 3: I added `overview` and `guide` because I want to make it clear what kind of content the reader is getting.
[Divio's Documentation System](https://documentation.divio.com/introduction/#the-secret) does a good job explaining why we want to structure our docs according to the four common types of documentation. I have been lax about this in the past but I'm going to start enforcing it more on the site.

Also, when I see a title that only says `Signed Exchanges` it's unclear what type of documentation I'm getting. Is it a reference? Guide? Overview?

Another option is to follow the content types more closely and actually split this content up into an overview and a guide. I tried doing that but it's a little awkward because there isn't much conceptual information. If you think it works, though, I'm open to this approach.

CC @mihajlija

src/site/content/en/fast/signed-exchanges/index.md
```diff
@@ -0,0 +1,328 @@
...
+SXGs are the first part of the Web Packaging spec that Chromium-based browsers will implement.
+
+## How to use SXGs
+
+### Loading SXGs
+
+Initially, the primary use case of SXGs will likely be as a delivery mechanism
+for a page's main document. For this use case, a SXG could be referenced using
+the `<link>` or `<a>` tags, as well as the [`Link`
+header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link).
```
kaycebasques, line 127: Do we mean this content to an explanation of how SXG works, or is this actually supposed to be the instructional content? Because the section heading is "How to use SXGs" yet it's currently phrased as more of a conceptual explainer. E.g. stuff like "The primary use case of SXGs will likely be as a delivery mechanism for a page's main document" doesn't sound like we're giving instructions to the reader.

If we're intending this to be a how-to guide, we should rephrase all of these "how-to" sections to speak directly to the reader. Something like this:

"To use SXG to deliver a page's main document:

[code samples]

Although it's theoretically possible to use `<script>` or `` tags, there isn't much tooling support for this use case yet…"

CC @mihajlija

kaycebasques (changes requested):
@khempenius there's two things I want to sort out before we launch (it may seem like a lot of work but I don't think it's much actually):

* In the email thread I mentioned that this guide was trying to do two things: be a conceptual explainer and a how-to guide. It still has that problem and we may want to split it into two docs
* The "How to use SXGs" section isn't phrased as instructions to the reader; it's phrased more like a spec describing how it works. We should rephrase it to speak to the reader.

Can you take this on @khempenius or does it feel like too much?

khempenius:
- This is not intended to be a how-to guide

khempenius:
Updated

kaycebasques:
(For the record) This content is still trying to do too many things IMO:

* Conceptual overview
* How-to guide

If we add more SXG content over time, we should split this content up. The conceptual overview should be optimized for people that want to understand SXG. The how-to guide should be optimized for people that want just want to implement (and therefore the conceptual information is just getting in their way).

src/site/content/en/fast/signed-exchanges/index.md
```diff
@@ -0,0 +1,326 @@
+---
+layout: post
+title: Signed Exchanges (SXGs) overview
```
kaycebasques, line 3:
```suggestion
title: Signed Exchanges (SXGs)
```
I've changed my mind on this. If we're going to have conceptual and how-to content in a single page, then we should leave the name more generic. Because I would never think to look for how-to content in a page called "Signed Exchanges overview".

kaycebasques (approved):

src/site/content/en/fast/signed-exchanges/index.md
```diff
@@ -0,0 +1,326 @@
+---
+layout: post
+title: Signed Exchanges (SXGs)
+subhead: |
+  An SXG is a delivery mechanism that makes it possible to authenticate the
+  origin of a resource independently of how it was delivered.
+authors:
+  - katiehempenius
+date: 2020-10-09
```
kaycebasques, line 9:
```suggestion
date: 2020-10-14
```

PR #4035, opened 2020-10-09
updates to animations guide

rachelandrew:

- Added some examples
- Moved the comparison of bad and good practice

Not complete yet as I need to do the other DevTools walkthroughs, but wanted to check this was the right direction with @kaycebasques


kaycebasques (approved):
Yup, looks like the right direction!

PR #4039, opened 2020-10-10
Update example from makeRequest() to handle()

chrisdwheatley:
This PR updates a code example in the post https://web.dev/app-shell-ux-with-service-workers/ to use `handle` rather than `makeRequest`.

Based on https://github.com/GoogleChrome/workbox/pull/2123 `makeRequest` has been removed and `handle` is what should be used instead.

I followed https://philipwalton.com/articles/smaller-html-payloads-with-service-workers/ for an example of how `handle` should be used.

### DIFF
--- a/src/site/content/en/reliable/app-shell-ux-with-service-workers/index.md
+++ b/src/site/content/en/reliable/app-shell-ux-with-service-workers/index.md
@@ -149,15 +149,21 @@ Next, tell Workbox how to use the strategies to construct a complete, streaming 
 
 ```javascript
 workbox.streams.strategy([
-  () => cacheStrategy.makeRequest({request: '/head.html'}),
-  () => cacheStrategy.makeRequest({request: '/navbar.html'}),
+  () => cacheStrategy.handle({
+      request: new Request(getCacheKeyForURL("/head.html")),
+    }),
+  () => cacheStrategy.handle({
+      request: new Request(getCacheKeyForURL("/navbar.html")),
+    }),
   async ({event, url}) => {
     const tag = url.searchParams.get('tag') || DEFAULT_TAG;
-    const listResponse = await apiStrategy.makeRequest(…);
+    const listResponse = await apiStrategy.handle(…);
     const data = await listResponse.json();
     return templates.index(tag, data.items);
   },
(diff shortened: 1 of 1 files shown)


samthor:
cc @philipwalton

src/site/content/en/reliable/app-shell-ux-with-service-workers/index.md
```diff
@@ -149,15 +149,21 @@ Next, tell Workbox how to use the strategies to construct a complete, streaming
 
 ```javascript
 workbox.streams.strategy([
-  () => cacheStrategy.makeRequest({request: '/head.html'}),
-  () => cacheStrategy.makeRequest({request: '/navbar.html'}),
+  () => cacheStrategy.handle({
```
philipwalton, line 152: In v6 these will also need to pass an event object, so we may as well include that now so it doesn't have to be changed again in a few weeks.

philipwalton (changes requested):

jpmedley:
@jeffposnick, please take a look.

src/site/content/en/reliable/app-shell-ux-with-service-workers/index.md
jeffposnick: Good point. Though if we're thinking about this with an eye towards v6, maybe the example should be updated to use `PrecacheStrategy` directly, instead of having to configure `CacheFirst`?

philipwalton: Yeah, that's a good idea, though not sure if that'll mean we have to wait to merge this until v6 comes out?

jeffposnick: We can just merge this now (without requiring `event`) and then file an issue to remind us to update the post once v6 is fully released.

jeffposnick:
Merging, with some additional changes that we'll make when Workbox v6 is fully released. Thanks for the contribution, @chrisdwheatley!

jpmedley (approved):

jpmedley:
@philipwalton I can't merge until you sign off.

philipwalton (approved):

PR #4066, opened 2020-10-14
Add cds banner.

robdodson:

- Adds a `.w-promo-banner` element which we can use for homepage promos
- Adds a banner for CDS 2020

note - I set the alt for the banner image to `alt=""` because the image content is duplicative of the text content so I figured we should just omit it entirely from the a11y tree.


src/site/_includes/homepage.njk
```diff
@@ -87,10 +87,23 @@ eleventyComputed:
...
+  </section> #}
+
+  <section class="w-promo-banner w-layout-container">
+    <img src="/images/modules/cds20-wide_960.jpg" alt="" width="480" height="270">
+    <p>
+      Chrome Dev Summit 2020 is back on December 9-10, but this time we're
+      coming to you. We’re bringing all the latest updates, just as much
+      content, and many more Chromies. Check out the
+      <a href="http://developer.chrome.com/devsummit/?utm_source=webdevhp&utm_medium=website">
+        <span class="w-visually-hidden">Chrome Dev Summit</span>website
```
samthor, line 99: Can we put a space before "website"? Screen readers will probably abut these words.

samthor (approved):

PR #4070, opened 2020-10-15
Fix fluid typography example in min-max-clamp blog post

darthmaim:
Fixes the fluid typography example in the [min(), max(), and clamp(): three logical CSS functions to use today](https://web.dev/min-max-clamp/) blog post.

```diff
p {
-  clamp(1.5rem, 5vw, 3rem);
+  font-size: clamp(1.5rem, 5vw, 3rem);
}
```


robdodson:
Thanks! I think this was just fixed in #4072

PR #4079, opened 2020-10-17
Clarify rel=preload tag use

refi64:

This updates the wording in the page describing the usage of `rel="preload"`. At a glance, the current wording implies that the rel=preload tags replace the existing ones, rather than supplementing them.

### DIFF
--- a/src/site/content/en/lighthouse-performance/uses-rel-preload/index.md
+++ b/src/site/content/en/lighthouse-performance/uses-rel-preload/index.md
@@ -59,7 +59,8 @@ should be downloaded as soon as possible.
 ## Declare your preload links
 
 Declare preload links in your HTML to instruct the browser to download key resources
-as soon as possible.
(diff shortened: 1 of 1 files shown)


kaycebasques:
@refi64 thanks for the PR. Can you give me an example of how someone might get confused here?

kaycebasques:
Also, can you let me know if this other page that we have on preloading properly explains the idea that you're getting at: https://web.dev/preload-critical-assets/

refi64:
I think it's more just that none of the examples ever show `rel="preload"` with the existing loading tags, only ever on there own, e.g.:

> The simplest way to implement `preload` is to add a `<link>` tag to the `<head>` of the document:
> ```html
> <head>
>   <link rel="preload" as="script" href="critical.js">
> </head>
> ```

Thus at a glance, the preload tag is the only one that ever appears.

PR #4080, opened 2020-10-17
Note on WCAG risk for fluid type

aardrian:
Added two sentences with two links total noting the risk with `clamp()` and `vw` for WCAG 1.4.4.

### DIFF
--- a/src/site/content/en/blog/min-max-clamp/index.md
+++ b/src/site/content/en/blog/min-max-clamp/index.md
@@ -232,6 +232,8 @@ p {
 }
 ```
 
+When you use `vw` units or limit how large text can get with `clamp()`, there is a chance a user may be unable to scale the text to 200% of its original size. If that happens, it is WCAG failure under [1.4.4 Resize text (AA)](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=144#resize-text) so be certain to [test the results with zoom](https://adrianroselli.com/2019/12/responsive-type-and-zoom.html).
+
 
 ## Conclusion
 The CSS math functions, `min()`, `max()`, and `clamp()` are very powerful, well


jpmedley (changes requested):
Thank you for the contribution. I have a suggestion about it and we need to get the approval of the post author.

@una what do you think?

aardrian:
@jpmedley Let's pretend I am new to GitHub. I cannot find what you changed or requested to be changed.

If it is adding an article to this sentence: "…it is _**a**_ WCAG failure under…", then I am totally on board with it.

src/site/content/en/blog/min-max-clamp/index.md
```diff
@@ -232,6 +232,8 @@ p {
 }
 ```
 
+When you use `vw` units or limit how large text can get with `clamp()`, there is a chance a user may be unable to scale the text to 200% of its original size. If that happens, it is WCAG failure under [1.4.4 Resize text (AA)](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=144#resize-text) so be certain to [test the results with zoom](https://adrianroselli.com/2019/12/responsive-type-and-zoom.html).
```
una, line 235: Suggested change:

{% Aside %}
Limiting how large text can get with `max()` or `clamp()` can cause a WCAG failure under [1.4.4 Resize text (AA)](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=144#resize-text), as a user may be unable to scale the text to 200% of its original size. Be certain to [test the results](https://adrianroselli.com/2019/12/responsive-type-and-zoom.html) with zoom.
{% endAside %}

una: As an aside, this sounds like a browser bug that should be resolved, as zoom should always override preferences.

aardrian: With one exception, I am totally fine with your edit. I would not pretend to be able to do your voice, so I tried to keep it as generic as possible.

The exception is that you replaced `vw` with `max()`, and while `max()` is a good add (thanks!), viewport units (including `vh`) have their own call-out in WCAG and are the most common reason today that I end up flagging this SC in audits: [F94: Failure of Success Criterion 1.4.4 due to incorrect use of viewport units to resize text](https://adrianroselli.com/2019/12/responsive-type-and-zoom.html).

As for the browser bug question, maybe? We already convinced browsers to ignore that meta tag that prevents zooming at all, so there may be a case for a method to zoom _without_ making users have to wade into the settings / guts. Would be happy to gather your thoughts separately (partly because I am too close to the issue, methinks).

jpmedley: My requested change seems to be gone completely. 

What I questioned was the need to say that this limitation is per the spec. Nearly every requirement of the web is because of a spec. We don't say it every time and don't need to because it's assumed. Is there more to this limitation than the fact that spec authors had to pick a number and it happens to be 200? Is so, then that's the explanation that should be here, not that it's in a spec.

aardrian: I may be a bit confused here. I cited the WCAG Success Criterion that mandates the ability to scale text by at least 200%. That SC is encoded in global accessibility legislation as well, so WCAG is now more than a set of guidelines (it is not a spec), but is law. I linked to the SC directly because it provides reasons, techniques, known failures, and more. To explain all of that in this article would be well outside the scope of the article.

Does that address your concern?

jpmedley (approved):

src/site/content/en/blog/min-max-clamp/index.md
aardrian: FWIW, I would have been fine with @una's edit (and as the author I think she would have final say). But I am happy to see this approved and thanks for accepting it.

jpmedley: Yeah, I messed that up. Corrected in #4112

PR #4087, opened 2020-10-19
prefer redirect

### LINKED ISSUE #4086: Prefer redirect over static page
As per email thread, we prefer a HTTP redirect over any existing static page.

samthor:
Fixes #4086

cc @MichaelSolati this probably affects your switch to Firebase Hosting. It might be worth clarifying the behavior there.


robdodson (approved):
LGTM but I'd like @devnook to approve as well to make sure we don't inadvertently mess up any of her language work.

kaycebasques (approved):
Likewise LGTM

PR #4094, opened 2020-10-20
Update Web OTP article

agektmr:
Small changes to the article content to provide more precise information.

### DIFF
--- a/src/site/content/en/blog/web-otp/index.md
+++ b/src/site/content/en/blog/web-otp/index.md
@@ -182,8 +182,8 @@ Your OTP is: 123456.
 Did you receive the SMS and see the prompt to enter the code to the input area?
 That is how the Web OTP API works for users.
 
-{% Aside 'caution' %}
-The dialog doesn't appear for you? Check out [the FAQ](#no-dialog).
+{% Aside 'gotcha' %}
+If the dialog doesn't appear for you, check out [the FAQ](#no-dialog).
 {% endAside %}
 
 Using the Web OTP API consists of three parts:
--- a/src/site/content/en/blog/web-otp/index.md
+++ b/src/site/content/en/blog/web-otp/index.md
@@ -4,7 +4,7 @@ subhead: Help users with OTPs received through SMS
 authors:
   - agektmr
 date: 2019-10-07
-updated: 2020-09-14
+updated: 2020-10-23
 hero: hero.png
 alt: A drawing of a woman using OTP to log in to a web app.
 
@@ -59,7 +59,7 @@ and
 [Safari](https://developer.apple.com/documentation/security/password_autofill/enabling_password_autofill_on_an_html_input_element).
 
 The Web OTP API lets your app receive specially-formatted messages bound to
-your app's origin. From this, you can programmatically obtain an OTP from an SMS
+your app's domain. From this, you can programmatically obtain an OTP from an SMS
 message and verify a phone number for the user more easily.
 
 {% Aside 'warning' %}
@@ -174,6 +174,8 @@ doesn't matter who the sender is when using the Web OTP API.
 1. Send your phone the following SMS text message from the another phone.
 
 ```text
+Your OTP is: 123456.
+
 @web-otp.glitch.me #12345
 ```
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/web-otp/index.md
```diff
@@ -174,19 +174,16 @@ doesn't matter who the sender is when using the Web OTP API.
...
 That is how the Web OTP API works for users.
 
 {% Aside 'caution' %}
-* If the sender's phone number is included in the receiver's contact list, this API
-  will not be triggered due to the design of the underlying [SMS
-User Consent  API](https://developers.google.com/identity/sms-retriever/user-consent/request#2_start_listening_for_incoming_messages).
-* If you are using a work profile on your Android device and the Web OTP does
-  not work, try installing and using Chrome on your personal profile instead
-  (i.e. the same profile in which you receive SMS messages).
+The dialog doesn't appear for you? Check out [the FAQ](#no-dialog).
```
jpmedley, line 186: Why not just explain it here?

src/site/content/en/blog/web-otp/index.md
```diff
@@ -368,9 +365,10 @@ know before using it. The message must be sent after
 where `get()` was called. Finally, the message must adhere to the following
 formatting:
 
-* The message begins with (optional) human-readable text leaving the last line
+* The message begins with (optional) human-readable text that contains a 3-10
```
jpmedley, line 368:
```suggestion
* The message begins with (optional) human-readable text that contains a three to ten
```

src/site/content/en/blog/web-otp/index.md
```diff
@@ -399,6 +397,17 @@ You may also fork it and create your version:
 
 ## FAQ
 
+### The dialog doesn't appear though I'm sending a properly formatted message. What's going wrong? {: #no-dialog}
```
jpmedley, line 400: Think about the order that you have implied by this. You didn't give these instruction above. Now the reader who thinks they got everything has to figure out why their code doesn't work. Then they find this and thinks, "Why didn't they provide this information to begin with?"

jpmedley (changes requested):

src/site/content/en/blog/web-otp/index.md
agektmr: I wanted to keep this in FAQ but wanted to avoid duplicate notes earlier. Yes, I know it should be noted earlier as well. Do you think it's ok to write the same thing twice in this article?

jpmedley: No. What I'm saying is, if the article covers this in a logical place to begin with, why would it need to be a frequently asked question?

agektmr: Because I'm assuming this will be asked frequently. It's easier for readers to look for FAQ than read from the beginning to find the information. I would keep it at FAQ.

src/site/content/en/blog/web-otp/index.md
```diff
@@ -174,19 +174,16 @@ doesn't matter who the sender is when using the Web OTP API.
...
 ```text
+Your OTP is: 123456.
+
 @web-otp.glitch.me #12345
 ```
 
 Did you receive the SMS and see the prompt to enter the code to the input area?
 That is how the Web OTP API works for users.
 
 {% Aside 'caution' %}
```
kaycebasques, line 185:
```suggestion
{% Aside 'gotcha' %}
```
Our gotcha callout is more appropriate now

src/site/content/en/blog/web-otp/index.md
kaycebasques:
```suggestion
If the dialog doesn't appear for you, check out [the FAQ](#no-dialog).
```

kaycebasques: I'm OK with how Eiji has handled it. Per https://kayce.basqu.es/blog/FAQs/ (which was inspired by you, Joe), the main thing is that Eiji has mentioned the topic in a contextually relevant place. I'm OK with us having an FAQ, so long as that isn't the only source of the information.

src/site/content/en/blog/web-otp/index.md
kaycebasques: I'm OK with the current organization. Eiji has mentioned the topic in a contextually relevant place within the main flow of the document. If this content was only in the FAQ, and nowhere else, I would side with Joe. FAQs are a very common pattern and it's plausible that some people find them useful. The main thing is to ensure that we don't get lazy and only use FAQs to present information. See https://kayce.basqu.es/blog/FAQs/

kaycebasques (approved):
I'm going to step in as content lead and play tiebreaker. My rationale is explained in the comments.

PR #4101, opened 2020-10-21
[Async Clipboard article] Fix iframe permissions policy attributes, add explanation

tomayac:

- Fixes the attributes that had an unfortunate line break.
- Add explanation about permissions policy.

CC: @dway123

### DIFF
--- a/src/site/content/en/blog/async-clipboard/index.md
+++ b/src/site/content/en/blog/async-clipboard/index.md
@@ -6,7 +6,7 @@ authors:
   - thomassteiner
 description: Async Clipboard API simplifies permissions-friendly copy and paste.
 date: 2020-07-31
-updated: 2020-07-31
+updated: 2020-10-21
 tags:
   - blog
   - capabilities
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/async-clipboard/index.md
```diff
@@ -311,6 +306,21 @@ setTimeout(async () => {
 }, 2000);
 ```
 
+## Permissions policy integration
+
+In order to use the API in iframes, you need to enable it by making use of the
```
jpmedley, line 311:
```suggestion
To use the API in iframes, you need to enable it with
```

src/site/content/en/blog/async-clipboard/index.md
```diff
@@ -311,6 +306,21 @@ setTimeout(async () => {
...
+[Permissions Policy](https://w3c.github.io/webappsec-permissions-policy/),
+which defines a mechanism that allows developers to selectively enable and
+disable various browser features and APIs. Concretely, you need to pass either
```
jpmedley, line 314:
```suggestion
which defines a mechanism that allows for selectively enabling and
disabling various browser features and APIs. Concretely, you need to pass either
```

jpmedley (changes requested):

jpmedley:
@tomayac Good move on adding the permission policy info.

tomayac:
Thanks, @jpmedley. PTAL.

jpmedley (approved):

dway123:
Thank you for this change! :)

PR #4112, opened 2020-10-22
Update min-max-clamp for real.

jpmedley:


src/site/content/en/blog/min-max-clamp/index.md
```diff
@@ -232,7 +232,11 @@ p {
 }
 ```
 
-When you use `vw` units or limit how large text can get with `clamp()`, there is a chance a user may be unable to scale the text to 200% of its original size. If that happens, it is WCAG failure under [1.4.4 Resize text (AA)](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=144#resize-text) so be certain to [test the results with zoom](https://adrianroselli.com/2019/12/responsive-type-and-zoom.html).
+{% Aside 'warning' %}
+Limiting how large text can get with max() or clamp() can cause a WCAG failure
```
kaycebasques, line 236:
```suggestion
Limiting how large text can get with `max()` or `clamp()` can cause a WCAG failure
```

kaycebasques (approved):

aardrian:
I know this is way way back in ancient history, but why did you pull the two links? Policy? Quality? Something else?

PR #4122, opened 2020-10-24
chore: add eslint-plugin-lit-a11y

thepassle:

- Adds [eslint-plugin-lit-a11y](https://github.com/open-wc/open-wc/tree/master/packages/eslint-plugin-lit-a11y) to catch accessibility mistakes in lit-html templates
- Fixes found accessibility issues

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,5 @@
 module.exports = {
-  extends: ['./node_modules/gts/.eslintrc.json'],
+  extends: ['./node_modules/gts/.eslintrc.json', 'plugin:lit-a11y/recommended'],
   env: {
     es6: true,
     node: true,
@@ -24,6 +24,7 @@ module.exports = {
         SwitchCase: 1,
       },
     ],
+    'lit-a11y/role-has-required-aria-attrs': 0,
     'new-cap': 0,
     'no-case-declarations': 0,
     'node/no-unpublished-import': 0,
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -24,7 +24,7 @@ module.exports = {
         SwitchCase: 1,
       },
     ],
-    'lit-a11y/role-has-required-aria-attrs': 0,
+    'lit-a11y/no-invalid-change-handler': 0,
     'new-cap': 0,
     'no-case-declarations': 0,
     'node/no-unpublished-import': 0,
--- a/src/lib/components/EventMap/index.js
+++ b/src/lib/components/EventMap/index.js
@@ -86,7 +86,7 @@ class EventMap extends BaseStateElement {
     // prettier-ignore
     /* eslint-disable indent */
     return html`
-      <select class="w-select--borderless" @blur="${this.onChange}">
+      <select class="w-select--borderless" @change="${this.onChange}">
         ${this.eventRegions.map((region) => {
           return html`<option value=${region}>${region}</option>`;
         })}
--- a/src/lib/components/EventMap/index.js
+++ b/src/lib/components/EventMap/index.js
@@ -86,7 +86,7 @@ class EventMap extends BaseStateElement {
     // prettier-ignore
     /* eslint-disable indent */
     return html`
-      <select class="w-select--borderless" @change="${this.onChange}">
+      <select class="w-select--borderless" @blur="${this.onChange}">
         ${this.eventRegions.map((region) => {
           return html`<option value=${region}>${region}</option>`;
         })}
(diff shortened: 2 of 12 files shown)


.eslintrc.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
         SwitchCase: 1,
       },
     ],
+    'lit-a11y/role-has-required-aria-attrs': 0,
```
thepassle, line 27: I disabled this because of this code here:

https://github.com/GoogleChrome/web.dev/blob/a40cb7161d4bf00147fefc63796b9c694b57c2a3/src/lib/components/Search/index.js#L99

According to https://www.w3.org/TR/wai-aria-1.1/#combobox , an element with `role="combobox"` requires `aria-controls`, but it seems like using `aria-owns` is valid as well, so I'm not entirely sure how to go about this, both in this project, as well as in the linting plugin... Maybe @robdodson could help shine a light?

src/lib/components/EventMap/index.js
```diff
@@ -86,7 +86,7 @@ class EventMap extends BaseStateElement {
     // prettier-ignore
     /* eslint-disable indent */
     return html`
-      <select class="w-select--borderless" @change="${this.onChange}">
+      <select class="w-select--borderless" @blur="${this.onChange}">
```
robdodson, line 89: hm, why change these to `blur` instead of `change?

src/lib/components/TableOfContents/index.js
```diff
@@ -76,7 +76,7 @@ class TableOfContents extends BaseStateElement {
         <button
           class="w-button w-button--secondary w-button--icon"
           data-icon="close"
-          aria-close="Close Table of Contents"
+          aria-label="Close Table of Contents"
```
robdodson, line 79: good catch

src/lib/components/SideNav/index.js
```diff
@@ -58,7 +58,11 @@ class SideNav extends BaseElement {
       }
     }
     return html`
-      <nav @click="${this.onBlockClicks}" class="web-side-nav__container">
+      <nav
```
robdodson, line 61: Unfortunately these changes break keyboard support for the responsive side nav. I think you can safely revert them.

.eslintrc.js
robdodson: it seems like you can enable this rule and our comboxbox should set aria-controls on `.web-search__input`

robdodson: ah actually we'll have to give web-search__input an id because aria only works with id references.

robdodson (changes requested):
This looks really neat. A few things:

- I think you can revert the `blur` changes. I think things work ok with `change` unless i'm missing something?
- I think you'll need to revert the changes to the side nav as they actually break keyboard support instead of helping with it.

thepassle:
Thanks for the review @robdodson ! Ive changed the things youve pointed out, PTAL 🙂

thepassle:
@robdodson friendly reminder 🙂

src/lib/components/Search/index.js
```diff
@@ -96,7 +96,7 @@ class Search extends BaseElement {
         class="web-search__input-wrapper"
         role="combobox"
         aria-expanded="${this.expanded}"
-        aria-owns="web-search-popout__list"
+        aria-controls="searchInput"
```
robdodson, line 99: oh sorry, I think we need to have both the aria-owns and aria-controls attributes.

robdodson (commented):
sorry for the delay. I've been heads down working on another project which will hopefully be shipping near the end of this month. Just one small comment and then I think we can merge.

src/lib/components/Search/index.js
thepassle: Just to make sure: The div should have both aria-owns and aria-controls that point to the searchinput, right?

thepassle:
> sorry for the delay. I've been heads down working on another project which will hopefully be shipping near the end of this month. Just one small comment and then I think we can merge.

No worries, all good 😊

src/lib/components/Search/index.js
robdodson: I think aria-owns should point at web-search-popout__list and aria-controls should point at searchInput

robdodson: ah sorry, this is a bit confusing.

the id shouldn't be `searchInput` (I don't think we use camel case ids anywhere).
I guess I'd rename it to `web-search__input` and just have it match the class name.

thepassle: oki, fixed now 🙂

robdodson (approved):

robdodson:
looks good. i think we can land it once the tests pass

thepassle:
Thanks for bearing with me 🙏

PR #4130, opened 2020-10-27
Corrected syntax error in code example

franktopel:
Corrected syntax error caused by an unnecessary and ill-placed semi-colon

This code has an unnecessary and wrong semi-colon at the end of the line that defines the `createHTML` property on the object:

```javascript
if (window.trustedTypes && trustedTypes.createPolicy) { // Feature testing
    trustedTypes.createPolicy('default', {
        createHTML: (string, sink) => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true});
    });
}
```

My PR only removes that semi-colon.

### DIFF
--- a/src/site/content/en/blog/trusted-types/index.md
+++ b/src/site/content/en/blog/trusted-types/index.md
@@ -260,7 +260,7 @@ Sometimes you can't change the offending code. For example, this is the case if 
 ```javascript
 if (window.trustedTypes && trustedTypes.createPolicy) { // Feature testing
   trustedTypes.createPolicy('default', {
-    createHTML: (string, sink) => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true});
+    createHTML: (string, sink) => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true})
   });
 }
 ```


beaufortfrancois:
@koto Please have a look

koto:
That's correct. Sorry for the syntax error, and thanks for the patch!

src/site/content/en/blog/trusted-types/index.md
```diff
@@ -260,7 +260,7 @@ Sometimes you can't change the offending code. For example, this is the case if
 ```javascript
 if (window.trustedTypes && trustedTypes.createPolicy) { // Feature testing
   trustedTypes.createPolicy('default', {
-    createHTML: (string, sink) => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true});
+    createHTML: (string, sink) => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true})
```
kaycebasques, line 263: While we're here: is `sink` unnecessary?

kaycebasques:
@koto PTAL

src/site/content/en/blog/trusted-types/index.md
koto: For this snippet, it's not used, but I wanted to show that the sink value would be passed by the API to the function, such that e.g. custom logging, or different behavior per-sink could be implemented.

jpmedley: If we removed it, we'd have to explain in text that there's a missing argument. Otherwise, readers might question whether the article was up to date with the spec. That's a reasonable concern given the cutting edge nature of our articles.

kaycebasques (approved):

kaycebasques:
Thank you @franktopel for the fix

PR #4133, opened 2020-10-28
Remove references to index.html

jakearchibald:
It's considered bad practice to navigate to `index.html`. Usually servers make it the default resource for a 'directory'. However, this is just convention, there's nothing that requires `/foo/` and `/foo/index.html` to be the same content.

If you prefetch `/foo/index.html` then navigate to `/foo/`, it will be a cache miss.

I've changes these references to `/articles/` to avoid this confusion.


kaycebasques (approved):
TIL. Thanks, @jakearchibald.

PR #4148, opened 2020-10-30
Run SVGMOMG on SVG icons

beaufortfrancois:
This PR is about running https://jakearchibald.github.io/svgomg/ on SVG icons for optimization purposes.


beaufortfrancois:
> hm interesting. I think we usually run svgomg before checking icons in. @beaufortfrancois are you able to confirm that the new files are smaller? I'm surprised that they are no longer on a single line.

I've disabled "prettify markup" on https://jakearchibald.github.io/svgomg/ at https://github.com/GoogleChrome/web.dev/pull/4148/commits/2511b2fefa5940dab96f72ecc23578040f913a79

Looking at lines diff below, it seems like files are smaller indeed.
image

robdodson (approved):

PR #4165, opened 2020-11-02
Use CDN for side nav logo

beaufortfrancois:
This CL make sure `/images/lockup.svg` is loaded only once from the CDN on web.dev website.
Currently it is downloaded twice, once from the web.dev domain **and** another time from the CDN.

image


robdodson (approved):

robdodson:
nice one, thank you!

PR #4173, opened 2020-11-03
Remove "clipboard" and "vr" from default iframes allow

beaufortfrancois:
This PR makes sure "clipboard" and "vr" unrecognized features are not set by default in iframes. For info, the complete list of
feature names allowed is available at https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/webexposed/feature-policy-features-expected.txt and https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/feature_policy/feature_policy_features.json5.

image

### DIFF
--- a/compile-css.js
+++ b/compile-css.js
@@ -109,6 +109,8 @@ function renderTo(result, fileName) {
 
   let out = result.css.toString('utf-8');
   if (!isProd) {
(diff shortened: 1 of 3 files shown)


robdodson (approved):

robdodson:
great thank you!

PR #4175, opened 2020-11-03
Lazyload algolia

beaufortfrancois:
As most people don't use web.dev search box, we should lazyload algolia lib only when user expresses intent to search.

This WIP PR is mostly about getting input from the eng team to know how to make this better.
1. I'd prefer to use `https://web.dev/algolia-[hash].js` instead of the CDN one but I'm not sure how to do this properly.
2. I'd rather use native ES Modules if possible but rollup complains when I'm using `await import('https://cdn.jsdelivr.net/npm/algoliasearch@4/dist/algoliasearch-lite.esm.browser.js');`.

Hopefully you'll be able to help there.

**Before**
image


samthor:
So our usage of Rollup does support dynamic import. Could you try calling `import('algoliasearch/dist/algoliasearch-lite.esm.browser')` (or possibly with the full node_modules/ and js extension it doesn't have to resolve it)?

samthor:
By the way, I am happy to look at this PR, but I think this is micro-optimizing.

Our site works largely fine without any JS at all. This is a feature. I would actually happily _delay_ loading JS to normalize that. (The measure page is a good exception, where clearly folks want to use an interactive tool.)

beaufortfrancois:
> So our usage of Rollup does support dynamic import. Could you try calling `import('algoliasearch/dist/algoliasearch-lite.esm.browser')` (or possibly with the full node_modules/ and js extension it doesn't have to resolve it)?

It works with `import('algoliasearch/dist/algoliasearch-lite.esm.browser')`. Thanks!

> By the way, I am happy to look at this PR, but I think this is micro-optimizing.

I'm not sure about the "micro" part.
Loading algolia only when needed shows respect about user's bandwidth.

src/lib/components/Search/index.js
```diff
@@ -7,18 +7,28 @@ import {BaseElement} from '../BaseElement';
...
-// Create an algolia client so we can get search results.
-// These keys are safe to be public.
-const applicationID = '2JPAZHQ6K7';
-const apiKey = '01ca870a3f1cad9984ed72419a12577c';
-const indexName = 'webdev';
-const client = algoliasearch(applicationID, apiKey);
-const index = client.initIndex(indexName);
+let algoliaSearchIndex;
+
+async function loadAlgoliaLibrary() {
```
samthor, line 16: Doing two searches in rapid succession would result in two `import()` statements being called, and two `algoliasearch` things being created.

src/lib/components/Search/index.js
```diff
@@ -7,18 +7,28 @@ import {BaseElement} from '../BaseElement';
...
+  }
+
+  const p = await import('algoliasearch/dist/algoliasearch-lite.esm.browser');
+  return p.then(() => {
+    // Create an algolia client so we can get search results.
+    // These keys are safe to be public.
+    const applicationID = '2JPAZHQ6K7';
+    const apiKey = '01ca870a3f1cad9984ed72419a12577c';
+    const indexName = 'webdev';
+    const client = window.algoliasearch(applicationID, apiKey);
```
samthor, line 28: can't we use `algoliasearch` from the resolved value of the `Promise`?

beaufortfrancois: Much better indeed.
Fixed in https://github.com/GoogleChrome/web.dev/pull/4175/commits/3e31747f816f83d1dfacbf47a3966bd19528ec48

beaufortfrancois:
> > So our usage of Rollup does support dynamic import. Could you try calling `import('algoliasearch/dist/algoliasearch-lite.esm.browser')` (or possibly with the full node_modules/ and js extension it doesn't have to resolve it)?
> 
> It works with `import('algoliasearch/dist/algoliasearch-lite.esm.browser')`. Thanks!

It actually does not work as import is overridden by "src/lib/utils/underscore-import-polyfill.js" which simply loads a script like I did.

src/lib/components/Search/index.js
beaufortfrancois: It's not true actually. See https://github.com/GoogleChrome/web.dev/pull/4175#issuecomment-721556584

src/lib/components/Search/index.js
samthor: This still isn't fixed.

My suggestion is to have two methods:

1. `internalLoadAlgoliaLibrary` which does everything, doesn't affect global state, and returns the result of `client.initIndex(...)`

2. `loadAlgoliaLibrary`, which _either_ sets `algoliaIndexPromise` to the result of the above method, or returns that result.

Basically if you call `loadAlgoliaLibrary` lots and lots of times you won't cause further requests.

samthor (commented):
So I tried this PR out and I'm actually quite happy with it. I think the time we take between clicking on search and using search is long enough that you just don't notice the lazy load.

Please fix up the issue with double-load before I approve though :)

samthor:
Also the preview still seems to load the JS from the CDN; loading a bundle from our own domain is preferred.

src/lib/components/Search/index.js
beaufortfrancois: Addressed in https://github.com/GoogleChrome/web.dev/pull/4175/commits/e7b8efa76b69b4c145ec153c97b11cd04333d678

beaufortfrancois:
As discussed offline, @samthor we should find a way to get rid of the `eval` at https://github.com/GoogleChrome/web.dev/pull/4175/files#diff-10473b8a553a43776e752214a6b8db0c85e6a2d1296247f4008966af3b52d16dR29

src/lib/utils/underscore-import-polyfill.js
```diff
@@ -26,7 +26,7 @@ window['_import'] = (src) => {
     const n = Object.assign(document.createElement('script'), {
       src: `/${src}`, // Rollup generates sources only in top-level
       type: 'module',
-      onload: () => resolve(),
+      onload: () => resolve(eval(`import('${n.src}');`)), // TODO: BAD!
```
samthor, line 29: I honestly think we should now drop the dynamic `import()` polyfill now. This removes about 0.5% support level; mostly legacy Edge. When we started a year ago this number was bigger.

@robdodson WDYT? It would make our lives easier.

samthor: @beaufortfrancois Can you remove all trace of the `_import` polyfill work?

beaufortfrancois: Sure. I'm ooo today but I'll take care of this tomorrow in a separate PR.

beaufortfrancois: @samthor Here it is: https://github.com/GoogleChrome/web.dev/pull/4211

samthor: Please rebase and remove eval now that we use a real `import` :)

beaufortfrancois: @samthor Done.

samthor (approved):

PR #4177, opened 2020-11-03
Add crop filter to GIF -> video FFMPEG commands

philwolstenholme:
This prevents the common 'height not divisible by 2' error when working with GIFs of odd dimensions

Solution taken from https://stackoverflow.com/a/29582287

- Updates example FFMPEG commands to include a crop filter to make the commands more resilient when used with GIFs that may have odd (in a numerical sense) dimensions

### DIFF
--- a/src/site/content/en/fast/replace-gifs-with-videos/index.md
+++ b/src/site/content/en/fast/replace-gifs-with-videos/index.md
@@ -54,7 +54,9 @@ ffmpeg -i my-animation.gif -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420
 This tells FFmpeg to take `my-animation.gif` as the **input**, signified by the
 `-i` flag, and to convert it to a video called `my-animation.mp4`.
 
-The libx264 encoder only works with files that have even dimensions, like 320px by 240px. If the input GIF has odd dimensions you can include a crop filter to avoid FFmpeg throwing a 'height/width not divisible by 2' error:
+The libx264 encoder only works with files that have even dimensions, like 320px
+by 240px. If the input GIF has odd dimensions you can include a crop filter to
+avoid FFmpeg throwing a 'height/width not divisible by 2' error:
 
 ```bash
 ffmpeg -i my-animation.gif -vf "crop=trunc(iw/2)*2:trunc(ih/2)*2" -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420p my-animation.mp4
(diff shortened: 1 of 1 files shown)


philwolstenholme:
@googlebot I signed it!

philwolstenholme:
Working with GIFs that don't meet FFMPEG's requirements might perhaps be beyond the scope of this article (it's a performance article, not a 'how to use FFMPEG' article), but showing a command that includes the crop filter would mean fewer users copying and pasting a command that may return an error depending on their source material.

jpmedley:
@housseindjirdeh, what do you think?

jpmedley:
@rkochman Can you help us out? I don't know enough about ffmpeg to judge the accuracy or value of this. Maybe somebody in your group can give us an assessment.

rkochman:
I'd recommend we accept this change but also add a note that describes what that option is actually doing. Alternatively, we could leave the command unchanged but add a note that it only works with even height.

philwolstenholme:
I've updated the PR to follow this suggestion:

> Alternatively, we could leave the command unchanged but add a note that it only works with even height.

This keeps the main examples uncluttered, but provides a paragraph that explains to readers what they should do if they run into a 'height/width not divisible by 2' error from FFmpeg.

src/site/content/en/fast/replace-gifs-with-videos/index.md
```diff
@@ -54,6 +54,12 @@ ffmpeg -i my-animation.gif -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420
 This tells FFmpeg to take `my-animation.gif` as the **input**, signified by the
 `-i` flag, and to convert it to a video called `my-animation.mp4`.
 
+The libx264 encoder only works with files that have even dimensions, like 320x240px. If the input GIF has odd dimensions we can include a crop filter to avoid FFmpeg throwing a 'height/width not divisible by 2' error:
```
jpmedley, line 57:
```suggestion
The libx264 encoder only works with files that have even dimensions, like 320px by 240px. If the input GIF has odd dimensions you can include a crop filter to avoid FFmpeg throwing a 'height/width not divisible by 2' error:
```

philwolstenholme:
Thanks for those suggestions @jpmedley, I wasn't sure about the tone of voice to use 👍

cleytontiagosorinho (approved):

cleytontiagosorinho (approved):

cleytontiagosorinho (approved):

src/site/content/en/fast/replace-gifs-with-videos/index.md
```diff
@@ -54,6 +54,12 @@ ffmpeg -i my-animation.gif -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420
 This tells FFmpeg to take `my-animation.gif` as the **input**, signified by the
 `-i` flag, and to convert it to a video called `my-animation.mp4`.
 
+The libx264 encoder only works with files that have even dimensions, like 320px by 240px. If the input GIF has odd dimensions you can include a crop filter to avoid FFmpeg throwing a 'height/width not divisible by 2' error:
```
kaycebasques, line 57:
```suggestion
The libx264 encoder only works with files that have even dimensions, like 320px by 240px.
If the input GIF has odd dimensions you can include a crop filter to avoid FFmpeg throwing
a 'height/width not divisible by 2' error:
```

kaycebasques (approved):
Thank you for the improvement @philwolstenholme

PR #4179, opened 2020-11-04
fix failure

samthor:
This fixes a build failure. I'm not incredibly happy that we didn't see this on PR submission, it should be more obvious.


beaufortfrancois (approved):

tomayac:
Thanks for catching this and thanks for the fix. I just assumed we had stopped deploys during these interesting political times… It's not the moment to make noise about this new article…

samthor:
Did you see a failure before submitting? Just trying to diagnose why we
ended up with the deploy failure.

tomayac:
I do now, not sure when this was introduced. It compiled locally just fine, but at some point I moved the edits over to the GitHub editor.

PR #4188, opened 2020-11-05
Update index.md

meistudioli:
typo

Fixes #SOME_ISSUE_NUMBER

- 
- 
-

### DIFF
--- a/src/site/content/en/blog/min-max-clamp/index.md
+++ b/src/site/content/en/blog/min-max-clamp/index.md
@@ -97,7 +97,7 @@ conjunction with `calc()` (i.e. `font-size: max(calc(0.5vw - 1em), 2rem)`),  or
 composed (i.e. `font-size: max(min(0.5vw, 1em), 2rem)`). 
 
 {% Aside %} When using a calculation inside of a `min()`, `max()`, or `clamp()`
-function, you can nix the `calc()`. For example, writing `font-size:
+function, you can mix the `calc()`. For example, writing `font-size:
 max(calc(0.5vw - 1em), 2rem)` would be the same as `font-size: max(0.5vw - 1em,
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/min-max-clamp/index.md
```diff
@@ -97,7 +97,7 @@ conjunction with `calc()` (i.e. `font-size: max(calc(0.5vw - 1em), 2rem)`),  or
 composed (i.e. `font-size: max(min(0.5vw, 1em), 2rem)`). 
 
 {% Aside %} When using a calculation inside of a `min()`, `max()`, or `clamp()`
-function, you can nix the `calc()`. For example, writing `font-size:
+function, you can mix the `calc()`. For example, writing `font-size:
```
jpmedley, line 100: Actually, I think 'nix' is the right word here. For this to make sense, the formulas in the following line need to be reversed.

@una, please weigh in.

jpmedley (changes requested):

src/site/content/en/blog/min-max-clamp/index.md
una: nix is correct

una: You can remove the calc() might be more clear?

kaycebasques: "Nix" is technically correct but this PR is evidence that using a less common word causes confusion. So, yes, let's replace.

kaycebasques:
```suggestion
function, you can remove the call to `calc()`. For example, writing `font-size:
```

kaycebasques (approved):

PR #4202, opened 2020-11-09
Fix removal of links.

jpmedley:
Addresses comment on #4112.


jpmedley:
@kaycebasques I forgot this was even open. Is there something else I need to do?

kaycebasques (approved):

PR #4216, opened 2020-11-13
Preload app.js

beaufortfrancois:
This PR makes sure `app.js` is preloaded to improve performance. 

image

### DIFF
--- a/src/site/_includes/layout.njk
+++ b/src/site/_includes/layout.njk
@@ -27,6 +27,7 @@ eleventyComputed:
     <link rel="preconnect" href="//firebaseinstallations.googleapis.com" crossorigin="" />
     <link rel="preconnect" href="//webdev.imgix.net" crossorigin="" />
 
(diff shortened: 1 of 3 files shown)


src/site/_includes/layout.njk
```diff
@@ -27,6 +27,7 @@ eleventyComputed:
     <link rel="preconnect" href="//firebaseinstallations.googleapis.com" crossorigin="" />
     <link rel="preconnect" href="//webdev.imgix.net" crossorigin="" />
 
+    <link rel="preload" as="script" crossorigin href="{{ preloadJS.path | stripQueryParamsDev }}">
```
yoavweiss, line 30: Unless the point where the script is used also has the crossorigin attribute, you probably don't want it here.

beaufortfrancois:
According to https://www.webpagetest.org/video/compare.php?tests=201113_DiPR_92a9388b592c3171f384b4b31d98731b,201113_DiMY_e11b23142c2048864a308017712dcaf0, it seems like it doesn't have the perf win I was expecting...
I'll continue investigating.

PR #4234, opened 2020-11-17
updated gcloud dep

### LINKED ISSUE #4233: Staging is broken because of add-env deprecation
We're seeing this in CI now:

```
::add-path::/opt/hostedtoolcache/gcloud/270.0.0/x64/bin
Error: Unable to process command '::add-path::/opt/hostedtoolcache/gcloud/270.0.0/x64/bin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

See [GitHub blog](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/).

samthor:
Fixes #4233.

We don't use the offending bits directly. I'm upgrading to the latest suggested guide for `setup-gcloud` which I think will fix us. [See guidance](https://github.com/google-github-actions/setup-gcloud/tree/master/setup-gcloud).

### DIFF
--- a/.github/workflows/cleanup-versions-workflow.yml
+++ b/.github/workflows/cleanup-versions-workflow.yml
@@ -8,15 +8,16 @@ jobs:
   cleanup:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
       - name: Setup gcloud
-        uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.1
+        uses: google-github-actions/setup-gcloud@master
         with:
-          version: '270.0.0'
+          version: '290.0.1'
           service_account_email: ${{ secrets.GCLOUD_EMAIL }}
           service_account_key: ${{ secrets.GCLOUD_KEY }}
-
-      - name: Checkout
-        uses: actions/checkout@v1.0.0
+          export_default_credentials: true
 
       - name: Setup node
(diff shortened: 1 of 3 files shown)


MichaelSolati (approved):

.github/workflows/cleanup-versions-workflow.yml
```diff
@@ -8,15 +8,16 @@ jobs:
   cleanup:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
       - name: Setup gcloud
-        uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.1
+        uses: google-github-actions/setup-gcloud@master
```
robdodson, line 15: @samthor @MichaelSolati should we switch this to 0.2.0? https://github.com/google-github-actions/setup-gcloud/releases

I think OSPO policy is that if we use actions we have to use a pinned version and not master

MichaelSolati: Even if it's not I guess it'd also be good for potential breaking changes, I didn't consider that...

PR #4236, opened 2020-11-17
Added "Using conversion measurement"

### LINKED ISSUE #4235: content: split up the conversion measurement article
https://web.dev/conversion-measurement/
https://web.dev/using-conversion-measurement/

More specifically: `/using-conversion-measurement/` is new.  _Some_ of its content is new but some is just substracted from `/conversion-measurement`.

Note: I did make a content proposal via the form (on Nov 5 with the title "Using the conversion measurement API"). But I haven't seen a new web.dev issue for it i.e. no explicit validation. I also haven't received an explicit rejection.

The reason we want to do this is that we have:
A. Readers who only want to get an overview about the API 
B. Partners who have already decided to use the API and now need more technical, specific how-to guidance. 
Hence the idea of splitting up the article to have tow focused articles adapted to these audiences: `/conversion-measurement` for A and `/using-conversion-measurement` for B.

### What's the deadline?
Earlier is better :)

maudnals:
Fixes #4235 and #4251

- Created `/using-conversion-measurement`
- Moved some content from `/conversion-measurement` to `/using-conversion-measurement`


src/site/content/en/blog/using-conversion-measurement/index.md
```diff
@@ -0,0 +1,207 @@
...
+For the conversion measurement API to be available on a page, it needs to be:
+
+- Enabled on the **origin**.
+- AND—this is specific to this API—Enabled on the **browser**.
+
+With this, the API can be enabled as follows:
+
+<figure class="w-figure">
+  <img src="./api-enable.jpg" alt="Diagram: how to enable the API">
+</figure>
```
kaycebasques, line 50: This use of images seem borderline inaccessible to me. The way I think about is this: the `alt` description is the only way for people who use screen readers to get the information from the image. If the contents of the image are explained elsewhere in the page, then we don't have a problem. Because the people who use screen reader can get the information there. But if we're embedding information into diagrams, and that information is not available anywhere else on the page, then the people who use a screen reader are now at a disadvantage.

I said "borderline inaccessible" because I sense that this is basically just a summary of the information that you provide below. I just want to share my perspective on use of images and accessibility.

src/site/content/en/blog/using-conversion-measurement/index.md
```diff
@@ -0,0 +1,207 @@
+---
+title: Using the Event Conversion Measurement API
+subhead: Must-dos and tips to use the Event Conversion Measurement API.
+authors:
+  - maudn
+hero: hero.jpg
+date: 2020-11-12
+updated: 2020-11-12
```
kaycebasques, line 8:
```suggestion
#updated: 2020-11-12
```
Nitpick: we can omit this until we actually do update the article sometime in the future

src/site/content/en/blog/conversion-measurement/index.md
```diff
@@ -152,13 +150,19 @@ timing](#report-timing)).
 
 ## Browser support and similar APIs
 
-### Chrome support
+## Browser support
```
kaycebasques, line 153:
```suggestion
### Browser support
```

kaycebasques (approved):

src/site/content/en/blog/using-conversion-measurement/index.md
maudnals: Thanks for merging and for pointing this out. 
Yes, the diagrams are indeed a sum-up of the article's info, but I think exactly for this reason their content *should* be made accessible to screen readers as well. 
I'll open a patch PR for this and will make sure to address this in future articles!

PR #4239, opened 2020-11-17
Add Steven Bingler author entry

rowan-m:
Adding author metadata and images for Steven Bingler.


samthor (approved):

mihajlija:
I checked the author page for Steven but it's not there: https://deploy-preview-4239--web-dev-staging.netlify.app/authors/bingler/ 

Since @samthor has already approved this, I'm guessing author pages are only generated after the author has published something on the site and I'm going to tentatively approve so merging is not blocked :)

mihajlija (approved):

rowan-m:
Checked on the recent preview with the article and the author page appears at least now that there's an article there to populate it.

PR #4253, opened 2020-11-17
question - What is "DOM Storage"?

beauxq:
What is "DOM Storage"? Does persist protect "Local Storage"?

This change could help clarify.

(I don't even know whether this is correct. I tried googling "DOM Storage", and some results make it look like it does include Local Storage, and other results make it look like it doesn't include Local Storage. If it doesn't, then reject this pull request, and instead add a note to say that Local Storage is not protected.)

### DIFF
--- a/src/site/content/en/blog/persistent-storage/index.md
+++ b/src/site/content/en/blog/persistent-storage/index.md
@@ -134,7 +134,7 @@ data stored in:
 
 - Cache API
 - Cookies
-- DOM Storage
+- DOM Storage (Local Storage)
 - File System API (browser-provided and sandboxed file system)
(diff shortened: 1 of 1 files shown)


beauxq:
@googlebot I signed it!

jpmedley:
@petele Can you please take a look at this?

petele (approved):
LGTM

beauxq:
I would like to see a more explicit answer to the question:

Does `navigator.storage.persist()` protect "Local Storage"?

I don't know how to test whether `persist()` is protecting something or not.

jpmedley (approved):

PR #4261, opened 2020-11-18
Change wording around Android zoom support.

beaufortfrancois:
This PR is about making it clear that Android used to support zoom before, but still no pan and tilt.


beaufortfrancois:
@mihajlija Can you have a look and merge?

mihajlija (approved):

PR #4264, opened 2020-11-18
Use correct event in the article on bfcache

myshov:
Just spotted a minor issue in the article on bfcache in the section on the restoration from bfcache


myshov:
cc: @philipwalton

philipwalton (approved):

philipwalton:
Good catch, thanks!

mihajlija (approved):

Issue #4263, opened 2020-11-18
Big CLS regression


robdodson:
Here's what we're seeing in search console:
![image](https://user-images.githubusercontent.com/1066253/99568304-1b936b00-2984-11eb-8bbc-255ed5249624.png)

Here's the current scores:
![image](https://user-images.githubusercontent.com/1066253/99568330-23530f80-2984-11eb-8e22-6e88934b7d25.png)

Here's the previous scores:
![image](https://user-images.githubusercontent.com/1066253/99568365-2d750e00-2984-11eb-8c50-28715901af59.png)

@beaufortfrancois It seems strange but I'm wondering if preloading the css maybe broke something? The dates seem to line up?

beaufortfrancois:
Weird... I'll have a look tomorrow.

beaufortfrancois:
Looking at https://web.dev/async-clipboard/ which went from 0.04 to 0.55 according to screenshots above from the Search Console, I can't reproduce in WebPageTest (0.004) but PageSpeed Insights indicates a CLS of 0.23 for Origin Summary while Lab Data shows 0.

I'll continue investigating.

![image](https://user-images.githubusercontent.com/634478/99660816-2d980b00-2a63-11eb-82bf-6e42c2c02609.png)
Source: https://webpagetest.org/result/201119_DiMM_903bf887287185a6a8ab4f1892e2cc0a/

![image](https://user-images.githubusercontent.com/634478/99660993-76e85a80-2a63-11eb-8299-59758f267024.png)
![image](https://user-images.githubusercontent.com/634478/99661190-c29b0400-2a63-11eb-9444-a93e54d9d9e5.png)
Source: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fweb.dev%2Fasync-clipboard%2F

beaufortfrancois:
@robdodson I've uploaded https://github.com/GoogleChrome/web.dev/pull/4266 to fix this. Can you have a look?

rviscomi:
Keep in mind that WPT doesn't scroll the page as a real user would, so if there are layout shifts further down the page it wouldn't report those. Could there possibly be issues lazily loading images in the middle of articles? Or maybe something to do with the hero images?

beaufortfrancois:
How would you diagnose those issues?

rviscomi:
A few options:

- Use WPT scripting to scroll the page and simulate a user experience
[...]

PR #4218, opened 2020-11-13
Preload app.css

beaufortfrancois:
This PR makes sure `app.css` file is preloaded before Material Icons font (43KB font blocking any layout) to improve rendering performance due to [chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=788757).

**TLDR;**

Time to start render: 1.1s faster
Web Vitals: 4.6 -> 3.5 
Time to interactive: 40 -> 18

**Without app.css preload**
webpagetest org_result_201113_Di8C_9b8ea21c0df163a2f4642f48feba819f_1_details_
webpagetest org_result_201113_Di8C_9b8ea21c0df163a2f4642f48feba819f_1_details_ (1)

**With app.css preload**
webpagetest org_result_201113_DiW6_b1a4559458aefb7524df621a3e91064e_2_details_
webpagetest org_result_201113_DiW6_b1a4559458aefb7524df621a3e91064e_2_details_ (1)

Full details https://webpagetest.org/video/compare.php?tests=201113_DiW6_b1a4559458aefb7524df621a3e91064e%2C201113_Di8C_9b8ea21c0df163a2f4642f48feba819f&thumbSize=100&ival=100&end=visual

PR #4266, opened 2020-11-19
Mark font-display:optional for Material Icons font

beaufortfrancois:
Following https://web.dev/preload-optional-fonts/ recommandations, marking Material Icons font as `font-display:optional` should improve greatly the CLS that has jumped recently with https://github.com/GoogleChrome/web.dev/pull/4218

image
Source: https://webpagetest.org/result/201119_DiJN_70db15a47be0291dfa7149a3beb4ad6c/

FIX: https://github.com/GoogleChrome/web.dev/issues/4263

### DIFF
--- a/src/styles/fonts/_material-icons.scss
+++ b/src/styles/fonts/_material-icons.scss
@@ -3,4 +3,5 @@
   font-style: normal;
   font-weight: 400;
   src: url('/fonts/material-icons/regular.woff2') format('woff2');
+  font-display: optional;
 }


robdodson:
hm... i'm not sure if this is what caused the regression. I looked into marking this font as optional in the past but it flashes words on the screen like `search` while it waits for the font to load in.

beaufortfrancois:
Can you reproduce nowadays with app.css now preloaded?

CLS score is now 0 with `font-display: optional` while it was 1 without (see https://webpagetest.org/result/201119_DiSV_db20e729190c447c524fd2c320723946/)

beaufortfrancois:
I'll look at other stuff that could impact it then.

robdodson:
I tested this and it seemed ok on the mobile viewport (I didn't see the hamburger menu flash "menu") so I think we can merge this.

I suspect one thing that might be going on is our web components seem to be taking a while to load in. I _think_ our javascript might not be loading until the page's load event fires and maybe that's taking a long time and when the custom elements finally load in perhaps they shift things around?

robdodson (approved):

rviscomi:
It seems that [WPT](https://webpagetest.org/video/compare.php?tests=201119_DiSV_db20e729190c447c524fd2c320723946-r:2-c:0) considers the cookie consent dialog to trigger 1.0 CLS and I'm not sure if it was font related.

image
image

beaufortfrancois:
@robdodson Can you revert? I see some font text on web.dev which I think is caused by this PR.
image

robdodson:
ah yep i'll revert.

@rviscomi good find re: the cookie banner! I'll have to think a bit about how to solve that.

beaufortfrancois:
FWIW I'm not able to get the cookie banner to increase CLS locally.

beaufortfrancois:
The `<web-feedback>` component though causes layout shift.
It is fixed when setting a proper height but I'm not sure how to fix this properly

robdodson:
> The <web-feedback> component though causes layout shift.

that's good to know. I haven't looked at that element in a while but I suspect it's a custom element that generates light DOM children so it would make sense that it would shift things around a bit. Definitely something worth looking into after we get developer.chrome.com out

beaufortfrancois:
> The `<web-feedback>` component though causes layout shift.

@robdodson gentle ping post CDS

robdodson:
Thanks for the nudge. There is _a lot_ of house cleaning I'd like to do on web.dev this quarter. We're still working on our 2021 planning but addressing our CWV on every page will be a big one.

PR #4300, opened 2020-11-25
Set width & height attributes images

beaufortfrancois:
To address CLS regression we've seen at https://github.com/GoogleChrome/web.dev/issues/4263, I believe we should set width and height HTML attributes to all images as documented at https://www.youtube.com/watch?v=YM3KszYmn58

Give it a try and let me know what you think.


beaufortfrancois:
@robdodson gentle ping post CDS

robdodson:
taking a look

src/site/_transforms/width-height-images.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  const $ = cheerio.load(content);
+  $('img').each((_, elem) => {
+    const $elem = $(elem);
+    const originalSrc = $elem.attr('src').trim();
+    const isLocal = !RegExp('^(https?://|/)').test(originalSrc);
+    if (!originalSrc || !isLocal) {
+      return;
+    }
```
robdodson, line 39: There's a transform that runs before this one called `responsive-images` which sets every image to come from our image CDN. As a result, this section of code causes the transform to opt-out of measuring most images.

image

I think you'd want to run your transform either before responsive-images, or perhaps as part of responsive-images (it could be a module that exports a function which expects the cheerio object). That way we wouldn't need to parse each page twice with cheerio.

robdodson (changes requested):
Thanks for working on this! I added a note about putting it before our other transform that changes all of the image paths to use our image CDN.

src/site/_transforms/width-height-images.js
beaufortfrancois: My latest patch hopefully addresses your feedback by merging this transform into responsive-images.

beaufortfrancois (commented):
@robdodson PTAL

robdodson (approved):
LGTM! Thank you for working on this change, it really helps!

beaufortfrancois:
Woohoo! I'm happy about this one ;)

PR #4322, opened 2020-12-01
Article: Enabling shared array buffer

jakearchibald:
Don't merge until:

- [x] Add correct origin trial ID

@kaycebasques @jpmedley otherwise this is ready for review.


src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
+---
+layout: post
+title: Enabling Shared Array Buffer
```
lutzvahl, line 3: On Desktop it's more a migration. As the post is covering both, I think the title is misleading.
What about Action to be taken to use SABs?

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+- `SharedArrayBuffer` will arrive in Android Chrome 88, but only for cross-origin isolated pages (details below).
+- `SharedArrayBuffer` is currently available in desktop Chrome, but from Chrome 91 it will be limited to cross-origin isolated pages.
+
+You can make a page _cross-origin isolated_ by serving the page with these headers:
+
+```http
+Cross-Origin-Embedder-Policy: require-corp
+Cross-Origin-Opener-Policy: same-origin
+```
+
```
lutzvahl, line 40: Might be good to mention the report-only mode as well to check for errors before enabling those headers.

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
+---
+layout: post
+title: Enabling Shared Array Buffer
+authors:
+  - jakearchibald
+description: >
+  The way to enable Shared Array Buffer is changing, but the good news is it works cross-browser and cross-platform.
```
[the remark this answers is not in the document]
jakearchibald: Isn't that what it says?

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+With these mitigations in place, we reintroduced `SharedArrayBuffer` in Chrome 68 (July 2018), but only on desktop. The extra process requirements meant we couldn't do the same on mobile devices. It was also noted that Chrome's solution was incomplete, as we were only blocking 'incorrect' data formats, whereas it's possible (although unusual) that valid CSS/JS/images at guessable URLs can contain private data.
+
+Web standards folks got together to come up with a more complete solution. The solution was to give pages a way to say "I hereby relinquish my ability to bring other-origin content into this process without their opt-in". This declaration is done via [COOP and COEP headers](/coop-coep/) served with the page. The browser enforces that, and in exchange the page gains access to `SharedArrayBuffer` and other APIs with similar powers. Other origins can opt-in to content embedding via [`Cross-Origin-Resource-Policy`](<https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)>) or [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
```
lutzvahl, line 77: more complete cross browser solution

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+Then, in December 2020, I wrote this article, and you read it. Hello.
+
+And that's where we are now. Chrome 88 brings `SharedArrayBuffer` back to Android for pages that are cross-origin isolated, and Chrome 91 brings the same requirements to desktop, both for consistency, and to achieve total cross-origin isolation.
```
lutzvahl, line 81: To be clear, only Chrome on Android and not WebView or CCT

lutzvahl (changes requested):
Thanks Jake, left some comments!

src/site/content/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: How about "How to enable shared array buffer"?

src/site/content/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: Ohh, so `SharedArrayBuffer` won't work in CCT? Does it work in PWAs added to homescreen (via the APK method)?

src/site/content/en/blog/enabling-shared-array-buffer/index.md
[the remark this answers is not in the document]
lutzvahl: Yeah kind of. It's more now a cross browser thing again and we're happy to align even if we need to adopt...

It's just my point of view. No need to change it..

src/site/content/en/blog/enabling-shared-array-buffer/index.md
lutzvahl: Sorry for the confusion! 
Chrome on Android and therefore CCT as well will support COOP/COEP and doing so SABs are available.
WebView supports only COOP yet. COEP discussion is ongoing,but it's unlikely to be shipped soon

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+- `SharedArrayBuffer` is currently available in desktop Chrome, but from Chrome 91 it will be limited to cross-origin isolated pages.
+
+You can make a page _cross-origin isolated_ by serving the page with these headers:
+
+```http
+Cross-Origin-Embedder-Policy: require-corp
+Cross-Origin-Opener-Policy: same-origin
+```
+
+Once you do this, your page will not be able to load cross-origin content unless the [`Cross-Origin-Resource-Policy` header](<https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)>) allows it, or if it's requested using [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), and the usual `Access-Control-Allow-*` headers allow it.
```
mikewest, line 41: How about:

> Setting these headers means that your page won't have access to cross-origin content unless it allows you to load it, either by setting [`Cross-Origin-Resource-Policy: cross-origin`], or via the existing CORS dance that you know and love (`Access-Control-Allow-*` headers and so forth).

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+This was a problem for us browser vendors, as we want to allow sites to execute code in the form of JavaScript and WASM, but strictly control the memory this code can access. If you arrive on my website, I shouldn't be able to read anything from the internet banking site you also have open. In fact, I shouldn't even know you have your internet banking site open. These are fundamentals of web security.
+
+To mitigate this, we reduced the resolution of our high resolution timers such as `performance.now()`. However, you can _create_ a high resolution timer using `SharedArrayBuffer` by modifying memory in a tight loop in a worker, and reading it back in another thread. This couldn't be effectively mitigated without heavily impacting well-intentioned code, so `SharedArrayBuffer` was disabled altogether.
+
+A general mitigation is to ensure a webpage's system process doesn't contain sensitive data from elsewhere. Chrome had invested in a mutliprocess architecture from the start ([remember the comic?](https://www.google.com/googlebooks/chrome/big_00.html)), but there were still cases where data from multiple sites could end up in the same process:
```
mikewest, line 60:
```suggestion
A general mitigation is to ensure a webpage's system process doesn't contain sensitive data from elsewhere. Chrome had invested in a multiprocess architecture from the start ([remember the comic?](https://www.google.com/googlebooks/chrome/big_00.html)), but there were still cases where data from multiple sites could end up in the same process:
```

src/site/content/en/blog/enabling-shared-array-buffer/index.md
mikewest: TL;DR: WebView's process model sucks.

mikewest (commented):
This is great, Jake!

src/site/content/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: I avoided giving a value for CORP, since there are multiple values that might be ok here, like same-site, and I didn't want to have to go into the difference.

I like the bit about acknowledging the joy of CORS. I'll work that in.

src/site/content/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: lol even with the diff it took me ages to see the error. Thanks!

0mutlí

jakearchibald:
@mikewest ugh, can you check you've signed the thing https://cla.developers.google.com/. I assume you have, but it seems to get confused when it comes to PRs with multiple authors.

mikewest:
@googlebot I fixed it.

lutzvahl:
I've requested the OT, as soon as I've the ID I'll add it to the article and let you know!

src/site/content/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -74,7 +76,9 @@ We worked around these legacy APIs by preventing content from entering the webpa
...
+Web standards folks got together to come up with a more complete cross-browser solution. The solution was to give pages a way to say "I hereby relinquish my ability to bring other-origin content into this process without their opt-in". This declaration is done via [COOP and COEP headers](/coop-coep/) served with the page. The browser enforces that, and in exchange the page gains access to `SharedArrayBuffer` and other APIs with similar powers. Other origins can opt-in to content embedding via [`Cross-Origin-Resource-Policy`](<https://developer.mozilla.org/en-US/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)>) or [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
+
+Firefox was the first to ship `SharedArrayBuffer` with this restriction, in version 79 (July 2020).
 
 Then, in December 2020, I wrote this article, and you read it. Hello.
```
lutzvahl, line 83: One more thing to add here. Starting M89 on Desktop we've the new flag RestrictSharedArrayBuffer to already gate SABs for test purposes. The behavior using this flag is the same as it will be in M91.

FYO starting in M89 a DevTools deprecation issue will be shown in case SABs are used in none gated context. Those warnings will point the users to this post. Do we need to add something to catch those users?

jakearchibald: Is that going to show up in chrome:flags?

lutzvahl: Yes, this new flag will show up in M89 (and is already in canary starting 89.0.4351.0)

jakearchibald: I can't find this flag in `chrome:flags` in Canary 89.0.4379.0. I've tried searching for:

- shared
- array
- buffer
- gate
- restrict
- coop
- coep

…but can't find anything relevant.

agektmr (approved):
LGTM.

src/site/content/en/blog/enabling-shared-array-buffer/index.md
lutzvahl: I've just checked the CL: https://chromium-review.googlesource.com/c/chromium/src/+/2579721
This feature is in Canary starting 89.0.4351.0 - https://chromiumdash.appspot.com/commit/5089ddb0638765b738ce9690ee00be44d2e19a9c

But the feature is not UI enabled, it need's to be set via command line parameter.

jakearchibald:
I've pinged @kaycebasques and @jpmedley about this (and another article) but I'm having trouble getting a response.

kaycebasques (changes requested):
This one seems like it should also be on developer.chrome.com. The big news is "Chrome is re-enabling shared array buffer". The history section is a bit more general but it also talks a lot about Chrome's specific experience related to shared array buffer. So can you also move this content to a PR on the developer.chrome.com repo please?

lutzvahl:
> This one seems like it should also be on developer.chrome.com. The big news is "Chrome is re-enabling shared array buffer". The history section is a bit more general but it also talks a lot about Chrome's specific experience related to shared array buffer. So can you also move this content to a PR on the developer.chrome.com repo please?

Not sure about that.. Big news is: Chrome is gating SABs behind COOP/COEP - same way as other browsers already did to re-enable them.

I agree some bits are Chrome specific, but the change is a WP change and not Chrome specific.

@jakearchibald: When do you plan to release it?

jakearchibald:
@kaycebasques here's the PR https://github.com/GoogleChrome/developer.chrome.com/pull/202

kaycebasques:
@lutzvahl I will kick off an email thread for us to quickly sort this out with relevant stakeholders.

Here's my rationale for putting this on developer.chrome.com:

1) The "triggering factor" that is causing us to publish this is the fact that it's re-enabled in Chrome.

2) The call-to-action of the post is essentially "try it out by registering for an origin trial".

3) All of these sentences are phrased from Chrome's perspective:

> If you don't think you can make these changes in time for Chrome 91, you can [register for an origin trial](#register-for-ot) to retain current desktop Chrome behavior until Chrome 93.

> `SharedArrayBuffer` arrived in Chrome 60 (that's July 2017, for those of you who think of time in dates rather than Chrome versions), and everything was great. For 6 months.

> Chrome had invested in a multiprocess architecture from the start ([remember the comic?](https://www.google.com/googlebooks/chrome/big_00.html)), but there were still cases where data from multiple sites could end up in the same process:

> With these mitigations in place, we reintroduced `SharedArrayBuffer` in Chrome 68 (July 2018), but only on desktop. The extra process requirements meant we couldn't do the same on mobile devices. It was also noted that Chrome's solution was incomplete, as we were only blocking 'incorrect' data formats, whereas it's possible (although unusual) that valid CSS/JS/images at guessable URLs can contain private data.

> And that's where we are now. Chrome 88 brings `SharedArrayBuffer` back to Android for pages that are cross-origin isolated, and Chrome 91 brings the same requirements to desktop, both for consistency, and to achieve total cross-origin isolation.

> And that's where we are now. Chrome 88 brings `SharedArrayBuffer` back to Android for pages that are cross-origin isolated, and Chrome 91 brings the same requirements to desktop, both for consistency, and to achieve total cross-origin isolation.

> This is a temporary exception in the form of an 'origin trial' that gives folks more time to implement cross-origin isolated pages. It enables `SharedArrayBuffer` without requiring the page to be cross-origin isolated. The exception expires in Chrome 93, and the exception only applies to desktop Chrome.

I hear what you're saying about the meat of the story being more focused on the web platform, though. Specifically:

a) Shared array buffer (SAB) is supported in Firefox and Chrome

b) Firefox also gates SAB behind COOP/COEP

c) I also just learned that [Firefox is rolling out site isolation](https://www.zdnet.com/article/firefox-site-isolation-feature-enters-user-testing-expected-next-year/)

For the sake of getting this out before Chrome Beta, I think the fast path is to put this content mostly as-is on developer.chrome.com/blog.

Over the long run, I think the "cleanest" path would be to:

* Cover the web platform side of this story on web.dev
* Leave the Chrome-specific release stuff on developer.chrome.com/blog

kaycebasques:
Posted at https://developer.chrome.com/blog/enabling-shared-array-buffer/

PR #4337, opened 2020-12-03
update Lighthouse a11y docs

connorjclark:
ref https://github.com/GoogleChrome/lighthouse/pull/11661

LH is deprecating 2 audits and adding 5 more. Each of the new audits are all variations of "this ARIA item is missing a name" (there were 2 before already like this, each have their own page already - `aria-input-field-name` and `aria-toggle-field-name`). We'd like to link to a single page for this. Should we also delete the others and redirect?


src/site/content/en/lighthouse-accessibility/aria-name/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+The easiest way to provide an accessible name for most elements
+is to include text content inside the element.
+
+For example, this custom checkbox will be announced as "Newspaper"
+to assistive technology users:
+
+```html
+<div id="checkbox1" role="checkbox">Newspaper</div>
+```
```
robdodson, line 60: Using the "clip pattern" described in this post (https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/) you can hide the inner text on screen, but still have it announced by assistive technology. This can be especially handy if you translate your pages for localization.

```html
<a href="https://web.dev/accessible">Learn more <span class="visually-hidden">about accessibility on web.dev</span></a>
```

src/site/content/en/lighthouse-accessibility/aria-name/index.md
```diff
@@ -0,0 +1,144 @@
...
+        aria-checked="true"></li>
+</ul>
+```
+
+## Example 2: How to add accessible names to your custom ARIA input fields
+
+The easiest way to provide an accessible name for most elements
+is to include text content in the element.
+However, custom input fields typically don't have inner text,
+so you can use one of two strategies instead.
```
robdodson, line 103:
```suggestion
However, custom input fields typically don't have inner text,
so you can use one of the following strategies instead.
```

robdodson: Originally I thought this was telling me to look up the page at the previously presented options

src/site/content/en/lighthouse-accessibility/aria-name/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Option 1: Add an `aria-label` attribute to the element
+
+Use the `aria-label` attribute to define the name for the current element.
+
+For example, this custom combobox will be announced as "country"
+to assistive technology users:
+
+```html
+<div id="combo1" aria-label="country" role="combobox">England</div>
```
robdodson, line 113: should the inner text be removed from this example?

connorjclark: I don't know–I just copied this example from the page being deleted. Deleting inner text seems fine tho.

src/site/content/_redirects.yaml
```diff
@@ -237,3 +237,9 @@ redirects:
 # Redirect for the renamed WebTransport API.
 - from: /quictransport
   to: /webtransport
+
+# Redirect for consolidated Lighthouse audit for a11y names.
+- from: /aria-toggle-field-name/
+  to: /aria-name/
+- from: /aria-input-field-name/
+  to: /aria-name/
```
robdodson, line 245: I can't remember if it affects things, but I noticed we always leave the trailing slash off of the redirects so maybe we should remove them from here as well.

robdodson (approved):

src/site/content/en/lighthouse-accessibility/aria-name/index.md
```diff
@@ -0,0 +1,150 @@
...
+
+<figure class="w-figure">
+  <img class="w-screenshot" src="aria-toggle-field-name.png"
+    alt="Lighthouse audit showing custom toggle elements without accessible names">
+</figure>
+
+There are 7 audits that check for accessible names, each one covers a different
+set of [ARIA roles](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_ex). Elements
+that have any of the following ARIA roles but don't have accessible names will cause this audit to fail:
+- `aria-command-name` - `button`, `link`, `menuitem`
```
kaycebasques, line 36:
```suggestion
that have any of the following ARIA roles but don't have accessible names will cause this audit to fail:

- `aria-command-name` - `button`, `link`, `menuitem`
```
I believe I've seen some markdown renderers fail to see this as a paragraph followed by a list, so I'll add a newline to defend against that

src/site/content/en/lighthouse-accessibility/aria-name/index.md
```diff
@@ -0,0 +1,150 @@
...
+is to include text content inside the element.
+
+For example, this custom checkbox will be announced as "Newspaper"
+to assistive technology users:
+
+```html
+<div id="checkbox1" role="checkbox">Newspaper</div>
+```
+
+Using the ["clip pattern"](https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/) you can hide the inner text on screen, but still have it announced by assistive technology. This can be especially handy if you translate your pages for localization.
```
kaycebasques, line 62:
```suggestion
Using the [clip pattern](https://www.a11yproject.com/posts/2013-01-11-how-to-hide-content/) you can hide the inner text on screen, but still have it announced by assistive technology. This can be especially handy if you translate your pages for localization.
```

kaycebasques (changes requested):
A couple formatting changes. I'll also push up a commit to change the list to a table.

kaycebasques (approved):

PR #4346, opened 2020-12-06
Push Notifications article and codelabs: minor fixes

samdutton:
I checked the article and did usability tests on the codelabs — nice work!

Made a few minor fixes in this PR.

I tested in Edge and Chrome on Windows 10 on a 'real' PC (and Windows 8 via BrowserStack). The codelabs worked as expected and I updated the compatibility information in the text.

* I noticed that [push-notifications-client-codelab-complete.glitch.me](https://push-notifications-client-codelab-complete.glitch.me/) is actually the server codelab, and [push-notifications-server-codelab-complete.glitch.me](https://push-notifications-server-codelab-complete.glitch.me/) has the heading **Push notifications client**. Am I missing something, or is this just a glitch?

* I added a note to one of the codelabs (using the text from the other) about allowing push notifications in the permission dialog when you subscribe.

* It might be worth adding a `favicon.ico` file to the codelabs, just to avoid the error message in the console. I've done this for [address-form.glitch.me](https://address-form.glitch.me) and other projects.

* Nit: I changed 'I.e. if the ...' to 'If the ...' just because it looks a bit odd with the i.e. at the start of the sentence. Same for e.g.

* Nit I moved `Scope: ` and `Endpoint: ` in the message examples back onto the previous line because that's how they're displayed when you run the codelab code.

### DIFF
--- a/src/site/content/en/blog/push-notifications-client-codelab/index.md
+++ b/src/site/content/en/blog/push-notifications-client-codelab/index.md
@@ -1,52 +1,53 @@
 ---
 layout: codelab
-title: "Codelab: Build a push notification client"
+title: "Codelab: Build a push notification server"
 authors: 
   - kaycebasques
   - katejeffreys
 description: >
   A step-by-step interactive tutorial that shows you how to build a
-  client that subscribes the user to push notifications,
-  displays push messages as notifications, and unsubscribes the user
-  from push notifications.
+  server that manages push notification subscriptions and sends
+  web push protocol requests to a push service.
 date: 2020-11-11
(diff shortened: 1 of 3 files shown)


kaycebasques (approved):
Thanks for testing it out @samdutton and the updates

kaycebasques:
> I noticed that push-notifications-client-codelab-complete.glitch.me is actually the server codelab, and push-notifications-server-codelab-complete.glitch.me has the heading Push notifications client. Am I missing something, or is this just a glitch?

It was intentional. I meant that that part of the overall system is the client. I see what you're saying though about it being confusing. Happy to change it.

PR #4355, opened 2020-12-09
feat(search): add highlighting

Haroenv:

- adds highlighting to the search 
- adds attributes to retrieve to the search (only request what's shown, makes the search a bit faster)

some questions that I need to check before it should be merged (but the main idea can already been reviewed)

- [ ] should the highlight tag be mark, or strong
- [x] is the code correctly indented (I committed from GitHub, didn't run prettier)
- [x] does every hit have a title (aka should ?. be used)

### DIFF
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -348,8 +348,8 @@ class Search extends BaseElement {
         hitsPerPage: 10,
         attributesToHighlight: ['title'],
         attributesToRetrieve: ['url'],
-        highlightPreTag: '<mark>',
-        highlightPostTag: '</mark>',
+        highlightPreTag: '<strong>',
+        highlightPostTag: '</strong>',
       });
       if (this.query === query) {
         this.hits = hits;
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -344,6 +344,7 @@ class Search extends BaseElement {
       const {hits} = await index.search(query, {
         hitsPerPage: 10,
         attributesToHighlight: ['title'],
+        attributesToRetrieve: ['url'],
         highlightPreTag: '<mark>',
         highlightPostTag: '</mark>',
       });
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -3,6 +3,7 @@
  */
 
 import {html} from 'lit-element';
+import {unsafeHTML} from 'lit-html/directives/unsafe-html';
 import {BaseElement} from '../BaseElement';
 import {store} from '../../store';
 import * as router from '../../utils/router';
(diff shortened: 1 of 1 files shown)


MichaelSolati (approved):

MichaelSolati:
@robdodson I think it's working well but may not fit the style of the site, wanna take a peek real quick?

robdodson:
yeah I'm curious if it might match better if we use `<strong>` instead of `<mark>`

robdodson:
thank you for working on this @Haroenv!

francoischalifour:
We usually recommend using `mark` for highlighting so I would keep it rather than `strong` but style the tag differently—the default browser styling for `mark` is not great.

Haroenv:
I can change it to whatever you prefer :)

robdodson:
@Haroenv would you be cool swapping it to `<strong>`?

robdodson:
oh sorry, I didn't see @francoischalifour's comment.

@francoischalifour why do you recommend mark over strong? is it for accessibility reasons?

We could ship a mark style, but I imagine we'd extend our strong style.

robdodson:
btw our strong style, which is in _type.scss is:

```
strong {
  font-weight: $FONT_WEIGHT_MEDIUM;
}
```

src/lib/components/Search/index.js
```diff
@@ -341,7 +344,13 @@ class Search extends BaseElement {
     }
     try {
       const index = await loadAlgoliaLibrary();
-      const {hits} = await index.search(query, {hitsPerPage: 10});
+      const {hits} = await index.search(query, {
+        hitsPerPage: 10,
+        attributesToHighlight: ['title'],
+        attributesToRetrieve: ['url'],
+        highlightPreTag: '<mark>',
+        highlightPostTag: '</mark>',
```
Haroenv, line 352:
```suggestion
        highlightPreTag: '<strong>',
        highlightPostTag: '</strong>',
```

going with strong here, it's consistent with google search (which uses `b`)

Haroenv:
I've swapped it @robdodson :)

Haroenv:
hey @robdodson! it's the way you prefer it now :)

robdodson:
great thank you! I think I may do a follow up to make the <strong> styles a bit more noticeable

Haroenv:
Yes, seems like the bold is quite subtle now, but nice to see it live @robdodson! Don't hesitate to reach out if you have any feedback around algolia by the way

robdodson:
@Haroenv actually I was wondering if you have any advice on reducing the number of search requests we're sending over on developer.chrome.com. Currently we are doing an autocomplete as you type search, but I recently had to debounce it to only send every 500ms because sending with every keystroke was causing our algolia bill to become very expensive 😅

PR #4381, opened 2020-12-11
build(firebase): reconfigure site to be hosted by firebase, fixes #2722, fixes #4382, fixes #3913, fixes #2687

### LINKED ISSUE #2687: Refactor site structure to make /en/ directory canonical.

### LINKED ISSUE #2722: Explore moving to Firebase Hosting.
We're currently on App Engine but there might be benefits to moving to Firebase Hosting. Since we're a static site, it would mean that all of our html files would be cached at the edge nodes better distributed. It'd be great to put together a plan to think through _how_ we would actually do this sort of migration. I think both @MichaelSolati and @samthor have expressed interest in this so tentatively assigning to you both as something to collaborate on.

### LINKED ISSUE #3913: Static files served from App Engine have a Last-Modified of 1980
If you take a look at any of the static files that are served by `web.dev`, they all have

```
last-modified: Tue, 01 Jan 1980 00:00:01 GMT
```

I believe that the underlying `mtime` values for those files "on disk" in App Engine are set to that 1980 date as well.

This is messing up both `Last-Modified` timestamps as well as weak Etag generation, which relies on a combination of `size` and `mtime` to [generate](https://github.com/jshttp/etag/blob/4664b6e53c85a56521076f9c5004dd9626ae10c8/index.js#L126-L131) a (hopefully, but apparently not) unique value that is then hashed.

It's probably worth following up with some folks who know more about how files get uploaded to App Engine to figure out whether there's some way of preserving the underlying timestamps and not resetting them all to 1980. Without that, I don't think that any form of `Cache-Control` for the `web.dev` static files is safe.

C.f. https://github.com/GoogleChrome/web.dev/pull/3912

### LINKED ISSUE #4382: Feed's Last-modified set to Tue, 01 Jan 1980 00:00:01 GMT
**Describe the bug**
The Atom feed is provided with a "Last-Modified" header always set to "Tue, 01 Jan 1980 00:00:01 GMT". Hitting it with a earlier date (like "Tue, 01 Jan 1980 00:00:0**2** GMT") specified in "If-Modified-since" header will always lead to a 304 Not Modified.

**To Reproduce**
1. hit https://web.dev/feed.xml with an empty set of headers -> 200 OK with "Last-Modified" set to "Tue, 01 Jan 1980 00:00:01 GMT" 
2. hit the same URL with "If-Modified-Since: Tue, 01 Jan 1980 00:00:02 GMT" in the headers
3. get a 304 Not Modified although we know the feed was updated several times since 1980

**Expected behavior**
- The date provided by "Last-Modified" should match the last publication in the feed
- We should get a 304 only if there's nothing new in the feed since the date specified in "If-Modified-Since" 

**Screenshots**
not applicable

Thank you very much for taking care of this

MichaelSolati:
Fixes #2722, fixes #4382, fixes #3913, fixes #2687

- 
- 
-


.github/workflows/lint-and-test-workflow.yml
```diff
@@ -1,8 +1,11 @@
 name: Lint and test
 
-# By default, runs when a pull request is opened, synchronized, or reopened.
-on: pull_request
-
+# By default, runs when pushed to master or a pull request is opened, synchronized, or reopened.
```
robdodson, line 3: I think I originally made this only run on PRs because it felt a duplicative to have a PR do these tests, pass, merge it, and then run the same tests again. Do you think we need it?

cloud-secrets.js
```diff
@@ -0,0 +1,47 @@
+const {SecretManagerServiceClient} = require('@google-cloud/secret-manager');
```
robdodson, line 1: It kind of feels like we should start exploring an option where we can have a shared place for things like this and other transforms/plugins/etc that we share between web.dev and d.c.c.

.cloudbuild/deploy.yaml
```diff
@@ -0,0 +1,12 @@
+steps:
+  - name: node
+    entrypoint: npm
+    args: ['ci']
```
robdodson, line 4: Why do we do some steps here (`ci` and `cloud-secrets`) but do other steps in the firebase predeploy `build` and `algolia`? Is it because `ci` and `cloud-secrets` aren't needed if someone is deploying from their local machine?

redirects.yaml
```diff
@@ -0,0 +1,245 @@
...
+  destination: /codelab-text-compression-brotli
+
+- source: /fast/incorporate-performance-budgets-into-your-build-tools/codelab-setting-performance-budgets-with-webpack
+  destination: /codelab-setting-performance-budgets-with-webpack
+
+# Redirect all pre-v1 paths to the project root.
+# Before: web.dev/fast/remove-unused-code
+# After: web.dev/remove-unused-code
+
+- source: /accessible/:patch
```
robdodson, line 122: Not sure how I feel about the name `:patch`. Would `:path` or `:part` be ok?

src/build/output-permalink.js
```diff
@@ -0,0 +1,40 @@
...
+  const localization = data.lang !== defaultLocale;
+
+  if (data.permalink) {
+    return data.permalink.replace(
+      /^\/{{lang}}/,
+      localization ? '/localized-files/{{lang}}' : '/',
+    );
+  }
+
+  return data.page.inputPath
```
robdodson, line 35: Could you add a comment showing what an example input path would look like, and what the expected output path will look like?

This also feels like a very testable piece of code.

src/site/content/en/handbook/redirects/index.md
```diff
@@ -10,26 +10,26 @@ This guide explains how to redirect content.
...
 
-1. Open [`src/site/content/_redirects.yaml`][source].
+1. Open [`redirects.yaml`][source].
 1. Add an entry similar to the following:
 
    ```yaml
-   from: /path/to/old/page
-   to: /path/to/new/page
+   source: /path/to/old/page
+   destionation: /path/to/new/page
```
robdodson, line 18:
```suggestion
   destination: /path/to/new/page
```

src/site/content/en/handbook/redirects/index.md
```diff
@@ -10,26 +10,26 @@ This guide explains how to redirect content.
...
 
-1. Open [`src/site/content/_redirects.yaml`][source].
+1. Open [`redirects.yaml`][source].
 1. Add an entry similar to the following:
 
    ```yaml
-   from: /path/to/old/subdirectory/...
-   to: /path/to/new/subdirectory/...
+   source: /path/to/old/subdirectory/:patch
+   destionation: /path/to/new/subdirectory/:patch
```
robdodson, line 31:
```suggestion
   destination: /path/to/new/subdirectory/:patch
```

robdodson: also if you rename `:patch` to `:path` or `:part` you'll want to update it here as well

src/site/content/en/newsletter/archive/2019/11/index.njk
```diff
@@ -1,5 +1,6 @@
 ---
 title: Performance, Chrome Dev Summit and more
+permalink: /{{lang}}/newsletter/archive/2019/11/index.html
```
robdodson, line 3: should the /archive/ dir have an 11ty data file that sets the computed data for these so you don't have to specify the permalink for each page?

robdodson (commented):
looking really good so far. Just a few small comments, nothing major.

.github/workflows/lint-and-test-workflow.yml
MichaelSolati: I only included it here because the deploy script also ran the tests, but im not running test on cloud build (i dont mind running tests on cloud build, but i will also need to up the runtime on cloud build).

cloud-secrets.js
MichaelSolati: Maybe create a public repo with shared code and just `npm` install that?

.cloudbuild/deploy.yaml
MichaelSolati: Yea, the cloud secrets only needs (can be) run when in a cloud build environment.

src/site/content/en/newsletter/archive/2019/11/index.njk
MichaelSolati: I'm looking at it and I'm actually not sure if it's necessary anymore. I'll tinker with it to see if I can remove it, otherwise yea ill compute them instead.

MichaelSolati:
Keep it alive bot!

firebase.json
```diff
@@ -0,0 +1,359 @@
+{
```
samthor, line 1: Did you want to check this file in or could it be autogenerated? Since we clobber the redirects every time and all.

cloud-secrets.js
samthor, line 1: Can we have a

```js
/**
 * @​fileoverview This generates blah blah blah
 */
```

Just to get a sense of what this is doing.

gulpfile.js
```diff
@@ -90,7 +97,7 @@ gulp.task('copy-content-assets', () => {
           // serve in production.
           // e.g. en/blog/foo/bar.jpg -> en/foo/bar.jpg
```
samthor, line 98: Is this comment now invalid? We strip the en/ too ?

src/build/output-permalink.js
```diff
@@ -0,0 +1,53 @@
...
+    return data.permalink.replace(
+      /^\/{{lang}}/,
+      localization ? '/i18n/{{lang}}' : '/',
+    );
+  }
+
+  return data.page.inputPath
+    .replace(/^.\/src\/site\/content/, localization ? 'i18n/' : '')
+    .replace(/index.(md|njk)$/, '')
+    .replace(/(md|njk)$/, 'html')
```
samthor, line 51: make \. part of this regexp

src/build/output-permalink.js
```diff
@@ -0,0 +1,53 @@
...
+
+  if (data.permalink) {
+    return data.permalink.replace(
+      /^\/{{lang}}/,
+      localization ? '/i18n/{{lang}}' : '/',
+    );
+  }
+
+  return data.page.inputPath
+    .replace(/^.\/src\/site\/content/, localization ? 'i18n/' : '')
```
samthor, line 49: is the leading dot intended to be a literal dot, or a catch-all? Why would we catch a single letter on the far left?

samthor:
As mentioned in chat I would like to rename the default "localized-files" to say, "intl" (which is what Google internally uses, see Santa Tracker).

I also wonder if, in another PR, it's worth moving all the "en" content to the top-level and then moving the "pl" and so forth stuff to that same "intl/pl" folder. We just avoid moving stuff around in the build which could get confusing. WDYT?

src/lib/app.js
```diff
@@ -82,6 +82,7 @@ function serviceWorkerIsSupported(hostname) {
   // included via importScripts as well as the SW itself. (This works from mid-2018 everywhere, but
   // it seems good to double-check.)
   const allowedHostnames = [
+    'auth.web.dev',
```
devnook, line 85: Is this temporary or here to stay?

firebase.json
robdodson: yeah i think we should check this file in and remove redirects.yml

src/lib/app.js
robdodson: oh yeah we should remove this.

MichaelSolati:
> As mentioned in chat I would like to rename the default "localized-files" to say, "intl" (which is what Google internally uses, see Santa Tracker).
> 
> I also wonder if, in another PR, it's worth moving all the "en" content to the top-level and then moving the "pl" and so forth stuff to that same "intl/pl" folder. We just avoid moving stuff around in the build which could get confusing. WDYT?

I'm biased to keep the folder structure. I think for a lay person coming into the project the way languages are broken up in the content folder is straight forward. Though I'm sure this added complexity will cause issues.

src/build/output-permalink.js
MichaelSolati: It's intended to be a literal `.` is there a better way I can be writing this?

MichaelSolati: Nevermind, I tested it in regex tester and immediately saw the problem.

src/site/content/en/handbook/redirects/index.md
```diff
@@ -10,26 +10,32 @@ This guide explains how to redirect content.
 
 ## Redirect a single page
 
-1. Open [`src/site/content/_redirects.yaml`][source].
-1. Add an entry similar to the following:
-
-   ```yaml
-   from: /path/to/old/page
-   to: /path/to/new/page
+1. Open [`firebase.json`][source].
```
robdodson, line 13: I think we'd need to revert this to your previous commit

robdodson:
So I think there are three action items remaining:

- [ ] Need to sort out the redirects.yaml stuff
- [ ] Need Sam to do a CL to get the DNS ready in g3
- [ ] Need to pick a date to go live and announce a code freeze to our team

I think Michael and I can work through the first item today, and Sam's CL should be doable next week. Depending on how long it takes to get a review done, we could conceivably go live as early as end of next week. But we'll need to give the team at least a day or two heads up about the code freeze.

robdodson (approved):

PR #4411, opened 2020-12-27
Update index.md

AlbertWhite:
Fixes #SOME_ISSUE_NUMBER

- Small fix on text in the article https://web.dev/your-first-performance-budget/


samthor:
This is intentionally a fake URL but I accept that it goes to a 'real' site.

Can we come up with fake URLs with TLDs that do not exist? cc @mihajlija

kaycebasques:
Thank you for the PR @AlbertWhite. I am going to close this without merging because the content is technically "working as expected". We are intentionally using fake URLs.

FYI @mihajlija (and CC @samthor) we are not supposed to come up with our own URLs. I don't think it's worth our time to change in this case but we should avoid it in the future. https://developers.google.com/style/examples

PR #4426, opened 2021-01-02
Update index.md - method to enable

bevand10:
Minor enhancement re enabling serial support in any Chromium-based browser.

- Simplify the language used to describe the process of enabling serial support.
- Provide a copy/paste opportunity to streamline the action of enabling serial support.

### DIFF
--- a/src/site/content/en/blog/serial/index.md
+++ b/src/site/content/en/blog/serial/index.md
@@ -92,7 +92,7 @@ communication between the website and the device that it is controlling.
 
 To experiment with the Serial API locally on all desktop platforms, without an
 origin trial token, paste `chrome://flags#experimental-web-platform-features` into
-the address bar and choose `Enabled` from the drop-down list.
+the address bar and choose **Enabled** from the drop-down list.
 
 
 ### Enabling support during the origin trial phase
--- a/src/site/content/en/blog/serial/index.md
+++ b/src/site/content/en/blog/serial/index.md
@@ -91,8 +91,9 @@ communication between the website and the device that it is controlling.
 ### Enabling via chrome://flags
 
 To experiment with the Serial API locally on all desktop platforms, without an
-origin trial token, enable the `#experimental-web-platform-features` flag in
-`chrome://flags`.
+origin trial token, paste `chrome://flags#experimental-web-platform-features` into
+the address bar and choose `Enabled` from the drop-down list.
+
 
 ### Enabling support during the origin trial phase
 


bevand10:
@googlebot I signed it!

samthor:
The link is actually "chrome://flags/#enable-experimental-web-platform-features". I'm not convinced we want to do this though, since it might change.

kaycebasques:
This change seems OK to me because the instructions seem more explicit. @beaufortfrancois WDYT?

beaufortfrancois:
I'd rather have a web.dev page that explains how to enable a flag and link to it since I believe it would require screenshots to give some context. See https://web.dev/nfc/#enabling-via-chrome:flags for instance

image

src/site/content/en/blog/serial/index.md
```diff
@@ -91,8 +91,9 @@ communication between the website and the device that it is controlling.
 ### Enabling via chrome://flags
 
 To experiment with the Serial API locally on all desktop platforms, without an
-origin trial token, enable the `#experimental-web-platform-features` flag in
-`chrome://flags`.
+origin trial token, paste `chrome://flags#experimental-web-platform-features` into
+the address bar and choose `Enabled` from the drop-down list.
```
kaycebasques, line 95:
```suggestion
the address bar and choose **Enabled** from the drop-down list.
```

kaycebasques:
@beaufortfrancois are you going to work on that? @bevand10's update is more descriptive than what we currently have so I don't see why we wouldn't merge it. If @bevand10 took the time to make this update to the docs, it probably means that they had trouble with our current instructions and thinks that this extra context is useful to our readers.

beaufortfrancois:
Regarding this specific PR, https://github.com/GoogleChrome/web.dev/pull/4436 is going to remove this entire flag section soon now.

As you can see at https://www.google.com/search?q=site%3Aweb.dev+%22experimental-web-platform-features%22, we have plenty of web.dev pages that would benefit from a consistent way of explaining how to enable flags in Chromium browser. If we're going to change this in this article, we should change it everywhere and make sure authors are aware of this new page OR use the same wording  when dealing with flags.

I appreciate @bevand10 effort and I'm simply trying to make sure every article benefits from it.
I'll let @kaycebasques decide what's best for web.dev

kaycebasques:
Closing the PR since the section no longer exists.

@beaufortfrancois thanks for the perspective. I agree. It seems like one of those low-priority things that would be good but we'll probably never get to.

PR #4435, opened 2021-01-05
Update browser support info for ::marker

tedw:
Minor edit to [Custom bullets with CSS ::marker](https://web.dev/css-marker-pseudo-element/) noting desktop and iOS Safari only support `color` and `font-*` (see https://bugs.webkit.org/show_bug.cgi?id=204163).


tedw:
@googlebot I signed it!

kaycebasques (approved):
Thank you for keeping our docs up-to-date @tedw

PR #4462, opened 2021-01-15
Update Node packages

samdutton:
For this PR I used npm-check-update. After updating, `npm audit` only identifies low severity vulnerabilities.

The only apparent code change required to run the site with the updated packages is too use `yaml.load()` rather than `yaml.safeLoad()`.

**Warning: the site appears to run OK locally with the updates, and `npm run test` doesn't identify any errors, but I haven't done any other testing, and this PR may break something that I'm not aware of.**

I've added a DO NOT MERGE label just in case.


samthor:
@samdutton can you rebase this PR and we can look again?

robdodson:
I think https://github.com/GoogleChrome/web.dev/pull/4820 could work but we now have a ton of linter errors to fix :\

Issue #4457, opened 2021-01-14
content: Update the LCP post for Chrome 88 changes


philipwalton:
* Use this template to request updates to existing content.
* Delete any fields that aren't relevant to your issue. 
* Update the title of this issue with a brief summary of the work to be done.

### What page(s) need to be updated?

https://web.dev/lcp/

### Why is this update needed?

That page currently includes the following aside:

> In the future, elements removed from the DOM may still be considered as LCP candidates. Research is currently being done to assess the impact of this change. You can follow the metrics CHANGELOG to stay up-to-date.

This change was made and will ship in Chrome 88. After Chrome 88 is released we should update the post with the new info and perhaps change the aside to mention that older Chrome versions had the original behavior.

See the [CHANGELOG](https://chromium.googlesource.com/chromium/src/+/master/docs/speed/metrics_changelog/lcp.md) for details.

### What's the deadline?

After Chrome 88 ships (~Jan 19)

#### Is it a hard deadline?

No.

PR #4500, opened 2021-01-21
Correct metrics content for LCP to include removed elements

runjak:
Hey 👋 

When reading [web.dev/lcp/](https://web.dev/lcp/) I've noted the aside section pointing to WICG/largest-contentful-paint/issues/41.

My current understanding is that the issue has been closed for a few months and that the decision was made to keep considering removed elements for LCP. Hence the adjustment.

I believe to have signed the CLS accordingly.
I'm a  bit unsure whether an issue should still be created to link it in the commit? Please advise.

### DIFF
--- a/src/site/content/en/metrics/lcp/index.md
+++ b/src/site/content/en/metrics/lcp/index.md
@@ -161,18 +161,7 @@ the DOM as new content becomes available. If any of these new elements is
 larger than the previous largest contentful element, a new `PerformanceEntry`
 will also be reported.
 
-If a page removes an element from the DOM, that element will no longer be
-considered. Similarly, if an element's associated image resource changes (e.g.
-changing `img.src` via JavaScript), then that element will stop being considered
-until the new image loads.
-
-{% Aside %}
-  In the future, elements removed from the DOM may still be considered as LCP
-  candidates. [Research is currently being
-  done](https://github.com/WICG/largest-contentful-paint/issues/41#issuecomment-583589387)
-  to assess the impact of this change. You can follow the metrics
-  [CHANGELOG](http://bit.ly/chrome-speed-metrics-changelog) to stay up-to-date.
-{% endAside %}
+If a page removes an element from the DOM, that element will still be considered.
 
 The browser will stop reporting new entries as soon as the user interacts with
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/lcp/index.md
```diff
@@ -161,18 +161,7 @@ the DOM as new content becomes available. If any of these new elements is
...
-until the new image loads.
-
-{% Aside %}
-  In the future, elements removed from the DOM may still be considered as LCP
-  candidates. [Research is currently being
-  done](https://github.com/WICG/largest-contentful-paint/issues/41#issuecomment-583589387)
-  to assess the impact of this change. You can follow the metrics
-  [CHANGELOG](http://bit.ly/chrome-speed-metrics-changelog) to stay up-to-date.
-{% endAside %}
+If a page removes an element from the DOM, that element will still be considered.
```
mihajlija, line 164:
```suggestion
In order to correctly compute LCP in cases such as image carousels, the LCP algorithm 
considers removed elements. If a page removes an element from the DOM, that element 
will still be considered.
```

mihajlija (approved):

mihajlija:
Thanks for the update @runjak!

philipwalton:
I think I would phrase this slightly differently. The main information we want to convey here is that an element that was removed from the DOM can continue to be considered the LCP candidate. I think we should lead with that. I don't think the bit about carousels is relevant enough to be the first thing mentioned (I think we can mention it in the aside).

How about something like:

"If an element that is currently the largest contentful element is removed from the viewport (or even removed from the DOM), it will remain the largest contentful element unless a larger element is rendered."

Then in an aside we can add the note:

"Prior to Chrome 88, removed elements were not considered as largest contentful elements, and removing the current candidate would trigger a new `largest-contentful-paint` entry to be dispatched. However, due to popular UI patterns such as image carousels that often removed DOM elements, the metric was updated to more accurately reflect what users experience. See the [CHANGELOG](https://chromium.googlesource.com/chromium/src/+/master/docs/speed/metrics_changelog/2020_11_lcp.md) for more details."

WDYT?

philipwalton:
Also we can close this issue once this change is finalized: https://github.com/GoogleChrome/web.dev/issues/4457

PR #4547, opened 2021-01-28
webrtc: fix gfn brand usage

fippo:
see e.g. first item on
 https://www.nvidia.com/en-us/geforce-now/faq/
I'll ask around if this needs more marketing related things from our end.


kaycebasques (approved):
Good catch, thanks

PR #4553, opened 2021-01-29
Added localhost warning note

maudnals:
A [developer reaction](https://github.com/WebGDEProgram/main/issues/130) to the article surfaced that we need to make it clearer that HTTPS is only for special cases.

- Added a disclaimer on top of the article
- Add one case to the image


maudnals:
@mihajlija PTAL when you have time, thanks!

kaycebasques (approved):

PR #4564, opened 2021-02-01
fix(meta): replace lt/gt sign with single l/r pointing angle quotation mark

MichaelSolati:


src/site/_includes/components/Meta.js
```diff
@@ -106,10 +106,21 @@ module.exports = (locale, page, collections, renderData = {}) => {
 
   function renderTwitterMeta() {
     const meta = getMetaByPlatform('twitter');
+    /**
+     * We replace the `<` and `>` characters for Twitter because HTML tags
```
robdodson, line 110: Maybe explicitly state that Twitter also strips out `&lt;` and `&gt;` since I think most folks would normally reach for those.

robdodson (approved):

PR #4570, opened 2021-02-02
feat: Copying a heading copies the link to the clipboard

tomayac:

- Copying a heading copies the link to the clipboard.


kaycebasques:
Oh, nice feature!

tomayac:
> Oh, nice feature!

Thanks :-) 

An additional idea might be to set `cursor: pointer` for all headings with an `id` (now, only the ‘#’ has that style).

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
...
+        // Only run once.
+        if (heading.dataset.toasted) {
+          return;
+        }
+        try {
+          navigator.clipboard.writeText(
+            `${location.origin}${location.pathname}#${heading.id}`,
+          );
+          const temp = heading.innerHTML;
+          heading.innerHTML += '&nbsp;<small>(📋 Copied.)</small>';
```
tomayac, line 57:
```suggestion
          heading.innerHTML += '&nbsp;<small>(📋 Copied)</small>';
```

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
   }
   return parts.join('/');
 }
+
+export function copyLinkToClipboard() {
+  if (!('clipboard' in navigator)) {
+    return;
+  }
+  document
```
robdodson, line 38:
```suggestion
  }

  document
```

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
   }
   return parts.join('/');
 }
+
+export function copyLinkToClipboard() {
```
robdodson, line 34: I think it might be better to give this utility its own file. I'm not sure if it makes sense in this url package.

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
...
+  if (!('clipboard' in navigator)) {
+    return;
+  }
+  document
+    .querySelector('main')
```
robdodson, line 39: Instead of adding a listener to each heading, what do you think about adding a single listener to the body and then doing:

```js
const link = e.target.closest('.w-headline-link');
if (!link) {
  return;
}
```

`.w-headline-link` is the link that we use on the hash anchor. I think I'd rather consolidate the functionality on that element since technically a heading is non-interactive and not clickable.

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
...
+    // ToDo: Use `:is(h2, h3, h4, h5, h6)[id]` once support is better.
+    .querySelectorAll('h2[id], h3[id], h4[id], h5[id], h6[id]')
+    .forEach((heading) => {
+      heading.addEventListener('click', (ev) => {
```
robdodson, line 43:
```suggestion
      heading.addEventListener('click', (e) => {
```

I don't know if we have a hard rule on this but I think we tend to use `e` for event names. Although I do see one other instance of `(ev)`. But maybe let's go with `e` in this case.

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
...
+        // Don't jump when the '#' is clicked.
+        if (ev.target.nodeName === 'A') {
+          ev.preventDefault();
+        }
```
robdodson, line 47: I'm not sure why we'd want to prevent this? I looked at css-tricks and it allows the jump.

src/lib/urls.js
```diff
@@ -30,3 +30,39 @@ export function getCanonicalPath(path) {
...
+          const temp = heading.innerHTML;
+          heading.innerHTML += '&nbsp;<small>(📋 Copied)</small>';
+          heading.dataset.toasted = 'toasted';
+          setTimeout(() => {
+            heading.innerHTML = temp;
+            delete heading.dataset.toasted;
+          }, 2000);
+        } catch (err) {
+          console.warn(err.name, err.message);
+        }
```
robdodson, line 65: I was thinking we should remove this. I understand the intent but I found the set timeout can look a bit odd if you mouse out or if you use a keyboard to click the hash. You end up with a sort of oddly spaced "(Copied)" string.


https://user-images.githubusercontent.com/1066253/107593356-06fdf200-6bc4-11eb-92f3-a90561052e04.mp4

robdodson (changes requested):
Cool feature! Added some thoughts.

robdodson:
Also we'll need to rebase this as `app.js` just went through some big changes.

src/lib/urls.js
tomayac: Preventing the jump is pretty common best practice since it avoids user confusion. FWIW, the implementation on WebFu ([sample link](https://developers.google.com/web/fundamentals/performance/critical-rendering-path#critical-rendering-path)) did this, too.



Screen Shot 2021-02-11 at 16 11 07

src/lib/urls.js
tomayac: Done.

src/lib/urls.js
tomayac: The advantage of my logic is that you can click the headline or the '#' with the same effect. You could get the same with just one global listener, but then the code to disregard irrelevant clicks and to find the closest headline or headline link to show the toast would add complexity. Per article there aren't that many headlines that the current approach would cause a serious issue.

src/lib/urls.js
tomayac: I just thought since it deals with URLs it might be the right place, but happy to move to its own file.

src/lib/urls.js
tomayac: The old WebFu had a toast in the lower left corner:

Screen Shot 2021-02-11 at 16 28 25

I kinda like the proximity of this solution to the user action for added context.

tomayac:
> Also we'll need to rebase this as `app.js` just went through some big changes.

Refactored to include the changes in `app.js`. PTAL.

robdodson:
Hey Thomas, just wanted to mention that I haven't forgotten about this PR. I'm trying to find some time to do a follow up review.

src/lib/urls.js
robdodson: I'm supportive of the PR but still not a big fan of the copy UI... We do have a snackbar similar to devsite but it's a little tricky to use. You can see it in Snackbar.js and how it's used in actions.js.

tomayac:
> Hey Thomas, just wanted to mention that I haven't forgotten about this PR. I'm trying to find some time to do a follow up review.

Sounds good. I'm converting this PR back to Draft PR and will work on it after some more ciritical big upcoming Google event work.

tomayac:
Active

robdodson:
Copying this comment down here in case it was missed:

I'm supportive of the PR but still not a big fan of the copy UI... We do have a snackbar similar to devsite but it's a little tricky to use. You can see it in Snackbar.js and how it's used in actions.js.

tomayac:
> Copying this comment down here in case it was missed:
> 
> I'm supportive of the PR but still not a big fan of the copy UI... We do have a snackbar similar to devsite but it's a little tricky to use. You can see it in Snackbar.js and how it's used in actions.js.

It's on my mind, I simply haven't gotten to it yet. I have kept and will keep the Issue active, fighting the @stale bot :-)

robdodson:
@tomayac I added the `accepted` label to make the stalebot chill

tomayac:
Closing and starting over.

PR #4572, opened 2021-02-02
Add a note about `registrationStrategy`

xylo04:
I was following this guide to add PWA functionality to an Angular 11 app, but the Lighthouse tool was claiming I [still had no service worker](https://web.dev/service-worker/?utm_source=lighthouse&utm_medium=devtools) despite the fact that it was there. I also noticed that the service worker would take a long time to be installed.

Lo and behold, the default strategy can [take up to 30 seconds](https://stackoverflow.com/a/61152966/587091) to install the service worker. I don't know if you want this article to recommend `registerImmediately`, but at least the `registrationStrategy` property should be discussed.

### DIFF
--- a/src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
+++ b/src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
@@ -44,6 +44,17 @@ This command will:
 * Add the [`theme-color`](/themed-omnibox) `<meta>` tag to `index.html`.
 * Create app icons in the `src/assets` directory.
 
+## Register the service worker at startup
+
+In `app.module.ts`, make sure the service worker starts immediately so it can serve pages when offline:
+
+```typescript
+ServiceWorkerModule.register('ngsw-worker.js', {
+  enabled: environment.production,
+  registrationStrategy: 'registerImmediately',
+}),
+```
+
 ## Customize your PWA
 
 The [Precaching with the Angular service worker](/precaching-with-the-angular-service-worker) post explains how to configure the Angular service worker. There you can find how to specify which resources you want the service worker to cache and what strategy it should use to do so.
--- a/src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
+++ b/src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
@@ -44,16 +44,7 @@ This command will:
 * Add the [`theme-color`](/themed-omnibox) `<meta>` tag to `index.html`.
 * Create app icons in the `src/assets` directory.
 
-## Register the service worker at startup
-
-In `app.module.ts`, make sure the service worker starts immediately so it can serve pages when offline:
-
-```typescript
-ServiceWorkerModule.register('ngsw-worker.js', {
-  enabled: environment.production,
-  registrationStrategy: 'registerImmediately',
-}),
-```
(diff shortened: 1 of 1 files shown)


kaycebasques:
@mgechev @jeffposnick WDYT?

src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
```diff
@@ -44,6 +44,17 @@ This command will:
 * Add the [`theme-color`](/themed-omnibox) `<meta>` tag to `index.html`.
 * Create app icons in the `src/assets` directory.
 
+## Register the service worker at startup
+
+In `app.module.ts`, make sure the service worker starts immediately so it can serve pages when offline:
```
jeffposnick, line 49: I am not super-familiar with `ngsw`, but I think this option just controls whether a first-time visitor will get the service worker **registered** immediately.

Once a service worker is registered, for repeat visitors, it will always start immediately prior to any network request made by an in-scope client.

jeffposnick (changes requested):
I have a nit about the phrasing of "starts immediately" vs. "registers immediately".

But beyond that, a bigger question is that if this is guidance that we expect all Angular developers to follow, especially in light of PWA installability checks, perhaps it means that the default behavior in `ngsw` should be reevaluated?

Our [general guidance](https://developers.google.com/web/fundamentals/primers/service-workers/registration) is to defer registration until you're confident that it won't interfere with more important network requests, but it sounds like `ngsw` might be overcompensating if it waits up to 30 seconds.

I'll defer to @mgechev and folks on the Angular team about that.

src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
```diff
@@ -44,6 +44,17 @@ This command will:
...
+
+```typescript
+ServiceWorkerModule.register('ngsw-worker.js', {
+  enabled: environment.production,
+  registrationStrategy: 'registerImmediately',
```
mgechev, line 54: I'd suggest not recommending this strategy. By default we use `registerWhenStable` with `30000ms` timeout to make sure the app is stable (i.e. no ongoing micro/macro tasks).

It might be worth rephrasing/changing the paragraph to just discuss briefly the different registration strategies from the [docs](https://angular.io/api/service-worker/SwRegistrationOptions#registrationStrategy), but I'd say the majority of folks (95%+) would want to go with the default.

xylo04: Fair enough, but I don't think my app (https://github.com/k0swe/forester) is doing anything too exotic, and yet service worker registration takes a long time and Lighthouse can't identify it as a PWA. It was incredibly difficult to find information about how to resolve that situation.

xylo04: Maybe this experience isn't common enough to warrant adding to docs here. Is there a way to see what tasks are delaying my SW registration?

xylo04: Changed, PTAL.

src/site/content/en/angular/creating-pwa-with-angular-cli/index.md
xylo04: Changed, PTAL.

jeffposnick (approved):
I'm fine with the language now. Thanks!

(I'll again defer on any `ngsw`-policy related questions to the Angular folks.)

jpmedley (approved):

PR #4608, opened 2021-02-06
Trust Tokens docs: clean up a few syntax errors, and add context to a…

davidvancleve:
… callout

This change makes two small changes to web.dev/trust-tokens:

* It updates examples' syntax to no longer use the `issuer` argument in issuance and redemption operations (it's not used in these operations), and to swap it out for `issuers`, a list, in signing operations.
* It replaces a reference to "version 2 of the API" in a "Caution" callout with "updated syntax available since Chrome 88"; the reference to "version 2" seems hard to interpret since we don't give the reader any definition of the different versions.

**Reviewer**: It struck me that the "Caution" callout style is a little alarming and might be better suited for some kind of security-relevant usage warning than for a notification about a minimum version necessary for compatibility with an update to the API's syntax. Is there a more appropriate parameter I could provide to the {%aside%} tag?

### DIFF
--- a/src/site/content/en/blog/trust-tokens/index.md
+++ b/src/site/content/en/blog/trust-tokens/index.md
@@ -113,7 +113,7 @@ The following is adapted from
 [sample code in the API explainer](https://github.com/WICG/trust-token-api#sample-api-usage).
 
 {% Aside 'caution' %}
-The code in this post uses version 2 of the API.
+The code in this post uses updated syntax available since Chrome 88.
 {% endAside %}
 
 Imagine that a user visits a news website (`publisher.example`) which embeds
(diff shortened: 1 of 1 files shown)


mihajlija:
cc @samdutton for a quick review of the proposed syntax.

For the aside, we can use the `note` style https://web.dev/handbook/web-dev-components/#note-asides.

samdutton (approved):
Thanks David!

samdutton:
Thank you both. Added approval.

davidvancleve:
Thanks, Milica! Done.

Thanks for the review, Sam.

davidvancleve:
> Code owner review required
> Waiting on code owner review from GoogleChrome/web-dev-approvers and/or GoogleChrome/web-dev-content. 

I'm still new to GitHub. Is it possible to see the members of these groups so I know who to ping? It seems like Sam's LGTM didn't make this check pass.

samdutton:
I think it needs a 'code owner review'. @mihajlija — can you do that?

PR #4613, opened 2021-02-07
Remove SPA and service worker

### LINKED ISSUE #4571: SPA refactor
- Need to figure out how to bundle web components on a per-page basis. This code lives in `lib/pages`.
- Redo rollup bundle
  - Maybe look into using WMR? ([issue](https://github.com/preactjs/wmr/issues/182))
- Refactor gulp tasks
- Only load firebase on the /measure page
- Injected styles should be as early as possible in the head

robdodson:
Fixes #4571

- Replaces comiple-css.js with the sass gulp task from d.c.c. and adds purgecss/minifyhtml steps from d.c.c.
  - In a follow up PR I'd like to refactor the gulpfile so it's more like the one from d.c.c. but I didn't want to add too much to this PR.
- Replaces build.js with rollup.config.js
- Removes webcomponents polyfills
- Uses rimraf for the npm clean task so things are a bit more Windows friendly
- Speed up integration tests a bit by removing the image copy/optimization step. I'd really like to refactor the gulpfile and these tests in a follow up PR. I had to remove a couple of the images from the test because copying images (in prod mode) is just super slow.
- Moved analytics to a footer include, similar to d.c.c.
  - This means we don't really track if someone is signed in but tbh we haven't used those stats for anything and we're discussing changes to sign-in anyway.
- Removes the sw and changes app.js to aggressively unregister service workers.
- Previously we dynamically loaded custom elements for special page types using entrypoint files in lib/pages. This changes things so if a page doesn't define an entrypoint we load default.js (which has been changed to only contain the bare minimum that a page needs). To declare an entrypoint, a page can use the `pageScripts` property in their YAML front matter.
- Adds a new entypoint file to lib/pages called `content.js`. This contains the components for posts and codelabs.

One thing I couldn't figure out is how to fix the broken test in unit/src/lib/index.js. It gives me:
```
TypeError: loadFirebase$1 is not a function.
```
Obviously it's something to do with my rollup setup but I can't figure out what. I tried a bunch of things with no luck. That's the only failing test.

**Note**: I noticed Netlify is doing a dev build (#4614) so if you want to see the inlining you should grab the PR, `npm ci`, and do `ELEVENTY_ENV=prod npm run build`. 
image


MichaelSolati (approved):

rollup.config.js
```diff
@@ -0,0 +1,127 @@
...
+  },
+  plugins: [
+    virtual(buildVirtualJSON(virtualImports)),
+    nodeResolve(),
+    commonjs(),
+    postcss(),
+  ],
+};
+
+const productionConfig = {
```
samthor, line 69: The configs are basically the same that I personally would build them programatically, e.g.:

```js
const config = {
  /* common stuff */
  plugins: [...];
};

// Build the config for our production or dev build.
if (isProduction) {
  config.plugins.push(terser());
} else {
  // etc
```

... but I'll confess it's a stylistic choice.

src/site/_includes/layout.njk
```diff
@@ -19,30 +32,26 @@ tags: []
...
-    <script async src="//www.google-analytics.com/analytics.js"></script>
-    <script defer src="{{ resourceJS.path | stripQueryParamsDev }}"></script>
+    <script type="module" src="{{ helpers.hashForProd('/js/app.js') }}"></script>
+    {# Add a facility for pages to declare an array of script paths. #}
+    {# If no scripts are declared we will use the default page script. #}
+    {# default.js loads the basic set of custom elements that all pages need. #}
+    {% if pageScripts %}
+      {% for item in pageScripts %}
+        <script type="module" src="{{ helpers.hashForProd(item) }}"></script>
+      {% endfor %}
```
samthor, line 49: I think nunjucks allows else from a for loop too (if nothing was output, do something else). You can streamline this.

PR #4617, opened 2021-02-08
[File System Access article] Write origin private file system without dash

tomayac:

- Write origin private file system without dash.


kaycebasques (commented):
Did they change it in the spec? `origin-private` with a dash seems grammatically correct to me. I'll defer to @jpmedley on this one.

https://en.wikipedia.org/wiki/Compound_modifier

tomayac:
The spec uses the term without a dash, I decided to put it (since it seems grammatically correct to me), but then realized I should go with what the spec does.

jpmedley (approved):

PR #4626, opened 2021-02-09
Clarify support matrix for `getInstalledRelatedApps`

petele:
- Clarifies support matrix for `getInstalledRelatedApps`
- Tells people to ask questions on StackOverflow

### DIFF
--- a/src/site/content/en/blog/get-installed-related-apps/index.md
+++ b/src/site/content/en/blog/get-installed-related-apps/index.md
@@ -55,8 +55,8 @@ relationship between the two, you can check if the app is installed.
   <table class="w-table--top-align">
     <thead>
       <tr>
-        <th>Can check for your</th>
-        <th>Can check from</th>
+        <th>App type</th>
+        <th>Checkable from</th>
       </tr>
     </thead>
     <tbody>
@@ -387,7 +387,7 @@ only available when served over **HTTPS**.
 
 ## Still have questions? {: #questions }
 
-Still have questions? Check [`getInstalledRelatedApps` on StackOverflow][so-gira]
+Still have questions? Check the [`getInstalledRelatedApps` tag on StackOverflow][so-gira]
 to see if anyone else has had similar questions. If not, ask your
 [question][so-ask-question] there, and be sure to tag it with the
 [`progressive-web-apps`][so-tagged-pwa] tag. Our team frequently monitors
--- a/src/site/content/en/blog/get-installed-related-apps/index.md
+++ b/src/site/content/en/blog/get-installed-related-apps/index.md
@@ -7,9 +7,10 @@ authors:
   - petelepage
 description: |
   The getInstalledRelatedApps() API is a web platform API that allows you
-  to check whether your iOS/Android/desktop app or PWA is installed on the user's device.
+  to check whether your iOS/Android/desktop app or PWA is installed on the
+  user's device.
 date: 2018-12-20
-updated: 2020-08-06
+updated: 2021-02-09
 tags:
   - blog
   - capabilities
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/get-installed-related-apps/index.md
```diff
@@ -44,15 +45,54 @@ For example, if your app is already installed:
...
-* [Windows (UWP) app](#check-windows) is installed
-* Progressive Web App is installed, if it's running in the
-  [same scope](#check-pwa-in-scope) or in a
-  [different scope](#check-pwa-out-of-scope)
+<div class="w-table-wrapper">
+  <table class="w-table--top-align">
+    <thead>
+      <tr>
+        <th>Can check for your</th>
+        <th>Can check from</th>
```
jpmedley, line 59: I'm not crazy about these heading. It looks like you're using a sentence style headings. Though that's a perfectly fine thing to do, these headings feel clunky to me. What about more functional headings like 'App type' and 'Checkable from'?

src/site/content/en/blog/get-installed-related-apps/index.md
```diff
@@ -44,15 +45,54 @@ For example, if your app is already installed:
...
+          <a href="#check-android">Android app</a>
+        </td>
+        <td>
+          Android only<br>
+          Chrome 80 or later
+        </td>
+      </tr>
+      <tr>
+        <td>
+          <a href="#check-windows">Windows (UWP) app</a>
```
jpmedley, line 74: UWP?

src/site/content/en/blog/get-installed-related-apps/index.md
```diff
@@ -66,7 +106,9 @@ Your website can check if your:
 
 Your website can check if your Android app is installed.
 
-* Android: Chrome 80 or later
+{% Compare 'better', 'Supported on' %}
+Android: Chrome 80 or later
+{% endCompare %}
```
jpmedley, line 111: @robdodson @kaycebasques Is this a correct use of the macro?

I would avoid an improper use a macro. If an automated update is ever deployed on the site's content, it could break the intent of this formatting and no one would catch it for weeks or months.

src/site/content/en/blog/get-installed-related-apps/index.md
```diff
@@ -337,6 +385,14 @@ taken into account.
 Like most other powerful web APIs, the `getInstalledRelatedApps()` API is
 only available when served over **HTTPS**.
 
+## Still have questions? {: #questions }
+
+Still have questions? Check [`getInstalledRelatedApps` on StackOverflow][so-gira]
```
jpmedley, line 390:
```suggestion
Still have questions? Check [`getInstalledRelatedApps()` on StackOverflow][so-gira]
```

jpmedley (changes requested):

src/site/content/en/blog/get-installed-related-apps/index.md
petele: Good suggestion, changed to app type/checkable from.

src/site/content/en/blog/get-installed-related-apps/index.md
petele: Technical term, Universal Windows Program - I think... Its how they refer to that type of app.

src/site/content/en/blog/get-installed-related-apps/index.md
petele: Made it more explict - "check the `getInstalledRelatedApps` tag", the tag doesn't include the brackets.

src/site/content/en/blog/get-installed-related-apps/index.md
petele: I'd argue it's a reasonable use for this. The compare widget is what developers should do, and this clearly calls out that they should test these types of apps.

robdodson: I think it's probably fine. Joe makes a good point that when we work on the CSS for the site we tend to think of the compare widget as a pair of do/don'ts so there might be _some_ risk that someday we'd make a change that could break this use case, but I don't think we have any immediate plans to radically alter the way this widget works.

So I guess my answer is something like "sure, but to joe's point note that there is a bit of incurred risk".

petele: Thanks - I've filed #4645 to track a new feature request for a 'Supported On' widget, and also included this bug in there so that if/when we get it, we know to come back and update this doc.

jpmedley (approved):

PR #4628, opened 2021-02-10
Escape HTML titles/descriptions when displaying them as search results

### LINKED ISSUE #4575: Weird site search results.
**To Reproduce**
1. Click the search box
2. Type "tr"

**Screenshots**
image

The articles that its linking to are:

- https://web.dev/video-description/
- https://web.dev/video-caption/

both of which contain brackets in their title so perhaps it's actually putting the html in the search result?

robdodson:
Fixes #4575

- Escape html entities (but allow `<strong>` tags) in search results. We allow strong tags because Algolia sends them back in the results to indicate which letters in the title match what the user was typing. The effect is really subtle and hard to see on the current site, but more noticeable on d.c.c.

### DIFF
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -9,6 +9,7 @@ import {store} from '../../store';
 import * as router from '../../utils/router';
 import {debounce} from '../../utils/debounce';
 import {trackError} from '../../analytics';
+import {allowHtml, escapeHtml} from '../../../lib/utils/escape-html';
 import 'focus-visible';
 import './_styles.scss';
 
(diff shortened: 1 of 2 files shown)


src/lib/components/Search/index.js
```diff
@@ -223,8 +224,19 @@ class Search extends BaseElement {
...
+        return html``;
+      }
+
+      let title = hit._highlightResult.title.value;
+      // Escape any html entities in the title except for <strong> tags.
+      // Algolia sends back <strong> tags in the title which help highlight
+      // the characters that match what the user has typed.
+      title = allowHtml(escapeHtml(title), 'strong');
+      // Strip backticks as they look a bit ugly in the results.
+      title = title.replace(/`/g, '');
```
MichaelSolati, line 238: Not saying it's necessary, but you could use the `strip` filter for this => https://github.com/GoogleChrome/web.dev/blob/76eff20f910842979fcc56fceac2dd3af896eece/src/site/_filters/strip.js

MichaelSolati (approved):

src/lib/components/Search/index.js
robdodson: oh cool i'll do that!

robdodson: ran into weird issues trying to get strip (which is CJS) to import correctly into this file so I'm just going to leave the little regex.

PR #4633, opened 2021-02-10
Add Identity collection

mihajlija:
Do not merge.


mihajlija:
@agektmr Please take a look at the restructured IA and the suggested aliases for external content. I organized sections by ascending complexity. 

I renamed "Learn authentication related APIs" to a bit more enthusiastic "Build authentication systems with modern APIs". I understand this section is more aimed at API explainers, however the content includes demos too so I think the new title fits better.

What do you think about dropping the "Build advanced authentication systems" section and moving that content into "Build authentication systems with modern APIs"? Since there's only one post under that section, I think it makes more sense to bundle it for now.

robdodson:
No worries @mihajlija. I'd like to try to think up ways that we can make the authoring experience have fewer of these gotchas so it's good to note.

Btw I went ahead and approved the Percy snapshot so y'all should be good to go :)

mihajlija:
@robdodson Thanks! For now, I'll just do my part to document the process of creating a new collection and write down the `layout=post` requirement :)

mihajlija:
@robdodson While you're at it, I think we'll also need your codeowner approval because I'm touching json files etc :)

robdodson (approved):

robdodson:
done!

src/site/_data/paths/identity.json
```diff
@@ -0,0 +1,53 @@
...
+        "title": "Strenghten security with two-factor authentication and two-step verification",
+        "pathItems": [
+            "sms-otp-form"   
+        ]
+      },
+      {
+        "title": "Build advanced authentication systems",
+        "pathItems": [
+          {
+            "title": "Set up a biometric re-authentication functionality with WebAuthn",
```
agektmr, line 33: What do you think appending this with `[Codelab]` to set the expectation right?

src/site/_data/paths/identity.json
```diff
@@ -0,0 +1,53 @@
...
+            "url": "https://developers.google.com/codelabs/webauthn-reauth/#0"
+          }
+        ]
+      },
+      {
+        "title": "Build authentication with identity federation",
+        "pathItems": [
+          {
+            "title": "Use OAuth 2.0 to access Google APIs",
```
agektmr, line 42: I requested this URL, but was wondering if it's okay to have "Google APIs" in the subject line. The content itself is a good overview of how OAuth 2.0 works in general, using Google (you can consider it as an example identity provider).  
Ideally we should have generic guidance on OAuth 2.0 and OpenID Connect without an affiliation to any particular services, but in reality there's dialects and helper libraries when adopting to identity providers. I think realistically this is an OK approach.

kaycebasques (commented):
- [ ] Images should be uploaded to the image CDN via the image uploader. @MichaelSolati has a script that might help you automate the process. https://web.dev/handbook/markup-media/

mihajlija:
@kaycebasques These articles were moved from blog, so the images have already been migrated over to GCP. I think the easiest solution is for me to rebase and move the updated content over from /blog again.

mihajlija:
Pushing this to Q3, closing the stale PR.

agektmr:
Closing this since IIUC this was merged with a separate pull request. Resurrect if I was wrong.

PR #4649, opened 2021-02-10
test if approvers get pinged

robdodson:
Just a test to see if the web.dev-approvers list gets emailed or not.


rowan-m:
Ping confirmed here.

petele:
also pinged.

PR #4662, opened 2021-02-12
[Img component] Prevent markup bloat

tomayac:
For images that are capped at 300px, don't create sizes up to 1,600px.

Screen Shot 2021-02-12 at 17 47 53

If the image is smaller than 200px, use the image width as the lower limit as to not scale up small images.


MichaelSolati (approved):
Good thinking, thanks.

PR #4702, opened 2021-02-19
chore(authors): migrate author images to gcp

MichaelSolati:
It looks big, but its not. Here's a breakdown

- I deleted all the author images after uploading them to GCP
- I updated the `_data/authorsData.json` with the image path in the bucket.
- I removed the check from `_collections/authors.js` to see if an image exists, as it just should in the JSON and assign `author.data.hero = authorsImage;`
- `_includes/components/Author.js` uses the `Img` shortcode for it's `` now.
- `_includes/components/AuthorCard.js` uses the `Img` shortcode for it's `` in the `renderThumbnail` method now.
- `_includes/components/AuthorsDate.js` uses the `Img` shortcode for it's `` now.
- Because the `Img` shortcode is now being called outside of templates I just check if there is a page to reference for the error message in `_includes/components/Img.js`.
- `content/en/authors/individual.njk` uses the `Img` shotcode for it's `` now.
- I updated `types/site/_data/authorsData.d.ts` to reflect the new `image` field.


samthor (commented):
Quick phone comments:

- Why are all our images in admin/?
- it seems like a missing author image will end up calling `Img` with an undefined image. Can you update the data to set a blank placeholder? The image.js file should do this, like it used to

MichaelSolati:
@samthor so first, about the `admin` all images i upload with my uploader is uploaded by admin and not a user. as far as the missing author img, i dont think we do anything like that for dcc, but ill add a fallback.

src/site/_collections/authors.js
```diff
@@ -130,17 +128,10 @@ module.exports = (collections) => {
...
-    if (process.env.ELEVENTY_ENV === 'prod') {
-      const authorsImageExists = fs.existsSync(path.join('src', authorsImage));
-      if (!authorsImageExists) {
-        console.warn(
-          `No 2x image was found for ${author.title} (${author.key}), replacing with placeholder.`,
-        );
-        authorsImage = path.join('/images', 'authors', 'no-photo.svg');
-      }
+    if (!author.image) {
+      author.image = 'image/admin/1v5F1SOBl46ZghbHQMle.svg';
```
devnook, line 132: Can we put it in a const, like PLACEHOLDER_IMG or so?

MichaelSolati: sure, ill do that now.

devnook (approved):

PR #4703, opened 2021-02-19
chore(hero): upload all hero images to gcp

MichaelSolati:
- I deleted all the hero and thumbnail images after uploading them to GCP.
- I updated the thumbnail and hero values in the frontmatter to the new GCP paths.
- Because the `Img` shortcode is now being called outside of templates I just check if there is a page to reference for the error message in `_includes/components/Img.js`.
- `_includes/components/BaserCard.js` uses the `Img` shortcode for it's `` in the `renderThumbnail` method now.
- `_includes/components/Hero.js` uses the `Img` shortcode for it's `` now.


src/site/_includes/components/Hero.js
```diff
@@ -13,26 +13,19 @@
...
-        ${hero}?auto=format&fit=max&w=${width} ${width}w,
-      `)}"
-      src="${hero}"
-      alt="${alt}"
-      loading="lazy"
-    />
-  `;
+  return Img({
+    src: hero,
+    alt: typeof alt === 'string' ? alt : 'Hero Image',
```
devnook, line 25: Even if empty string?

devnook (approved):

src/site/_includes/components/Hero.js
MichaelSolati: This only exists because some post(s) send an object that breaks the type check (i cant pin down the file cause i dont have access to the page object). Separately the short code requires an alt tag, so this was a decent one.

robdodson: @MichaelSolati if you change the shortcode to use `function() {}` instead of arrow functions then you can inspect `this.page` inside of it to see which page is breaking.

PR #4709, opened 2021-02-20
Improve CLS on Authors page regression

robdodson:

- Adds dimensions to images
- Improves the `sizes` for the author image.

A note about the `Img` shortcode. In this example the image is 128px on small screens and then 192 px on large screens. The `Img` shortcode has this line `minWidth: Math.min(MIN_WIDTH, widthAsNumber),` where MIN_WIDTH is 200. This means that the minWidth would be 192px even though the true minimum size this image is displayed at is 128px. I passed in an explicit minWidth option to fix this, but I'm curious if that line of code in the shortcode is correct. I'm not sure what a better alternative would be off the top of my head.

### DIFF
--- a/src/site/content/en/authors/individual.njk
+++ b/src/site/content/en/authors/individual.njk
@@ -29,7 +29,8 @@ pagination:
         height="192",
         width="192",
         class="w-author-page__image",
-        sizes="(min-width: 481px) 192px, 128px"
+        sizes="(min-width: 481px) 192px, 128px",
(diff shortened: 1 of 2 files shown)


robdodson:
@jpmedley @mihajlija you can ignore this PR. I think pullapprove is being overly aggressive with the rules and I need to tweak it.

robdodson:
@MichaelSolati PTAL

PR #4713, opened 2021-02-21
[Streams article] Various fixes

MattiasBuelens:
First of all: great article! This will surely help make ~2016~ **2021** the year of web streams! 😄

I did spot a couple of small inaccuracies, which this PR aims to fix:
* A "controller" is associated with a readable/writable stream, not with a reader/writer.
* The queuing strategy for a readable byte stream cannot have a `size(chunk)` function. ([Step 4.1 of the constructor](https://streams.spec.whatwg.org/commit-snapshots/4b5b3cdb0f22a791fc1a8e734296a32a94ad753a/#rs-constructor) forbids this. 🤓) The `highWaterMark` is *always* interpreted as a number of bytes, and the size of each chunk is *always* determined by its `byteLength`. (Interestingly, this also means that you cannot pass a `ByteLengthQueuingStrategy`, since that *does* have a `size(chunk)` method.)
* In one of the examples, the `value` from a `reader.read()` call is printed *before* `done` is checked. This would print `undefined` when the stream closes, which is probably not what you'd want. I moved that line *after* the `done` check.

### DIFF
--- a/src/site/content/en/blog/streams/index.md
+++ b/src/site/content/en/blog/streams/index.md
@@ -460,20 +460,18 @@ property returns the desired size to fill the controlled stream's internal queue
 ### The `queuingStrategy`
 
 The second, likewise optional, argument of the `ReadableStream()` constructor is `queuingStrategy`.
-It is an object that optionally defines a queuing strategy for the stream, which takes two
-parameters:
+It is an object that optionally defines a queuing strategy for the stream, which takes one
+parameter:
 
-- `highWaterMark`: A non-negative number indicating the high water mark of the stream using this queuing strategy.
-- `size(chunk)`: A function that computes and returns the finite non-negative size of the given chunk value.
-  The result is used to determine backpressure, manifesting via the appropriate `ReadableByteStreamController.desiredSize` property.
+- `highWaterMark`: A non-negative number of bytes indicating the high water mark of the stream using this queuing strategy.
+  This is used to determine backpressure, manifesting via the appropriate `ReadableByteStreamController.desiredSize` property.
   It also governs when the underlying source's `pull()` method is called.
 
+Note that a queuing strategy for a readable byte stream does not have a `size(chunk)` function.
+The size of each chunk is always determined by its `byteLength` property.
+
 {% Aside %}
-  You could define your own custom `queuingStrategy`, or use an instance of
-  [`ByteLengthQueuingStrategy`](https://developer.mozilla.org/en-US/docs/Web/API/ByteLengthQueuingStrategy)
-  or [`CountQueuingStrategy`](https://developer.mozilla.org/en-US/docs/Web/API/CountQueuingStrategy)
-  for this object value. If no `queuingStrategy` is supplied, the default used is the same as a
-  `CountQueuingStrategy` with a `highWaterMark` of `1`.
+  If no `queuingStrategy` is supplied, the default used is one with a `highWaterMark` of `0`.
 {% endAside %}
 
 #### The `getReader()` and `read()` methods
--- a/src/site/content/en/blog/streams/index.md
+++ b/src/site/content/en/blog/streams/index.md
@@ -7,7 +7,7 @@ description: |
 authors:
   - thomassteiner
 date: 2021-02-19
-updated: 2021-02-22
+updated: 2021-02-25
 hero: image/8WbTDNrhLsU0El80frMBGE4eMCD3/TuciUuOQOd3u7uMgDZBi.jpg
 alt: A forest stream with colored fallen leaves.
 tags:
--- a/src/site/content/en/blog/streams/index.md
+++ b/src/site/content/en/blog/streams/index.md
@@ -469,8 +469,9 @@ parameter:
   It also governs when the underlying source's `pull()` method is called.
 
 {% Aside %}
-  A queuing strategy for a readable byte stream does not have a `size(chunk)` function.
-  The size of each chunk is always determined by its `byteLength` property.
+  Unlike queuing strategies for other stream types, a queuing strategy for a readable byte stream
+  does not have a `size(chunk)` function. The size of each chunk is always determined by its
+  `byteLength` property.
 {% endAside %}  
 
 {% Aside %}
--- a/src/site/content/en/blog/streams/index.md
+++ b/src/site/content/en/blog/streams/index.md
@@ -1079,8 +1079,9 @@ This article was reviewed by
 [François Beaufort](https://github.com/beaufortfrancois),
 [Sam Dutton](https://samdutton.com/),
 [Mattias Buelens](https://github.com/MattiasBuelens),
-[Surma](https://surma.dev/), and
-[Joe Medley](https://github.com/jpmedley).
+[Surma](https://surma.dev/),
+[Joe Medley](https://github.com/jpmedley), and
+[Adam Rice](https://github.com/ricea).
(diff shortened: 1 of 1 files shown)


MattiasBuelens:
There's also a problem with the example that enqueues a timestamp every second (the first one in [this section](https://web.dev/streams/#readable-stream-code-samples)). The `cancel()` function tries to use the `interval` variable, but that variable lives inside `startEnqueuing` and is not accessible here.

One possible solution that preserves the "privateness" of `interval` would be to use a private class field:
```javascript
class TimestampSource {
  #interval

  start(controller) {
    this.#interval = setInterval(() => {
      const string = new Date().toLocaleTimeString();
      // Add the string to the stream.
      controller.enqueue(string);
      console.log(`Enqueued ${string}`);
    }, 1_000);

    setTimeout(() => {
      clearInterval(this.#interval);
      // Close the stream after 10s.
      controller.close();
    }, 10_000);
  }

  cancel() {
    // This is called if the reader cancels.
    clearInterval(this.#interval);
  }
}

const stream = new ReadableStream(new TimestampSource());
```
Alternatively, you could have `startEnqueuing` return a `{ start, cancel }` underlying source object (instead of just being a `start()` function) that "closes over" an `interval` variable.

I wasn't sure how you'd want to see this being solved, so I haven't changed it in the PR yet.

src/site/content/en/blog/streams/index.md
```diff
@@ -460,20 +460,18 @@ property returns the desired size to fill the controlled stream's internal queue
...
 
-- `highWaterMark`: A non-negative number indicating the high water mark of the stream using this queuing strategy.
-- `size(chunk)`: A function that computes and returns the finite non-negative size of the given chunk value.
-  The result is used to determine backpressure, manifesting via the appropriate `ReadableByteStreamController.desiredSize` property.
+- `highWaterMark`: A non-negative number of bytes indicating the high water mark of the stream using this queuing strategy.
+  This is used to determine backpressure, manifesting via the appropriate `ReadableByteStreamController.desiredSize` property.
   It also governs when the underlying source's `pull()` method is called.
 
+Note that a queuing strategy for a readable byte stream does not have a `size(chunk)` function.
+The size of each chunk is always determined by its `byteLength` property.
```
tomayac, line 471:
```suggestion
{% Aside %}
  A queuing strategy for a readable byte stream does not have a `size(chunk)` function.
  The size of each chunk is always determined by its `byteLength` property.
{% endAside %}  
```

We have a special "note" aside widget.

tomayac (approved):
LGTM, approval, with one "use the widget" nit. 

Thanks so much for taking the time to review this!

tomayac:
I went with your suggestion to use a private class field in https://github.com/GoogleChrome/web.dev/pull/4713/commits/c9f06be306e2037ff76ecb4adacf875ed082c977. I quite like how concise it is.

src/site/content/en/blog/streams/index.md
```diff
@@ -273,38 +273,38 @@ console.log(`The stream is ${locked ? 'indeed' : 'not'} locked.`);
 ### Readable stream code samples
 
 The code sample below shows all the steps in action. You first create a `ReadableStream` that in its
-`underlyingSource` argument defines a `start` method that tells the stream's `controller` to
+`underlyingSource` argument (that is, the `TimestampSource` class) defines a `start` method
```
jpmedley, line 276:
```suggestion
`underlyingSource` argument (that is, the `TimestampSource` class) defines a `start()` method
```

jpmedley: I think this tries to convey too much in a single sentence, which I'm having trouble following. There doesn't have to be a 1 to 1 relationship between lines of text and lines of code, but it helps if single gestures in the code get at least their own suggestions.

src/site/content/en/blog/streams/index.md
```diff
@@ -273,38 +273,38 @@ console.log(`The stream is ${locked ? 'indeed' : 'not'} locked.`);
...
-  }, 1_000);
-
-  setTimeout(() => {
-    clearInterval(interval);
-    // Close the stream after 10s.
-    controller.close();
-  }, 10_000);
-};
+class TimestampSource {
+  #interval
```
jpmedley, line 284: Please stick to plain JavaScript.

src/site/content/en/blog/streams/index.md
```diff
@@ -460,20 +460,20 @@ property returns the desired size to fill the controlled stream's internal queue
...
+  This is used to determine backpressure, manifesting via the appropriate `ReadableByteStreamController.desiredSize` property.
   It also governs when the underlying source's `pull()` method is called.
 
 {% Aside %}
-  You could define your own custom `queuingStrategy`, or use an instance of
-  [`ByteLengthQueuingStrategy`](https://developer.mozilla.org/en-US/docs/Web/API/ByteLengthQueuingStrategy)
-  or [`CountQueuingStrategy`](https://developer.mozilla.org/en-US/docs/Web/API/CountQueuingStrategy)
-  for this object value. If no `queuingStrategy` is supplied, the default used is the same as a
-  `CountQueuingStrategy` with a `highWaterMark` of `1`.
+  A queuing strategy for a readable byte stream does not have a `size(chunk)` function.
```
jpmedley, line 471: I want to make sure there isn't something else to explain. Why would I expect a `size(chunk)` function?

jpmedley (changes requested):

src/site/content/en/blog/streams/index.md
tomayac: This is the new [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) notation.

src/site/content/en/blog/streams/index.md
tomayac: Because all the other stream variants have one to set a custom queuing strategy.

tomayac:
@jpmedley, I have split up the long sentence into three shorter ones. PTAL.

src/site/content/en/blog/streams/index.md
jpmedley: maybe add , "like other stream types do"? 

This is the conceptual version of a broken link. If I'm not already familiar with streams, I can't work out for myself why this information is here or why I care.

jpmedley:
The split sentence looks great. I still have a concern about `size(chunk)`.

src/site/content/en/blog/streams/index.md
MattiasBuelens: How about:
> Unlike queuing strategies for other stream types, a queuing strategy for a readable byte stream does not have a `size(chunk)` function.

src/site/content/en/blog/streams/index.md
```diff
@@ -7,7 +7,7 @@ description: |
 authors:
   - thomassteiner
 date: 2021-02-19
-# updated: 2021-02-19
+updated: 2021-02-22
```
tomayac, line 10:
```suggestion
updated: 2021-02-25
```

src/site/content/en/blog/streams/index.md
tomayac: I went with this variant in https://github.com/GoogleChrome/web.dev/pull/4713/commits/cdab862037056c1084b3a841c7cb55a421ffe1fa.

tomayac:
@jpmedley This should be good to go now.

jpmedley (approved):

PR #4718, opened 2021-02-22
Article on JavaScript `this`

jakearchibald:
@mathiasbynens I'd love a review of this if you've got time.

TODOs:

- [x] Add correct date
- [x] Find/create a good banner image
- [x] _sighhhhhhhh_ hard wrap everything to 100 chars

### DIFF
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -302,7 +302,7 @@ const $ = document.querySelector;
 const el = $('.some-element');
 ```
 
-This breaks because the implementation of `querySelector` looks it its own `this` value and expects it to be a DOM node of sorts, and the above breaks that connection. To achieve the above correctly:
+This breaks because the implementation of `querySelector` looks at its own `this` value and expects it to be a DOM node of sorts, and the above breaks that connection. To achieve the above correctly:
 
 ```js
 const $ = document.querySelector.bind(document);
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -4,10 +4,10 @@ subhead: "Figuring out the value of `this` can be tricky in JavaScript, here's h
 description: "Figuring out the value of `this` can be tricky in JavaScript, here's how to do it…"
 authors:
   - jakearchibald
-date: 2021-02-22
+date: 2021-02-23
 # updated: 2021-02-19
-hero: image/8WbTDNrhLsU0El80frMBGE4eMCD3/TuciUuOQOd3u7uMgDZBi.jpg
-alt: A forest stream with colored fallen leaves.
+hero: image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/cePCOGeXNFT6WCy85gb4.png
+alt: this 🤔
 tags:
   - blog # blog is a required tag for the article to show up in the blog.
   - javascript
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -132,6 +132,20 @@ class Whatever {
 }
 ```
 
+Alternative pattens involve binding an existing function in the constructor, or assigning the
+function in the constructor. If you can't use class fields for some reason, assigning functions in
+the constructor is a reasonable alternative:
+
+```js
+class Whatever {
+  constructor() {
+    this.someMethod = () => {
+      // …
+    };
+  }
+}
+```
+
 ## Otherwise, if the function/class is called with `new`: {: #new }
 
 ```js
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -35,7 +35,7 @@ const arrowFunction = () => {
 };
 ```
 
-In this case, the value of `this` will _always_ be the same as `this` in the parent scope:
+In this case, the value of `this` is _always_ the same as `this` in the parent scope:
 
 ```js
 const outerThis = this;
@@ -260,9 +260,9 @@ I avoid using `this` in cases like above, and instead:
 
 ```js
 element.addEventListener('click', (event) => {
-  // Grab it from the parent scope:
+  // Ideally, grab it from a parent scope:
   console.log(element);
-  // Or from the event object:
+  // But if you can't do that, get it from the event object:
   console.log(event.currentTarget);
 });
 ```
@@ -312,6 +312,8 @@ const $ = document.querySelector.bind(document);
 const $ = (...args) => document.querySelector(...args);
 ```
 
+Fun fact: Not all APIs use `this` internally. Console methods like `console.log` were changed to avoid `this` references, so `log` doesn't need to be bound to `console`.
+
 {% Aside 'warning' %}
 Don't transplant a function onto an object just to set `this` to some value unrelated to the parent object; it's usually unexpected and it's why `this` gets such a bad reputation. Consider passing the value as an argument instead; it's more explicit, and works with arrow functions.
 {% endAside %}
@@ -328,7 +330,7 @@ function someFunction() {
 console.log(someFunction() === undefined);
 ```
 
-In this case, the value of `this` is undefined. `'use strict'` isn't needed in the function if the parent scope is in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).
+In this case, the value of `this` is undefined. `'use strict'` isn't needed in the function if the parent scope is in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) (and all modules are in strict mode).
 
 {% Aside 'warning' %}
 Don't rely on this. I mean, there are easier ways to get an `undefined` value 😀.
@@ -345,12 +347,18 @@ function someFunction() {
 console.log(someFunction() === globalThis);
 ```
 
-In this case, the value of `this` is the global object.
+In this case, the value of `this` is the same as `globalThis`.
+
+{% Aside %}
+Most folks (including me) call `globalThis` the global object, but this isn't 100% technically correct. Here's [Mathias Bynens with the details](https://mathiasbynens.be/notes/globalthis#terminology), including why it's called `globalThis` rather than simply `global`.
+{% endAside %}
 
 {% Aside 'warning' %}
-Avoid using `this` to reference the global object. Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
+Avoid using `this` to reference the global (yes, I'm still calling it that). Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
 {% endAside %}
 
 ## Phew!
 
 And that's it! That's everything I know about `this`. Any questions? Something I've missed? Feel free to [tweet at me](https://twitter.com/jaffathecake).
+
+Thanks to [Mathias Bynens](https://twitter.com/mathias) and [Ingvar Stepanyan](https://twitter.com/RReverser) for reviewing.
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -22,7 +22,7 @@ I'm going to start with the most specific situation, and end with the least-spec
 1. [If the function is defined as an arrow function](#arrow-functions)
 1. [Otherwise, if the function/class is called with `new`](#new)
 1. [Otherwise, if the function has a 'bound' `this` value](#bound)
-1. [Otherwise, if `this` is set all call-time](#call-apply)
+1. [Otherwise, if `this` is set at call-time](#call-apply)
 1. [Otherwise, if the function is called via a parent object (`parent.func()`)](#object-member)
 1. [Otherwise, if the function or parent scope is in strict mode](#strict)
 1. [Otherwise](#otherwise)
--- a/src/site/content/en/blog/javascript-this/index.md
+++ b/src/site/content/en/blog/javascript-this/index.md
@@ -1,0 +1,356 @@
+---
+title: 'JavaScript: What is the meaning of this?'
+subhead: "Figuring out the value of `this` can be tricky in JavaScript, here's how to do it…"
+description: "Figuring out the value of `this` can be tricky in JavaScript, here's how to do it…"
+authors:
+  - jakearchibald
+date: 2021-02-22
+# updated: 2021-02-19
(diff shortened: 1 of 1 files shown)


jakearchibald:
I'll add the hard-wraps just before commiting. They're a pain to maintain and don't make things easier to read.

jakearchibald:
https://deploy-preview-4718--web-dev-staging.netlify.app/javascript-this/

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+
+JavaScript's `this` is the butt of many jokes, and that's because, well, it's pretty complicated. However, I've seen developers do much-more-complicated and domain-specific things to avoid dealing with this `this`. If you're unsure about `this`, hopefully this will help. This is my `this` guide.
+
+I'm going to start with the most specific situation, and end with the least-specific. This article is kinda like a big `if (…) … else if () … else if (…) …`, so you can go straight to the first section that matches the code you're looking at.
+
+1. [If the function is defined as an arrow function](#arrow-functions)
+1. [Otherwise, if the function/class is called with `new`](#new)
+1. [Otherwise, if the function has a 'bound' `this` value](#bound)
+1. [Otherwise, if `this` is set all call-time](#call-apply)
+1. [Otherwise, if the function is called via a parent object (`parent.func()`)](#object-member)
```
RReverser, line 26: Nit: if possible, I'd reorder sections by popularity. For example, I'm sure that `parent.func()` is far more common than e.g. `bound this value` so I'd push it up.

jakearchibald: @RReverser I don't think that's possible. Right now it's ordered by specificity, and moving `parent.func()` would break that, requiring more rules and exceptions to make sense of it all.

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+// Logs `true`:
+console.log(anotherObj.someMethod() === anotherObj);
+```
+
+`someMethod() === obj` is false because `someMethod` _isn't_ called as a member of `obj`. You might have encountered this gotcha when trying something like this:
+
+```js
+const $ = document.querySelector;
+// TypeError: Illegal invocation
+const el = $('.some-element');
```
mathiasbynens, line 304: Great example! Other examples that came to mind, that have since been fixed but that folks might still remember struggling with:

- Good old `console.log` (and other `console` methods)
- `requestAnimationFrame` (had to be invoked as `window.requestAnimationFrame` when it first shipped)

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+function someFunction() {
+  'use strict';
+  return this;
+}
+
+// Logs `true`:
+console.log(someFunction() === undefined);
+```
+
+In this case, the value of `this` is undefined. `'use strict'` isn't needed in the function if the parent scope is in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).
```
mathiasbynens, line 331: Worth spelling out that modules are implicitly in strict mode?

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+```js
+function someFunction() {
+  return this;
+}
+
+// Logs `true`:
+console.log(someFunction() === globalThis);
+```
+
+In this case, the value of `this` is the global object.
```
mathiasbynens, line 348: This gets into [a somewhat pedantic terminology nitpick](https://mathiasbynens.be/notes/globalthis#terminology) which the article probably shouldn't get into, but perhaps we can use a more accurate term than "the global object"? Perhaps:

```
In this case, the value of `this` matches the value of `this` in the global scope.
```

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+
+{% Aside 'warning' %}
+Avoid using `this` to reference the global object. Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
```
mathiasbynens, line 351: Same here

mathiasbynens, line 351: IMHO `globalThis` deserves a more prominent mention — there’s an opportunity to explain its name and how it relates to `this`, which (based on the flamewars that happened when the proposal renamed `global` to `globalThis`) lots of developers are confused about.

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+element.addEventListener('click', (event) => {
+  // Grab it from the parent scope:
+  console.log(element);
+  // Or from the event object:
+  console.log(event.currentTarget);
+});
+```
+
+{% endCompare %}
+
```
mathiasbynens, line 271: How about adding a quick sentence listing some more common examples of built-in Web Platform APIs that force a specific `this` value (e.g. `setTimeout` & `setInterval`, whose forced `this` is IMHO even more surprising than event handlers)?

mathiasbynens (commented):
Thanks for including me! This looks awesome so far. I left some suggestions already. At a high level, the most important things that IMHO are currently missing or not featured prominently enough are:

- modules being strict, and thus having `this === undefined` in the global scope
- `globalThis` (already mentioned, but deserves some elaboration)

WDYT?

src/site/content/en/blog/javascript-this/index.md
jakearchibald: Yes! I was intending on mentioning `console.log` as an exception here, but forgot. I didn't know the `requestAnimationFrame` thing, hah!

src/site/content/en/blog/javascript-this/index.md
jakearchibald: Yeahhhh

src/site/content/en/blog/javascript-this/index.md
jakearchibald: I'm trying to discourage folks from using `this` to get the global. Maybe I should just mention `globalThis` here.

src/site/content/en/blog/javascript-this/index.md
mathiasbynens: It made polyfilling more surprising/painful than expected: https://www.paulirish.com/2011/requestanimationframe-for-smart-animating/#comment-918142887

src/site/content/en/blog/javascript-this/index.md
mathiasbynens: Agreed on discouraging it 👍  But specifically, I was suggesting to avoid saying "the global object" when referring to something that's observably different

jakearchibald: Actually, I might just acknowledge that it isn't 100% accurate and link to your article

mathiasbynens: The smallest possible delta to make it 100% accurate is to replace "the global object" with "the global `this`", I think, but +1 to mentioning that this is commonly known as "the global object" which isn't actually correct + linking to more info seems useful 👍

src/site/content/en/blog/javascript-this/index.md
jakearchibald: I've never really been surprised by `this` in `setTimeout`. It's never a useful value so it doesn't really get used. Do you have an example in mind?

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,356 @@
...
+## If the function is defined as an arrow function: {: #arrow-functions }
+
+```js
+const arrowFunction = () => {
+  console.log(this);
+};
+```
+
+In this case, the value of `this` will _always_ be the same as `this` in the parent scope:
```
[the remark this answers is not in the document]
jakearchibald: Good advice!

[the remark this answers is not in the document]
jakearchibald: I think I was trying to avoid the sound of "this is", but it doesn't really matter.

src/site/content/en/blog/javascript-this/index.md
mathiasbynens: > It's never a useful value so it doesn't really get used.

This is exactly why it's surprising, IMHO. At least for event listeners, I can see the point of giving easy access to the affected node. I have no idea why you'd need a shorthand for `globalThis` within `setTimeout`/`setInterval`, which makes the API design more baffling.

src/site/content/en/blog/javascript-this/index.md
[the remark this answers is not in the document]
mathiasbynens: I think it reads more nicely, not just here, but for all occurrences of "will" throughout the article.

jakearchibald:
@kaycebasques @jpmedley this is ready for final review and publish.

mathiasbynens (approved):

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+In this case, the value of `this` is the same as `globalThis`.
+
+{% Aside %}
+Most folks (including me) call `globalThis` the global object, but this isn't 100% technically correct. Here's [Mathias Bynens with the details](https://mathiasbynens.be/notes/globalthis#terminology), including why it's called `globalThis` rather than simply `global`.
+{% endAside %}
+
+{% Aside 'warning' %}
+Avoid using `this` to reference the global (yes, I'm still calling it that). Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
```
tomayac, line 355:
```suggestion
Avoid using `this` to reference the global object (yes, I'm still calling it that). Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
```

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+
+`someMethod() === obj` is false because `someMethod` _isn't_ called as a member of `obj`. You might have encountered this gotcha when trying something like this:
+
+```js
+const $ = document.querySelector;
+// TypeError: Illegal invocation
+const el = $('.some-element');
+```
+
+This breaks because the implementation of `querySelector` looks it its own `this` value and expects it to be a DOM node of sorts, and the above breaks that connection. To achieve the above correctly:
```
tomayac, line 305:
```suggestion
This breaks because the implementation of `querySelector` looks at its own `this` value and expects it to be a DOM node of sorts, and the above breaks that connection. To achieve the above correctly:
```

I think this is what you wanted to type.

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+new Whatever();
+```
+
+The above will call `Whatever` (or its constructor function if it's a class) with `this` set to the result of `Object.create(Whatever.prototype)`.
+
+```js
+class MyClass {
+  constructor() {
+    console.log(
+      this.constructor === Object.create(MyClass.prototype).constructor,
```
tomayac, line 135:
```suggestion
      this.constructor === Object.create(MyClass.prototype).constructor
```

It's probably prettier that did this, but it really looks weird here I think.

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+// Logs `true`:
+new MyClass();
+```
+
+The same is true for older-style constructors:
+
+```js
+function MyClass() {
+  console.log(
+    this.constructor === Object.create(MyClass.prototype).constructor,
```
tomayac, line 149:
```suggestion
    this.constructor === Object.create(MyClass.prototype).constructor
```

It's probably prettier that did this, but it really looks weird here I think.

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+With arrow functions, the value of `this` _can't_ be changed by calling the function as a constructor:
+
+```js
+// TypeError: arrowFunction is not a constructor
+new arrowFunction();
+```
+
+### 'Bound' instance methods
+
+With instance methods, if you want to ensure `this` always refers to the class instance, the best way is to use arrow functions and [class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields):
```
tomayac, line 84: Curious. You write "the best way", which suggests not the only way. What other alternatives exist? That being said, I love this pattern!

tomayac (changes requested):
Just some minor nits…

src/site/content/en/blog/javascript-this/index.md
jakearchibald: Yeah, I've been sticking to the prettier rules, and I disagree with the lot of them (especially the way we diverge from the defaults), but it felt right to stick to the rules.

src/site/content/en/blog/javascript-this/index.md
jakearchibald: I see this pattern a lot:

```js
class Whatever {
  someMethod() {
    // …
  }
  constructor() {
    this.someMethod = this.someMethod.bind(this);
  }
}
```

And:

```js
class Whatever {
  constructor() {
    this.someMethod = () => {
      // …
    };
  }
}
```

src/site/content/en/blog/javascript-this/index.md
```diff
@@ -0,0 +1,362 @@
...
+
+{% Aside 'warning' %}
+Avoid using `this` to reference the global object (yes, I'm still calling it that). Instead, use [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis), which is much more explicit.
+{% endAside %}
+
+## Phew!
+
+And that's it! That's everything I know about `this`. Any questions? Something I've missed? Feel free to [tweet at me](https://twitter.com/jaffathecake).
+
+Thanks to [Mathias Bynens](https://twitter.com/mathias) and [Ingvar Stepanyan](https://twitter.com/RReverser) for reviewing.
```
jakearchibald, line 362:
```suggestion
Thanks to [Mathias Bynens](https://twitter.com/mathias), [Ingvar Stepanyan](https://twitter.com/RReverser), and [Thomas Steiner](https://twitter.com/tomayac) for reviewing.
```

jakearchibald:
@tomayac can you do the CLA thing?

tomayac:
@googlebot I fixed it.

jakearchibald:
@google-cla I hate you so much

tomayac (approved):
LGTM

Maybe add a small note after “the best way” that basically includes your answer in https://github.com/GoogleChrome/web.dev/pull/4718#discussion_r581034659

surma (approved):
LGTM

PR #4741, opened 2021-02-25
Add course flag

robdodson:
Adds a quick and dirty eleventyComputed permalink to the course directory which prevents the course from going live during prod builds.

### DIFF
--- a/src/site/content/en/learn/css/css.11tydata.js
+++ b/src/site/content/en/learn/css/css.11tydata.js
@@ -1,0 +1,18 @@
+module.exports = function () {
+  return {
+    eleventyComputed: {
+      permalink: (data) => {
+        if (process.env.ELEVENTY_ENV === 'prod') {
+          return false;
+        }
+
+        // I'm not sure why the output-permalinks.js eleventyComputed function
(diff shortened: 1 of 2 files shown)


src/site/content/en/learn/css/css.11tydata.js
```diff
@@ -0,0 +1,18 @@
+module.exports = function () {
+  return {
+    eleventyComputed: {
+      permalink: (data) => {
+        if (process.env.ELEVENTY_ENV === 'prod') {
```
samthor, line 5: I feel like we want an `IGNORE_...` set of flags like as we do for d.c.c.. But this is fine if you want.

samthor (approved):

src/site/content/en/learn/css/css.11tydata.js
robdodson: Oh I think this'll be for a different purpose than what we do on d.c.c.

Kayce asked if I could add Andy to repo so he could work on the course. I wanted to make it so he could run the site locally and test the course, but nothing would go live until we remove this file.

robdodson: thanks for the review!

PR #3445, opened 2020-07-07
Update variable fonts article

RoelN:
This is an update to [Introduction to variable fonts on the web](https://web.dev/variable-fonts/) by @davelab6 and @roeln. Dave and I paired on this, and I added some animations afterwards, and now submitting this PR.

This PR needs one more revision by @davelab6 to address the last TODOs.

Changes proposed in this pull request:

- General updates and rewrites
- Updated `@font-face` code examples
- Updated examples to modern GF fonts
- Updated fallbacks and browser support
- Add CSS custom properties workaround for `font-variation-settings`
- Images and movies have been optimised

## TODO:
- [ ] TODOs (see "TODO" and "RN" comments in document)
  - [ ] Add Dave Crossland and Roel Nieskens to authors
  - [ ] @davelab6 TODO: [Print industry reference](https://github.com/GoogleChrome/web.dev/blame/7ac29dbea9e4e7bd47faac2f775cbf272a87595f/src/site/content/en/blog/variable-fonts/index.md#L53)
  - [ ] @davelab6 TODO: [Simplify language of intruction](https://github.com/GoogleChrome/web.dev/blame/7ac29dbea9e4e7bd47faac2f775cbf272a87595f/src/site/content/en/blog/variable-fonts/index.md#L95)
  - [ ] @davelab6 TODO: ["Restart here"](https://github.com/GoogleChrome/web.dev/blame/7ac29dbea9e4e7bd47faac2f775cbf272a87595f/src/site/content/en/blog/variable-fonts/index.md#L139)
  - [ ] @davelab6 TODO: [explain opsz, and px/pt snakepit](https://github.com/GoogleChrome/web.dev/blame/7ac29dbea9e4e7bd47faac2f775cbf272a87595f/src/site/content/en/blog/variable-fonts/index.md#L139)
- [x] Add animation for Amstelvar "Finesse" example
  - [ ] Check text explaining Amstelvar "Finesse" example, not sure my placeholder covers everything
- [ ] Move image/movie media to https://storage.googleapis.com/web-dev-assets/variable-fonts

PR #4746, opened 2021-02-25
Fix Axis Praxis movie in Variable Fonts article

RoelN:
This MP4 movie has been reencoded to match the other videos on the
page, as this movie otherwise wouldn't play in Firefox:

afbeelding

This MP4 file should be moved to your cloud storage bucket, as done when we wrote this article: https://github.com/GoogleChrome/web.dev/pull/3445#issuecomment-692241527 @kaycebasques Is that something I can bother you for? :-)


RoelN:
Note that the staging will not show the fixed movie, as it will need to be moved to the cloud storage bucket first.

mihajlija:
@RoelN Thank you for the fix, adding this to my TODO list and I'll make sure the video is uploaded by Monday.

davelab6:
Thank you @mihajlija!

src/site/content/en/blog/variable-fonts/index.md
```diff
@@ -643,9 +643,9 @@ You can view the working example and source code for the above sample
 ### Animation
 
 <figure class="w-figure">
-  <video controls autoplay loop muted class="w-screenshot">
-    <source src="https://storage.googleapis.com/web-dev-assets/variable-fonts/axis-praxis.mp4" type="video/mp4">
-  </video>
+  {% Video src="video/vgdbNJBYHma2o62ZqYmcnkq3j0o1/2Du2L0Ii5nUqz8n6S3Vz.mp4",
+  class="w-screenshot"
```
mihajlija, line 647:
```suggestion
  {% Video src="video/vgdbNJBYHma2o62ZqYmcnkq3j0o1/2Du2L0Ii5nUqz8n6S3Vz.mp4",
  class="w-screenshot",
  controls=false,
  autoplay=true,
  loop=true,
  muted=true
```

mihajlija (approved):

RoelN:
@mihajlija Thanks picking this up!

Re: https://github.com/GoogleChrome/web.dev/pull/4746/commits/ee6b781b044e9db927a38cccedea768586b6b868, I think the `{% video %}` component sets these automatically so they're not set for other videos in this document. So they might not be needed? Unless you explicitly want to turn off controls, which the other videos still have. Not sure what's here but figured I'd point this out for consistency's sake.

mihajlija:
The `{% Video %}` component is a new feature on web.dev, it creates a video that doesn't loop, or autoplay, with the controls visible by default, so the attributes are necessary to make this video behave like all the others on the page. From what I can tell, the controls are visible on hover on all the videos.

RoelN:
@mihajlija Gotcha now, thanks for the explanation. Thanks for picking this up!

PR #4749, opened 2021-02-25
[Display Override article] Fix hero image

tomayac:

- Fix hero image


MichaelSolati (approved):
Not my area of expertise, but the image CDN stuff is, so LGTM.

PR #4750, opened 2021-02-25
feat(Img): add test to see if images passed into shortcode have correct syntax from uploader

MichaelSolati:


robdodson (approved):

robdodson:
generally LGTM but rerunning the percy test to see if it'll complete succesfully

robdodson (changes requested):
Looks like we're getting some errors being reported now. It may be the shortcode working as intended, but we'll need to fix them before we merge.

src/site/_includes/components/Img.js
```diff
@@ -8,6 +8,30 @@ const MAX_WIDTH = 800;
...
+/**
+ * @param {string} src
+ * @returns {boolean}
+ */
+const IS_UPLOADED_IMG = (src) => {
+  /**
+   * Because file extensions may be upper case, we split the string based on
+   * a `.`, which we expect to signify the files extension. We then check if
+   * new array has a length less than 2. If it does then there was no `.` and
+   * therefore there was not extension. Then we lower case the last element
```
robdodson, line 20:
```suggestion
   * therefore there was no extension. Then we lower case the last element
```

robdodson (approved):
LGTM. I think it's ok to go ahead and merge, but I'd also like to get @samthor's thoughts because he's _much_ better at regex than me 😁  

So we don't block, I was thinking if Sam has comments we can do another PR on Monday.

PR #4772, opened 2021-02-28
Add Agrofy case study

mihajlija:
Fixes ?

cc @kaycebasques


kaycebasques (approved):
LGTM but let's confirm with Agrofy before merging

PR #4773, opened 2021-02-28
[Handbook update] Add instructions for inlcuding image captions

mihajlija:
I was working on a PR where I wanted to include captions with an image and had to do a little deducing to figure out how it works with the new image shortcodes. This should save our contributors some time.

Also, instructions with CSS classes for images (w-screenshot etc.) are gone from the handbook and I think we should bring them back. They still work and I find them very useful.

cc @kaycebasques


robdodson:
LGTM! Feel free to merge when you're ready @mihajlija

mihajlija:
@robdodson Thanks! I just opened an issue for this, i'd appreciate it if you take a look :)

PR #4776, opened 2021-03-01
Update the close port example in the serial article

beaufortfrancois:
This PR updates an example in the serial article to make the design pattern for closing a serial port clearer as raised in https://github.com/WICG/serial/issues/126.

### DIFF
--- a/src/site/content/en/blog/serial/index.md
+++ b/src/site/content/en/blog/serial/index.md
@@ -4,7 +4,7 @@ subhead: The Web Serial API allows websites to communicate with serial devices.
 authors:
   - beaufortfrancois
 date: 2020-08-12
-updated: 2021-02-23
+updated: 2021-03-01
 hero: image/admin/PMOws2Au6GPLq9sXSSqw.jpg
 thumbnail: image/admin/8diipQ5aHdP03xNuFNp7.jpg
 alt: |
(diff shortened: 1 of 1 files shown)


mihajlija (approved):

src/site/content/en/blog/serial/index.md
```diff
@@ -296,26 +296,45 @@ case, calling `reader.cancel()` will force `reader.read()` to resolve
...
-  // value is a Uint8Array.
-  console.log(value);
+
+  await port.close();
 }
 
+const closed = readUntilClosed();
+
+
+// Sometimes later...
```
reillyeon, line 332: "Sometime later"

It might be helpful to break this example into two code blocks with text in between making it even clearer that the code in this last section is expected to be invoked by something like a button click.

jpmedley: @beaufortfrancois For what it's worth, I've made this exact suggestion on lots of code samples.

beaufortfrancois: Is https://github.com/GoogleChrome/web.dev/pull/4776/commits/887fdfdb3bda7a8460b2e30026e3541aebd92e05 better?

src/site/content/en/blog/serial/index.md
```diff
@@ -323,18 +323,19 @@ async function readUntilClosed() {
     }
   }
 
-  await port.close();
+  return port.close();
```
jpmedley, line 326: This still returns a promise. Did you remove `await` because you don't care about the promise? Since someone might infer behavior from the code instead of reading the text, I'd suggest adding a comment that you don't care about the returned promise.

jpmedley (changes requested):

src/site/content/en/blog/serial/index.md
beaufortfrancois: I think it's clearer this way.
See https://jakearchibald.com/2017/await-vs-return-vs-return-await/

jpmedley: It's clear if you know that `close()` returns a promise and that your script doesn't need it. If I'm reading this article, it's presumed that I don't know anything about the api or the method.

reillyeon: Since `close()` returns `Promise<undefined>` I prefer using `await` over `return` to highlight the fact that it is an asynchronous operation.

reillyeon (approved):

src/site/content/en/blog/serial/index.md
beaufortfrancois: Addressed.

PR #4798, opened 2021-03-02
Tabs post patch: adds missing alt text

argyleink:


argyleink:
i have rewrap, tried using it on these alt text areas, didnt work? should i manually do it for alt text?

robdodson (approved):

robdodson:
@kaycebasques do you think we could have the review bot look for "ALT_TEXT_HERE" and fail if it sees it.
similarly should we have it look for alt="" and warn if it sees that? There are some scenarios where alt="" is ok (like if an image is purely decorative) but I don't think most folks are using images like that so it would almost always be a mistake I bet.

PR #4806, opened 2021-03-04
fix linter and remove all smart quote styles

samthor:
We were matching single smart-quotes (‘’) but not doubles (“”).

This has actually fixed issues with inline samples already: e.g., the fix to "blog/link-prefetch/index.md".

This also fixes some issues with em/en dashes. We didn't hit these before as we considered all lines separately, but this would actually create space between content.

### DIFF
--- a/src/site/content/en/blog/min-max-clamp/index.md
+++ b/src/site/content/en/blog/min-max-clamp/index.md
@@ -1,4 +1,4 @@
-k---
+---
 title: 'min(), max(), and clamp(): three logical CSS functions to use today'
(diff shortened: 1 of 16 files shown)


src/site/content/en/blog/min-max-clamp/index.md
```diff
@@ -1,4 +1,4 @@
----
+k---
```
robdodson, line 1:
```suggestion
---
```

robdodson (approved):

robdodson:
good catch, thank you!

PR #4812, opened 2021-03-05
Chore/courses toc

### LINKED ISSUE #4580: ToC for courses
The ToC looks similar to the one we're using on developer.chrome.com. I think we should migrate that over to web.dev and not only use it for these courses but maybe in the rest of our articles as well. Definitely we should prioritize landing it in the courses first, especially if there are blockers which make it difficult to easily update the rest of our content to use it.

### LINKED ISSUE #4716: nvm and "engines" out of sync
https://github.com/GoogleChrome/web.dev/blob/main/.nvmrc specifies Node 10
https://github.com/GoogleChrome/web.dev/blob/main/package.json#L161 specifies Node 12.

robdodson:
Fixes #4580, fixes #4716

- Adds gorko utility css from d.c.c.
- Adds icon macro (and supporting svg normalization filter) from d.c.c.
- Removes some outdated sasslint rules. We're going to migrate from $SCREAMING_SNAKE_CASE variables to $dash-case (which is what d.c.c. uses)
- Adds a random string to CSS and JS files when eleventy rebuilds in dev to avoid caching
- Adds the `toc` filter that Sam wrote for d.c.c.
- Adds the `px-to-rem` and `px-to-em` sass functions from d.c.c.
- Adds a `.flow` utility. On d.c.c. we have a somewhat more elaborate version of this called `.stack` which sets everything to display flex + column. I think it's handy just having `.flow` by itself, though I may add `.stack` to web.dev at some point as well.
- Revs the `.nvmrc` and `engines` property in `package.json` because the `toc.js` filter uses optional chaining and it was causing the build to fail.
- Begins using the [GoogleChrome/webdev-infra](http://github.com/googlechrome/webdev-infra) shared repo for a couple of filters.


samthor (approved):
I'm sad about the technical debt we're leaving to be merged in future but I'm generally fine with this PR.

src/styles/tools/_functions.scss
```diff
@@ -15,3 +15,13 @@
 @function calculate_unitless_line_height($line_height, $font_size) {
   @return $line_height / $font_size;
 }
+
+// Pixels to ems.
+@function px-to-em($px) {
+  @return $px / $base-font-size * 1em;
+}
+
+// Pixels to rems.
```
samthor, line 24: Do you think this is worth bringing across? Can we just keep using px instead?

robdodson: I'm using it in the toc widget and we use it a lot in d.c.c.'s css so it seemed like it's worth pulling across.

robdodson:
> I'm sad about the technical debt we're leaving to be merged in future but I'm generally fine with this PR.

Sam and I chatted and we're starting to address this by having a few of the filters use the webdev-infra widget for shared code betweeen web.dev and dcc

PR #4826, opened 2021-03-08
Handbook updates

kaycebasques:
WIP


samthor:
@robdodson Kayce is clearly not going to work on this PR; but it also deletes a bunch of stuff, so I don't actually understand the goal here.

Thoughts? Shall we revert this PR?

robdodson:
I think this is something that we do want to keep, as it does a good job of explaining our content process.

I think a big chunk of stuff that was deleted is because it's duplicative of the Google style guide. The one exception is the 'use media' section—I'm not sure why that was deleted.

@mihajlija and I have been talking about making some handbook updates so I'm going to ask that we keep this open for now. We'll discuss it on our tech writer sync tomorrow.

robdodson:
Quick update that @rachelandrew and @mihajlija are going to work on this a bit in separate PRs. I'm going to leave this one open for a bit just in case we decide to steal the process explanation stuff.

PR #4831, opened 2021-03-08
Add social tags content to path landing pages.

### LINKED ISSUE #4815: some pages do not have descriptive titles
**Describe the bug**
quite a few pages have just "web.dev" in the title.

looks like those are contents under https://github.com/GoogleChrome/web.dev/tree/main/src/site/_data/paths

**To Reproduce**
1. Go to https://web.dev/metrics/
2. Look at the title

**Expected behavior**
It should show something like "Metrics - web.dev"

### LINKED ISSUE #4828: Twitter cards markup for collection landing pages is incorrect
Example page: https://web.dev/learn-web-vitals/ (all pages under [/learn/](https://web.dev/learn/) actually)
Output from [Twitter Card Validator](https://cards-dev.twitter.com/validator): ```ERROR: Required meta tag missing (twitter:text:title)```

devnook:
Since Learning path landing pages are generated from _data json files, they did not have page.title and page.description attributes. This Pr adds these fields to renderData so that they can be used by the Meta component to render social data in the head.

Fixes #4828, Fixes #4815


MichaelSolati (approved):

src/site/_includes/path.njk
```diff
@@ -1,5 +1,9 @@
 ---
 layout: default
+renderData:
```
robdodson, line 3: I think renderData is deprecated in favor of using [elventyComputed](https://www.11ty.dev/docs/data-computed/)

I think you might be able to do this with eleventyComputed in this way:

```js
eleventyComputed:
  title: {{ paths[pathName].title }}
```

robdodson (changes requested):
Looks good, I think we just need to make a small change to using `eleventyComputed` instead of `renderData`

src/site/_includes/path.njk
devnook: done

devnook:
ptal

robdodson:
just trying to figure out why everything is failing. It looks like maybe js-yaml updated

devnook:
@MichaelSolati I added a whole new part with learning path images migration (it was needed to make Twitter cards display the image properly). Can you please take another look?

MichaelSolati:
So we now have svgs rendering correctly in twitter cards

image


However I've noticed that not all of these svgs are the same size, and as a result they're getting warped on pages like => https://deploy-preview-4831--web-dev-staging.netlify.app/angular/

image

maybe we can store the height and width in the paths/*.json and use them for the path's cover image?

devnook:
@robdodson ptal

robdodson:
I think this is blocked by https://github.com/GoogleChrome/web.dev/issues/4890

MichaelSolati:
@googlebot I consent.

robdodson (approved):

PR #4839, opened 2021-03-09
fix(en): copy default locale's (en) content to i18n folder on prod builds, fixes #4642

### LINKED ISSUE #4642: INTL: Serve EN content when user has set English as the highest priority language
No English content is served when other supported language is present in the Accept-Language header, even if english has higher priority.

MichaelSolati:
I tried to get clever with this, but gulp globs ended up being the easiest solution.

As a weirdness I noticed that I wanna bring to @robdodson's attention. 

Content in `./src/site/content/en/handbook/sandbox` all have permalink turned off. As a result they still get built to  `./dist/en/handbook/sandbox`. Which I don't think we want, and then with this pr they also get added to `./dist/i18n/en/en/handbook/sandbox`.

Should I remove the `permalink: false` in the frontmatter?


gulpfile.js
```diff
@@ -109,6 +109,16 @@ gulp.task('copy-fonts', () => {
   return gulp.src('src/fonts/**/*').pipe(gulp.dest('dist/fonts/'));
 });
 
+// Copies content default locale into it's own folder on prod.
```
devnook, line 112: Let's mention here we need it because of FB hosting lang redirects.

robdodson:
> Content in ./src/site/content/en/handbook/sandbox all have permalink turned off. As a result they still get built to ./dist/en/handbook/sandbox

hm, that seems odd. Setting `permalink: false` should prevent files from being generated.

I don't think we use the /sandbox/ stuff anymore so we can remove that whole directory of stuff.

gulpfile.js
robdodson, line 112:
```suggestion
// Copies content default locale into its own folder on prod.
```

🤓

gulpfile.js
```diff
@@ -109,6 +109,16 @@ gulp.task('copy-fonts', () => {
...
+gulp.task('default-locale', () => {
+  if (isProd) {
+    return gulp
+      .src(['./dist/**/*.html', '!./dist/i18n/**/*.html'])
```
robdodson, line 116: This copies the html files but not the images. If firebase were to (for whatever reason) use this directory would the relative image paths all break?

gulpfile.js
robdodson: +1

gulpfile.js
MichaelSolati: No, images would be fine. The top level is the default and will fallback to that when no language is provided.

PR #4852, opened 2021-03-09
Add i18n filter for templates

devnook:
This is based on the approach in dcc. I only translate 1 file (footer.njk) so that in case there are changes to the approach I don't have to update many files. The next translations will be added in consecutive PR.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -67,6 +67,7 @@ const tags = require(`./${collectionsDir}/tags`);
 
 const filtersDir = 'src/site/_filters';
 const consoleDump = require(`./${filtersDir}/console-dump`);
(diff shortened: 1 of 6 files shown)


robdodson (approved):

devnook:
Yes, it is the same as dcc. I have some doubts about the ux of this approach, and tried a few other options, but in the end I decided that done is better perfect and went with the dcc code to unblock the i18n project.

robdodson (approved):

robdodson:
> I have some doubts about the ux of this approach

I'd be interested to know your thoughts. I feel like a lot of the i18n stuff has been pretty tricky for me to find a good solution for. Almost every option seems to have its own unique drawbacks.

PR #4859, opened 2021-03-10
Proposal to update information on lazy-loading

danieltxok:
Chrome in Android stops the autoplay but does not stop from downloading the video, so strictly speaking, it is not lazy-loading it. I tested this on Android. I am not aware of a way of doing this natively. Thanks

CC @addyosmani @rachelandrew @malchata

### DIFF
--- a/src/site/content/en/fast/lazy-loading-video/index.md
+++ b/src/site/content/en/fast/lazy-loading-video/index.md
@@ -90,9 +90,8 @@ The `autoplay`, `muted`, and `loop` attributes are self-explanatory.
 [`playsinline` is necessary for autoplaying to occur in
 iOS](https://webkit.org/blog/6784/new-video-policies-for-ios/). Now you have a
 serviceable video-as-GIF replacement that works across platforms. But how to go
-about lazy-loading it? [Chrome will lazy-load video for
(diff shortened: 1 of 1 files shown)


malchata:
This is just my opinion, but I see any form of autoplaying video without explicit opt-in to be an anti-pattern, and for what it's worth, I've considered removing the ability from [yall.js](https://github.com/malchata/yall.js) to allow it. However, my opinion is obviously not Google's, so I'd be interested in discussing this and—if warranted—modifying our guidance accordingly.

mihajlija:
Paging @khempenius for input.

cc @philipwalton

khempenius:
I agree that this section needs to be updated. I would remove these two sentences:

```
Chrome will lazy-load video for you, but you can't count on all browsers to provide this optimized behavior. Depending on your audience and application requirements, you may need to take matters into your own hands.
```

khempenius:
With regards to autoplay guidance: my opinion (FWIW) is that I'm generally in favor of avoiding the use of autoplay, but as a platform feature I think it has its place - if nothing else, because it provides as a way to create "GIF style" media experiences without the ridiculous file sizes of actual GIFs (and smaller file sizes than WebP).

PR #4872, opened 2021-03-11
Simplify Aside and use i18n filter on it

### LINKED ISSUE #1617: Markdown reference links in `{% Aside %}` aren't recognized
**Describe the bug**
If I define a reference link (eg: `[my-link]: https://google.com`) at the bottom of my doc, then try to use it in an `{% Aside %}`, it doesn't recognize the reference link.

**To Reproduce**
1. At the bottom of your doc add `[my-link]: https://google.com`
2. In your main doc, add:

```
{% Aside %}
This is a test link to [Google][my-link], did it work?
{% endAside %}
```

**Expected behavior**
I would expect the link to be rendered.

**Actual behavior**
```
This is a test link to [Google][my-link], did it work?
```

### LINKED ISSUE #4787: {% Aside %} component breaks when nested in a numbered list
**Describe the bug**
Issues:
- All text within the `{% Aside %}` renders in "code-block" style - even if it is not a code block.
- The `{% Aside %}` breaks list numbering - the next list item after the `{% Aside %}` is always "1.".

**To Reproduce**
```
1. One

2. Two
  {% Aside %}
    Blah
  {% endAside %}

3. Three 
```

For more examples, see https://github.com/GoogleChrome/web.dev/pull/4764. This has a bunch of places where I used `**Note:**` or `**Optional**` headers to act as a defacto aside.

devnook:
This also removes the md.render calls in favour of 11ty native markdown parser (forced with the use of newlines).

Fixes #1617, Closes #4787


src/site/_filters/i18n.js
```diff
@@ -76,4 +76,13 @@ const i18n = (path, locale = 'en') => {
   }
 };
 
-module.exports = {i18n};
+/**
+ * Infer the page locale using page's filePathStem.
```
robdodson, line 80: Would you mind adding a note that this is only needed when we need to use the i18n filter inside of the implementation of a shortcode? Otherwise the 11tydata.js file at the root of each language directory should set the locale so normally you can do `'i18n.some.thing' | i18n` and it'll "just work"

robdodson:
I was able to confirm that markdown inside of the aside (lists and code) as well as reference links work.

````md
{% Aside %}
Use the note aside to provide supplemental information.

- Lorem ipsuom dolor set
- Lorem ipsuom dolor set
- Lorem ipsuom dolor set

```js
function hello() {
  console.log('hello world');
}
```

With a [reference style link][google] outside of the Aside
{% endAside %}

[google]: https://google.com
````

image

I was also able to confirm that this snippet works with the **Gotchas** variation.

image

robdodson (approved):

robdodson:
LGTM, great work @devnook !

src/site/_filters/i18n.js
devnook: I actually see this as a helper function that can be used in different types of context. If you already have a locale, either from the top of the data cascade, or passed as a param, then great and don't use this function. But I'd rather not define it's usage here. wdyt?

PR #4877, opened 2021-03-11
Add translations to _data/paths

devnook:


robdodson (approved):
This change LGTM. I think we'll want to communicate this to the tech writing team so they understand that learning paths need to use the i18n stuff otherwise they'll get errors. We could communicate it on the web.dev-discuss@ email list and we should also think about writing up some guidance in the handbook. We have guidance in the dcc handbook https://developer.chrome.com/docs/handbook/how-to/add-a-project/#add-i18n-paths-for-your-titles

devnook:
I'll update the docs and communicate in our channels, thanks!

PR #4886, opened 2021-03-12
Attempt to fix broken rendering of coop-coep article

developit:
The there's an extra end tag that causes the page's markup to be rendered outside of the center column:
https://web.dev/coop-coep/#:~:text=Understanding%20same-site%20and%20same-origin

> **Update:** this did not fix the issue.


robdodson:
I was able to use `git bisect` to verify that we introduced this bug in https://github.com/GoogleChrome/web.dev/pull/4872

robdodson:
I'm going to see if we can do a quick fix, if not I may need to roll back that PR

PR #4887, opened 2021-03-12
Change relative path to assets path for actions CTAs

### LINKED ISSUE #4807: Share/subscribe CTAs icons don't load in subfolder
On a page [like this](https://web.dev/handbook/web-dev-components), the CTA buttons look like:

Screen Shot 2021-03-04 at 15 24 43

The URLs are relative to the current page URL, which seems to assume a single-level. Granted, this is just the handbook page, but we might have more of these around in future.

adzo261:
Fixes #4807

- Share and Subscribe action CTAs used in https://web.dev/handbook/web-dev-components/ seemed to be using assets from the images subfolder in `src`.
- Changed the relative path to image cdn assets path.

### DIFF
--- a/src/styles/components/_actions.scss
+++ b/src/styles/components/_actions.scss
@@ -119,7 +119,7 @@ $W_FAB_LARGE_RADIUS: 28px;
 //------------------------------------------------------------------------------
 
 .w-actions__fab--share::before {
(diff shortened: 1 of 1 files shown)


robdodson (approved):

robdodson:
nice one @adzo261, thank you!

PR #4889, opened 2021-03-12
Make sure Asides and Compare widgets dont break markdown-it

robdodson:
Fixes #4886

- Remove newlines from shortcodes which were introducing closing </p> tags and breaking the page.


MichaelSolati (approved):
This feels gross, we should probably wrap these in some sort of comment that prevents prettier from formatting them => https://prettier.io/docs/en/ignore.html#range-ignore

PR #4892, opened 2021-03-13
Learn css side nav

### LINKED ISSUE #4579: Need slightly tweaked header for courses
Course headers are very similar to our other content, but they include a little number at the top. Maybe there's an easy way for us to sneak this in.

### LINKED ISSUE #4583: Side nav for Courses
This is probably one of the trickier pieces of the UI. The drawer is visible on desktop and hides on mobile with the classic hamburger pattern. The header gets quite large on mobile and I think it's supposed to reduce in height when you scroll. It also contains a search box so we'll have to think through how that works.

robdodson:
Fixes #4583, Fixes #4579

This PR is pretty massive because to build the navigation drawer and header for the course section I ended up needing to refactor a number of parts of the site to make them more flexible.

At a high level this renames the SideNav to NavigationDrawer (web-navigation-drawer) so it can match the naming used in the Material spec. It also refactors the drawer so the _contents_ are not part of the drawer's CSS. This makes it easy to add new menus for the site if we need to in the future. So now we have:

- _web-navigation-drawer.scss
  - _drawer_default.scss
  - _drawer-course.scss

Similarly the header was refactored so it's just a sticky position box at the top of the screen and the content is styled separately. So now we have:

- _web-header.scss
  - _header-default.scss
  - _header-course.scss

I scope creeped this PR by also adding course configuration to it. Values defined in `_data/courses/css/meta.yml` get injected into the templates to define things like the course name, the text gradient used on the logo, and the background image used on the drawer/header.


.eleventy.js
```diff
@@ -87,6 +87,7 @@ const stripBlog = require(`./${filtersDir}/strip-blog`);
 const stripQueryParamsDev = require(`./${filtersDir}/strip-query-params-dev`);
 const getPaths = require(`./${filtersDir}/get-paths`);
 const navigation = require(`./${filtersDir}/navigation`);
+const padStart = require(`./${filtersDir}/pad-start`);
```
robdodson, line 90:
```suggestion
const padStart = require(`./${filtersDir}/pad-start`);
```

src/lib/components/Header/index.js
```diff
@@ -15,32 +15,33 @@
...
-    store.subscribe(this.onStateChanged);
+    this.openDrawerBtn = this.querySelector('[data-open-drawer-button]');
+    if (this.openDrawerBtn) {
+      this.openDrawerBtn.addEventListener('click', openNavigationDrawer);
+    }
   }
 
   disconnectedCallback() {
-    this.hamburgerBtn.removeEventListener('click', expandSideNav);
+    if (this.openDrawerBtn) {
```
MichaelSolati, line 42: Excessive complaining, but could we do optional chaining here?

`this.openDrawerBtn?.removeEventListener()`

src/lib/components/Header/index.js
```diff
@@ -72,12 +73,14 @@ export class Header extends HTMLElement {
...
   /**
-   * This is called by the SideNav to return focus to this control when the
-   * user closes the SideNav.
+   * This is called by the NavigationDrawer to return focus to this control when
+   * the user closes the NavigationDrawer.
    * This is important for accessibility.
    */
   manageFocus() {
-    this.hamburgerBtn.focus();
+    if (this.openDrawerBtn) {
```
MichaelSolati, line 81: Same as above

src/lib/components/Header/index.js
robdodson: yep since this isn't client side we should be ok to use it.

PR #4903, opened 2021-03-15
Add instruction to productionize CertURLBase

twifkak:

- Updates the web packager "Serve signed exchanges using a `CanSignHttpExchanges` certificate" section to mention need to remove or update the `CertURLBase = "data:"` that's only for dev.

/cc @khempenius


khempenius:
@twifkak I think this would solve the same issue and is probably good to mention anyway in case people skip the instructions for local setup. WDYT?
https://github.com/GoogleChrome/web.dev/compare/twifkak-patch-4

twifkak:
@khempenius Ah, OK, yeah that makes sense. Maybe add an adjective for emphasis so people don't miss it, like "Create a fresh `webpkgsever.toml`" (or "brand new", etc.)

Issue #4873, opened 2021-03-11
A broken link


AM-77:
**Describe the bug**
A broken link that redirect to a 404 page in github

**To Reproduce**
1. Go to https://web.dev/offline-start-url/#resources
2. Click on the link  [Source code for start_url does not respond with a 200 when offline audit](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/audits/offline-start-url.js)

lucalves:
Hello @AM-77, I opened PR #4906 to fix this. Let's wait for the review.

mihajlija:
Fixed #4906

PR #4906, opened 2021-03-16
Removes a broken audit link

lucalves:
Hello everyone, everything good?

According to this [Pull Request](https://github.com/GoogleChrome/lighthouse/pull/11806), in the lighthouse repository where @paulirish remove works-offline and offline-start-url audits, the current link is pointing to a broken page.

Therefore, I believe that it would be the best solution to remove the link, since the file no longer exists.

This fixes Issue #4873.


mihajlija:
@lucalves Thank you for the fix!

mihajlija (approved):

PR #4911, opened 2021-03-17
Fix analysis anchor in cross origin isolation guide

CYBAI:

- The link of `Analyze the impact of cross-origin isolation` in https://web.dev/cross-origin-isolation-guide/ is 404 now because it's a typo with leading colon. Instead, it should use `#` to be an anchor.


mihajlija (approved):

mihajlija:
@CYBAI Thank you for the fix!

PR #4934, opened 2021-03-21
fix base card anchor tags

### LINKED ISSUE #4794: Cards in blog sections are interactive but doesn't work when clicked in whitespaces
**Describe the bug**
Card in blog sections are not entirely clickable while there is visual feedback for the click.

**To Reproduce**
1. Go to https://web.dev/blog/
2. Click on any card but in whitespace
4. Clicking should take to the article but nothing happens while providing visual cues that card can be clicked anywhere.

**Expected behavior**
Clicking anywhere on the card should take to the article.

**Screenshots**
https://user-images.githubusercontent.com/5694967/109661816-489a1100-7b62-11eb-8fd9-0ecb56b03c5c.mov

MarianaPicolo:
Fixes #4794

### DIFF
--- a/src/site/_includes/components/AuthorsDate.js
+++ b/src/site/_includes/components/AuthorsDate.js
@@ -52,9 +52,9 @@ const renderAuthorImages = (limit, pairs) => {
       });
       return html`
         <div class="w-author__image--row-item">
-          <a href="${info.href}">
+          <span>
             ${img}
-          </a>
+          </span>
         </div>
       `;
     })
@@ -74,9 +74,7 @@ const renderAuthorNames = (pairs) => {
 
   const inner = pairs
     .map(({info}) => {
-      return html`
-        <a class="w-author__name-link" href="${info.href}">${info.title}</a>
-      `;
+      return html` <span class="w-author__name-link"> ${info.title}</span> `;
     })
     .join(', ');
 
--- a/src/site/_includes/components/BaseCard.js
+++ b/src/site/_includes/components/BaseCard.js
@@ -83,11 +83,9 @@ class BaseCard {
     }
 
     return html`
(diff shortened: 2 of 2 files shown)


MarianaPicolo:
@googlebot I signed it!

robdodson:
Hey @MarianaPicolo, thank you for taking a look at this. I think one issue that I noticed is that making the entire card an anchor means you can no longer click on the author or click the tags at the bottom of the card.

Though it seems a little hacky, I'm kind of wondering if we could do something like this instead:

```
<div onclick="window.location.href=${this.url}" class="w-card
```

That would make it so the links inside of the card still go to different destinations, but if someone happens to click on the card with a mouse then they'll get the expected behavior.

robdodson:
I asked in our team chat about this and it lead to a very interesting discussion. Some folks said that they like that some of the card is not clickable because they like to drag to select the text and use text-to-speech to read it out loud.

It sounds like to do this without breaking that use case would be a bit complicated. We may need to use pointer events, instead of click, and abort if pointerdown and pointerup happen within ~5px of each other. 😅

Quite tricky!

Another option is to say that it's a feature and not a bug. Essentially having some of the card not clickable means we have (accidentally) enabled a good accessibility use case for users who like text-to-speech.

MarianaPicolo:
@robdodson thanks for reviewing my first contribution! 🥳 

It's indeed complicated. I actually found tricky to remove the anchor tags because the author info will not be clickable in the `/live` page anymore. 

Maybe we can accept that the card not entire clickable is not a bug, and close the issue 😄

robdodson:
> Maybe we can accept that the card not entire clickable is not a bug, and close the issue 😄

yeah I think that's a good call. I noticed the cards on css-tricks are also not entirely clickable so we are in good company 😁

image

PR #4942, opened 2021-03-22
Translate Instructions component

devnook:


src/site/_includes/components/Instruction.js
```diff
@@ -16,18 +16,21 @@
...
  * This helps ensure consistency in our docs and makes it easy
  * to respond when Glitch changes their UI.
+ * @this {EleventyPage}
  * @param {string} type The type of instruction to print.
  * @param {string} listStyle The list style to use. Defaults to 'ul'.
  * @return {string} A list of instructions.
  */
-module.exports = (type, listStyle = 'ul') => {
+function Instruction(type, listStyle = 'ul') {
+  const locale = getLocaleFromPath(this.page && this.page.filePathStem);
```
robdodson, line 31: I'm kind of wondering if `this.page && this.page.filePathStem` is useful here... I think if `this.page` is undefined it's still going to call the function but may pass in `undefined` or `false` as the argument, right? The function has a default return value so that would be fine, but it does feel a little sketchy to potentially pass an undefined into a function.

src/site/_data/i18n/instructions.yml
```diff
@@ -0,0 +1,60 @@
...
+
+console:
+  click_tools:
+    en: Click **Tools**.
+
+  click_logs:
+    en: Click **Logs**.
+
+  click_console:
+    en: Click **Console**.
```
robdodson, line 12: I wonder if this nesting is something the translation vendor might mess up.

src/site/_includes/components/Instruction.js
```diff
@@ -138,13 +137,13 @@ module.exports = (type, listStyle = 'ul') => {
...
-        // prettier-ignore
+      instruction = getInstruction('devtools.open');
+      let tab = type.substring('devtools-'.length);
+      if (tab) {
+        tab = capitalize(tab);
         instruction = html`
-          ${instruction}
-          ${bullet}Click the **${capitalize(substitution)}** tab.
+          ${getInstruction('devtools.open')}
+          ${getInstruction('devtools.open_tab').replace(/<TAB>/g, tab)}
```
robdodson, line 146: this seems a bit confusing because `instruction` gets redefined (whereas in the old version it was reused).

Should we remove line 140?

robdodson:
@devnook I'm doing some bug sheriffing and wanted to check-in on this one. Do you want to keep working on it?

PR #4820, opened 2021-03-06
Rev dependencies. Fix npm audit grumps.

robdodson:
Fixes #4462, fixes #478

Revs npm deps to latest versions.

PR #4949, opened 2021-03-23
allow strikethrough

samthor:
This allows strikethrough (denoted by "~~").

We still trigger on "no-duplicate-headings-in-section", but a number of blog posts repeat headings. Can we turn this one off—or perhaps limit it to only top-level headings? @robdodson

"fixes" part of #4820.


robdodson:
> We still trigger on "no-duplicate-headings-in-section", but a number of blog posts repeat headings. Can we turn this one off—or perhaps limit it to only top-level headings? @robdodson

really we should fix these because I think it's also a lighthouse accessibility failure.

robdodson (approved):

PR #4958, opened 2021-03-24
Correct the performance gain explanation

akshayagarwal:
The author most probably accidentally missed the negation. If a website uses only a single font such as Roboto Regular, they would most likely not see any meaningful gain in font file size  if they switch to a variable font with many axes.

- Correct the performance gain explanation when using variable font against only a single style font


rachelandrew (approved):
lgtm - thanks :)

PR #4959, opened 2021-03-24
First pass at using strict CSP

adamraine:
We are working on a new Lighthouse audit to recommend using a strict CSP. It is important to practice what we preach, so this PR aims to introduce a strict CSP to web.dev.

Following our own advice [here](https://web.dev/strict-csp/#step-1:-decide-if-you-need-a-nonce-or-hash-based-csp), CSP hashes seemed like the best approach because web.dev is served statically. There are a few complications, but overall this solution is less destructive than an approach using nonces.

@samthor


.eleventy.js
```diff
@@ -256,6 +257,11 @@ module.exports = function (config) {
     config.addTransform('minifyHtml', minifyHtml);
   }
 
+  config.on('afterBuild', () => {
+    const {hashList} = require('./src/site/_data/helpers');
+    fs.writeFileSync('script-hash-list.json', JSON.stringify([...hashList]));
```
adamraine, line 262: Could we generate `firebase.json` in this function to eliminate the need for a temporary file and changes to `npm run dev`?

package.json
```diff
@@ -16,7 +16,7 @@
     "debug:test:lib": "TEST_MODE=debug npm-run-all clean rollup --parallel watch:rollup karma",
     "debug:test:unit": "node --inspect-brk ./node_modules/.bin/mocha ./test/unit",
     "deploy": "ELEVENTY_ENV=prod npm run build && node index-algolia.js && firebase deploy --non-interactive",
-    "dev": "npm-run-all clean rollup gulp --parallel watch:rollup watch:eleventy watch:gulp firebase-config start",
+    "dev": "npm-run-all clean rollup gulp eleventy --parallel watch:rollup watch:gulp firebase-config start",
```
adamraine, line 19: Change ensures hashes are generated before they are used in `firebase-config.js`.

package.json
```diff
@@ -66,7 +66,7 @@
     "wicg-inert": "^3.0.1"
   },
   "devDependencies": {
-    "@11ty/eleventy": "^0.11.0",
+    "@11ty/eleventy": "^0.11.1",
```
adamraine, line 69: Need access to `config.on('afterBuild', () => ...)`

src/site/_data/lib/hash.js
```diff
@@ -52,9 +62,9 @@ const hashForProdCache = {};
  * @return {string}
  */
 function hashForProd(file) {
-  if (!isProduction) {
+  /*if (!isProduction) {
     return `${file}?v=${randomHash()}`;
-  }
+  }*/
```
adamraine, line 67: If this isn't commented out, there would be a unique script hash for every single page. This would still technically work, but it would make the CSP unnecessarily long.

adamraine:
Before merge, make sure reporting works at this destination:
https://csp.withgoogle.com/csp/webdev

connorjclark:
guess LHCI doesn't work because this is a fork?

adamraine:
@robdodson We are noticing an improvement in LH performance scores (~84 -> ~94) on the home page with this patch.

Looking a the traces from each run it looks like the LCP and FCP elements are painted sooner because scripts loaded with this script loader are run async:

Branch main (no patch):
Screen Shot 2021-03-24 at 4 51 00 PM

With CSP patch:
Screen Shot 2021-03-24 at 4 51 48 PM

Is this improvement to the LH score worthwhile if we are changing the way these scripts load in?

connorjclark:
LCP/FCP are dependent on how soon that blocking CSS script loads. The LCP element is a large text element which is styled by the CSS. LCP improves w/ the script loader because the browser only sees the CSS and loads that with no contention before any JS.

An option is to move the blocking CSS to after the script loader code. That might keep the loading behavior the same (haven't tested). I think the LH score is probably high enough (99) that a slightly better LCP value is not worth loading the scripts a bit slower.

firebase-config.js
```diff
@@ -18,5 +20,11 @@ firebaseJson.hosting.redirects = parsedRedirects.reduce(
   },
   [],
 );
+firebaseJson.hosting.headers[0].headers.push({
```
samthor, line 23: Can we remove when not in production mode? It'll make the other production stuff easier to remove.

(To be clear, you can test production mode locally. It's just a longer build.)

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,34 @@
...
+  s.src = url;
+  if (type) s.type = type;
+  if (type === 'module') {
+    const pre = document.createElement('link');
+    pre.rel = 'modulepreload';
+    pre.href = url;
+    document.head.appendChild(pre);
+    window.addEventListener('DOMContentLoaded', () => {
+      document.head.appendChild(s);
+    })
```
adamraine, line 47: I was experiencing errors on `/live/`, my hypothesis was that modules are not deferred when they are dynamically loaded like this.

Changing this code to load the modules after `DOMContentLoaded` appears to have fixed the issues.

adamraine:
Moving this to https://github.com/GoogleChrome/web.dev/pull/5008

PR #4962, opened 2021-03-24
Edit the authors note on Lowe's case study

mihajlija:
Edit the authors note on Lowe's case study


rachelandrew (commented):
we are keeping the other authors in the metadata so they show at the top?

mihajlija:
@rachelandrew Yep, this is just to highlight which of the authors are associated with Lowe's. (Safwan is from Google.)

rachelandrew (approved):

PR #4964, opened 2021-03-24
render compare caption with md

### LINKED ISSUE #4963: Markdown not rendering correctly on strict CSP article
**Describe the bug**
Markdown is not rendering properly on https://web.dev/strict-csp/
Screen Shot 2021-03-24 at 5 17 50 PM

**To Reproduce**
1. Go to https://web.dev/strict-csp/#step-2:-set-a-strict-csp-and-prepare-your-scripts
2. Open summary titled "Option B: Hash-based CSP..."
3. Scroll down just before the "Gotchas!" section.

**Expected behavior**
Markdown links are properly rendered.

**Screenshots**
Markdown links are handled as plaintext.

samthor:
Fixes #4963.

### DIFF
--- a/src/site/_includes/components/CompareCaption.js
+++ b/src/site/_includes/components/CompareCaption.js
@@ -14,10 +14,16 @@
  * limitations under the License.
  */
 
+// Need markdown-it for nested shortcodes to render as expected
+const md = require('markdown-it')();
+const {html} = require('common-tags');
+
 /**
  * @param {string} content Markdown with the content for the compare element.
  * @return {string}
  */
 module.exports = (content) => {
-  return `<figcaption class="w-compare__caption">${content}</figcaption>`;
+  return html`<figcaption class="w-compare__caption">
+    ${md.renderInline(content)}
+  </figcaption>`;
 };
--- a/src/site/_includes/components/CompareCaption.js
+++ b/src/site/_includes/components/CompareCaption.js
@@ -14,16 +14,13 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


samthor:
FWIW the Percy change was relevant—we had a line in a caption that wasn't formatted, and now it is.

robdodson (changes requested):
@samthor I think you can do this without using markdown-it. If we use markdown-it then it'll break syntax highlighting on code blocks and you can't use reference links.

I think if we introduce a newline it'll signal to the markdown parser to kick back in. I'll try it real quick and if it works I'll update the PR.

samthor:
Oh, maybe I was confused because Details does the same thing?

robdodson:
> Oh, maybe I was confused because Details does the same thing?

Yeah the web.dev components are a bit outdated. I learned two things:
- Adding the newline will let the markdown parser kick back in
- common-tags will often remove this spacing, so I try not to use it anymore

so if you look at the components used on d.c.c. I think many/most of them don't use common-tags, and they have this sort of funky whitespace which looks odd, but lets the markdown parser keep its context from the parent page which is really helpful.

robdodson (approved):

robdodson:
I checked the article and it works with this change. I know it looks a bit odd but it allows people to do things like this:

```
{% Details %}
Some [link][google]
{% endDetails %}

[google]: https://google.com
```

and I think it prevents the syntax highlighting from breaking because that's something that eleventy does as a sort of customization of markdown-it, but we don't get it if we call markdown-it ourselves.

PR #4969, opened 2021-03-25
Asyncify blog post

### LINKED ISSUE #4817: Asyncify
* [Draft](https://docs.google.com/document/d/10LI7Ip16HB3kUH2B2hwpbJWvBEpcUZPV/edit#heading=h.wqlb16cpmov)

RReverser:
Fixes #4817

- Add blog post about Asyncify.

### DIFF
--- a/src/site/content/en/blog/asyncify/index.md
+++ b/src/site/content/en/blog/asyncify/index.md
@@ -362,7 +362,7 @@ Pass `--asyncify` to enable the transform, and then use `--pass-arg=…` to prov
 list of asynchronous functions, where the program state should be suspended and later resumed.
 
 All that's left is to provide supporting runtime code that will actually do that—suspend and resume
-our code. Again, in the C / C++ case this would be included by Emscripten, but now we need custom
+WebAssembly code. Again, in the C / C++ case this would be included by Emscripten, but now you need custom
 JavaScript glue code that would handle arbitrary WebAssembly files. We've created a library just for
 that.
 
@@ -410,7 +410,7 @@ This is quite similar to async-await feature in JavaScript that I showed earlier
 JavaScript one, doesn't require any special syntax or runtime support from the language, and instead
 works by transforming plain synchronous functions at compile-time.
 
-For simplicity, let's take our asynchronous sleep example:
+When compiling the earlier shown asynchronous sleep example:
 
 ```js
 puts("A");
@@ -418,7 +418,7 @@ async_sleep(1);
 puts("B");
 ```
 
-Asyncify takes this code and transforms it to roughly like the following (pseudo-code, real
+Asyncify takes this code and transforms it to roughly like the following one (pseudo-code, real
 transformation is more involved than this):
 
 ```js
@@ -482,7 +482,7 @@ WebAssembly I/O in the console and the server-side. It was designed as a compila
 system languages, and exposes all sorts of filesystem and other operations in a traditional
 synchronous form.
 
-What if we could map one to another? Then we could compile any application in any source language
+What if you could map one to another? Then you could compile any application in any source language
 with any toolchain supporting the WASI target, and run it in a sandbox on the web, while still
 allowing it to operate on real user files! With Asyncify, you can do just that.
 
--- a/src/site/content/en/blog/asyncify/index.md
+++ b/src/site/content/en/blog/asyncify/index.md
@@ -20,7 +20,7 @@ Asyncify.
 
 ## I/O in system languages
 
-Let's start by taking a look at a simple example in C. Let's say, you want to read user's name from
+I'll start with a simple example in C. Say, you want to read the user's name from
 a file, and greet them with a "Hello, (username)!" message:
 
 ```cpp
@@ -74,7 +74,7 @@ single invocation:
 ```rust
 fn main() {
     let s = std::fs::read_to_string("name.txt");
-    println!("Hello, {}!", s);
+    println!("Hello, {}!", s);
 }
 ```
 
@@ -85,7 +85,7 @@ need to read data from some storage.
 ## Asynchronous model of the web
 
 The web has a variety of different storage options you could map to, such as in-memory storage (JS
-objects), localStorage, IndexedDB, server-side storage, and a new [File System Access
+objects), [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), server-side storage, and a new [File System Access
 API](https://web.dev/file-system-access/).
 
 However, only two of those APIs—the in-memory storage and the localStorage—can be used
@@ -177,7 +177,7 @@ console.log("B");
 
 …and, in fact, that's exactly what Emscripten does in [its default implementation of
 "sleep",](https://github.com/emscripten-core/emscripten/blob/16d5755a3f71f27d0c67b8d7752f94844e56ef7c/src/library_pthread_stub.js#L47-L52)
-but that's very inefficient, will block the entire UI, won't allow any other events to get handled
+but that's very inefficient, will block the entire UI, won't allow any other events to be handled
 meanwhile, and, generally, don't do that in production code :)
 
 Instead, a more idiomatic version of "sleep" in JavaScript would involve calling
@@ -206,7 +206,7 @@ width="800", height="200" %}
 
 ### Usage in C / C++ with Emscripten
 
-If you wanted to use it to implement an asynchronous sleep for the last example, you could do it
+If you wanted to use Asyncify to implement an asynchronous sleep for the last example, you could do it
 like this:
 
 ```cpp
@@ -229,9 +229,9 @@ is a macro that allows defining JavaScript snippets as if they were C functions.
 function
 [`Asyncify.handleSleep()`](https://emscripten.org/docs/porting/asyncify.html#making-async-web-apis-behave-as-if-they-were-synchronous)
 which tells Emscripten to suspend the program and provides a `wakeUp()` handler that should be
-called when the asynchronous operation has finished. In the example above, we pass that handler to
-`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, we call
-the `async_sleep()` function anywhere we want just like regular `sleep()` or any other synchronous
+called once the asynchronous operation has finished. In the example above, the handler is passed to
+`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, you can call
+`async_sleep()` anywhere you want just like regular `sleep()` or any other synchronous
 API.
 
 When compiling such code, you need to tell Emscripten to activate the Asyncify feature. Do that by
@@ -261,8 +261,8 @@ B
 You can [return values from
 Asyncify](https://emscripten.org/docs/porting/asyncify.html#returning-values) functions too. What
 you need to do is return the result of `handleSleep()`, and pass the result to the `wakeUp()`
-callback. For example, if, instead of reading from a file, you want to fetch some numeric answer
-from a remote resource, you can use a snippet like below to issue a request, suspend the C code, and
+callback. For example, if, instead of reading from a file, you want to fetch a number
+from a remote resource, you can use a snippet like the one below to issue a request, suspend the C code, and
 resume once the response body is retrieved—all done seamlessly as if the call was synchronous.
 
 ```js
@@ -299,8 +299,8 @@ int answer = get_answer();
 
 #### Awaiting complex values
 
-But this example still limits you only to numbers. What if you want to implement our original
-example, where we tried to get a user's name from a file as a string? Well, you can do that too!
+But this example still limits you only to numbers. What if you want to implement the original
+example, where I tried to get a user's name from a file as a string? Well, you can do that too!
 
 Emscripten provides a feature called
 [Embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html) that allows
@@ -376,7 +376,7 @@ difference is that, under a regular WebAssembly API you can only provide synchro
 imports, while under the Asyncify wrapper, you can provide asynchronous imports as well:
 
 ```js
-let … = await Asyncify.instantiateStreaming(fetch('app.wasm'), {
+const { instance } = await Asyncify.instantiateStreaming(fetch('app.wasm'), {
     env: {
         async get_answer() {
             let response = await fetch("answer.txt");
@@ -437,14 +437,14 @@ puts("B");
 ```
 
 Initially `mode` is set to `NORMAL_EXECUTION`. Correspondingly, the first time such transformed code
-is executed, only the part leading up to "async_sleep" will get evaluated. As soon as the
+is executed, only the part leading up to `async_sleep()` will get evaluated. As soon as the
 asynchronous operation is scheduled, Asyncify saves all the locals, and unwinds the stack by
 returning from each function all the way to the top, this way giving control back to the browser
 event loop.
 
 Then, once `async_sleep()` resolves, Asyncify support code will change `mode` to `REWINDING`, and
-call our function again. This time, the "normal execution" branch is skipped - since it already did
-the job last time and we want to avoid printing "A" twice - and instead it comes straight to the
+call the function again. This time, the "normal execution" branch is skipped - since it already did
+the job last time and I want to avoid printing "A" twice - and instead it comes straight to the
 "rewinding" branch. Once it's reached, it restores all the stored locals, changes mode back to
 "normal" and continues the execution as if the code was never stopped in the first place.
 
@@ -459,7 +459,7 @@ uncompressed code size.
 {% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/Im4hmQOYRHFcsg8UTfTR.png", alt="A graph showing code size overhead for various benchmarks, from near-0% under fine-tuned conditions to over 100% in worst cases",
 width="800", height="494" %}
 
-It's not ideal, but in many cases acceptable when the alternative is not having the functionality
+This isn't ideal, but in many cases acceptable when the alternative is not having the functionality
 altogether or having to make significant rewrites to the original code.
 
 Make sure to always enable optimizations for the final builds to avoid it going even higher. You can
@@ -471,20 +471,20 @@ to the cost of the actual work, it's usually negligible.
 
 ## Real-world demos
 
-Now that we've looked at the simple examples, let's move on to more complicated scenarios.
+Now that you've looked at the simple examples, I'll move on to more complicated scenarios.
 
 As mentioned in the beginning of the article, one of the storage options on the web is an
 asynchronous [File System Access API](https://web.dev/file-system-access/). It provides access to a
 real host filesystem from a web application.
 
-On the other hand, there is a de-facto standard—[WASI](https://github.com/WebAssembly/WASI) - for
+On the other hand, there is a de-facto standard called [WASI](https://github.com/WebAssembly/WASI)  for
 WebAssembly I/O in the console and the server-side. It was designed as a compilation target for
 system languages, and exposes all sorts of filesystem and other operations in a traditional
 synchronous form.
 
 What if we could map one to another? Then we could compile any application in any source language
 with any toolchain supporting the WASI target, and run it in a sandbox on the web, while still
-allowing it to operate on real user files! With Asyncify, we can do just that.
+allowing it to operate on real user files! With Asyncify, you can do just that.
 
 In this demo, I've compiled Rust [coreutils](https://github.com/RReverser/coreutils) crate with a
 few minor patches to WASI, passed via Asyncify transform and implemented asynchronous
--- a/src/site/content/en/blog/asyncify/index.md
+++ b/src/site/content/en/blog/asyncify/index.md
@@ -20,8 +20,8 @@ Asyncify.
 
 ## I/O in system languages
 
-I'll start with a simple example in C. Say, you want to read the user's name from
-a file, and greet them with a "Hello, (username)!" message:
+I'll start with a simple example in C. Say, you want to read the user's name from a file, and greet
+them with a "Hello, (username)!" message:
 
 ```cpp
 #include <stdio.h>
@@ -38,18 +38,18 @@ int main() {
 }
 ```
 
-While the example doesn't do much, it already demonstrates something you'll find in an application of any
-size: it reads some inputs from the external world, processes them internally and writes outputs
-back to the external world. All such interaction with the outside world happens via a few functions
-commonly called input-output functions, also shortened to I/O.
+While the example doesn't do much, it already demonstrates something you'll find in an application
+of any size: it reads some inputs from the external world, processes them internally and writes
+outputs back to the external world. All such interaction with the outside world happens via a few
+functions commonly called input-output functions, also shortened to I/O.
 
-To read the name from C, you need at least two crucial I/O calls: `fopen`, to open the
-file, and `fread` to read data from it. Once you retrieve the data, you can use another I/O
-function `printf` to print the result to the console.
+To read the name from C, you need at least two crucial I/O calls: `fopen`, to open the file, and
+`fread` to read data from it. Once you retrieve the data, you can use another I/O function `printf`
+to print the result to the console.
 
 Those functions look quite simple at first glance and you don't have to think twice about the
-machinery involved to read or write data. However, depending on the environment, there can
-be quite a lot going on inside:
+machinery involved to read or write data. However, depending on the environment, there can be quite
+a lot going on inside:
 
   - If the input file is located on a local drive, the application needs to perform a series of
     memory and disk accesses to locate the file, check permissions, open it for reading, and then
@@ -58,8 +58,8 @@ be quite a lot going on inside:
   - Or, the input file might be located on a mounted network location, in which case, the network
     stack will now be involved too, increasing the complexity, latency and number of potential
     retries for each operation.
-  - Finally, even `printf` is not guaranteed to print things to the console and might be redirected to
-    a file or a network location, in which case it would have to go via the same steps above.
+  - Finally, even `printf` is not guaranteed to print things to the console and might be redirected
+    to a file or a network location, in which case it would have to go via the same steps above.
 
 Long story short, I/O can be slow and you can't predict how long a particular call will take by a
 quick glance at the code. While that operation is running, your whole application will appear frozen
@@ -68,8 +68,8 @@ and unresponsive to the user.
 This is not limited to C or C++ either. Most system languages present all the I/O in a form of
 synchronous APIs. For example, if you translate the example to Rust, the API might look simpler, but
 the same principles apply. You just make a call and synchronously wait for it to return the result,
-while it performs all the expensive operations and eventually returns the result in a
-single invocation:
+while it performs all the expensive operations and eventually returns the result in a single
+invocation:
 
 ```rust
 fn main() {
@@ -85,8 +85,9 @@ need to read data from some storage.
 ## Asynchronous model of the web
 
 The web has a variety of different storage options you could map to, such as in-memory storage (JS
-objects), [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), server-side storage, and a new [File System Access
-API](https://web.dev/file-system-access/).
+objects), [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage),
+[IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), server-side storage,
+and a new [File System Access API](https://web.dev/file-system-access/).
 
 However, only two of those APIs—the in-memory storage and the localStorage—can be used
 synchronously, and both are the most limiting options in what you can store and for how long. All
@@ -132,8 +133,8 @@ async function main() {
 }
 ```
 
-Even though it looks synchronous, under the hood each `await` is essentially syntax sugar
-for callbacks:
+Even though it looks synchronous, under the hood each `await` is essentially syntax sugar for
+callbacks:
 
 ```js
 function main() {
@@ -154,8 +155,8 @@ Thanks to the asynchronous nature of those steps, the original function can retu
 browser as soon as the I/O has been scheduled, and leave the entire UI responsive and available for
 other tasks, including rendering, scrolling and so on, while the I/O is executing in background.
 
-As a final example, even simple APIs like "sleep", which makes an application wait a specified number of
-seconds, are also a form of an I/O operation:
+As a final example, even simple APIs like "sleep", which makes an application wait a specified
+number of seconds, are also a form of an I/O operation:
 
 ```cpp
 #include <stdio.h>
@@ -180,8 +181,8 @@ console.log("B");
 but that's very inefficient, will block the entire UI and won't allow any other events to be handled
 meanwhile. Generally, don't do that in production code.
 
-Instead, a more idiomatic version of "sleep" in JavaScript would involve calling
-`setTimeout()`, and subscribing with a handler:
+Instead, a more idiomatic version of "sleep" in JavaScript would involve calling `setTimeout()`, and
+subscribing with a handler:
 
 ```js
 console.log("A");
@@ -190,10 +191,10 @@ setTimeout(() => {
 }, 1000);
 ```
 
-What's common to all these examples and APIs? In each case, the idiomatic code in the
-original systems language uses a blocking API for the I/O, whereas an equivalent example for the web
-uses an asynchronous API instead. When compiling to the web, you need to somehow transform between
-those two execution models, and WebAssembly has no built-in ability to do so just yet.
+What's common to all these examples and APIs? In each case, the idiomatic code in the original
+systems language uses a blocking API for the I/O, whereas an equivalent example for the web uses an
+asynchronous API instead. When compiling to the web, you need to somehow transform between those two
+execution models, and WebAssembly has no built-in ability to do so just yet.
 
 ## Bridging the gap with Asyncify
 
@@ -201,13 +202,14 @@ This is where [Asyncify](https://emscripten.org/docs/porting/asyncify.html) come
 compile-time feature supported by Emscripten that allows pausing the entire program and
 asynchronously resuming it later.
 
-{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/VSMrdTiQ7PubW6vfE6WZ.svg", alt="A call graph describing a JavaScript -> WebAssembly -> web API -> async task invocation, where Asyncify connects the result of the async task back into WebAssembly",
-width="800", height="200" %}
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/VSMrdTiQ7PubW6vfE6WZ.svg", alt="A call graph
+describing a JavaScript -> WebAssembly -> web API -> async task invocation, where Asyncify connects
+the result of the async task back into WebAssembly", width="800", height="200" %}
 
 ### Usage in C / C++ with Emscripten
 
-If you wanted to use Asyncify to implement an asynchronous sleep for the last example, you could do it
-like this:
+If you wanted to use Asyncify to implement an asynchronous sleep for the last example, you could do
+it like this:
 
 ```cpp
 #include <stdio.h>
@@ -224,15 +226,13 @@ async_sleep(1);
 puts("B");
 ```
 
-[`EM_JS`](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=em_js#c.EM_JS)
-is a macro that allows defining JavaScript snippets as if they were C functions. Inside, use a
-function
+[`EM_JS`](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=em_js#c.EM_JS) is a
+macro that allows defining JavaScript snippets as if they were C functions. Inside, use a function
 [`Asyncify.handleSleep()`](https://emscripten.org/docs/porting/asyncify.html#making-async-web-apis-behave-as-if-they-were-synchronous)
 which tells Emscripten to suspend the program and provides a `wakeUp()` handler that should be
 called once the asynchronous operation has finished. In the example above, the handler is passed to
-`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, you can call
-`async_sleep()` anywhere you want just like regular `sleep()` or any other synchronous
-API.
+`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, you can
+call `async_sleep()` anywhere you want just like regular `sleep()` or any other synchronous API.
 
 When compiling such code, you need to tell Emscripten to activate the Asyncify feature. Do that by
 passing `-s ASYNCIFY` as well as [`-s ASYNCIFY_IMPORTS=[func1,
@@ -249,9 +249,8 @@ emcc -O2 \
 This lets Emscripten know that any calls to those functions might require saving and restoring the
 state, so the compiler will inject supporting code around such calls.
 
-Now, when you execute this
-code in the browser you'll see a seamless output log like you'd expect, with B coming after a short
-delay after A.
+Now, when you execute this code in the browser you'll see a seamless output log like you'd expect,
+with B coming after a short delay after A.
 
 ```text
 A
@@ -261,8 +260,8 @@ B
 You can [return values from
 Asyncify](https://emscripten.org/docs/porting/asyncify.html#returning-values) functions too. What
 you need to do is return the result of `handleSleep()`, and pass the result to the `wakeUp()`
-callback. For example, if, instead of reading from a file, you want to fetch a number
-from a remote resource, you can use a snippet like the one below to issue a request, suspend the C code, and
+callback. For example, if, instead of reading from a file, you want to fetch a number from a remote
+resource, you can use a snippet like the one below to issue a request, suspend the C code, and
 resume once the response body is retrieved—all done seamlessly as if the call was synchronous.
 
 ```js
@@ -325,8 +324,8 @@ Okay, so this all works great in Emscripten. What about other toolchains and lan
 Say that you have a similar synchronous call somewhere in your Rust code that you want to map to an
 async API on the web. Turns out, you can do that too!
 
-First, you need to define such a function as a regular import via `extern` block (or your
-chosen language's syntax for foreign functions).
+First, you need to define such a function as a regular import via `extern` block (or your chosen
+language's syntax for foreign functions).
 
 ```rust
 extern {
@@ -362,9 +361,9 @@ Pass `--asyncify` to enable the transform, and then use `--pass-arg=…` to prov
 list of asynchronous functions, where the program state should be suspended and later resumed.
 
 All that's left is to provide supporting runtime code that will actually do that—suspend and resume
-WebAssembly code. Again, in the C / C++ case this would be included by Emscripten, but now you need custom
-JavaScript glue code that would handle arbitrary WebAssembly files. We've created a library just for
-that.
+WebAssembly code. Again, in the C / C++ case this would be included by Emscripten, but now you need
+custom JavaScript glue code that would handle arbitrary WebAssembly files. We've created a library
+just for that.
 
 You can find it on Github at
 [https://github.com/GoogleChromeLabs/asyncify](https://github.com/GoogleChromeLabs/asyncify) or npm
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+            .then(response => response.text())
+            .then(text => wakeUp(Number(text)));
+    });
+});
+puts("Getting answer...");
+int answer = get_answer();
+printf("Answer is %d\n", answer);
+```
+
+In fact, for Promise-based APIs like `fetch()`, you can even combine Asyncify with JavaScript
```
jpmedley, line 281:
```suggestion
In fact, for Promise-based APIs like `fetch()`, you can even combine Asyncify with JavaScript's
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+        return Number(text);
+    });
+});
+
+int answer = get_answer();
+```
+
+#### Awaiting complex values
+
+But this example still limits you only to numbers. What if you want to implement our original
```
jpmedley, line 302:
```suggestion
But this example still limits you only to numbers. What if you want to implement my original
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+example, where we tried to get a user's name from a file as a string? Well, you can do that too!
```
jpmedley, line 303:
```suggestion
example, where I tried to get a user's name from a file as a string? Well, you can do that too!
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+Emscripten provides a feature called
+[Embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html), that allows
```
jpmedley, line 306:
```suggestion
[Embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html) that allows
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+you to handle conversions between JavaScript and C++ values. It has support for Asyncify as well, so
+you can call `.await()` on external `Promise`s and it will act just like `await` in an async-await
```
jpmedley, line 308:
```suggestion
you can call `await()` on external `Promise`s and they will act just like `await` in async-await
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+```
+
+When using this method, you don't even need to pass `ASYNCIFY_IMPORTS` as a compile flag, as it's
+already included by default.
+
+Okay, so this all works great in Emscripten. What about other toolchains and languages?
+
+### Usage from other languages
+
+Let's say you have a similar synchronous call somewhere in our Rust code that you want to map to an
```
jpmedley, line 325:
```suggestion
Say that you have a similar synchronous call somewhere in your Rust code that you want to map to an
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+async API on the Web. Turns out, you can do that too!
+
+First of all, you need to define such a function as a regular import via `extern` block (or your
```
jpmedley, line 328:
```suggestion
First, you need to define such a function as a regular import via `extern` block (or your
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+      --pass-arg=asyncify-imports@env.get_answer \
+      [...]
+```
+
+Pass `--asyncify` to enable the transform, and then use `--pass-arg=…` to provide a comma-separated
+list of asynchronous functions, where the program state should be suspended and later resumed.
+
+All that's left is to provide supporting runtime code that will actually do that—suspend and resume
+our code. Again, in the C / C++ case this would be included by Emscripten, but now we need custom
+JavaScript glue code that would handle arbitrary WebAssembly files. We've created a library just for
```
jpmedley, line 366:
```suggestion
JavaScript glue code that can handle arbitrary WebAssembly files. We've created a library just for
```

jpmedley: 'We' is OK in this case.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+that.
+
+You can find it on Github at
+[https://github.com/GoogleChromeLabs/asyncify](https://github.com/GoogleChromeLabs/asyncify) or npm
+under the name [`asyncify-wasm](https://www.npmjs.com/package/asyncify-wasm)`.
```
jpmedley, line 371:
```suggestion
under the name [`asyncify-wasm`](https://www.npmjs.com/package/asyncify-wasm).
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+It simulates a standard [WebAssembly instantiation
+API](https://developer.mozilla.org/en-US/docs/WebAssembly), but under its own namespace. The only
+difference is that, under a regular WebAssembly API you can only provide synchronous functions as
+imports, while under the Asyncify wrapper, you can provide asynchronous imports as well:
+
+```js
+let … = await Asyncify.instantiateStreaming(fetch('app.wasm'), {
```
jpmedley, line 379: Please put an actual variable name here.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+            let text = await response.text();
+            return Number(text);
+        }
+    }
+});
+…
+await instance.exports.main();
+```
+
+Once you try to call such asynchronous function - like `get_answer()` in the example above - from
```
jpmedley, line 392:
```suggestion
Once you try to call such an asynchronous function - like `get_answer()` in the example above - from
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+the WebAssembly side, the library will detect the returned `Promise`, suspend and save the state of
+the WebAssembly application, subscribe to the promise completion, and later, once it's resolved,
+seamlessly restore the call stack and state and continue execution as if nothing has happened.
+
+Since any function in the module might make an asynchronous call, all the exports become potentially
+asynchronous, too, so they get wrapped as well. You might have noticed in the example above that you
```
jpmedley, line 398:
```suggestion
asynchronous too, so they get wrapped as well. You might have noticed in the example above that you
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+need to `await` the result of `instance.exports.main()` to know when the execution is truly
+finished.
+
+### How does this all work under the hood?
+
+When Asyncify detects a call to one of the `ASYNCIFY_IMPORTS` functions, it starts an asynchronous
+operation, saves the entire state of the application, including the call stack and any temporary
+locals, and later, when that operation is finished, restores all the memory and call stack and and
```
jpmedley, line 406:
```suggestion
locals, and later, when that operation is finished, restores all the memory and call stack and
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+resumes from the exact same place and with the exact same state as if the program has never stopped.
```
jpmedley, line 407:
```suggestion
resumes from the exact place and with the exact state as if the program has never stopped.
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+}
+if (mode == REWINDING) {
+    restoreLocals();
+    mode = NORMAL_EXECUTION;
+}
+puts("B");
+```
+
+Initially `mode` is set to `NORMAL_EXECUTION`. Correspondingly, the first time such transformed code
+is executed, only the part leading up to "async_sleep" will get evaluated. As soon as the
```
jpmedley, line 440:
```suggestion
is executed, only the part leading up to `async_sleep()` will get evaluated. As soon as the
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+asynchronous operation is scheduled, Asyncify saves all the locals, and unwinds the stack by
+returning from each function all the way to the top, this way giving control back to the browser
+event loop.
+
+Then, once `async_sleep()` resolves, Asyncify support code will change `mode` to `REWINDING`, and
+call our function again. This time, the "normal execution" branch is skipped - since it already did
```
jpmedley, line 446:
```suggestion
call the function again. This time, the "normal execution" branch is skipped - since it already did
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+the job last time and we want to avoid printing "A" twice - and instead it comes straight to the
```
jpmedley, line 447:
```suggestion
the job last time and I want to avoid printing "A" twice - and instead it comes straight to the
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+"rewinding" branch. Once it's reached, it restores all the stored locals, changes mode back to
+"normal" and continues the execution as if the code was never stopped in the first place.
```
jpmedley, line 449:
```suggestion
"normal" and continues the execution as if the code were never stopped in the first place.
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+### Transformation costs
+
+Unfortunately, Asyncify transform isn't completely free, since it has to inject quite a bit of
+supporting code for storing and restoring all those locals, navigating the call stack under
+different modes and so on. It tries to modify only functions marked as asynchronous on the command
+line, as well as any of their potential callers, but that still might result in ~50% overhead to the
```
jpmedley, line 456:
```suggestion
line, as well as any of their potential callers, but that still might result in approximately fifty percent overhead to the
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+uncompressed code size.
```
jpmedley, line 457: >fifty percent overhead to the uncompressed code size

Should this be 'fifty percent overhead **as compared** to the uncompressed code size'.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/Im4hmQOYRHFcsg8UTfTR.png", alt="A graph showing code size overhead for various benchmarks, from near-0% under fine-tuned conditions to over 100% in worst cases",
+width="800", height="494" %}
+
+It's not ideal, but in many cases acceptable when the alternative is not having the functionality
```
jpmedley, line 462:
```suggestion
This isn't ideal, but in many cases acceptable when the alternative is not having the functionality
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+Make sure to always enable optimizations for the final builds to avoid it going even higher. You can
+also check [Asyncify-specific optimization
+options](https://emscripten.org/docs/porting/asyncify.html#optimizing) to reduce the overhead by
+limiting transforms only to specified functions and/or only direct function calls. There is also a
+minor cost to runtime performance, but it's limited to the async calls themselves. However, compared
+to the cost of the actual work, it's usually negligible.
+
+## Real-world demos
+
+Now that we've looked at the simple examples, let's move on to more complicated scenarios.
```
jpmedley, line 474:
```suggestion
Now that you've looked at the simple examples, I'll move on to more complicated scenarios.
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+As mentioned in the beginning of the article, one of the storage options on the Web is an
+asynchronous [File System Access API](https://web.dev/file-system-access/). It provides access to a
+real host filesystem from a Web application.
+
+On the other hand, there is a de-facto standard—[WASI](https://github.com/WebAssembly/WASI) - for
```
jpmedley, line 480:
```suggestion
On the other hand, there is a de-facto standard called [WASI](https://github.com/WebAssembly/WASI)  for
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+WebAssembly I/O in the console and the server-side. It was designed as a compilation target for
+system languages, and exposes all sorts of filesystem and other operations in a traditional
```
jpmedley, line 482:
```suggestion
system languages, and exposes all sorts of file system and other operations in a traditional
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+synchronous form.
+
+What if we could map one to another? Then we could compile any application in any source language
+with any toolchain supporting the WASI target, and run it in a sandbox on the Web, while still
+allowing it to operate on real user files! With Asyncify, we can do just that.
```
jpmedley, line 487:
```suggestion
allowing it to operate on real user files! With Asyncify, you can do just that.
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+In this demo, I've compiled Rust [coreutils](https://github.com/RReverser/coreutils) crate with a
```
jpmedley, line 489: >I've compiled Rust crate

What?

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+few minor patches to WASI, passed via Asyncify transform and implemented asynchronous
+[bindings](https://github.com/GoogleChromeLabs/wasi-fs-access/blob/main/src/bindings.ts) from WASI
+to File System Access API on the JavaScript side. Once combined with xterm.js, this provides a
```
jpmedley, line 492: I'm not sure what this sentence is saying.

jpmedley, line 492: What's xterm.js? You haven't mentioned it before.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+realistic shell running in the browser tab and operating on real user files - just like an actual
+terminal.
+
+{% Video src="video/9oK23mr86lhFOwKaoYZ4EySNFp02/4244yB6c9RbMCjGjP8ZW.mp4" %}
+
+Check it out live at [https://wasi.rreverser.com/](https://wasi.rreverser.com/).
+
+It's not limited just to timers and filesystems, either. You can go further and use more niche APIs
```
jpmedley, line 500: Is 'It' referring to WASI or Asyncify?

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+devices right in the sandbox of a Web page.
+
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/2rscL8dyhOVMacuq54Ad.jpg", alt="Screenshot of libusb debug output on a Web page, showing information about the connected Canon camera",
+width="375", height="548" %}
+
+It's probably a story for another blog post though.
+
+Those examples demonstrate just how powerful Asyncify can be for bridging the gap and porting all
+sorts of applications to the Web, allowing you to gain cross-platform access, sandboxing, and better
+security, all without losing in functionality.
```
jpmedley, line 516:
```suggestion
security, all without losing functionality.
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+  - Finally, even printf is not guaranteed to print things to the console and might be redirected to
+    a file or a network location, in which case it would have to go via the same steps above.
+
+Long story short, I/O can be slow and you can't predict how long a particular call will take by a
+quick glance at the code. While that operation is running, your whole application will appear frozen
+and unresponsive to the user.
+
+This is not limited to C or C++ either. Most system languages present all the I/O in a form of
+synchronous APIs. For example, if you translate the example to Rust, the API might look simpler, but
+the same principles apply. You just make a call and synchronously wait for it to return the result,
```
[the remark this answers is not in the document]
RReverser: > I almost suggested 'procedures'. The point is that I think you need something more descriptive of activity than 'principles'.

How about "but the workflow is similar"?

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+while it performs all the expensive operations and eventually returns the result back to you in a
+single invocation:
+
+```rust
+fn main() {
+    let s = std::fs::read_to_string("name.txt");
+    println!("Hello, {}!", s);
```
RReverser, line 77: Huh, I wonder what happened here...

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+}
+```
+
+But what happens when you try to compile any of those samples to WebAssembly and translate them to
+the Web? Or, to provide a specific example, what could "file read" operation translate to? It would
```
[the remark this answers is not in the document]
RReverser: I think it's more of a "could" than "would", because there are many possibilities, and the one I'm describing is not the one currently used in toolchains.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+need to read data from some storage.
+
+## Asynchronous model of the Web
+
+The Web has a variety of different storage options you could map to, such as in-memory storage (JS
+objects), localStorage, IndexedDB, server-side storage, and a new [File System Access
```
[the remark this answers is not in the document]
RReverser: > Can you provide links to these as well? Is localStorage actually Window.localStorage?

I mean, it does live on `window`, but it's commonly accessed as a global `localStorage` object.

> I think this is the first time I've heard of an in-memory object instance referred to as storage.

It's somewhat common to use storage backed by memory (like filesystem emulation for Wasm like described in this article, or general RAM disks or other similar implementations).

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+the other options provide only asynchronous APIs.
+
+This is one of the core properties of executing code on the Web: any time-consuming operation, which
+includes any I/O, has to be asynchronous.
+
+The reason is that the Web is historically single-threaded, and any user code that touches the UI,
+has to run on the same thread as the UI. It has to compete with the other important tasks like
+layout, rendering and event handling for the CPU time. You wouldn't want a piece of JavaScript or
+WebAssembly to be able to start a "file read" operation and block everything else—the entire tab,
+or, in the past, the entire browser—for a range from milliseconds to a few seconds, until it's over.
```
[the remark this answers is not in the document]
RReverser: I don't think there are any browsers where this is still true... Last one was IE AFAIK.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+```js
+function main() {
+  return fetch("name.txt")
+    .then(response => response.text())
+    .then(name => console.log("Hello, %s!", name));
+}
+```
+
+From the desugared example, it's a bit clearer that it starts a request and subscribes to the
```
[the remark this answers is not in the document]
RReverser: I wasn't referring to `fetch()` but to the example itself. `fetch` on its own doesn't subscribe to the response, so updated version of the sentence would be incorrect.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+response with the first callback. Once the browser receives the initial response—just the HTTP
+headers—it asynchronously invokes this callback. The callback stars reading the body as text using
```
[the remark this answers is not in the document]
RReverser: It's important to explain why we need a separate step/callback, as I know this sometimes confuses readers less familiar with `fetch` - that reason is that HTTP headers arrive in a separate step from the actual body.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+#include <stdio.h>
+#include <unistd.h>
+// ...
+printf("A\n");
+sleep(1);
+printf("B\n");
+```
+
+Sure, you could translate it in a very straightforward manner that would block the current thread
+until the time expires…
```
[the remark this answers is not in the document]
RReverser: Do you think this is necessary? I was trying to make this look as a single uninterrupted sentence.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+```js
+console.log("A");
+for (let start = Date.now(); Date.now() - start < 1000;);
+console.log("B");
+```
+
+…and, in fact, that's exactly what Emscripten does in [its default implementation of
+"sleep",](https://github.com/emscripten-core/emscripten/blob/16d5755a3f71f27d0c67b8d7752f94844e56ef7c/src/library_pthread_stub.js#L47-L52)
+but that's very inefficient, will block the entire UI, won't allow any other events to get handled
+meanwhile, and, generally, don't do that in production code :)
```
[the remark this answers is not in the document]
RReverser: :(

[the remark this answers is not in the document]
RReverser: But, if we split it this way, then we probably need "and" instead of comma before "won't allow..." above.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+console.log("A");
+setTimeout(() => {
+    console.log("B");
+}, 1000);
+```
+
+What's common for all those different examples and APIs? In each case, the idiomatic code in the
+original systems language uses a blocking API for the I/O, whereas an equivalent example for the Web
+uses an asynchronous API instead. When compiling to the Web, you need to somehow transform between
+those two execution models, and WebAssembly has no built-in capability to do so just yet.
```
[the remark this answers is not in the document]
RReverser: Yes, there are plans and proposals.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+This is where [Asyncify](https://emscripten.org/docs/porting/asyncify.html) comes in. Asyncify is a
+compile-time feature supported by Emscripten, that allows pausing the entire program and
+asynchronously resuming it later.
+
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/VSMrdTiQ7PubW6vfE6WZ.svg", alt="A call graph describing a JavaScript -> WebAssembly -> Web API -> async task invocation, where Asyncify connects the result of the async task back into WebAssembly",
+width="800", height="200" %}
+
+### Usage in C / C++ with Emscripten
+
+If you wanted to use it to implement an asynchronous sleep for the last example, you could do it
```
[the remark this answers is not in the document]
RReverser: Asyncify mentioned in the previous sentence.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+which tells Emscripten to suspend the program and provides a `wakeUp()` handler that should be
+called when the asynchronous operation has finished. In the example above, we pass that handler to
+`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, we call
+the `async_sleep()` function anywhere we want just like regular `sleep()` or any other synchronous
+API.
+
+When compiling such code, you need to tell Emscripten to activate the Asyncify feature. Do that by
+passing `-s ASYNCIFY` as well as [`-s ASYNCIFY_IMPORTS=[func1,
+func2]`](https://emscripten.org/docs/porting/asyncify.html#more-on-asyncify-imports) with a
+comma-separated list of functions that might be asynchronous.
```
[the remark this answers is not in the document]
RReverser: Idk, I feel like for "array literal" I would need to clarify which language it's in, but it's not JS or JSON or anything like that... Maybe I still could, just not sure.

src/site/content/en/blog/asyncify/index.md
RReverser: Hm, isn't filesystem commonly written as a single word?

src/site/content/en/blog/asyncify/index.md
RReverser: Numeric versions seems more readable IMO, do we need to spell out "50%"?

src/site/content/en/blog/asyncify/index.md
RReverser: Now it's unclear which exact place / state it's referring to - something we pass as an argument or ..? IMO "same" is important to make it clear it's still the same place/state as before.

src/site/content/en/blog/asyncify/index.md
RReverser: 1) Isn't `.` better because it shows that it's a method not a bare function like in all other examples we've shown so far?
2) it's "it", as in `.await()`, not "they" as in "Promises".

src/site/content/en/blog/asyncify/index.md
RReverser: What's the question?

src/site/content/en/blog/asyncify/index.md
RReverser: Asyncify, the feature in general.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+  - webassembly
+  - file-system
+---
+
+The I/O APIs on the Web are asynchronous, but they're synchronous in most system languages. When
+compiling code to WebAssembly, you need to bridge one kind of APIs to another - and this bridge is
+Asyncify.
+
+---
+
```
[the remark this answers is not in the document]
jpmedley: Please reword it.

src/site/content/en/blog/asyncify/index.md
jpmedley: Is 'Rust crate' a typo?

src/site/content/en/blog/asyncify/index.md
jpmedley: Please link to it. 

I call this a broken conceptual link. The fix is either to make it an actual link or remove it completely.

src/site/content/en/blog/asyncify/index.md
jpmedley: Please reword this to make it clear.

src/site/content/en/blog/asyncify/index.md
RReverser: No, that's how they're called - Rust crates.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+## I/O in system languages
+
+Let's start by taking a look at a simple example in C. Let's say, you want to read user's name from
```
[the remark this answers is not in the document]
RReverser: The intent of the article is to be read as if I'm walking with the reader through the process where we try and make something work together. Hence most usages of "we" except for places where I explain what they might want / not want ("you" in those places).

Replacing all those "we" with "I" / "you" creates an unnecessary division and changes the tone of the article to something very different from the intended workshop-like walkthrough.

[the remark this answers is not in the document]
jpmedley: I get what you're trying to do, but 'we' is not absolutely necessary to creating a conversational tone. Addressing the reader directly ('you') is adequate. 

Also, this is part of the requirements for web.dev: https://web.dev/handbook/grammar/#instructions

src/site/content/en/blog/asyncify/index.md
RReverser:
```suggestion
    println!("Hello, {}!", s);
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+```rust
+fn main() {
+    let s = std::fs::read_to_string("name.txt");
+    println!("Hello, {}!", s);
+}
+```
+
+But what happens when you try to compile any of those samples to WebAssembly and translate them to
+the Web? Or, to provide a specific example, what could "file read" operation translate to? It would
+need to read data from some storage.
```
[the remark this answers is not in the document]
RReverser: Which part of the statement?

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
RReverser:
```suggestion
objects), [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage), [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), server-side storage, and a new [File System Access
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+once it's finished. Such callbacks are then executed as part of the browser's event loop. I won't be
+going into details here, but if you're interested in learning how event loop works under the hood,
+check out Jake Archibald's
+[article](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) which explains this
+topic in-depth.
+
+The short version is that the browser runs all the pieces of code in sort of an infinite loop, by
+taking them from the queue one by one. When some event is triggered, the browser queues the
+corresponding handler, and on the next loop iteration it's taken out from the queue and executed.
+This mechanism allows to simulate concurrency and run lots of parallel operations while still using
```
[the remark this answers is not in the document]
RReverser: The sentence doesn't sound correct after rewording (particularly, "and run").

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+  return fetch("name.txt")
+    .then(response => response.text())
+    .then(name => console.log("Hello, %s!", name));
+}
+```
+
+From the desugared example, it's a bit clearer that it starts a request and subscribes to the
+response with the first callback. Once the browser receives the initial response—just the HTTP
+headers—it asynchronously invokes this callback. The callback stars reading the body as text using
+`response.text()`, and subscribes to the result with another callback. Finally, once `fetch` has
```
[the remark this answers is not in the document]
RReverser: Is "via" meant to be instead of "using"? The proposed change would result in "using via", I don't think it's correct?

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
RReverser:
```suggestion
If you wanted to use Asyncify to implement an asynchronous sleep for the last example, you could do it
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+puts("A");
+async_sleep(1);
+puts("B");
+```
+
+Here, [`EM_JS`](https://emscripten.org/docs/api_reference/emscripten.h.html?highlight=em_js#c.EM_JS)
+is a macro that allows defining JavaScript snippets as if they were C functions. Inside, use a
+function
+[`Asyncify.handleSleep()`](https://emscripten.org/docs/porting/asyncify.html#making-async-web-apis-behave-as-if-they-were-synchronous)
+which tells Emscripten to suspend the program and provides a `wakeUp()` handler that should be
```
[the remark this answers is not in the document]
RReverser: There is now two mentions of Emscripten in a row, I'd need to replace second one with "it" to make it easier to read. Would it be better than the current sentence though?

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+called when the asynchronous operation has finished. In the example above, we pass that handler to
```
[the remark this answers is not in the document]
RReverser:
```suggestion
called once the asynchronous operation has finished. In the example above, the handler is passed to
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, we call
```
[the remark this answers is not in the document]
RReverser: "handler", mentioned 2 words ago.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+```text
+A
+B
+```
+
+You can [return values from
+Asyncify](https://emscripten.org/docs/porting/asyncify.html#returning-values) functions too. What
+you need to do is return the result of `handleSleep`, and pass the result to the `wakeUp()`
+callback. For example, if, instead of reading from a file, you want to fetch some numeric answer
+from a remote resource, you can use a snippet like below to issue a request, suspend the C code, and
```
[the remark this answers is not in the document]
RReverser: Maybe rather

```suggestion
from a remote resource, you can use a snippet like the one below to issue a request, suspend the C code, and
```

src/site/content/en/blog/asyncify/index.md
RReverser:
```suggestion
But this example still limits you only to numbers. What if you want to implement the original
```

src/site/content/en/blog/asyncify/index.md
RReverser:
```suggestion
const { instance } = await Asyncify.instantiateStreaming(fetch('app.wasm'), {
```

src/site/content/en/blog/asyncify/index.md
RReverser: Sgtm.

```suggestion
line, as well as any of their potential callers, but that still might result in ~50% overhead as compared to the
uncompressed code size.
```

src/site/content/en/blog/asyncify/index.md
RReverser: Is this resolved after linking Xterm or the issue is orthogonal to that one?

src/site/content/en/blog/asyncify/index.md
jpmedley: I'm not aware of a standard, but two words is more common web.dev than one word by about 10 to 1. Please change.

jpmedley:
@RReverser Thanks for your patience. There's one outstanding issue, then we're done.

RReverser:
> @RReverser Thanks for your patience. There's one outstanding issue, then we're done.

Thanks. Please also check the conversations Github has hidden under "[...] hidden conversation" - I've left some questions there in response to your comments.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,514 @@
...
+```js
+function main() {
+  return fetch("name.txt")
+    .then(response => response.text())
+    .then(name => console.log("Hello, %s!", name));
+}
+```
+
+From the desugared example, it's a bit clearer that it starts a request and subscribes to the
+response with the first callback. Once the browser receives the initial response—just the HTTP
```
jpmedley, line 145:
```suggestion
In this example, which is a bit clearer, a request is started and responses are subscribed to with the first callback. Once the browser receives the initial response—just the HTTP
```

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
jpmedley: I took another shot at this. It's in a different comment (because it affects more lines) so I'm closing this.

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
jpmedley: In English ellipses indicate breaks in quotes where some part has been excised. You're introducing a code sample. That requires a colon.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+```
+
+This lets Emscripten know that any calls to those functions might require saving and restoring the
+state, so the compiler will inject a supporting code around such calls.
+
+Now, when you execute this
+code in the browser you'll see a seamless output log like you'd expect, with B coming after a short
+delay after A.
+
+```text
```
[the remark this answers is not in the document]
jpmedley: Because `puts()` writes to stdout.

src/site/content/en/blog/asyncify/index.md
jpmedley: 1. This is the convention. I've never heard of it causing confusion.
2. My bad. Fixed.

src/site/content/en/blog/asyncify/index.md
jpmedley: I was trying to remove some wordiness. What about removing 'exact' instead?

src/site/content/en/blog/asyncify/index.md
jpmedley: A conditional clause requires subjunctive mood. https://grammar.collinsdictionary.com/us/easy-learning/conditional-clauses

src/site/content/en/blog/asyncify/index.md
jpmedley: I hypercorrect. The digits should have stayed, but rest needs to be spelled out. (This site uses a more formal writing style than you would use on your own blog.)

src/site/content/en/blog/asyncify/index.md
jpmedley: >50% overhead to the uncompressed code size

I'm having trouble working out what I'm getting 50% of. Do you mean that the asyncified code is 50% larger than the original before compression?

jpmedley (changes requested):
Sorry about the last review. There wasn't more than one. It clearly was not my day.

src/site/content/en/blog/asyncify/index.md
RReverser: Yes.

src/site/content/en/blog/asyncify/index.md
RReverser: O_o Okay...

src/site/content/en/blog/asyncify/index.md
RReverser: That works.

```suggestion
resumes from the same place and with the same state as if the program has never stopped.
```

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
RReverser: More precisely, to browser console or Node.js console when compiled with Emscripten as I do. Either way, it's a plain text and should be never highlighted, so `text` is more appropriate here.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,516 @@
...
+
+The reason is that the Web is historically single-threaded, and any user code that touches the UI,
+has to run on the same thread as the UI. It has to compete with the other important tasks like
+layout, rendering and event handling for the CPU time. You wouldn't want a piece of JavaScript or
+WebAssembly to be able to start a "file read" operation and block everything else—the entire tab,
+or, in the past, the entire browser—for a range from milliseconds to a few seconds, until it's over.
+
+Instead, code is only allowed to schedule an I/O operation together with a callback to be executed
+once it's finished. Such callbacks are then executed as part of the browser's event loop. I won't be
+going into details here, but if you're interested in learning how event loop works under the hood,
```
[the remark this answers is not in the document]
RReverser: `the` makes more sense in that case

```suggestion
going into details here, but if you're interested in learning how the event loop works under the hood,
```

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
RReverser: I see. Looking at the sentence, it seems like it became harder to read / more complex due to extra indirection than the original. Is this change necessary to accomodate the guidelines?

src/site/content/en/blog/asyncify/index.md
RReverser: I'd like to keep "de-sugared" here to reiterate that it's not a new example, but a de-sugared version of the previous one. How does this sound?

```suggestion
In this de-sugared example, which is a bit clearer, a request is started and responses are subscribed to with the first callback. Once the browser receives the initial response—just the HTTP
```

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
jpmedley: My suggestion was a little funky so I've edited it. The original is grammatically incorrect, so it has to change.

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
jpmedley: >it's a bit clearer that **it** starts

The problem I was trying to fix is that I don't know what the second 'it' refers to.

src/site/content/en/blog/asyncify/index.md
jpmedley: Your rewrite looks good.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,513 @@
...
+A
+B
+```
+
+You can [return values from
+Asyncify](https://emscripten.org/docs/porting/asyncify.html#returning-values) functions too. What
+you need to do is return the result of `handleSleep()`, and pass the result to the `wakeUp()`
+callback. For example, if, instead of reading from a file, you want to fetch a number from a remote
+resource, you can use a snippet like the one below to issue a request, suspend the C code, and
+resume once the response body is retrieved—all done seamlessly as if the call was synchronous.
```
jpmedley, line 261:
```suggestion
resume once the response body is retrieved—all done seamlessly as if the call were synchronous.
```

jpmedley (changes requested):

src/site/content/en/blog/asyncify/index.md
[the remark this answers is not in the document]
RReverser: That sounds better, added.

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,513 @@
...
+[`Asyncify.handleSleep()`](https://emscripten.org/docs/porting/asyncify.html#making-async-web-apis-behave-as-if-they-were-synchronous)
+which tells Emscripten to suspend the program and provides a `wakeUp()` handler that should be
+called once the asynchronous operation has finished. In the example above, the handler is passed to
+`setTimeout()`, but it could be used in any other context that accepts callbacks. Finally, you can
+call `async_sleep()` anywhere you want just like regular `sleep()` or any other synchronous API.
+
+When compiling such code, you need to tell Emscripten to activate the Asyncify feature. Do that by
+passing `-s ASYNCIFY` as well as [`-s ASYNCIFY_IMPORTS=[func1,
+func2]`](https://emscripten.org/docs/porting/asyncify.html#more-on-asyncify-imports) with a
+comma-separated list of functions that might be asynchronous.
```
RReverser, line 236:
```suggestion
func2]`](https://emscripten.org/docs/porting/asyncify.html#more-on-asyncify-imports) with an
array-like list of functions that might be asynchronous.
```

jpmedley (changes requested):

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,513 @@
...
+"rewinding" branch. Once it's reached, it restores all the stored locals, changes mode back to
+"normal" and continues the execution as if the code were never stopped in the first place.
+
+### Transformation costs
+
+Unfortunately, Asyncify transform isn't completely free, since it has to inject quite a bit of
+supporting code for storing and restoring all those locals, navigating the call stack under
+different modes and so on. It tries to modify only functions marked as asynchronous on the command
+line, as well as any of their potential callers, but that still might result in approximately 50 percent overhead to the
+uncompressed code size.
```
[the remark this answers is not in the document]
RReverser: Hm, it sounds fine to me, but then, admittedly, I don't have Anglophone ears so I trust your judgment :) If you have a better suggestion, I'd be happy to accept it to unblock this.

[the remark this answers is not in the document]
RReverser: Ping? This post has been in review for 2 months now, and if there are any specific last suggestions, I'm happy to accept them, just to make sure it's published.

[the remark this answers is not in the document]
kaycebasques: I also don't quite understand where the performance hit will show up but I think the general message is clear enough: there may be performance costs to using Asyncify

kaycebasques (approved):

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,513 @@
+---
+title: Using asynchronous web APIs from WebAssembly
+description: Learn how to invoke asynchronous web APIs when compiling traditionally synchronous languages to WebAssembly.
+date: # TODO
```
kaycebasques, line 4:
```suggestion
date: 2021-04-22
```

src/site/content/en/blog/asyncify/index.md
```diff
@@ -0,0 +1,513 @@
...
+and a new [File System Access API](https://web.dev/file-system-access/).
+
+However, only two of those APIs—the in-memory storage and the localStorage—can be used
+synchronously, and both are the most limiting options in what you can store and for how long. All
+the other options provide only asynchronous APIs.
+
+This is one of the core properties of executing code on the web: any time-consuming operation, which
+includes any I/O, has to be asynchronous.
+
+The reason is that the web is historically single-threaded, and any user code that touches the UI,
```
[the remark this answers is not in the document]
RReverser: Makes sense, thanks. I usually use the same rule, but something went wrong here 😅

src/site/content/en/blog/asyncify/index.md
RReverser, line 452: @kaycebasques Just one more attempt, if this doesn't work either, I'll leave it as-is. What do you think?

```suggestion
different modes and so on. It tries to modify only functions marked as asynchronous on the command
line, as well as any of their potential callers, but the code size overhead still adds up to approximately 50% before compression.
```

PR #4993, opened 2021-03-30
Add rel=alternate hreflang links

### LINKED ISSUE #4638: INTL: Add canonical and alternate tags for original and translated posts

devnook:
Fixes #4638


src/site/_includes/components/Meta.js
```diff
@@ -130,6 +135,31 @@ module.exports = (locale, page, collections, renderData = {}) => {
...
+    if (!pageUrl) {
+      return;
+    }
+    const url = pageUrl.startsWith('/i18n/')
+      ? pageUrl.replace(i18nRegex, '')
+      : pageUrl;
+
+    // Find i18n equivalents of the current url and heck if they exist.
+    const langhrefs = i18nLocales
+      .map((locale) => [locale, `/i18n/${locale}${url}`])
```
robdodson, line 148: should this line use `path.join('i18n', locale, url)` to make it a little more resilient? Just in case `locale` or `url` ever changes and no longer has a slash

src/site/_includes/components/Meta.js
```diff
@@ -130,6 +135,31 @@ module.exports = (locale, page, collections, renderData = {}) => {
...
+      .filter((langhref) => findByUrl(langhref[1]))
```
robdodson, line 149: I was a little confused by this because I thought calling `findByUrl` on something that hasn't been translated would either throw an error or return undefined. I didn't realize that Array.filter() coerces undefined to `false`, so it's being used intentionally here.

Would you mind adding a comment above line 149 saying something like "If a translated version of the page does not exist then findByUrl will return undefined which coerces to false and removes the untranslated page from the array. By the end we have an array that only contains translated pages."

robdodson (approved):
LGTM with two small comments.

jpmedley:
@devnook We went to a system some time ago where DevRel submitters merge their own PRs. Feel free to merge this.

robdodson:
@jpmedley before merging she should address the nits in the comments. I think Fri-Mon have been holidays in Zurich though.

PR #5008, opened 2021-03-31
First pass at using strict CSP

adamraine:
Moving https://github.com/GoogleChrome/web.dev/pull/4959 here so we aren't using a fork anymore.

-----

We are working on a new Lighthouse audit to recommend using a strict CSP. It is important to practice what we preach, so this PR aims to introduce a strict CSP to web.dev.

Following our own advice [here](https://web.dev/strict-csp/#step-1:-decide-if-you-need-a-nonce-or-hash-based-csp), CSP hashes seemed like the best approach because web.dev is served statically. There are a few complications, but overall this solution is less destructive than an approach using nonces.


.eleventy.js
```diff
@@ -272,6 +288,13 @@ module.exports = function (config) {
   // Make .yml files work in the _data directory.
   config.addDataExtension('yml', (contents) => yaml.safeLoad(contents));
 
+  // Make CSP hashes accessible to firebase config.
+  if (isProd) {
+    config.on('afterBuild', () => {
+      fs.writeFileSync('script-hash-list.json', JSON.stringify([...hashList]));
```
adamraine, line 294: Could we generate `firebase.json` in this function to eliminate the need for a temporary file and changes to `npm run dev`?

package.json
```diff
@@ -16,7 +16,7 @@
     "debug:test:lib": "TEST_MODE=debug npm-run-all clean rollup --parallel watch:rollup karma",
     "debug:test:unit": "node --inspect-brk ./node_modules/.bin/mocha ./test/unit",
     "deploy": "ELEVENTY_ENV=prod npm run build && node index-algolia.js && firebase deploy --non-interactive",
-    "dev": "npm-run-all clean rollup gulp --parallel watch:rollup watch:eleventy watch:gulp firebase-config start",
+    "dev": "npm-run-all clean rollup gulp eleventy --parallel watch:rollup watch:gulp firebase-config start",
```
adamraine, line 19: Change ensures hashes are generated before they are used in `firebase-config.js`.

package.json
```diff
@@ -68,7 +68,7 @@
     "wicg-inert": "^3.0.1"
   },
   "devDependencies": {
-    "@11ty/eleventy": "^0.11.0",
+    "@11ty/eleventy": "^0.11.1",
```
adamraine, line 71: Need access to `config.on('afterBuild', () => ...)`

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,34 @@
...
+  s.src = url;
+  if (type) s.type = type;
+  if (type === 'module') {
+    const pre = document.createElement('link');
+    pre.rel = 'modulepreload';
+    pre.href = url;
+    document.head.appendChild(pre);
+    window.addEventListener('DOMContentLoaded', () => {
+      document.head.appendChild(s);
+    })
```
adamraine, line 47: I was experiencing errors on `/live/`, my hypothesis was that modules are not deferred when they are dynamically loaded like this.

Changing this code to load the modules after `DOMContentLoaded` appears to have fixed the issues.

adamraine (commented):
Moving comments here. Also some Lighthouse performance questions https://github.com/GoogleChrome/web.dev/pull/4959#issuecomment-806182206.

src/site/_includes/partials/head.njk
samthor: That seems right. I'd probably add a comment explaining it since the logic might be confusing to a future reader.

```
{# We use DOMContentLoaded as the loader script is running sync, and inserting a module script here doesn't defer. This brings back normal type="module" behavior. #}
```

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,34 @@
 <link rel="icon" type="image/png" sizes="16x16" href="/images/favicon-16x16.png">
 <link rel="apple-touch-icon" sizes="180x180" href="/images/apple-touch-icon.png">
 <link rel="mask-icon" color="#0054ff" href="/images/safari-pinned-tab.svg">
-<script type="module" src="{{ helpers.hashForProd('/js/app.js') }}"></script>
+{% set scriptLoader %}
+function loadScript(url, type) {
+  const s = document.createElement('script');
+  s.src = url;
+  if (type) s.type = type;
```
samthor, line 39: nit: `{}`'s please

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,34 @@
...
+  if (type === 'module') {
+    const pre = document.createElement('link');
+    pre.rel = 'modulepreload';
+    pre.href = url;
+    document.head.appendChild(pre);
```
samthor, line 44: nit: you can use `.append` instead of `.appendChild`. (If you're using arrow functions, you can definitely use the more modern append).

.eleventy.js
samthor: We could, but, ordering in build systems like this has always been a pain.

I don't mind your `npm run dev` change. My main thing would be potentially to put this file inside the `dist/` folder, so it's clear it's an intermediate file and not something real that needs to be committed forever.

src/site/_includes/partials/head.njk
```diff
@@ -36,17 +36,20 @@
...
+  if (type) {
+    s.type = type;
+  }
   if (type === 'module') {
     const pre = document.createElement('link');
     pre.rel = 'modulepreload';
     pre.href = url;
-    document.head.appendChild(pre);
+    document.head.append(pre);
+    {# We use DOMContentLoaded as the loader script is running sync, and inserting a module script here doesn't defer. This brings back normal type="module" behavior. #}
```
samthor, line 47: This looks like it shouldn't work because it's in the middle of JS but I guess it does, the `{#` rule wins.

This is a super minor nit but could this be a JS comment, since we're minifying now and it'll be removed at that point? (It might be useful to disable minification in dev... which actually I might suggest.)

samthor (approved):

package.json
adamraine: So it looks like this upgrade is causing issues with `pngquant`
```
[11:49:07] Using gulpfile ~/src/web.dev/gulpfile.js
[11:49:07] Starting 'build'...
[11:49:07] Starting 'copy-global-images'...
[11:49:07] Starting 'copy-misc'...
[11:49:07] Starting 'copy-content-assets'...
[11:49:07] Starting 'copy-fonts'...
[11:49:07] Starting 'sass'...
[11:49:08] Finished 'sass' after 1.1 s
[11:49:08] 'copy-global-images' errored after 1.45 s
[11:49:08] Error in plugin "gulp-imagemin"
Message:
    
Details:
    errno: -2
    code: ENOENT
    syscall: spawn /Users/asraine/src/web.dev/node_modules/pngquant-bin/vendor/pngquant
    path: /Users/asraine/src/web.dev/node_modules/pngquant-bin/vendor/pngquant
    spawnargs: -,--quality,100-100
    stdout: 
    stderr: 
    failed: true
    signal: null
    cmd: /Users/asraine/src/web.dev/node_modules/pngquant-bin/vendor/pngquant - --quality 100-100
    timedOut: false
    killed: false
    fileName: /Users/asraine/src/web.dev/src/images/android-chrome-192x192.png
    domainEmitter: [object Object]
    domainThrown: false

[11:49:08] 'build' errored after 1.46 s
[11:49:08] The following tasks did not complete: copy-misc, copy-content-assets, copy-fonts
[11:49:08] Did you forget to signal async completion?
```
I was using `npm i` to upgrade the package. Is there another way I should be doing it?

robdodson (changes requested):
Just want to make sure I get a chance to review so adding a Request change flag.

package.json
adamraine: I think this was an issue, with how I was bumping the version. https://github.com/GoogleChrome/web.dev/commit/83fd829250eb364d4d73e67ed5ffc571e547722d appears to have fixed this.

package.json
robdodson: note to self: we will have to change this, otherwise it will break eleventy in dev mode (it won't rebuild when you save).

.eleventy.js
```diff
@@ -174,6 +178,16 @@ module.exports = function (config) {
...
+  config.addFilter('minifyJs', minifyJs);
+
+  const hashList = new Set();
+  config.addFilter('cspHash', (raw) => {
+    if (isProd) {
+      const hash = `'sha256-${sha256base64(raw)}'`;
+      hashList.add(hash);
+    }
+    return raw;
+  });
```
robdodson, line 190: Could we move this into its own filter file, similar to the other filters?

src/site/_includes/partials/analytics.njk
```diff
@@ -11,4 +11,5 @@ ga('set', 'anonymizeIp', true);
 ga('set', 'page', window.location.pathname);
 ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version}}');
 ga('send', 'pageview');
-</script>
+{% endset %}
+<script>{{ analyticsScript|minifyJs|cspHash|safe }}</script>
```
robdodson, line 15:
```suggestion
<script>{{ analyticsScript | minifyJs | cspHash | safe }}</script>
```

style nit

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,37 @@
 <link rel="icon" type="image/png" sizes="16x16" href="/images/favicon-16x16.png">
 <link rel="apple-touch-icon" sizes="180x180" href="/images/apple-touch-icon.png">
 <link rel="mask-icon" color="#0054ff" href="/images/safari-pinned-tab.svg">
-<script type="module" src="{{ helpers.hashForProd('/js/app.js') }}"></script>
+{% set scriptLoader %}
```
robdodson, line 35: It might be good to have a big comment here explaining what is going on and why we're doing it.

src/site/_includes/partials/head.njk
```diff
@@ -32,15 +32,37 @@
...
+  loadScript('{{ helpers.hashForProd('/js/default.js') }}', 'module');
 {% endfor %}
 {% if process.env.ELEVENTY_ENV === 'prod' %}
-  <script async src="https://www.google-analytics.com/analytics.js"></script>
-{% endif %}
\ No newline at end of file
+  loadScript('https://www.google-analytics.com/analytics.js', null);
+{% endif %}
+{% endset %}
+<script>{{ scriptLoader|minifyJs|cspHash|safe }}</script>
```
robdodson, line 68:
```suggestion
<script>{{ scriptLoader | minifyJs | cspHash | safe }}</script>
```

style nit

robdodson (approved):

robdodson:
I left it so the CSP stuff only happens during a prod build. This is because we're currently only doing CSP reporting—we're not actually blocking scripts.

If we switch to blocking scripts then I think it would be good to consider enforcing CSP in dev mode so we don't accidentally ship a CSP breaking change.

robdodson:
ok so like immediately after I wrote this Connor pointed out that it does still console.warn if you've broken CSP in reporting mode—which seems like useful info to have during development! So I've created https://github.com/GoogleChrome/web.dev/pull/5184 to run it in dev.

PR #5013, opened 2021-04-01
[Notification Triggers article] Update article now that the OT is over

tomayac:

- Update article now that the OT is over.


rachelandrew (approved):
lgtm :)

PR #5024, opened 2021-04-06
Add audio player for courses section

### LINKED ISSUE #4581: Audio player for courses
The design for this is quite fancy. I'm curious if we want to go that elaborate since other courses likely will not have a big audio counterpart and we don't have audio used elsewhere on the site... Originally I thought we might get away with an `<audio>` element. Need to see what folks think in our next sync.

robdodson:
Fixes #4581

Adds an audio player macro and an audio FAB macro. The audio FAB is a custom element which uses an intersection observer to show itself if the audio player is offscreen. Clicking the audio FAB will scroll the page back up to the audio player.

### DIFF
--- a/src/site/_includes/macros/audio-fab.njk
+++ b/src/site/_includes/macros/audio-fab.njk
@@ -1,0 +1,16 @@
+{% macro audioFab(thumbnail, target="audio-player") %}
+<web-audio-fab target="{{ target }}">
+  <button class="web-audio-fab__button button" data-playing aria-label="Back to audio player">
+    {% Img class="web-audio-fab__thumbnail", src=thumbnail, alt="", width="56", height="56" %}
+    <div class="web-audio-fab__equalizer-container">
+      <svg class="web-audio-fab__equalizer" xmlns="http://www.w3.org/2000/svg" width="13" height="8" viewBox="0 0 13 8">
+        <g>
+          <rect class="web-audio-fab__equalizer-left" width="3" height="8" />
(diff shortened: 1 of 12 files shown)


sfddddh (approved):

sfddddh (approved):

src/site/_includes/macros/audio-fab.njk
```diff
@@ -0,0 +1,16 @@
+{% macro audioFab(thumbnail, target="audio-player") %}
+<web-audio-fab target="{{ target }}">
+  <button class="web-audio-fab__button button" data-playing aria-label="Back to audio player">
+    {% Img class="web-audio-fab__thumbnail", src=thumbnail, alt="", width="56", height="56" %}
+    <div class="web-audio-fab__equalizer-container">
+      <svg class="web-audio-fab__equalizer" xmlns="http://www.w3.org/2000/svg" width="13" height="8" viewBox="0 0 13 8">
```
MichaelSolati, line 6: No pressure, but do we want to start to handle svgs like we do on dcc, `{% include 'icons/audio-fab.svg' %}`

MichaelSolati (approved):

src/site/_includes/macros/audio-fab.njk
robdodson: In general yes, and I've been using the icons macro elsewhere in web.dev.

In this case I didn't because I wanted to apply classes to the svg elements themselves and the macro doesn't support that.

PR #5040, opened 2021-04-07
feat(updated): sort posts by updated and show updated date, fixes #2283

### LINKED ISSUE #2283: Have a way to promote older items to the front of the blog page.
There are valid reasons why an old article might need to be on the main page of web.dev/blog. Here are a few use cases:
* New platform features are often dribbled out over multiple releases. In many cases the changes are small enough to be added to an existing article without going over the web.dev word limit. For the sake of web developers it would be better if all the information about a new feature in the same article. At the same time, *it would be nice if an updated article could be shown on the mian page without altering the URL or the publish date, which would affect site stats.*
* Imagine a feature we implemented two years ago. Today the last major browser to add support was released. It's to our advantage to republicize the article we wrote two years ago.
* Imagine we've polished off the last feature of a platform API we started several years ago and we've timed it for IO or CDS. Furthermore, imagine that we've also told attendees articles about our announcements will be posted on the blog. 

A flag in article metadata to indicate that something should be promoted to the main page of the feed. It's exact position could be based on the `updated:` field.

Changing the date affects article statistics as does changing the URL. Splitting related content is bad for readers.

**Additional context**
In discussing this with various people over the last few days, it was suggested that promoting/featuring an old article has minimal value because stats show that more people go to search than they do the landing page of the blog. I suggest that this is because we are not using or promoting the landing page to our full benefit and in the nature of landing pages. 

For example, the beta release post I write every six weeks could include a line such as "For information on these items and other platform news visit our blog" (or something like that). We also tweet the blog post to web developers, but the blog post is intended for the tech press. Would it make more sense to tweet the blog every six weeks?

WebFu had a similar feature, though it was admittedly underused.

MichaelSolati:
Fixes #2283

- Prefer sorting of posts in author and tag collections by their `updated` date rather than their posted date.
- Display the updated date on cards.
- Add the term "Updated" to `i18n.common`.

### DIFF
--- a/src/site/_collections/authors.js
+++ b/src/site/_collections/authors.js
@@ -17,7 +17,7 @@
 const authorsData = require('../_data/authorsData.json');
 const {livePosts} = require('../_filters/live-posts');
 const setdefault = require('../_utils/setdefault');
-const {sortByDate} = require('../_utils/sort-by-date');
+const {sortByUpdated} = require('../_utils/sort-by-updated');
 
 /** @type Authors */
 let processedCollection;
@@ -80,7 +80,7 @@ module.exports = (collections) => {
 
   if (collections) {
     // Find all posts, sort and key by author. Don't yet filter to live posts.
-    allPosts = collections.getFilteredByGlob('**/*.md').sort(sortByDate);
+    allPosts = collections.getFilteredByGlob('**/*.md').sort(sortByUpdated);
   }
 
   const authorsPosts = findAuthorsPosts(allPosts);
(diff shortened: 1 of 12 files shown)


src/site/_collections/authors.js
```diff
@@ -17,6 +17,7 @@
 const authorsData = require('../_data/authorsData.json');
 const {livePosts} = require('../_filters/live-posts');
 const setdefault = require('../_utils/setdefault');
+const {sortByDate} = require('../_utils/sort-by-date');
```
robdodson, line 20: Should this function be renamed to something like `sortByUpdated` as that's the primary thing it's trying to sort on. It only uses `date` as a fallback.

robdodson (approved):
LGTM with one comment

src/site/_collections/authors.js
MichaelSolati: I'll rename it then merge

PR #5061, opened 2021-04-09
Update nginx-sxg-module to latest versions

twifkak:
Update instructions to download the latest releases of nginx and
nginx-sxg-module, and remove the extra `--without-http_rewrite_module
--with-http_ssl_module` arguments as those are not necessary.

/cc @kumagi @khempenius


twifkak:
Nevermind, closing this because nginx is particular about [version mismatches](https://www.nginx.com/blog/failed-nginx-plus-upgrade-module-version-x-instead-of-y/).

PR #5063, opened 2021-04-10
Add course card to /learn/ page

### LINKED ISSUE #4577: Three column layout for course cards on /learn
We'll need a layout that can support up to three columns of cards on the learn page. I'd need to research if flexbox or grid would be preferable here—maybe we ask Adam/Una/Andy :D

In addition, these cards are going to need to know how many pages are in each course so they can display it on the card.
image

IMO it's probably best for this to be an attribute that gets set on the card. Something like:

```html
<div class="course-card" data-resources="30">
```

Another question worth thinking through is how we want to do the actual course titles. The design has some fancy gradients but I'm wondering if we want to try to do it all in CSS (since each course might be different) or do the title as SVG with alternative text?

robdodson:
Fixes #4577 

- Adds a three card layout to the /learn/ page for course cards. The appearance of these cards is a bit different from the existing grid items because if there's only one course it should span 50% of the available space so I added a configurable `auto-fit-grid` utility class.
- Made a lot of changes to how we were displaying the SVGs for the course background and title. Previously I was trying to do this by using gradient text for the course logo and svg for the background. But taking another look a the mocks I realized it wouldn't be possible for every course title to use the same gradient text style, so instead I switched to using SVG for the course title as well.
- Added a new `svg` macro that's like a lower level version of the `icon` macro. The `svg` macro just inlines an svg from anywhere so long as you provide a file path relative to the _include directory.

Since we're using SVG for the logo text I realized I need to fix an issue in the webdev-infra repo https://github.com/GoogleChrome/webdev-infra/issues/2 but that shouldn't block this PR.


jamescrush123:
https://www.420primeshop.com/

src/site/_data/courses/css/meta.yml
```diff
@@ -1,3 +1,9 @@
+# Used to organize the courses on the /learn page
+date: 2021-05-01
+url: /learn/css/
+draft: true
 title: i18n.courses.learn_css.title
-gradient: "96.46deg, #B323F4 5.23%, #2BA1FF 33.16%, #02E895 68.79%, #6DDB00 90.93%"
-background: "url('/images/courses/learn-css.svg')"
+description: i18n.courses.learn_css.description
+# This path should be relative to the _include folder.
+logo: '../../../images/courses/css/logo.svg'
```
MichaelSolati, line 8: Do these work appropriately? I ask cause if they're like the hero images then it should be in the bucket. i doubt there is an issue because the imgix stuff would have thrown an error, but id check to be sure.

src/site/_includes/macros/icon.njk
```diff
@@ -1,10 +1,25 @@
 {#
   Finds an icon in the _includes/icons folder.
   Adds a class="icon" to the icon, as well as any additional classes passed in.
+  @param {string} name - The filename for the icon, minus the .svg extension
```
MichaelSolati, line 4: I like this commenting params in the njk 👍

MichaelSolati (approved):

src/site/_data/courses/css/meta.yml
robdodson: yeah right now the SVG are local. Let's do a follow up to consider if we want ot CDN them

PR #5070, opened 2021-04-12
Added information about who runs the back-end FLoC service

samdutton:
Also added line breaks to 'How does ...' section.

@tomayac fyi


mihajlija (approved):

tomayac (approved):
LGTM

PR #5119, opened 2021-04-20
Add progress tracking to course drawer

### LINKED ISSUE #4601: Track user state in Courses
After a user has viewed a page we need to show that in the side nav using a check mark. Likely we can do this with localstorage.

devnook:
Fixes #4601

adds a data-complete attribute to completed course modules. 
This does not touch the css as I need to catch up with Rob on this.


src/lib/components/DrawerCourse/index.js
```diff
@@ -0,0 +1,53 @@
...
+  render() {
+    return html`${this.childElements}`;
+  }
+
+  onStateChanged({courseProgress}) {
+    this.courseProgress = courseProgress;
+  }
+}
+
+customElements.define('drawer-course', DrawerCourse);
```
robdodson, line 53: I think we're still using the `web-` prefix for our custom elements so we may want to do that here as well and call it `web-drawer-course`

src/lib/components/DrawerCourse/index.js
```diff
@@ -0,0 +1,53 @@
...
+  static get properties() {
+    return {
+      current: {type: String},
+    };
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    markAsCompleted(this.current);
```
robdodson, line 35: nit. To fix typescript complaining about `this.current` being undefined I think you can set it in the constructor.

```
constructor() {
  this.current = '';
}
```

src/lib/components/DrawerCourse/index.js
```diff
@@ -0,0 +1,53 @@
...
+      if (this.courseProgress.indexOf(child.getAttribute('href')) > -1) {
+        child.setAttribute('data-complete', 'true');
+      }
+      return child;
+    });
+  }
+
+  render() {
+    return html`${this.childElements}`;
+  }
```
robdodson, line 46: I think you can get rid of this call to render because you're already updating the children in connectedCallback

src/lib/components/DrawerCourse/index.js
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {BaseStateElement} from '../BaseStateElement';
+import {html} from 'lit-element';
+import {markAsCompleted} from '../../actions';
```
robdodson, line 19: I think we may be able to simplify this since we don't currently have a use case for sharing the course state with the rest of the site.

So I'd suggest removing the changes to actions.js and store.js and making this element extend from BaseElemetn instead of BaseStateElement. In this element's constructor you can run the steps to read the cookies and set this.courseProgess. Then in connectedCallback you can have it mutate its children.

I'm leaning toward keeping things as simple as possible until we have an actual use case for shared state.

devnook:
ptal

robdodson (approved):

robdodson:
looks good! I think we can merge this

PR #5120, opened 2021-04-20
Add badge for new content in courses

### LINKED ISSUE #4792: "New"/"Updated" indicators for course modules
Allow for a "New" badge to distinguish newly added modules in web.dev Courses, and "Updated" badge for updated modules.

devnook:
Fixes #4792

### DIFF
--- a/src/site/_data/i18n/courses.yml
+++ b/src/site/_data/i18n/courses.yml
@@ -3,6 +3,3 @@ learn_css:
     en: Learn CSS!
   description:
     en: "Let’s gleefully breakdown complex aspects of CSS into digestible episodes covering everything from accessibility to z-index."
-
-new:
-  en: new
--- a/src/site/_data/i18n/courses.yml
+++ b/src/site/_data/i18n/courses.yml
@@ -3,3 +3,6 @@ learn_css:
     en: Learn CSS!
   description:
     en: "Let’s gleefully breakdown complex aspects of CSS into digestible episodes covering everything from accessibility to z-index."
+
+new:
+  en: new
--- a/src/styles/components/_drawer-course.scss
+++ b/src/styles/components/_drawer-course.scss
@@ -72,6 +72,17 @@
       background: $BLUE_50;
       color: $WEB_PRIMARY_COLOR;
     }
+
+    &[data-new='true']::after {
+      align-self: normal;
+      background: $WEB_PRIMARY_COLOR;
+      border-radius: 50%;
+      content: '';
+      display: inline-block;
+      height: 0.5em;
+      margin-left: 0.2em;
+      width: 0.5em;
+    }
   }
 
   &__link-counter {
--- a/src/styles/components/_drawer-course.scss
+++ b/src/styles/components/_drawer-course.scss
@@ -72,27 +72,6 @@
       background: $BLUE_50;
       color: $WEB_PRIMARY_COLOR;
     }
-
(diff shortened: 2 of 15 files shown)


robdodson:
This looks good. I sent a quick email to Jiwoong and Una/Adam to get their thoughts. Folks may request that we add the word `New` to the badge, but that's the only thing that popped to mind for me.

devnook:
I added NEW keyword and updated dot, as per design discussion.

src/styles/components/_drawer-course.scss
```diff
@@ -82,6 +82,19 @@
...
+
+    &::after {
+      display: inline-block;
+      height: 0.5em;
+      margin-left: 0.6em;
+    }
+
+    &[data-badge='new']::after {
+      color: $WEB_PRIMARY_COLOR;
+      content: 'NEW';
```
robdodson, line 94: Could we use https://developer.mozilla.org/en-US/docs/Web/CSS/attr() to set the content here? Maybe the markup would look like:

```
<a data-new data-new-label="{{ 'i18n.courses.new' | i18n(locale) }}">
```

test/unit/src/site/_filters/is-new-content.js
```diff
@@ -0,0 +1,22 @@
+const assert = require('assert');
```
robdodson, line 1: yay tests!

src/site/_data/i18n/courses.yml
```diff
@@ -3,3 +3,6 @@ learn_css:
     en: Learn CSS!
   description:
     en: "Let’s gleefully breakdown complex aspects of CSS into digestible episodes covering everything from accessibility to z-index."
+
+new:
+  en: new
```
robdodson, line 8: maybe we should put this in i18n.common since it seems like something we may use elsewhere on the site

robdodson:
From a UX perspective, it feels like the New badge should be hidden after I visit a module.

image

Could we update the css so if something is completed by the user it hides the new badge?

src/styles/components/_drawer-course.scss
devnook: done

src/site/_data/i18n/courses.yml
devnook: done

devnook:
PTAL

robdodson (approved):

robdodson (changes requested):
There was a merge conflict so I pulled the PR down to rebase it and now I'm noticing some small things that need fixing.

robdodson:
Ok this started as a small change and become something bigger 😅  Initially I was just doing a rebase and wanted to fix the background color on the links, but then I noticed there was a lot of FOUC because an element will render the `NEW` badge and then remove it when the completed attribute is set. It looked a bit odd so I decided to see if we could fix it by inlining the element.

So the changes are:

- Adds an `inlinePageScripts` njk property. This lets a layout inline custom elements if they need to render to avoid FOUC
- Refactors the course-links element so it doesn't have any dependencies. Instead we inline it so it can set attributes on its children during initial render and avoid FOUC.
- Adds analytics to course-links so it will fire for every 10% that a user has completed.
- Adds a course key to the course_progress localstorage so it can support multiple courses.

MichaelSolati (approved):

devnook (commented):
lgtm

PR #5123, opened 2021-04-20
Update gitlocalize link

devnook:


robdodson (approved):

robdodson:
@devnook I think the PR needs a quick rebase but it should be good to go after that.

PR #5124, opened 2021-04-20
build(staging): add staging environment

MichaelSolati:


src/site/content/en/learn/css/css.11tydata.js
```diff
@@ -7,7 +7,10 @@ module.exports = function () {
     projectKey: 'css',
     eleventyComputed: {
       permalink: (data) => {
-        if (process.env.ELEVENTY_ENV === 'prod') {
+        if (
+          process.env.ELEVENTY_ENV === 'prod' ||
+          process.env.ELEVENTY_ENV === 'staging'
```
robdodson, line 12: We only want to hide these on prod. They should be allowed for staging

src/site/_filters/strip-query-params-dev.js
```diff
@@ -15,9 +15,10 @@
  */
```
robdodson, line 15: Is this filter actually used?

robdodson (approved):

PR #5137, opened 2021-04-21
build(git): only deploy site if it's a new commit, fixes #3655

### LINKED ISSUE #3655: Only deploy if HEAD is different from what's currently deployed
We recently switched to deploying the site hourly. This is great, but it also means we're often deploying the same version of the site.

We should only deploy if HEAD is different from the current live site.

Pete mentioned that on webfu they used a git-head.txt file to compare versions. I think the only downside to that approach is needing to check the git-head.txt file into source control with every PR.

Other options:

- Write the version to a meta tag on the homepage and fetch the homepage to do the comparison during a build. Pete pointed out that this would mean invalidating the homepage whenever any other part of the site changes, which could cause our service worker to throw out the cached file.
- Have Eleventy generate a file that is git ignored that contains the version number and gets served by the server at `https://web.dev/version.json`. Then compare HEAD to that file.
- Use GitHub Action's artifacts feature to upload the version to GitHub. When an Action runs it can fetch the artifact and compare it against HEAD to see if it needs to deploy.

MichaelSolati:
Fixes #3655


HEAD.js
```diff
@@ -0,0 +1,36 @@
+const request = require('request');
+const {execSync} = require('child_process');
+const {CloudBuildClient} = require('@google-cloud/cloudbuild');
+
+const client = new CloudBuildClient();
+
+/**
+ * @returns {Promise<string>}
+ */
+const getDeployedHead = () => {
```
robdodson, line 10: Random thought, rather than use the word "Head" throughout the code, I wonder if it might be quicker to grok if we referred to this as the "version"? I realize it's not the same as the package.json version, but IMO it makes some of the code easier to read at a glance, e.g. `if (deployedVersion === currentVersion)`

robdodson (approved):

PR #5144, opened 2021-04-22
fix error

ZYinMD:
I believe the correct signature should be:

`MediaQueryList.addEventListener('change', callback)`. See [doc](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList).

The old version does work, but VSCode complains about it.


ZYinMD:
@googlebot I signed it!

jpmedley (approved):

kaycebasques:
Thanks for the fix @ZYinMD

PR #5157, opened 2021-04-22
Run eleventy once before starting the server, then watch for changes.

### LINKED ISSUE #3605: Clean up CLI output so authors know when to refresh the page.

robdodson:
Fixes #3605, unblocks #5008

`eleventy --watch` will run eleventy, which means if you have an npm script (like our dev script) where you want to 1st run eleventy, and then watch for changes, you can't really do it without running eleventy twice ([issue](https://github.com/11ty/eleventy/issues/1336)).

The solution is to write our own `chokidar` file watcher (which is what eleventy is using under the hood) to watch the `src/site` directory and re-run eleventy ourselves. Importantly, `chokidar` does **not** run your command when you first call the npm script.

- Replace `eleventy --watch` with our own chokidar watcher.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "debug:test:lib": "TEST_MODE=debug npm-run-all clean rollup --parallel watch:rollup karma",
     "debug:test:unit": "node --inspect-brk ./node_modules/.bin/mocha ./test/unit",
     "deploy": "ELEVENTY_ENV=prod npm run build && node index-algolia.js && firebase deploy --non-interactive",
-    "dev": "npm-run-all clean rollup gulp --parallel watch:rollup watch:eleventy watch:gulp firebase-config start",
+    "dev": "npm-run-all clean rollup gulp eleventy --parallel watch:rollup watch:eleventy watch:gulp firebase-config start",
     "eleventy": "eleventy --quiet",
     "firebase-config": "node firebase-config.js",
     "gulp:content": "npx gulp copy-content-assets",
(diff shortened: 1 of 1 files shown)


samthor:
My concerns were discussed in chat but I'll list them here:

re, Eleventy:
- Eleventy is already slow in that it is O(n) to the number of pages we have. This is part of its premise; it's imperative so it needs to rebuild the entire site every time. I sent them a PR to only render the page you care about (since rendering should have no side effects and Nunjucks is very slow) but that's not been merged either.
- It does however load a bunch of JS (its config) which is not reloaded on every rebuild, so there's some efficiencies there
- We preclude ourselves from any future optimizations by going down this route.

re, making every user interact with prod things:
- Average post author doesn't care about CSP or SWs or often even our JS build.
- But Eleventy is so slow for our huge site anyway that probably those things are really minor compared to the ever-increasing O(n) of all our pages.
- Perhaps we solve this by making editing part of CMS that lets you write your post in isolation, which is 90% of what people do when they interact with our repo(s), and then our repo is just for site _engineers_, not authors.

As @robdodson has pointed out, we can also roll this back of it's bad. But I wanted to get my thoughts on paper.

samthor (approved):

robdodson:
> We preclude ourselves from any future optimizations by going down this route.

It doesn't seem like Eleventy has any immediate plans to do incremental builds but if they did we would replace this.

> Average post author doesn't care about CSP or SWs or often even our JS build.

That's true, but the eng team also tends to not do prod builds before we merge stuff and deploy it. So we might shoot ourselves in the foot with a CSP change in dev, not realize it, and deploy. This kind of thing already happens because we don't run purgeCss in dev so we occasionally deploy broken UI. So it seems like it may actually be a good safe guard to do these in dev.

> Perhaps we solve this by making editing part of CMS that lets you write your post in isolation, which is 90% of what people do when they interact with our repo(s), and then our repo is just for site engineers, not authors.

I'm really supportive of this and I think it's the direction eleventy is heading in. See [this demo from Zach's talk](https://youtu.be/6XtVDfJKFzs?t=812) at Sanity conf where he shows the Eleventy 1.0 progammatic API which lets you invoke eleventy and pass explicit content.
- https://github.com/11ty/demo-eleventy-cloud/

robdodson:
Also a somewhat minor point. The current build script is a bit broken because we start the server while eleventy is still running. So I occasionally here from authors "I tried to run the site and was getting 404s...oh wait, now it works!" This would fix that issue at least.

PR #5159, opened 2021-04-23
Add thumbnail for css course

### LINKED ISSUE #5127: Add social media thumb to course
Jiwoong produced a new thumbnail which should be available in figma. We could maybe set it in the 11tydata file at the root of the course so all pages can use it.

devnook:
Fixes #5127

### DIFF
--- a/src/site/_data/courses/css/meta.yml
+++ b/src/site/_data/courses/css/meta.yml
@@ -7,3 +7,4 @@ description: i18n.courses.learn_css.description
 # This path should be relative to the _include folder.
 logo: '../../../images/courses/css/logo.svg'
 background: "url('/images/courses/css/background.svg')"
+thumbnail: 'image/jxu1OdD7LKOGIDU7jURMpSH2lyK2/fs5FxWQTb8apvFViRifm.svg'
--- a/src/site/content/en/learn/css/css.11tydata.js
+++ b/src/site/content/en/learn/css/css.11tydata.js
@@ -1,6 +1,7 @@
 module.exports = function () {
   return {
     layout: 'course',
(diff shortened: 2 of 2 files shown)


robdodson (approved):

robdodson (commented):
What do you think about doing something like this?

```js
module.exports = function () {
  return {
    layout: 'course',
    // The key is used to look up the data for the course in _data.
    // e.g. A course with a key of 'a11y' would have a corresponding
    // _data/courses/a11y directory.
    projectKey: 'css',
    eleventyComputed: {
      thumbnail: (data) => {
        // If the page has explicitly defined a thumbnail in its frontmatter,
        // use that.
        if (data.thumbnail) {
          return data.thumbnail;
        }

        // If the course has defined a thumbnail in its meta.yml,
        // use that.
        const {projectKey} = data;
        return data?.courses?.[projectKey]?.meta?.thumbnail;
      },
      permalink: (data) => {
        if (process.env.ELEVENTY_ENV === 'prod') {
          return false;
        }

        // I'm not sure why the output-permalinks.js eleventyComputed function
        // isn't stripping this for me, maybe it has to do with the order that
        // eleventy runs these in. But it's not a big deal as this is a short
        // lived 11tydata.js file that we're going to remove when the course
        // ships.
        return (data.page.filePathStem + '.html').replace(/^\/en\//, '/');
      },
    },
  };
};
```


That way we keep as much of the course config as we can in the `meta.yml` file`. Maybe the thumbnail function should be pulled out to its own file and placed in something like a `_computed` directory since it'll be reused by all courses. @MichaelSolati wdyt?

MichaelSolati:
Seems fine to me, should work.

robdodson (changes requested):
Hey @devnook, what do you think about updating the PR to match my comment?

devnook:
I updated the PR - ptal

robdodson (approved):

PR #5161, opened 2021-04-23
Implement share button

### LINKED ISSUE #4578: App bar with course breadcrumbs and share button
Each page in a course will need to have breadcrumbs (based on the page title) and a share button. We built something similar to this for d.c.c. so we can look to see if we might be able to repurpose it here.

devnook:
Fixes #4578


src/styles/components/_app-bar.scss
```diff
@@ -38,3 +38,20 @@
...
+
+  &::before {
+    background-size: 22px 22px;
+    content: "";
+    display: inline-block;
+    height: 22px;
+    margin-right: 4px;
+    width: 22px;
+    vertical-align: middle;
+    background-image: url(/images/icons/share.svg);
```
MichaelSolati, line 54: What's with the two background images?

MichaelSolati (approved):

src/site/_includes/partials/course-app-bar.njk
```diff
@@ -10,5 +10,15 @@
...
   </ul>
-  <button class="w-ml--auto">Share</button>
+  <share-action class="gc-analytics-event app-bar__share w-ml--auto"
+    authors="{{ authors | expandAuthors(collections.authors) }}"
+    data-category="web.dev"
+    data-label="share"
+    data-action="click"
+    data-icon="share"
+    tabindex="0"
+    role="button"
```
robdodson, line 20: hm it's interesting that the existing share action is implemented this way. It seems like it would be a lot better for it to wrap a `<button>`

There's an existing `.w-button--svg` class that I created for when we want to inline svg. I wonder if we could do something like:

```
{% from 'macros/icon.njk' import icon with context %}
<share-action>
  <button class="w-button--svg">
    {{ icon('share') }}
  </button>
</share-action>
```

src/site/_includes/partials/course-app-bar.njk
```diff
@@ -10,5 +10,15 @@
...
+  >
+    <span>Share</span>
```
robdodson, line 22: we want to i18n this text

src/site/_includes/partials/course-app-bar.njk
robdodson: oh i just realized w-button--svg is for buttons that only contain svg but this also contains text. let me fiddle with it a bit

src/images/icons/share.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
robdodson, line 1: I think this should go in _includes/icons so we can inline it with the `icon()` macro

robdodson:
I pushed a few changes:

- Moved the icon into the _inculdes/icons directory and optimized it with svgo
- Switched to inlining the icon with the `icon()` macro
- Tweaked the CSS a bit to add hover/focus states
- i18n'd the app-bar text

robdodson (approved):

PR #5166, opened 2021-04-23
Adds color assessment

argyleink:
Adds 2 questions to section conclusion

https://deploy-preview-5166--web-dev-staging.netlify.app/learn/css/color/

also, fixes a few invalid hsl color examples provided in the text 👍🏻

### DIFF
--- a/src/site/content/en/learn/css/color/color.assess.yml
+++ b/src/site/content/en/learn/css/color/color.assess.yml
@@ -1,0 +1,38 @@
+setLeader: Test your knowledge of color
+tabLabel: question
+questions:
+  - type: multiple-choice
+    cardinality: "1+"
+    correctAnswers: "1,3,4,5"
+    stem: Which of the following are valid colors?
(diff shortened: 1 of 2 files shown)


robdodson (approved):

robdodson:
@davegaeddert I noticed pullapprove didn't tag argyleink or una as the expected reviewers for this but I'm not sure why. Is this glob syntax incorrect? https://github.com/GoogleChrome/web.dev/blob/main/.pullapprove.yml#L20-L21

davegaeddert:
@robdodson I don't think that we have braces enabled, but we probably could and should. I can look into adding that but I'll need to test some things out so it will probably be next week sometime. Thanks for the heads up! (The immediate fix would be to write it without the braces.)

robdodson:
ah got it, yeah I recently added the braces which is probably what broke it. Good to know! I'll update it to the previous syntax.

PR #5184, opened 2021-04-27
Run csp reporting in dev.

robdodson:

- Changes CSP reporting to run in dev so we get console warnings.

### DIFF
--- a/src/site/_filters/csp-hash.js
+++ b/src/site/_filters/csp-hash.js
@@ -2,11 +2,8 @@ const {sha256base64} = require('../_data/lib/hash');
 const hashList = new Set();
 
 function cspHash(raw) {
-  if (process.env.ELEVENTY_ENV === 'prod') {
-    const hash = `'sha256-${sha256base64(raw)}'`;
-    hashList.add(hash);
-  }
-  return raw;
+  const hash = `'sha256-${sha256base64(raw)}'`;
+  hashList.add(hash);
 }
 
 function getHashList() {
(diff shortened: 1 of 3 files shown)


MichaelSolati (approved):

src/site/_filters/csp-hash.js
```diff
@@ -2,11 +2,8 @@ const {sha256base64} = require('../_data/lib/hash');
 const hashList = new Set();
 
 function cspHash(raw) {
-  if (process.env.ELEVENTY_ENV === 'prod') {
-    const hash = `'sha256-${sha256base64(raw)}'`;
-    hashList.add(hash);
-  }
-  return raw;
```
adamraine, line 9: This return value is necessary no, since we are chaining filters together.

robdodson: oh! yes.

robdodson:
I'm going to close this as it breaks the site :(

When we do a dev build we generate a random string for our script files (?v=[random string]). We do this so they aren't cached when you're doing dev work.

But the way the CSP stuff is implemented, the script loader is an inline script whose content gets fed to the cspHash function. This inline script _includes_ these random numbers.

image

So we're basically generating CSP hashes for the hundreds of pages for the site. When you go to load the page in dev, for whatever reason, this breaks saying that the CSP is blocking the script (you can see the massive number of hashes generated by all the random numbers).

image

It's not clear to me why this breaks on the _initial_ dev build. It makes sense that it would break on subsequent dev builds because we don't restart the firebase server, so it doesn't pull in the new list of CSP hashes.

I'm going to close this PR for now and we can keep investigating it if we decide to try to run CSP in dev again.

PR #5186, opened 2021-04-28
Preload fonts and fix layout shifts on learn pages.

robdodson:

- Preloads google sans because LH kept complaining about it.
- Fixes a CLS issue on all of the learning path landing pages
- Gives the site logo an explicit width/height juuuuuust in case
- Changes fonts to font-display: optional so the browser will just ditch the font if it's taking too long to load. I think this is now recommended over `swap`.
- Deletes some old devsite cruft


samthor (approved):

robdodson:
btw here's the explanation on the font change.
https://web.dev/preload-optional-fonts/

PR #5187, opened 2021-04-28
feat(youtube): migrate YouTube shortcode to use `webdev-infra`

MichaelSolati:


src/styles/components/_youtube.scss
```diff
@@ -5,6 +5,7 @@
 //
 // =============================================================================
 
+.youtube,
 .w-youtube {
```
samthor, line 9: Should we drop `w-youtube` since the shared code doesn't generate that?

samthor (approved):
I think you keep clobbering `package-lock.json` because you have an old version of NPM, right? I don't really care but it keeps getting bigger/smaller.

MichaelSolati:
> I think you keep clobbering `package-lock.json` because you have an old version of NPM, right? I don't really care but it keeps getting bigger/smaller.

Lol, you're right. I think it's cause i swap nvm also. Uhm, should i upgrade and update the package lock?

src/styles/components/_youtube.scss
MichaelSolati: It was used for the livestream player, but ill modify that.

PR #5190, opened 2021-04-28
Pl translations

devnook:


jpmedley (approved):

robdodson:
@devnook should we close this?

jpmedley:
I approved this on 4/30. Why is it a draft again?

robdodson:
> I approved this on 4/30. Why is it a draft again?

We're not planning to ship Polish translations on the site just yet. So I think we would close this until we actually want to ship Polish docs.

PR #5193, opened 2021-04-28
Fix and harmonize links to caniuse for browser-fs-access

Schweinepriester:
No issue created yet given the low complexity of the changes. But can create one if you want me to.

- The first link, <https://caniuse.com/#feat=file-system-access-api>, was broken, linking to an empty page basically
- The second link, <https://caniuse.com/?search=file%20system%20access%20api>, was linking to a search rather than the single feature (which might lead to different results in the future)
- Therefore exchange both links with the correct one, <https://caniuse.com/native-filesystem-api>
- (The URL is still derived from the previous name of the File System Access API, because https://github.com/Fyrd/caniuse/pull/5622#issuecomment-701910863)

### DIFF
--- a/src/site/content/en/blog/browser-fs-access/index.md
+++ b/src/site/content/en/blog/browser-fs-access/index.md
@@ -167,13 +167,13 @@ const saveFile = async (blob) => {
 ## Introducing browser-fs-access
 
 As perfectly fine as the File System Access API is,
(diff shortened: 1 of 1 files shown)


tomayac (approved):
LGTM, approval 

Thank you very much!

kaycebasques (approved):

PR #5222, opened 2021-05-01
Update documentation on render-blocking stylesheet

ekafyi:
This PR clarifies that `media=all` in a stylesheet link tag does _not_ constitute "matching the user's device" and thus is still considered render-blocking. I believe this could be potentially confusing if not clarified.

I realize the wording might not be very smooth / matches the overall writing style. Feel free to edit with better wording.

### DIFF
--- a/src/site/content/en/lighthouse-performance/render-blocking-resources/index.md
+++ b/src/site/content/en/lighthouse-performance/render-blocking-resources/index.md
@@ -35,7 +35,7 @@ A `<link rel="stylesheet">` tag that:
 
 * Does not have a `disabled` attribute. When this attribute is present,
   the browser does not download the stylesheet.
-* Does not have a `media` attribute that matches the user's device.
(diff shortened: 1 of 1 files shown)


kaycebasques:
Thanks for the update @ekafyi. Your wording looks good to me. Do you have a source for this?

ekafyi:
Hi @kaycebasques, my source is https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-blocking-css, specifically this part:

screenshot of linked article

This quote shows `media=all` is equivalent to not having a `media` attribute at all.

> The first declaration is render blocking and matches in all conditions.
>
> The second declaration is also render blocking: "all" is the default type so if you don’t specify any type, it’s implicitly set to "all". Hence, the first and second declarations are actually equivalent.

(I'm aware this article refers to browser behaviour in general, not Lighthouse specifically. In case this is not applicable or I understood this incorrectly, feel free to close this PR!)

robdodson (approved):

PR #5229, opened 2021-05-03
Decouple search and search results

devnook:


src/lib/components/Search/index.js
```diff
@@ -165,128 +169,26 @@ class Search extends BaseStateElement {
...
     this.inputEl = this.renderRoot.querySelector('.web-search__input');
   }
 
   /**
-   * Keep track of cursor changes and reflect them to aria-activedescendant.
-   * This ensures screen readers properly announce the current search result.
-   * We do this because focus never leaves the search input box, so when the
-   * user is arrowing through results, we have to tell the screen reader about
-   * it.
+   * Passes on updated search properties to the searh results element.
```
robdodson, line 181:
```suggestion
   * Passes on updated search properties to the search results element.
```

src/site/_includes/partials/header-default.njk
```diff
@@ -53,7 +53,8 @@
         {{ 'i18n.header.about' | i18n(locale) }}
       </a>
     </div>
-    <web-search></web-search>
+    <web-search resultsId="search-main-results"></web-search>
```
robdodson, line 56: HTML usually doesn't camel case attributes so maybe we should rename this. I think if you do it dash cased, `results-id`, Lit is smart enough to convert that to camel case for you. Or you could use a different attribute like `for=""` or `target=""`.

src/site/_includes/course.njk
```diff
@@ -1,5 +1,7 @@
 ---
 layout: base
+tags:
+  - course
```
robdodson, line 4: should this be removed? it doesn't seem to do anything

src/site/content/en/learn/css/css.11tydata.js
```diff
@@ -5,6 +5,8 @@ module.exports = function () {
     // e.g. A course with a key of 'a11y' would have a corresponding
     // _data/courses/a11y directory.
     projectKey: 'css',
+    tags: ['course-learn-css'],
+    searchTag: 'course-learn-css',
```
robdodson, line 9: We may want to put this in the _data/courses/css/meta.yml and then fetch them the same way we do the thumbnail down below in the eleventyComputed block. If the author has defined tags already on a course module you would likely need to merge them. Something like:

```
eleventyComputed: {
  tags: (data) => {
    const {projectKey} = data;
    return [...data.courses[projectKey].meta.tags, ...data.tags];
  }
}
```

I think it may also be good to see if we can consolidate these into a single property instead of two (tags and searchTag). Is that possible?

src/site/_includes/partials/navigation-drawer-course.njk
```diff
@@ -16,7 +16,8 @@
       </div>
 
       <div class="drawer-course__search">
-        <div class="input"></div>
+        <web-search resultsId="search-course-results" tag="{{ searchTag }}"></web-search>
+        <web-search-results id="search-course-results"></web-search-results>
```
robdodson, line 20: I wonder if we should exclude the courses stuff so it doesn't block landing the PR. I think we may need to do some animation and whatnot to fade in the search results and include some other stuff to match Jiwoong's mocks and we could do all of that in a separate PR so we can land this.

robdodson (changes requested):

robdodson:
One thing to note is that I added `eleventyExcludeFromCollections` to the css.11tydata.js file to fix issue with the course content showing up on the authors and tags pages ([PR](https://github.com/GoogleChrome/web.dev/pull/5265)). But I think using eleventyExcludeFromCollections will also make it so this content doesn't get indexed by Algolia. So if you're testing and noticing it's not showing up then that's probably why.

We may need to tell course authors to not use tags or put themselves down as authors if they don't want stuff to show up on the authors/tags pages.

robdodson:
OK ignore my previous comment. It led to bugs so I've changed it to be more specific and only exclude course content from authors and tags pages (https://github.com/GoogleChrome/web.dev/pull/5266)

src/site/content/en/learn/css/css.11tydata.js
devnook: I jus wonder- what is the rationale of having courses content defined in .yaml instead of .md and 11tydata, like for all the other types of content?

src/site/_includes/partials/header-default.njk
devnook: Lit didn't work its magic, so I just use the attribute directly (it does not need to be a property).

src/site/_includes/course.njk
devnook: I though of using it for search faceting, but its not needed for now.

src/site/content/en/learn/css/css.11tydata.js
devnook: I updated the tags to pull from meta.yml

devnook:
PTAL

robdodson:
I think everything is working well, but one thing I noticed one issue:

- Reload the site
- Throttle your network to Slow 3G
- Click on the text field and type a single letter

The results will do an initial render saying that there are no results:
image

Then if you wait a while the response from Algolia will come back and you'll see the actual results.

This differs from the current live site which will not show the search results at all until it gets an initial response. Because we have a lot of users in India on slower networks this seems like something we should fix. Apologies for not noticing this sooner :(

src/site/content/en/learn/css/css.11tydata.js
```diff
@@ -5,11 +5,22 @@ module.exports = function () {
     // e.g. A course with a key of 'a11y' would have a corresponding
     // _data/courses/a11y directory.
     projectKey: 'css',
+    searchTag: 'course-learn-css',
```
robdodson, line 8: I think this should go in meta.yml as well. Ideally the only value in here that's project specific would be the projectKey and I'm even wondering if we should just infer that from the directory name (that's an issue for a different PR tho 😛 ). But I'd like to remove project specific stuff so authors don't forget to configure it.

src/lib/components/Search/index.js
```diff
@@ -85,11 +87,14 @@ class Search extends BaseStateElement {
     super.connectedCallback();
     window.addEventListener('resize', this.onResize);
     this.onResize();
+    this.resultsEl = document.getElementById(this.getAttribute('results-id'));
+    this.resultsEl.addEventListener('select', this.onResultSelect);
```
robdodson, line 91: maybe do `if (this.resultsEl) { // add listener } else { console.warn() }`

src/lib/components/Search/index.js
```diff
@@ -85,11 +87,14 @@ class Search extends BaseStateElement {
...
   }
 
   disconnectedCallback() {
     super.disconnectedCallback();
     window.removeEventListener('resize', this.onResize);
+    this.resultsEl.removeEventListener('select', this.onResultSelect);
```
robdodson, line 97: `if (this.resultsEl) { // remove listener`

src/lib/components/Search/index.js
```diff
@@ -165,128 +170,26 @@ class Search extends BaseStateElement {
...
-
-    this.inputEl.setAttribute(
-      'aria-activedescendant',
-      `web-search-popout__link--${this.cursor}`,
-    );
+    // Reflect changed properties to the results el.
+    const sharedProperties = ['query', 'hits', 'showHits'];
+    sharedProperties.forEach((property) => {
+      if (changedProperties.has(property)) {
+        this.resultsEl[property] = this[property];
```
robdodson, line 190: I think we have to do a `if (this.resultsEl)` check here as well... pretty much anytime we want to access it we probably need to guard against a case where it doesn't exist.

src/lib/components/Search/index.js
```diff
@@ -300,42 +203,44 @@ class Search extends BaseStateElement {
...
+   * Keep track of which result is selected in the search results element and
+   * reflect them to aria-activedescendant.
+   * This ensures screen readers properly announce the current search result.
+   * We do this because focus never leaves the search input box, so when the
+   * user is arrowing through results, we have to tell the screen reader about
+   * it.
+   * @param {Event} event Select event fired by search results element.
+   */
+  onResultSelect(event) {
+    if (event.target instanceof HTMLElement) {
```
robdodson, line 216: I think you can remove this line and instead do `event.currentTarget` which will always be the results element because that's the one you've added the listener to.

src/lib/components/Search/index.js
```diff
@@ -300,42 +203,44 @@ class Search extends BaseStateElement {
...
+      const selected = event.target.querySelector('[aria-selected="true"]');
```
robdodson, line 217: I'm a little concerned with reaching into the other component in this way. As an alternative, would it be possible for the results element to fire two events, `selected` and `deselected`. When `selected` is fired it would contain the id of the selected item. And when we hear `deselected` we know to remove aria activedescendant.

src/lib/components/SearchResults/index.js
```diff
@@ -0,0 +1,217 @@
...
+    this.cursor = -1;
+    this.query = '';
+  }
+
+  set hits(hits) {
+    this.hits_ = hits;
+    this.cursor = -1;
+  }
+
+  navigate(key) {
```
robdodson, line 48: Can you add a comment explaining why this method exists and how it's used by another element to essentially drive this element.

src/site/content/en/learn/css/css.11tydata.js
devnook: I don't want this to be editable by course authors, as it has nothing to do with the content of the course and is used just to technically adjust our search infra. But you're right, it can also be simply derived from the projectkey and not editable at all.

src/lib/components/Search/index.js
devnook: This is actually here only to make typescript happy, otherwise it complains that EventTarget does not have HTMLElement interface.

devnook:
Good catch with the flash of empty search results, thanks!

PTAL

src/lib/components/Search/index.js
robdodson: oh I think you can do a cast with a comment. It's something like...

if ((/** @type {HTMLElement} */ event).target)

src/site/content/en/learn/css/css.11tydata.js
robdodson: Mostly so we can consolidate into the _data directory. It's really helpful to have stuff at that top level when you need to, for instance, fetch data about a course from some other part of the site that isn't a direct descendant of the /learn/css/ folder.

devnook:
PTAL

robdodson (approved):
Looks good! Feel free to merge when you're ready

PR #5232, opened 2021-05-03
Adds pseudo-element assessment

argyleink:
2 pop quiz questions at the end 👍🏻

### DIFF
--- a/src/site/content/en/learn/css/pseudo-elements/pseudo-elements.assess.yml
+++ b/src/site/content/en/learn/css/pseudo-elements/pseudo-elements.assess.yml
@@ -23,7 +23,7 @@ questions:
   - type: multiple-choice
     cardinality: "1"
     correctAnswers: "1"
-    stem: Psuedo elements can be found in the HTML
+    stem: Psuedo elements can be found in an HTML file
     options:
       - content: "True"
         rationale: "While DevTools may show pseudo elements in the Elements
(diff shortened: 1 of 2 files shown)


src/site/content/en/learn/css/pseudo-elements/pseudo-elements.assess.yml
```diff
@@ -0,0 +1,33 @@
...
+        rationale: "This is the bullet element when you use a list element or
+        display type"
+      - content: "`::pencil`"
+        rationale: "This doesn't exist in CSS"
+      - content: "`:active`"
+        rationale: "This is a pseudo class not a pseudo element"
+  - type: multiple-choice
+    cardinality: "1"
+    correctAnswers: "1"
+    stem: Psuedo elements can be found in the HTML
```
rachelandrew, line 26: I'm not sure I understand what this means. Do you mean "can be seen when you view source?" "Exist in the HTML document that you created".

(it is very very common that devs don't understand that the DOM they see in DevTools is different to what they get if they view source... )

rachelandrew (changes requested):
a question to look at in here.

src/site/content/en/learn/css/pseudo-elements/pseudo-elements.assess.yml
argyleink: "Exists in the HTML document you created or are looking inside of"

it def is trying to poke this grey space, to harden that these "elements" aren't in the document, in case they hadnt realized it fully yet.

```suggestion
    stem: Psuedo elements can be found in an HTML file
```

thoughts on that adjustment?

rachelandrew (approved):

PR #5269, opened 2021-05-06
Fix course description

### LINKED ISSUE #5253: Improve description text for course subheadings
image

robdodson:
Fixes #5253 

@una @argyleink @andy-piccalilli can I get y'alls thoughts on these changes? I didn't bump the text down to 0.9rem because then it would be smaller than the body copy below it (which is 18px). But I can bump it down if you think it would look better.


andy-piccalilli:
Yeh that looks decent!

argyleink (approved):
lgtm 

Screen Shot 2021-05-06 at 7 42 25 AM

PR #5273, opened 2021-05-06
fix(algolia): trim text of algolia content

MichaelSolati:


robdodson (approved):
LGTM but also discussed opening a ticket to address the core issue.

Currently we combine the english results into the translated record, so the spanish record for /vitals/ contains essentially two articles. We should rethink this approach because we have some articles that are going to be way longer than /vitals/.

Let's merge the PR now to unbreak the build, but also work on a better fallback option.

PR #5299, opened 2021-05-10
[Async Clipboard article] Document Safari's special way

tomayac:

- Safari requires async work to be done differently.


tomayac:
> Is WebKit going to change to match? I'm trying to work out whether we need to put this information on MDN.

I don't think they will change. It's just that they treat user activation differently. So this should probably go up on MDN, yes.

PR #5304, opened 2021-05-10
soft launch learn css

robdodson:

- Soft launch learn css course
- Add redirect for learnCSS -> /learn/css. It looks like our redirects are case sensitive.


argyleink (approved):
🎉

PR #5307, opened 2021-05-10
Remove CLS from the sign in element in /measure

### LINKED ISSUE #5272: Sign in button on web.dev/measure causes CLS
**Describe the bug**
Take a look at the two screenshots in this image to see that the sign-in button takes a moment to display and causes layout shift.

image

devnook:
Fixes #5272

### DIFF
--- a/src/site/content/en/measure/index.njk
+++ b/src/site/content/en/measure/index.njk
@@ -22,12 +22,14 @@ to remove them as part of the v1 migration. #}
           Sign in to track your progress over time.
         </span>
       </p>
-      <web-signin-button></web-signin-button>
-      <p class="w-page-header__copy w-unsupported-notice">
-        Sorry, the measure tool does not work with legacy browsers or with
-        JavaScript disabled. Please try again using an up-to-date version of
-        Edge, Firefox, Safari, or Chrome.
-      </p>
+      <div class="w-measure-signing-button">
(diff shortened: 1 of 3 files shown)


robdodson:
Thanks for taking a look at this Ewa! I think the solution in the PR may leave too much whitespace because it's reserving the area for the button with a fixed height.

image

One thing Sam added a while back for analytics was a little localStorage flag which tells us if the user was signed in or not
https://github.com/GoogleChrome/web.dev/blob/main/src/lib/app.js#L23

So I'm curious if we could have an inline script which reads from localStorage and hides these elements if that flag is already set. This is similar to the approach used in that Zach Leatherman article I shared with you https://www.zachleat.com/web/layout-shift/ and we do it on d.c.c. https://github.com/GoogleChrome/developer.chrome.com/blob/d69a8c2d626ee6f35be7267ee7a71fa7d233ce1b/site/_includes/partials/script.js#L41-L50

robdodson (changes requested):
See comment above

devnook:
@robdodson The whitespace is there right now as well on web.dev. Do you man we should make it smaller?

devnook:
PTAL

robdodson:
oh! I was thinking we would make the button display none. But we could explore doing that in a separate PR I guess.

robdodson (approved):

PR #5318, opened 2021-05-11
ci(firebase-preview): generate firebase previews on prs

MichaelSolati:


.github/workflows/firebase-hosting-pull-request.yml
```diff
@@ -0,0 +1,28 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build site
+        env:
+          ELEVENTY_ENV: prod
```
robdodson, line 19: We may want to leave this as a dev build because it will be faster and it will show `draft` posts. This was something we ran into recently while working on the Learn CSS course.

PR #5351, opened 2021-05-13
Add imgix query param to improve the webperf

nucliweb:
## Changes proposed in this pull request:

Following the IMGIX documentation https://docs.imgix.com/apis/rendering/format/fm we can improve the Web Performance adding the query param `auto=format` into the images requests, to serve the better image format.

## Before

Before fix

## After

After fix

### DIFF
--- a/src/site/_includes/components/Author.js
+++ b/src/site/_includes/components/Author.js
@@ -41,7 +41,6 @@ module.exports = ({id, author, showSocialMedia = false, small = false}) => {
     height: '64',
     class: `w-author__image${small ? ' w-author__image--small' : ''}`,
     params: {
-      auto: 'format',
       fit: 'crop',
       h: '64',
       w: '64',
--- a/src/site/_includes/components/Author.js
+++ b/src/site/_includes/components/Author.js
@@ -41,6 +41,7 @@ module.exports = ({id, author, showSocialMedia = false, small = false}) => {
     height: '64',
     class: `w-author__image${small ? ' w-author__image--small' : ''}`,
     params: {
+      auto: 'format',
       fit: 'crop',
       h: '64',
       w: '64',
--- a/src/site/_includes/components/AuthorCard.js
+++ b/src/site/_includes/components/AuthorCard.js
@@ -35,7 +35,6 @@ class AuthorCard extends BaseCard {
       height: '192',
       class: 'w-card-author__image',
       params: {
-        auto: 'format',
         fit: 'crop',
         h: '192',
         w: '192',
--- a/src/site/_includes/components/AuthorCard.js
+++ b/src/site/_includes/components/AuthorCard.js
@@ -35,6 +35,7 @@ class AuthorCard extends BaseCard {
       height: '192',
       class: 'w-card-author__image',
       params: {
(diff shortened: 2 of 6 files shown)


argyleink:
Looks nice!

Should changes be made upstream on the shortcode too? https://github.com/GoogleChrome/webdev-infra/blob/main/shortcodes/Img.js

nucliweb:
> Looks nice!
> 
> Should changes be made upstream on the shortcode too? https://github.com/GoogleChrome/webdev-infra/blob/main/shortcodes/Img.js

Really the `Img` component use the `params` (via [props](https://github.com/GoogleChrome/webdev-infra/blob/main/shortcodes/Img.js#L70)) to add to `src` and `srcset` 
https://github.com/GoogleChrome/webdev-infra/blob/main/shortcodes/Img.js#L70

MichaelSolati:
I think we may have a concern with how by doing this our SVGs will get rasterized, which we dont always want.

I'm not opposing this isn't of change/pr, but i think we need to check how it interacts with SVGs, or possibly change it to ignore SVGs.

I'll get back to y'all on Monday or Tuesday about this.

nucliweb:
Good point!

I can see that if we add the param to an SVG file, the service return an bitmap image, so if you add `auto=format` to an SVG image, the service return a WebP image format with Chrome browser.

Screen Shot 2021-05-16 at 23 01 00

We can do something like this:

```js
  const autoFormat = (src) => (/\.svg$/i.test(src) ? {auto: 'format'} : {});

  const img = Img({
    src: author.image,
    alt: author.title,
    width: '64',
    height: '64',
    class: `w-author__image${small ? ' w-author__image--small' : ''}`,
    params: {
      ...autoFormat(src),
      fit: 'crop',
      h: '64',
      w: '64',
    },
  });
```

MichaelSolati:
So I was thinking that the [Img component](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/components/Img.js) can be updated to turn on auto format for all of the images, we do a check if it's a SVG, if it is we don't turn on auto format (unless the user specifically sets auto format).

Hopefully that makes sense, you'd want to turn it into a function, but it'll be a lil messy to understand because the typescript types are broken ATM for this file so it may not make sense.


If you don't mind/want to explore that would be awesome, but if not i think I'd want to accept this as it is for the time.

MichaelSolati:
Also here's the function from infra if you'd like it for reference... https://github.com/GoogleChrome/webdev-infra/blob/03f0100af32a3757b50d456308fbd88e2edac475/shortcodes/Img.js#L60

nucliweb:
I'll take a look to do it your proposal, make sense, we'll have the `auto=format` only in one point.

Thanks @MichaelSolati

nucliweb:
@MichaelSolati new approach.

Now, the `Img` component add `auto=format` if isn't a SVG image and if `params` don't have it.

nucliweb:
Btw, I don't know why the build has an error with the `Percy` 😓

MichaelSolati:
I think it's just a security issue with GitHub Actions, thank you for this awesome contribution @nucliweb!

PR #5353, opened 2021-05-14
Remove focus outline from course cards

### LINKED ISSUE #5337: Learn CSS card shouldn't show focus ring on mouse click.
Mouse clicking the Learn CSS card on the /learn/ page will show a focus ring. We already have a focus indicator for the card so showing the ring is duplicative.

devnook:
Fixes #5337


robdodson (approved):
This works but I'm also curious if we could abbreviate it and the __description override using something like this:

```
.course-card :focus { outline: none; }
```

robdodson:
looking at this again, I think i actually filed a bad issue and no change is needed. I'm going ot suggest that we close this

PR #5401, opened 2021-05-19
Handle empty arrays in reduce calls in "Debug Web Vitals in the field" article

bazzadp:
Page with proposed change: https://web.dev/debug-web-vitals-in-the-field/

For very simple pages it is possible the `entires` array is empty. This prevents `reduce` working properly with the following error:

```
(index):51 Uncaught TypeError: Reduce of empty array with no initial value
    at Array.reduce (<anonymous>)
    at getLargestLayoutShiftEntry ((index):51)
    at sendWebVitalsGAEvents ((index):80)
    at web-vitals.js?module:1
    at web-vitals.js?module:1
    at n (web-vitals.js?module:1)
```
This will prevent 0 CLS values being logged for very simple pages.

This pull request adds an [`initialValue`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce#:~:text=Calling%20reduce()%20on%20an%20empty%20array%20without%20an%20initialValue%20will%20throw%20a%20TypeError) of `null` to prevent this error:

> Calling reduce() on an empty array without an initialValue will throw a TypeError.

Demo of original code with the fault: https://www.tunetheweb.com/experiments/core-web-vitals-debug/
Demo of updated code with no fault: https://www.tunetheweb.com/experiments/core-web-vitals-debug/index2.html
Demo of updated code with no fault for page with a CLS: https://www.tunetheweb.com/experiments/core-web-vitals-debug/index3.html

CC: @philipwalton

### DIFF
--- a/src/site/content/en/blog/debug-web-vitals-in-the-field/index.md
+++ b/src/site/content/en/blog/debug-web-vitals-in-the-field/index.md
@@ -133,12 +133,12 @@ to CLS and returns the largest source element from the largest shift:
 function getCLSDebugTarget(entries) {
   const largestShift = entries.reduce((a, b) => {
     return a && a.value > b.value ? a : b;
-  });
+  }, null);
   if (largestShift && largestShift.sources) {
     const largestSource = largestShift.sources.reduce((a, b) => {
       return a.node && a.previousRect.width * a.previousRect.height >
           b.previousRect.width * b.previousRect.height ? a : b;
-    });
+    }, null);
     if (largestSource) {
       return largestSource.node;
     }
@@ -347,14 +347,14 @@ function getSelector(node, maxLen = 100) {
 }
 
 function getLargestLayoutShiftEntry(entries) {
-  return entries.reduce((a, b) => a && a.value > b.value ? a : b);
+  return entries.reduce((a, b) => a && a.value > b.value ? a : b, null);
 }
 
 function getLargestLayoutShiftSource(sources) {
   return sources.reduce((a, b) => {
     return a.node && a.previousRect.width * a.previousRect.height >
         b.previousRect.width * b.previousRect.height ? a : b;
-  });
+  }, null);
 }
 
 function wasFIDBeforeDCL(fidEntry) {
--- a/src/site/content/en/blog/debug-web-vitals-in-the-field/index.md
+++ b/src/site/content/en/blog/debug-web-vitals-in-the-field/index.md
@@ -138,7 +138,7 @@ function getCLSDebugTarget(entries) {
     const largestSource = largestShift.sources.reduce((a, b) => {
       return a.node && a.previousRect.width * a.previousRect.height >
           b.previousRect.width * b.previousRect.height ? a : b;
-    }, null);
+    });
     if (largestSource) {
       return largestSource.node;
     }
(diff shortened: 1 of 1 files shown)


philipwalton:
Thanks for the reduced test cases, I think I see the issue now.

In the [full code example in this post](https://web.dev/debug-web-vitals-in-the-field/#:~:text=if%20(entries.length)%20%7B), the initial calling function has an `if (entries.length) { ... }` conditional, which I'm not seeing in your test cases. That explains why I wasn't seeing the error on any of the sites where I'm using this code (which typically don't have any CLS either).

I can see an argument for always making sure `.reduce()` is called with an initial value, but in those cases `null` will be returned, which could also lead to errors if someone was expecting a entry to be returned.

bazzadp:
Oh good spot! OK I'm happy to close this then as user error :-) Thanks for getting back so quickly.

ArgonAlex:
There is actually a legitimate bug with getLargestLayoutShiftEntry(). The `sources` array does not have a length check in the rest of the example. I just noticed this showing up in our client error logs.

philipwalton:
> The `sources` array does not have a length check in the rest of the example

The code in the article has the following conditional before ever referencing the `sources` property:

```js
if (largestEntry && largestEntry.sources) { 
```

ArgonAlex:
It needs to be:
```js
if (largestEntry && largestEntry.sources && largestEntry.sources.length) { 
```

philipwalton:
Hmmm, looks like I'm seeing a few reports of this error in my logs as well.

Given that it's happening in the wild, I agree it's worth updating the code example, but I also think it may be a Chrome bug. I don't think it should be possible to have a layout shift with no sources. (Adding @npm to confirm.)

yoavweiss:
Test cases/crbug may be interesting here. Digging around the code, I see that the list can be empty if the shifted element is a [TextNode created under certain conditions](https://source.chromium.org/search?q=kInvalidDOMNodeId%20f:list_marker_painter.cc&sq=&ss=chromium%2Fchromium%2Fsrc) or if the list of attributions is empty, which I don't think should happen...

^^ npm@ who may know more

yoavweiss:
^^ @npm1 who may know more

npm1:
Is there a website where we can easily repro this problem? It would be nice to investigate if this is an implementation problem or working as intended.

philipwalton:
I noticed this on my website in my RUM data. That is, I noticed instances of this error, which points to the code referencing `entry.sources`:

```
TypeError: Reduce of empty array with no initial value
  at Array.reduce (<anonymous>)
  at https://philipwalton.com/modules/log.mjs:1:4963
  at https://philipwalton.com/modules/web-vitals.mjs:1:927
  at https://philipwalton.com/modules/web-vitals.mjs:1:1119
  at n (https://philipwalton.com/modules/web-vitals.mjs:1:463)
```

However, I can't reproduce this error myself. It's only happened 3 times on my site in the past 28 days (all on Chrome for Android).

ArgonAlex:
Over the 5 days I had this code running without an extra length check, the error was logged a few thousand times. 96% of them were from Chrome on Android, the rest were from Chrome on a mix of desktop Linux, Windows, ChromeOS.

npm1:
Based on this it does sound like a Chrome bug, though we can only speculate on what the bug is until we can repro :S If you ever find a way, feel free to file a bug: https://bugs.chromium.org/p/chromium/issues/entry?template=CoreWebVitals+User+Report

(Or if you have a sample URL where this impacts a large % of the data on Android, since that could also help narrow down the cause)

philipwalton:
The code example on [web.dev/debug-web-vitals-in-the-field/](https://web.dev/debug-web-vitals-in-the-field/) has been updated to check for `sources.length`, so hopefully this won't catch people in the future. But yeah, it'd be great to understand how this can happen in the wild.

PR #5420, opened 2021-05-21
Mention and add reference to attribute selectors

richardkmichael:
- Adjust the heading so selector names order corresponds to the section content (keeping class and pseudo-class together seems helpful to the reader)
- Mention provide a reference to atttribute selectors in the discussion of specificity points


rachelandrew (approved):
lgtm - thanks :)

PR #5433, opened 2021-05-21
Fix bug in pseudo-classes markdown.

### LINKED ISSUE #5430: pseudo-classes - incorrect code formatted text
**Describe the bug**
A large chunk of a paragraph is code-formatted (monospaced) where it shouldn't be.

**To Reproduce**
1. Go to https://web.dev/learn/css/pseudo-classes/#:focus-:focus-within-and-:focus-visible
2. See that the formatted text includes all of this text: "when the tab. With :focus-visible you can present a focus style when an element receives focus via the keyboard, while also using the outline: none rule to prevent it when a pointer device interacts with it."

**Expected behavior**
Only `:focus-visible` and `outline: none` should be code-formatted.

**Screenshots**
https://drive.google.com/file/d/1r4hx9vxsar0Hdi4-Hzr_bavSrVwKDF6c/view?resourcekey=0-3OnlCvH6A-QZycOH-PJtpQ

I82Much:
Instead of nested `<kbd>` elements, close the `<kbd>` element. The nested elements were breaking the layout. 

Fixes #5430

### DIFF
--- a/src/site/content/en/learn/css/pseudo-classes/index.md
+++ b/src/site/content/en/learn/css/pseudo-classes/index.md
@@ -98,7 +98,7 @@ button:focus {
 This CSS removes the default browser focus ring when an element receives focus,
 which presents an accessibility issue for users who navigate a web page with a keyboard.
 If there is no focus style,
-they won't be able to keep track of where focus currently is when the <kbd>tab<kbd>.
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):
thank you!

PR #5509, opened 2021-06-02
Fixes size inaccuracy in box-model breakdown

### LINKED ISSUE #5488: Box model size incorrect
In the *Box Model* tutorial, this sentence has the wrong size for the box:

> The content would break out of your element and it would be 140px wide, rather than 100px.

Given the CSS provided:

```css
p {
    width: 100px;
    height: 50px;
    padding: 20px;
    border: 1px solid;
}
```
The `<p>` element will have a width of 142px (due to the border), not 140px. Pedantic, I know, but it stood out to me as soon as I saw the example.

**To Reproduce**
1. Go to [Box Model](https://web.dev/learn/css/box-model/)
2. Scroll to the first example under the `<p>` CSS shown above
4. Fix the 140px to be 142px

**Expected behavior**
The text should be:

> The content would break out of your element and it would be 142px wide, rather than 100px.

**Screenshots**
para-css
para-element

argyleink:
Fixes #5488

### DIFF
--- a/src/site/content/en/learn/css/box-model/index.md
+++ b/src/site/content/en/learn/css/box-model/index.md
@@ -29,7 +29,7 @@ p {
 }
 ```
 
-The content would break out of your element and it would be 140px wide,
+The content would break out of your element and it would be 142px wide,
 rather than 100px.
 Why is that?
 The box model is a core foundation of CSS and understanding how it works,


rachelandrew (approved):
lgtm

PR #5520, opened 2021-06-03
Factor our ScrollSpy from Toc into separate component

### LINKED ISSUE #5215: Course ToC should have scrollspy
Una said that it would be nice if the ToC on the right hand side of the course could do a scrollspy effect and style the link when the current heading is in the viewport. We do this with the ToC used on the blog.

image

I think we could use IntersectionObserver to add this to the course ToC

devnook:
Fixes #5215

### DIFF
--- a/src/styles/components/_toc.scss
+++ b/src/styles/components/_toc.scss
@@ -40,10 +40,6 @@ web-table-of-contents {
 
   .w-scroll-spy__active {
     position: relative;
-    -webkit-margin-before: 0em;
-    -webkit-margin-after: 0em;
-    -webkit-margin-start: 0px;
-    -webkit-margin-end: 0px;
     display: inline-block;
   }
 
@@ -132,15 +128,15 @@ web-table-of-contents-button {
   @include bp(sm) {
     height: calc(100% - 32px);
   }
+
+  a {
+    display: inline-block;
+  }
 }
 
 .w-toc__header {
   margin-bottom: 12px;
   margin-top: 12px;
-
-  a {
-    display: inline-block;
-  }
 }
 
 .w-toc__header--link {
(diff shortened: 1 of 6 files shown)


src/styles/components/_toc.scss
```diff
@@ -40,6 +34,31 @@ main.w-toc-open {
...
+  h2 {
+    line-height: 32px;
+  }
+
+  .w-scroll-spy__active {
+    position: relative;
+    -webkit-margin-before: 0em;
+    -webkit-margin-after: 0em;
+    -webkit-margin-start: 0px;
+    -webkit-margin-end: 0px;
```
robdodson, line 46: are these vendor prefixed logical properties? https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties

robdodson:
I think this is generally looking good. But I noticed one issue where the line height is changing between active states.

https://user-images.githubusercontent.com/1066253/120874501-c7d12380-c55b-11eb-93da-b9fbe00d10f5.mp4

src/styles/components/_toc.scss
devnook: No, this is rather to control the border height on the list-item element.

devnook:
PTAL

robdodson:
It looks like there is still a line height issue, but this only happens in Firefox. I was testing with this article. http://localhost:8080/excalidraw-and-fugu/


https://user-images.githubusercontent.com/1066253/121262108-77620a80-c868-11eb-815e-50e4e847540a.mp4

src/styles/components/_toc.scss
robdodson: I'm a bit confused... when I deleted them it didn't seem to change anything.
The reason I'm bringing it up is because I've tended to avoid using vendor prefixed properties if I can help it and it seems like these wouldn't work in Firefox. But I feel like I'm missing something.. can you show an example of what they do on the site?

devnook:
Ok, should be corrected now!

MichaelSolati (approved):

PR #5524, opened 2021-06-03
Add a better ignore script

### LINKED ISSUE #5060: Add .env ignore feature from d.c.c.
d.c.c. has a simple utility that lets you ignore parts of the site ([link](https://github.com/GoogleChrome/developer.chrome.com/blob/main/tools/eleventyignore/index.js)) using environment variables. This _dramatically_ speeds up build times when you're developing.

robdodson:
Fixes #5060

This replaces the old `isolate`/`integrate` feature and makes it so you can ignore the entire site and only include specific directories. The syntax looks like this:

```
# .env

ELEVENTY_IGNORE=true
ELEVENTY_INCLUDE=["learn", "about"]
```


MichaelSolati (approved):
Very cool, lgtm

Issue #5490, opened 2021-05-30
Wrong Codepen example for justify-items / align-items


gnehcwu:
**Describe the bug**
Properties "justify-items / align-items" should be used in the example.

**To Reproduce**
1. Go to https://web.dev/learn/css/grid/#moving-content-around
2. See mis-set properties

**Expected behavior**
Correct properties should be used for demoing.

**Screenshots**
![image](https://user-images.githubusercontent.com/5228723/120098393-8ee51900-c12d-11eb-9d64-450477933346.png)

argyleink:
I went looking through the Codepens and I dont think an align-items and justify-items demo was created. We'll need to fork and update the post.

PR #5539, opened 2021-06-04
docs(frontend): correct unpacking of `flex: 1` (#5537)

### LINKED ISSUE #5537: Possible error in "Learn CSS" course chapter 9 regarding unpacking of "flex: 1"
**Describe the bug**
Thank you for providing the "Learn CSS" course!
While reading [chapter 9 about Flexbox](https://web.dev/learn/css/flexbox/), I came across the [example about the unpacking of `flex: 1`](https://web.dev/learn/css/flexbox/#controlling-space-inside-flex-items), which might be incorrect.
The course material states

> To force all of the items to be a consistent size and ignore the size of the content change flex:auto to flex: 1. in the demo.
>
> This unpacks to:
>
>  * flex-grow: 1: items can grow larger than their flex-basis.
>  * flex-shrink: 0: items can't shrink smaller than their flex-basis.
>  * flex-basis: 0: items have a base size of 0.

which by my understanding of the [spec](https://drafts.csswg.org/css-flexbox-1/#flex-common) should be  

* `flex-shrink: 1`: items can shrink smaller than their flex-basis`. 

(see also the [MDN documentation regarding the `flex` property](https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax)).

**To Reproduce**
1. Go to https://drafts.csswg.org/css-flexbox-1/#flex-common
2. Compare it to the spec at https://drafts.csswg.org/css-flexbox-1/#flex-common

**Expected behavior**

* The second list item should probably be changed to  
    `flex-shrink: 1`: items can shrink smaller than their flex-basis` 
* The sentence before the list seems incorrect and could be changed to "In the demo this unpacks to:"

remigius42:
Change the unpacking of `flex: 1` with respect to `flex-shrink` to
`flex-shrink: 1` to be consistent with the specification and fix wording
of the sentence referring to the list detailing the unpacked values.

For further details please refer to:

- https://drafts.csswg.org/css-flexbox-1/#flex-common
- https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax

Fixes #5537

- Changing the explanation of the `flex: 1` unpacking from `flex-shrink: 0` to `flex-shrink: 1`.
- Rewording the sentence explaining the unpacking of `flex: 1`.

### DIFF
--- a/src/site/content/en/learn/css/flexbox/index.md
+++ b/src/site/content/en/learn/css/flexbox/index.md
@@ -317,12 +317,12 @@ as the space that is shared between the items is shared out _after_ each item is
 max-content size.
 So a large item will gain more space.
 To force all of the items to be a consistent size and ignore the size of the content change
-`flex:auto` to `flex: 1`. in the demo.
+`flex:auto` to `flex: 1` in the demo.
 
 This unpacks to:
 
 - `flex-grow: 1`: items can grow larger than their `flex-basis`.
-- `flex-shrink: 0`: items can't shrink smaller than their `flex-basis`.
+- `flex-shrink: 1`: items can shrink smaller than their `flex-basis`.
 - `flex-basis: 0`: items have a base size of `0`.
 
 Using `flex: 1` says that all items have zero size,


remigius42:
@googlebot I fixed it.

src/site/content/en/learn/css/flexbox/index.md
```diff
@@ -317,12 +317,10 @@ as the space that is shared between the items is shared out _after_ each item is
 max-content size.
 So a large item will gain more space.
 To force all of the items to be a consistent size and ignore the size of the content change
-`flex:auto` to `flex: 1`. in the demo.
-
-This unpacks to:
+`flex:auto` to `flex: 1`. In the demo, this unpacks to:
```
rachelandrew, line 320: the text is asking the reader to make the change in the demo, so this change doesn't make sense.

rachelandrew (changes requested):
the change to flex-shrink is right, the copy change changes the meaning of this, as the suggestion is for the reader to change the code in the example.

Then going on to explain the shorthand.

src/site/content/en/learn/css/flexbox/index.md
remigius42: Ah I see, I got confused by the period after `flex: 1`. How about
```suggestion
`flex:auto` to `flex: 1` in the demo.

This unpacks to:
```

rachelandrew: I don't think I can commit this @argyleink can you commit and approve this PR?

argyleink: i cant commit this, i think it's because it's on  @remigius42's branch?

remigius42: Yes thought it would be better if I fork the repository, add the suggestions and then open a PR from there. I think the merge is blocked due to the requested changes. For the next steps I would suggest:

- [ ] I amend [my commit](https://github.com/GoogleChrome/web.dev/pull/5539/commits/1834c313bf675555da4b74c1e0acf7d4ae25a1e5) with the [suggested change above](https://github.com/GoogleChrome/web.dev/pull/5539#discussion_r645692294) and change the commit message to

        docs(frontend): correct unpacking of `flex: 1` (#5537)

        Change the unpacking of `flex: 1` with respect to `flex-shrink` to
        `flex-shrink: 1` to be consistent with the specification and fix 
        punctuation of the sentence referring to the list detailing the
        unpacked values.

        For further details please refer to:

        - https://drafts.csswg.org/css-flexbox-1/#flex-common
        - https://developer.mozilla.org/en-US/docs/Web/CSS/flex#syntax

        Fixes #5537

    and force-push onto my fork of the repo, which should then update this pull request.
- [ ] By adding the `Fixes #5537` trailer, the bug would automatically be closed upon merge. If you prefer to do so yourself or would prefer [`closes` or something else for the reference](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), please let me know and I will change or remove the trailer.

Please note that I am more familiar with the team blue from Down Under when it comes to Git hosting. Therefore, I am not sure whether the force-push will automatically update the pull request, but at the same time I think it makes sense to keep the Git history clean. If you have more experience on that matter, I would be grateful, if you let me know.
Additionally, the [type `docs` and scope `frontend`](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) in the commit subject line are both educated guesses. Maybe using something like `learn-css` as scope would be more appropriate.

If I get a thumbs up on this comment, I will assume the two suggested changes are alright with you and implement them.

argyleink: yep, update your fork/brand and it'll update this PR 👍🏻  
working on a fork was the right way to go! 
you may be able to use the github ui to merge the above suggestion too, save yourself a force push. 

once the changes are reflected here, we'll look at merging. thanks for hacking on this!

remigius42: Perfect, thank you for the update. I [squashed the changes into one commit](https://github.com/GoogleChrome/web.dev/pull/5539/commits/669c0f784e40d4657247d17bd7d674841d29138c) and changed the conventional commit scope to `learn-css`. After the [requested changes](https://github.com/GoogleChrome/web.dev/pull/5539#pullrequestreview-676447383) are accepted by @rachelandrew, the merge should be good to go, which should also close the ticket [#5537](https://github.com/GoogleChrome/web.dev/issues/5537#). In order to make sure, that all changes are reflected as intended, I would leave it to either of you to confirm this by resolving this conversation.

I continued reading the course material and might have stumbled over two possible issues in the [following chapter about CSS Grid](https://web.dev/learn/css/grid/). However, I assume that both of you are quite busy and therefore I would like to ask first, whether you have the time to look into it and if so, what the preferred communication scheme would be, ie. should I compile multiple possible issues into one bug ticket combined with a pull request, do you want a separate bug ticket or just a pull request or if you prefer any other means of communication. One way to go would be, if I simply collected everything until I have read the complete course and afterwards open a pull request (with or without a bug ticket).

For one of the two issues regarding the CSS Grid chapter I will likely be unable to suggest a fix in a pull request, because it might involve extending the CodePens: by the [end of the chapter](https://web.dev/learn/css/grid/#moving-content-around), I think the referenced CodePen is not the one it should be, because the last two referenced CodePens are identical and by my understanding the last referenced CodePen does not feature any adjustments related to `justify-items`or `align-items`. Looking at [the existing CodePens](https://codepen.io/web-dot-dev/pens/public?grid_type=list&sort_col=created_at&sort_order=asc&cursor=ZD0xJm89MSZwPTEmdj01MTMyOTI1NA==), I would assume that the one in question is missing, because as far as I understand it, it would probably have been created around the 31st of March.

Thanks again for providing this course and best regards
Andreas

rachelandrew (approved):
lgtm :)

src/site/content/en/learn/css/flexbox/index.md
rachelandrew: Please do raise individual issues for anything you spot! It's easier to deal with one thing at a time rather than a whole list some of which by the time you raise them may have already changed elsewhere.

remigius42: Makes perfect sense, especially given that the [issue I detailed above about the CodePen mismatch](https://github.com/GoogleChrome/web.dev/pull/5539#discussion_r648021614) has already been reported as bug https://github.com/GoogleChrome/web.dev/issues/5490 :). I opened https://github.com/GoogleChrome/web.dev/issues/5581 regarding the second issue and opened the corresponding pull request https://github.com/GoogleChrome/web.dev/pull/5582.

The [build pipeline seems to have had a small glitch](https://github.com/GoogleChrome/web.dev/runs/2782050295?check_suite_focus=true) while checking the updated commit (at least I would be surprised, if the Markdown change would have been the cause of `Error: Input required and not supplied: firebaseServiceAccount`). Maybe it is a permission issue, but I cannot trigger the pipeline again manually. Therefore, if either of you could trigger it again, that would be perfect. Otherwise, if manually triggering is not an option, I could update the timestamp on my commit and force-push again, thereby initiating a new pipeline run.

argyleink (approved):
👍🏻

PR #5565, opened 2021-06-08
Add crossorigin attribute to preconnect hint on https://web.dev/css-web-vitals/

piotr-bajer:

https://web.dev/css-web-vitals/

The preconnect hint in the example does not have `crossorigin` attribute.
It is used to download fonts, according to MDN it should use `crossorigin` - https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#what_requests_use_cors


mihajlija:
@khempenius PTAL

khempenius:
Changes are in this PR: https://github.com/GoogleChrome/web.dev/pull/5713

PR #5576, opened 2021-06-09
Content edit of image cdn outro

z-svoboda-mob:

https://web.dev/image-cdns/#choosing-an-image-cdn

- This removes a line to look at the codelab below, the codelab described is no longer there. The only codelab relates to a previous paragraph about self-managed image cdns

Note how it says "several" while the codelab is only about the single self-hosted option.
Screen Shot 2021-06-09 at 10 08 49 AM

I don't think this is the ideal content edit, the article now ends abruptly. 
Please advise.


z-svoboda-mob:
Unfortunately I will not be able to sign the CLA, apologies.

PR #5582, opened 2021-06-09
docs(learn-css): correct reserved area keywords (#5581)

### LINKED ISSUE #5581: Possible error in "Learn CSS" course chapter 10 regarding reserved keywords for grid areas
**Describe the bug**
While reading the [Learn CSS chapter 10 about Grid](https://web.dev/learn/css/grid/), I came across the [reserved keywords for grid-template-areas](https://web.dev/learn/css/grid/#grid-template-areas), mentioning that only the name `span` would be invalid:

> The name can be anything you like other than the keyword `span`. Once all of your items are named, use the [`grid-template-areas`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas) property to define which grid cells each item will span. Each row is defined within quotes.

Based on my understanding of the [specification](https://drafts.csswg.org/css-grid/#named-lines), `auto` would be a reserved keyword as well.

**To Reproduce**
1. Go to https://web.dev/learn/css/grid/#grid-template-areas
2. Compare it to the spec at https://drafts.csswg.org/css-grid/#named-lines

**Expected behavior**

The first sentence of the paragraph should probably be changed to something like
> The name can be anything you like other than the keywords `auto` and `span`.

remigius42:
Additionally mention the keyword `auto` as being unavailable as a custom
grid area name, because `auto` is a keyword with special semantics.

For further details please refer to:

- https://drafts.csswg.org/css-grid/#named-lines

Fixes #5581

- The first sentence of the paragraph about reserved area keywords would be changed to
  > The name can be anything you like other than the keywords `auto` and `span`.

### DIFF
--- a/src/site/content/en/learn/css/grid/index.md
+++ b/src/site/content/en/learn/css/grid/index.md
@@ -644,7 +644,7 @@ footer {
 }
 ```
 
-The name can be anything you like other than the keyword `span`.
+The name can be anything you like other than the keywords `auto` and `span`.
 Once all of your items are named,
 use the
 [`grid-template-areas`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-areas)


argyleink (approved):
doesnt hurt to add an additional example sure 👍🏻

PR #5587, opened 2021-06-10
fix(frontend-learn-css): add notes to drop-shadow usage and text-shad…

### LINKED ISSUE #5586: [BUG] incorrect notes for text-shadow and drop-shadow in the shadows chapter of "Learn CSS"
**Describe the bug**

1. In the shadows chapter, all shadow values except `spread` value are said to be applicable for `drop-shadow`:

> The drop-shadow filter has the same values as box-shadow without a spread value. You can add as many shadows as you like, by adding multiple instances of drop-shadow values to the filter property. Each shadow will use the last shadow as a positioning reference point

Accroding to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow()#syntax), the `inset` keyword and `spread` value are not allowed for `drop-shadow()` . I think using both keyword and value together for explaining is more approachable.

2. Also it is said that `text-shadow` accepts all values but `inset` and `spread`:  
> The values for text-shadow are the same as box-shadow and in the same order. The only difference is that text-shadow has no spread value and no inset keyword.

In the quiz section, when you mark `inset` as one of the correct answers, it shows a wrong rationale:
> Try again! text-shadow has inset as well.

**To Reproduce**
1. Go to MDN docs for [drop-shadow](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow()#syntax) and [text-shadow](https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow#values)
2. Comapre with [the second paragraph of the Drop shadow](https://web.dev/learn/css/shadows/#drop-shadow) and [the second paragraph of the Text shadow](https://web.dev/learn/css/shadows/#text-shadow)
3. Answer the chapter quiz.

**Expected behavior**

1. The rationale for `text-shadow` should be changed to something like:

> Try again! `inset` is a **keyword** which is also unique to `box-shadow`.

2. This sentence:
> The drop-shadow filter has the same values as box-shadow without a spread value.

can change to which is more readable (keyword and value):
> The drop-shadow filter has the same values as box-shadow but the `inset` keyword and `spread` value are not allowed .

raminjafary:
Fixes #5586

### DIFF
--- a/src/site/content/en/learn/css/shadows/index.md
+++ b/src/site/content/en/learn/css/shadows/index.md
@@ -245,8 +245,7 @@ We cover CSS [filters](/learn/css/filters) in another module,
 but in short, filters allow you to apply multiple graphical effects to the pixels of an element.
 {% endAside %}
 
-The `drop-shadow` filter has the same values as `box-shadow` **without** a spread value.
-You can add as many shadows as you like,
+The `drop-shadow` filter has the same values as `box-shadow` **but** the `inset` keyword and `spread` value are not allowed. You can add as many shadows as you like,
 by adding multiple instances of `drop-shadow` values to the `filter` property.
 Each shadow will use the last shadow as a positioning reference point.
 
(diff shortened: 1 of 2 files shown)


argyleink (approved):
noice, I dig it

PR #5593, opened 2021-06-11
Fix default scope description to match spec

alancutter:

- Fix default scope description to match spec: https://www.w3.org/TR/appmanifest/#scope-member

### DIFF
--- a/src/site/content/en/progressive-web-apps/add-manifest/index.md
+++ b/src/site/content/en/progressive-web-apps/add-manifest/index.md
@@ -236,7 +236,7 @@ to the `<a>` tag. On Android, links with `target="_blank"` will open in a
 A few other notes on `scope`:
 
 * If you don't include a `scope` in your manifest, then the default implied
-  `scope` is the directory that your web app manifest is served from.
+  `scope` is the parent directory of your `start_url`.
 * The `scope` attribute can be a relative path (`../`), or any higher level
   path (`/`) which would allow for an increase in coverage of navigations
   in your web app.


jpmedley:
@petele Can you please confirm this.

src/site/content/en/progressive-web-apps/add-manifest/index.md
```diff
@@ -236,7 +236,7 @@ to the `<a>` tag. On Android, links with `target="_blank"` will open in a
 A few other notes on `scope`:
 
 * If you don't include a `scope` in your manifest, then the default implied
-  `scope` is the directory that your web app manifest is served from.
+  `scope` is the parent directory of your `start_url`.
```
petele, line 239: I don't see this in the spec, maybe I'm missing something. Can you clarify?

I *thought* the scope would be the directory where the PWA is served... 

For example:
```
scope: '/pwa/',
start_url: '/pwa/start.html',
```

The scope would be `/pwa/` and everything below that, and when launched it would be launched from `/pwa/start.html`.

petele (changes requested):
@alancutter - can you clarify here. I may be confused on this.

src/site/content/en/progressive-web-apps/add-manifest/index.md
alancutter: The web app manifest can be served from anywhere, doesn't have to be the same origin.
The scope defaults to the parent directory of the start_url.

Spec: https://www.w3.org/TR/appmanifest/#scope-member
> Set manifest["scope"] to the result of parsing "." with manifest["start_url"] as the base URL.

robdodson:
@alancutter friendly ping

alancutter:
Replied inline: https://github.com/GoogleChrome/web.dev/pull/5593#discussion_r655865921

robdodson:
Thanks @alancutter. Assigning to @petele to get his thoughts.

PR #5643, opened 2021-06-16
Add draft PR as a separate condition to pullapprove

### LINKED ISSUE #5632: Pullapprove Should Ignore Drafts
Pullapprove autoassigns reviewers when a PR is opened. It does so without regard to the draft status of the PR. This means we get review requests for PRs that are clearly not ready for review.

image

Pullapprove checks the draft status of a PR and does not send review requests for items in draft state. If pullapprove cannot do this, please direct me to their website and I'll open a feature request there.

NA

**Additional context**
I suspect that for this to work in every case, contributors will need to open the PRs as drafts when appropriate. I suspect that most of those who use draft status are doing so already. (The open PR button is a drop-down that gives you this option. I don't have an image.) 

I don't think we need it to work every time. If just DevRel folks take advantage of this, it would cut down on spam.

devnook:
Fixes #5632

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -11,9 +11,12 @@ version: 3
 # https://docs.pullapprove.com/config/pullapprove-conditions/
 pullapprove_conditions:
 # Don't tag in any reviewers if the PR is work in progress.
-- condition: "'WIP' not in title or not draft"
+- condition: '"WIP" not in title'
   unmet_status: pending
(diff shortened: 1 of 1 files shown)


devnook:
It seems that "or" was not supported in the "conditions" field - i tested with separate fields and it worked.

robdodson (approved):

robdodson:
good one, thank you @devnook !

PR #5652, opened 2021-06-17
fix(frontend-learn-css): typo of property's value

phulengo:

- animation-play-state: `pause` -> `paused`


argyleink (approved):
excellent catch, ty so much for taking the time to suggest the change here!

PR #5657, opened 2021-06-17
feat(i18n-virtual-collections): refactor authors and tags to use `i18n` data for titles and descriptions

MichaelSolati:
The purpose of the PR is to pull out the title and description of author and tag collections from their collection items and to an i18n yml.

The i18n yml will house the titles and descriptions for every author and tag for any and all languages we support.

The i18n paths are inferred `i18n.authors.${authorslug}.title` and are added right before they are used.

In this PR we will add the proper title/description for the language in places liked the paginated pages, however this will change in time to be more efficient especially as we refactor the cards.


robdodson:
@MichaelSolati this seems to be breaking the build:

```
Eleventy is building, please wait…
Problem writing Eleventy templates: (more in DEBUG output)
> Expected input to be a string
    TypeError: Expected input to be a string

        at parse (/home/runner/work/web.dev/web.dev/node_modules/strip-comments/lib/parse.js:14:11)
`TypeError` was thrown:
        at module.exports (/home/runner/work/web.dev/web.dev/node_modules/strip-comments/index.js:35:18)
        at limitText (/home/runner/work/web.dev/web.dev/src/site/_collections/pages.js:47:13)
        at Object.module.exports [as pages] (/home/runner/work/web.dev/web.dev/src/site/_collections/pages.js:145:16)
        at TemplateMap.getUserConfigCollection (/home/runner/work/web.dev/web.dev/node_modules/@11ty/eleventy/src/TemplateMap.js:453:47)
        at TemplateMap.initDependencyMap (/home/runner/work/web.dev/web.dev/node_modules/@11ty/eleventy/src/TemplateMap.js:238:54)
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (internal/process/task_queues.js:95:5)
        at async TemplateMap.cache (/home/runner/work/web.dev/web.dev/node_modules/@11ty/eleventy/src/TemplateMap.js:282:5)
        at async TemplateWriter._createTemplateMap (/home/runner/work/web.dev/web.dev/node_modules/@11ty/eleventy/src/TemplateWriter.js:170:5)
```

MichaelSolati:
Fuuuudge, I know what's happening... Ok, I'll fix that (that collection only builds on prod, so I guess I wouldn't have noticed it)

robdodson:
@MichaelSolati this seems like a pretty big change but there's no description for this PR. In general all PRs should have a description so the reviewer knows what they're supposed to be reviewing.

src/site/_collections/hooks/utils.js
```diff
@@ -20,10 +20,41 @@
...
 
 /**
- * @param {AuthorsItem[]|TagsItem[]} items
- * @return {TODO[]}
+ * @param {VirtualCollectionItem[]} items
+ * @param {string} path
+ * @param {string} [lang]
+ * @return {VirtualCollectionExpandedItem[]}
+ */
+const addFields = (items, path, lang = defaultLocale) => {
```
robdodson, line 32: why do you need to translate the title and description here as opposed to in the njk templates? Is this because of the cards?

src/site/_collections/authors.js
```diff
@@ -81,23 +58,38 @@ module.exports = (collections) => {
...
+    }
+
+    // Set created on date and updated date
+    if (elements.length > 0) {
+      date = elements.slice(-1).pop().data.date;
+      const tempUpdated = elements.slice(0, 1).pop().data.date;
+      if (date !== tempUpdated) {
+        updated = tempUpdated;
+      }
     }
```
robdodson, line 71: I see that we had this before but it's kind of confusing. What is it actually doing?

src/site/_collections/authors.js
```diff
@@ -38,37 +38,14 @@ module.exports = (collections) => {
 
   Object.keys(authorsData).forEach((key) => {
     const authorData = authorsData[key];
```
robdodson, line 40: If I understand correctly, it seems like the author's title and description are removed from here, and instead moved into the addFields hook and added at a later time, is that correct? If so, why do it that way?

src/site/_collections/hooks/utils.js
MichaelSolati: Yes it is. That's the primary use, though I suspect it makes it's easier this way to set the title of paginated pages from these virtual collections. I'll add a comment though.

src/site/_collections/authors.js
MichaelSolati: The title and description are not removed here per se, they used to exist in the `authorsData.json` and no longer exist there. Instead further down the pipeline we'll just use the key of the author and look up their name/title and description using our i18n filter. So we know an authors title will always be `i18n.authors.${key}.title`, so when we need the title we just pass that into the filter.

src/site/_collections/pages.js
```diff
@@ -116,43 +108,25 @@ module.exports = (collections) => {
     }
 
     const defaultUrl = getDefaultUrl(item.url);
+    const paginated = !!item.data.pagination;
+    const content =
+      paginated && !!item.data?.renderData?.description
+        ? item.data.renderData.description
```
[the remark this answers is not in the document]
robdodson: Chatted with Michael — We use renderData because we need to use the i18n filter in the frontmatter for these pages. We should refactor at least the author and tag pages to use eleventyComputed.

robdodson:
Before landing the changes to the author file we would need to give the writing team a heads up and we would need to rebase inflight PRs for them.

src/site/_data/tagsData.json
```diff
@@ -1,253 +1,82 @@
...
+  "performance": {},
+  "permissions": {},
+  "progressive-web-apps": {},
+  "privacy": {},
+  "puppeteer": {},
+  "rendering": {},
   "scale-on-web": {
-    "title": "Scale on web",
-    "description": "A collection of short case studies from India and Southeast Asia celebrating the success stories of Google partners across multiple aspects of modern web development. Learn how these partners achieved scale while building amazing web experiences and delighting their customers.",
     "hero": {
```
robdodson, line 61: We discussed making tags.yml the source of truth. We'll delete everything in this file, except for this one quirky thing. And if folks want to give tag pages heros, then we support that by unioning this object with tags.yml

src/site/content/es/authors/index.njk
```diff
@@ -0,0 +1,37 @@
+---
```
robdodson, line 1: I think having these .njk files is going to be an issue because for every language we will need to duplicate this file, and if we make changes to the en version, we have to remember to change it in every language.

On d.c.c. we have all of our .njk files in the _includes directory and site/content only contains markdown files. Could we do that here for these files?

src/site/content/es/authors/index.11tydata.js
```diff
@@ -0,0 +1,23 @@
+/*
```
robdodson, line 1: hm, it feels not great needing to duplicate these .11tydata.js files for every language...

src/site/_collections/authors.js
MichaelSolati: Not on the author pages (though we could), but it is used by for algolia and eventually firebase indexing for push notifications.

src/site/_filters/filter-by-lang.js
```diff
@@ -1,11 +1,25 @@
...
+module.exports = function filterByLang(posts, lang = defaultLocale) {
+  /** @type Map<string, EleventyCollectionItem> */
+  const filteredPosts = new Map();
+  for (const post of posts) {
+    if (post.data && [lang, defaultLocale].includes(post.data.lang)) {
+      const defaultUrl = getDefaultUrl(post.url);
+      if (filteredPosts.has(defaultUrl) && post.data.lang === lang) {
+        filteredPosts.set(defaultUrl, post);
+      } else if (!filteredPosts.has(defaultUrl)) {
+        filteredPosts.set(defaultUrl, post);
```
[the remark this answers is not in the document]
MichaelSolati: Before the filter would filter only posts belonging to the language passed in. As a result, if we applied the spanish filter, only spanish articles would show up. I think we've since decided we want to show english (default language) articles and if there is a version of the article in the selected language then we show that one instead of the english version.

So basically this function now shows all articles of the default language, except when we have a translated article, then we display that one instead.

src/site/_filters/filter-by-lang.js
```diff
@@ -1,11 +1,25 @@
...
- * @param {Array} posts An array of posts.
- * @param {string} lang Target language.
+ * @param {EleventyCollectionItem[]} posts An array of posts.
+ * @param {string} [lang] Target language.
+ * @return {EleventyCollectionItem[]}
  */
-module.exports = function filterByLang(posts, lang) {
-  if (!lang) {
-    return posts;
+module.exports = function filterByLang(posts, lang = defaultLocale) {
```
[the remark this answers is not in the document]
MichaelSolati: Ah the comment above explains it better.

src/site/_includes/components/AuthorInfo.js
```diff
@@ -18,7 +18,7 @@ const {html} = require('common-tags');
 
 /* eslint-disable require-jsdoc */
 
-module.exports = ({author, id, showSocialMedia = false}) => {
+module.exports = ({author, title, showSocialMedia = false}) => {
```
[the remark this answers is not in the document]
MichaelSolati: It's used at the very end of the function.

src/site/content/es/authors/index.njk
MichaelSolati: Ok, I just broke them into layouts, which works just fine for this!

robdodson (commented):
LGTM but waiting on team feedback before unlocking.

robdodson (approved):

PR #5658, opened 2021-06-17
Blog post on new responsiveness metric

npm1:

- Add blog post on ideas for new responsiveness metric
- Add authors entry for myself


mihajlija (approved):

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,326 @@
...
+
+1. The input from the user occurs. The time at which this occurs is the event’s `timeStamp`.
+2. The browser performs hit testing to decide which HTML frame (main frame or some iframe) an event belongs to. Then the browser sends the event to the appropriate renderer process in charge of that HTML frame.
+3. The renderer receives the event and queues it so that it can process when it becomes available to do so.
+4. The renderer processes the event by running its handlers. These handlers may queue additional asynchronous work, such as `setTimeout` and fetches, that are part of the input handling. But at this point, the synchronous work is complete.
+5. A frame is painted to the screen that reflects the result of event handlers running. Note that any asynchronous tasks queued by the event handlers may still be unfinished.
+
+The time between steps (1) and (3) above is an event's ***delay***, which is what FID measures.
+
+The time between steps (1) and (5) above is an event's ***duration***. This is what our new metric will measure.
```
mihajlija, line 92:
```suggestion
The time between steps (1) and (3) above is an event's _delay_, which is what FID measures.

The time between steps (1) and (5) above is an event's _duration_. This is what our new metric will measure.
```

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,326 @@
...
+
+For these types of interactions, we’d like the latency to involve the durations of all events associated with them. Since event durations for each "down" and "up" part of the interaction can overlap, the simplest definition of interaction latency that achieves this is the maximum duration of any event associated with it. Referring back to the keyboard diagram from earlier, this would be the `keydown` duration, as it is longer than the `keyup`:
+
+{% Img src="image/kns0INkO77RkiEStzHWYrugyWj32/WFPm4W86CqhFsNc1UZuW.jpeg", alt="Keyboard interaction with maximum duration highlighted", width="800", height="311" %}
+
+The `keydown` and `keyup` durations may overlap as well. This may happen for instance when the frame presented for both events is the same, as in the following diagram:
+
+{% Img src="image/kns0INkO77RkiEStzHWYrugyWj32/e7htYAZb44AW4UeplBwJ.jpeg", alt="Keyboard interaction where press and release occur in the same frame", width="800", height="311" %}
+
+There's are pros and cons to this approach of using the maximum, and we're interested in [hearing your feedback](#Feedback):
```
mihajlija:
```suggestion
There's are pros and cons to this approach of using the maximum, and we're interested in [hearing your feedback](#feedback):
```

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,326 @@
...
+
+In the diagram above, the durations are disjoint because the frame from `keydown` updates is presented before the `keyup` occurs, but this does not need to be the case always. In addition, note that a frame can be presented in the middle of a task in the renderer process since the last steps required to produce the frame are done outside of the renderer process.
+
+The `keydown` and `keypress` occur when the user presses the key, while the `keyup` occurs when the user releases the key. Generally the main content update occurs when the key is pressed: text appears on the screen, or the modifier effect is applied. That said, we want to capture the more rare cases where `keyup` would also present interesting UI updates, so we want to look at the overall time taken.
+
+In order to capture the overall time taken by the keyboard interaction, we can compute the maximum of the duration of the `keydown` and the `keyup` events. The `keypress` event is deprecated, and it should be fired in the same task as the `keydown` event, so its duration will always overlap with `keydown`'s.
```
mihajlija, line 204:
```suggestion
In order to capture the overall time taken by the keyboard interaction, we can compute the maximum of the duration of the `keydown` and the `keyup` events. The `keypress` event is deprecated, and it should be fired in the same task as the `keydown` event, so its duration will always overlap with the duration of the`keydown`.
```

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+
+{% DetailsSummary %}
+A note on measuring TTI in the field
+{% endDetailsSummary %}
+
+Measuring TTI on real users in the field is problematic because it occurs very late in the page
+load. A 5-second network quiet window is required before TTI can even be computed. In the lab, you
+can choose to unload the page whenever you have all the data that you need, but that’s not the case
+with real-user monitoring in the field. A user may choose to leave the page or interact with it at
+any time. In particular, users may choose to leave pages that take a long time to load, and TTI will
```
mmocny, line 101: Nit: `and TTI will not be recorded` --> `and accurate TTI will cannot be recorded`.

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+not be recorded in those cases. When we measured TTI for real users in Chrome, we found that only
+about half of page loads reached TTI. {% endDetails %}
+
+## What improvements are we considering?
+
+Given the shortcomings of FID described above, we would like to develop a new metric that extends
```
mmocny, line 107: I don't really feel like the shortcomings of FID are clearly described above...

We talk a lot about why it was better than alternatives, then compare it to lab metrics, and only in the second half of the last paragraph do we touch on it only being "portion of the experience".

The description of FID does have a section (`FID does not include the time spent running those event handlers, nor any work done by the browser afterwards to update the screen.`) but that just feels like a description.  Its not necessarily obvious this is undesirable.

And, I don't think it is emphasized at all why it is an issue to only captures the first input, for example.  (this should be self-evident, but we probably should call it out with its own header?)

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+what FID measures today yet still retains its strong connection to user experience.
+
+We want the new metric to:
+
+1. Consider the responsiveness of all user inputs (not just the first one), and capture each event's
```
mmocny, line 112: Could this be split into two points?  I feel both are meaty enough.

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+   do so.
+4. The renderer processes the event by running its handlers. These handlers may queue additional
+   asynchronous work, such as `setTimeout` and fetches, that are part of the input handling. But at
+   this point, the synchronous work is complete.
+5. A frame is painted to the screen that reflects the result of event handlers running. Note that
+   any asynchronous tasks queued by the event handlers may still be unfinished.
+
+The time between steps (1) and (3) above is an event's _delay_, which is what FID measures.
+
+The time between steps (1) and (5) above is an event's _duration_. This is what our new metric will
```
mmocny, line 149: These two lines are brilliant.  They should be bold.

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+that is meant to take a long time to be finished. Thus, our initial effort will look at step 5 as
+the end point: it will only consider synchronous work and the amount of time it takes to paint after
+such work is completed. That is, we’re not going to apply heuristics to guess the work that would be
+kicked off asynchronously in step 4 in our initial effort.
+
+It’s worth noting that, in many cases, work should be executed synchronously. In fact, this may be
+unavoidable because events are sometimes dispatched one after the other and the event handlers need
+to be executed in order. That said, we will still miss important work, like events which trigger
+fetching or which rely on important work to be done at the next `requestAnimationFrame` callback,
+for example. {% endDetails %}
```
mmocny, line 178: One more point: getting that sync portion+first paint to finish quickly is still uniquely important, even in situations of async follow-on work.  Typically there is *some* initial visual indication of an interaction (blue links change color, buttons look pressed, etc)

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+        <td><code>keypress</code></td>
+        <td><code>keypress</code></td>
+      </tr>
+      <tr>
+        <td>Key released</td>
+        <td><code>keyup</code></td>
+        <td><code>keyup</code></td>
+      </tr>
+      <tr>
+        <td rowspan="7">Tap / drag</td>
```
mmocny, line 227: Nit: `Tap / Drag` feel like this is a single interaction type with alternative names, or two steps.

I think this is meant to be two logical rows in the table, which happen to have the same events?
Would it be too much effort to make it two rows?

mmocny: ...looking at the full table, it's probably too much effort.

Still, I would consider doing something like `<b>Tap</b> or, <b>Drag</b>`

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+required to produce the frame are done outside of the renderer process.
+
+The `keydown` and `keypress` occur when the user presses the key, while the `keyup` occurs when the
+user releases the key. Generally the main content update occurs when the key is pressed: text
+appears on the screen, or the modifier effect is applied. That said, we want to capture the more
+rare cases where `keyup` would also present interesting UI updates, so we want to look at the
+overall time taken.
+
+In order to capture the overall time taken by the keyboard interaction, we can compute the maximum
+of the duration of the `keydown` and the `keyup` events. The `keypress` event is deprecated, and it
```
mmocny, line 302: This detail about keypress feels like it should be in brackets, or some addendum?

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+##### Tap
+
+Another important user interaction is when the user taps or clicks on a website. Similar to
+`keypress`, some events are fired as the user presses down, and others as they release, as shown in
+the diagram above, Note the events associated with a tap are a little different on desktop vs
+mobile.
+
+For a tap or click, the release is generally the one which triggers the majority of reactions, but,
+as with keyboard interactions, we want to capture the full interaction. And in this case it’s more
+important to do so because having some UI updates upon tap press is not actually that uncommon.
```
mmocny, line 327: I'm not sure this paragraph helps.

For one thing, tap press side-effects are VERY common.  Like every single form element common?

For another, this is backwards from above, which is easy to overlook in the wording, as there end up being a string of logical double negatives (`unlike with the above... the (opposite of the above) is actually import... because its not uncommon`)

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+`mouseup`, and `click` wait for the delay before being dispatched. This means that if we only looked
+at `pointerdown` and `pointerup` then we’d miss the duration from the synthetic events, which is
+large due to the browser tap delay and should be included. So we should measure `pointerdown`,
+`pointerup`, and `click` to cover the full interaction. {% endDetails %}
+
+##### Drag
+
+We decided to include dragging as well since it has similar associated events and since it generally
+causes important UI updates to sites. But for our metric we intend to only consider the drag start
+and the drop - the initial and final parts of the drag. This is to make it easier to reason about as
```
mmocny, line 354: Nit: do all drags lead to "drops"?  I think many drags are custom scrollers?  Perhaps just call it `drag start and stop`?

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+We’re also not considering drags implemented via the [Drag and Drop
+API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API) because they only work
+on desktop.
+
+##### Scrolling
+
+One of the most common forms of interacting with a site is via scrolling. For our new metric, we’d
+like to measure the latency for the initial scrolling interaction of the user. In particular, we
+care about the initial reaction of the browser to the fact that the user requested a scroll. In
+particular, we will not cover the whole scrolling experience. That is, scrolling produces many
```
mmocny, line 367: Two `In particular`s in a row

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+frames, and we’ll focus our attention on the initial frame produced as a reaction to the scroll.
+
+Why just the first one? For one, subsequent frames may be captured by a separate smoothness
+[proposal](https://docs.google.com/presentation/d/1VwGIzypntWNosCTXWMsUI6ifw4sEKSRQgnwx3P_wqVg/edit#slide=id.p).
```
mmocny, line 371: 😊

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+experiences that have clear points in time associated with them and for which we can easily compute
+their latency. Scrolling as a whole is a continuous experience, so we do not intend to measure all
+of it in this metric.
+
+So why measure scrolls? The scrolling performance we’ve gathered in Chrome shows that scrolling is
+generally very fast. That said, we still want to include initial scroll latencies in our new metric
+for various reasons. First, scrolling is fast only because it has been optimized so much, because it
+is so important. But there are still ways for a website to bypass some of the performance gains that
+the browser offers. The most common one in Chrome is to force scrolling to happen on the main
+thread. So our metric should be able to say when this happens and causes poor scrolling performance
```
mmocny, line 384: Nit: `So out metric should be able to say` --> `We would like the responsiveness metric to clearly detect`

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+The `keydown` and `keyup` durations may overlap as well. This may happen for instance when the frame
+presented for both events is the same, as in the following diagram:
+
+{% Img src="image/kns0INkO77RkiEStzHWYrugyWj32/e7htYAZb44AW4UeplBwJ.jpeg", alt="Keyboard interaction
+where press and release occur in the same frame", width="800", height="311" %}
+
+There's are pros and cons to this approach of using the maximum, and we're interested in [hearing
+your feedback](#feedback):
+
+* **Con**: It does not capture the full wait time of the user. For instance, it will capture the
```
mmocny, line 421: I feel this Con is so much less important than the Pro's.  I hope I am not just biased here, but I'm trying to evaluate all of the considerations for a metric and evaluating the various tradeoffs.  Here is how I would put it:

1: Represents what users actually see/feel
2. Is not affected by User behaviour (like physically long interactions with device)
3. Consistent with other metric definitions (simplicity)
4. In cases of interactions with multiple components which are each painful, captures cases with most overall user pain.


Regarding (1), I would say that using `max` has one more pro that isn't listed here:
`Whenever there is a frame painted between the components of one interaction, it does not feel representative of UX to consider any time longer than any one of the parts.  (And, if there is only one frame, max already covers the full duration for both.)`

Regarding (4), it can be conflicting with (1) and (3).

Finally, (4) feels conflicting with the overall aggregation/normalization strategy, to me.

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+* **Pro**: It aims to reduce noise for cases like keyboard interactions, where the `keyup` usually
+  does nothing and where the user may execute the key press and release quickly or slowly.
+
+For scrolling (which just has a single associated event) we’d like to define its latency as the time
+it takes for the browser to produce the first frame as a result of scrolling. That is, the latency
+is the delta between the event `timeStamp` of the first DOM event (like `touchmove`, if using a
+finger) that is large enough to trigger a scroll and the first paint which reflects the scrolling
+taking place.
+
+### Aggregate all interactions per page
```
mmocny, line 434: To be honest... this section reads like it was an afterthought of the blogpost, which I guess it is.

The text below which explores some options and asks for feedback, I'm not sure are fleshed out enough to be actionable.  Not sure, but, I wonder if we can just say something like "Next steps: Aggregating these interactions into a single metric value, stay tuned!"

src/site/content/en/blog/better-responsiveness-metric/index.md
```diff
@@ -0,0 +1,508 @@
...
+
+```js
+let maxTapOrDragDuration = 0;
+let maxKeyboardDuration = 0;
+const observer = new PerformanceObserver(list => {
+  list.getEntries().forEach(entry => {
+    switch(entry.name) {
+      case "keydown":
+      case "keyup":
+        maxKeyboardDuration = Math.max(maxKeyboardDuration,
```
mmocny, line 489: This is returns a _normalized max_, across all interactions?  I worry that this syntax may confuse folks to assume its the max across one interaction...

I know you highlight that interactionID is not available, but, I wonder if therefore it's better to just console.log all the durations instead of printing a single max?

(Maybe that also depends if you take my suggestion to drop aggregation topic from the post)

mmocny (commented):
I left a lot of feedback, but a lot of it is truly optional.

This is a phenomenal post!

src/site/content/en/blog/better-responsiveness-metric/index.md
npm1: These seem like a rewording of the existing text, except for (1) and it's not clear to me that (1) is particular to max. I did move the con to last, but other than that kept this.

src/site/content/en/blog/better-responsiveness-metric/index.md
npm1: It's not an afterthought. I already mention that `We’re exploring and evaluating several options.` to make it clear that this is ongoing work. I do see how it's not fleshed out enough to be actionable, but it's important to note this crucial step in the metric definition so I'm keeping this section.

src/site/content/en/blog/better-responsiveness-metric/index.md
npm1: This is the actionable bit from the page-aggregation section :) It's easy to log all durations, but less clear to me how you'd convert those into a number.

anniesullie:
@googlebot I consent.

debugtive0517:
@googlebot I consent.

PR #5660, opened 2021-06-18
Fix DEMO link and capture_links values

sisidovski:
I noticed the DEMO link URL is wrong and `capture_links` values here are written with snake case. https://web.dev/declarative-link-capturing/
@tomayac Could you take a look at this?


tomayac (approved):
LGTM, approval

Good catch!

PR #5679, opened 2021-06-21
Obfuscate error message

### LINKED ISSUE #5513: web.dev/measure error message leaks API key
**Describe the bug**
The API key used by web.dev/measure is leaked in an error message. (I would recommend it be regenerated once this bug is fixed.)

**To Reproduce**
1. Go to https://web.dev/measure/
2. Look up any URL during an API outage.
3. See API key in error

**Expected behavior**
The error message should not contain Google's internal API key.

**Screenshots**
image

devnook:
Fixes #5513

### DIFF
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -60,7 +60,7 @@ export const requestRunLighthouse = store.action((state, url) => {
   })();
 
   return p.catch((err) => {
-    const errMsg = (err.name === 'FetchError') ? err.name : err.toString();
+    const errMsg = err.name === 'FetchError' ? err.name : err.toString();
     console.warn('failed to run Lighthouse', url, errMsg);
 
     const update = {
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -60,10 +60,11 @@ export const requestRunLighthouse = store.action((state, url) => {
   })();
 
   return p.catch((err) => {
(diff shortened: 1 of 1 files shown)


MichaelSolati (approved):

src/lib/actions.js
```diff
@@ -60,10 +60,11 @@ export const requestRunLighthouse = store.action((state, url) => {
   })();
 
   return p.catch((err) => {
-    console.warn('failed to run Lighthouse', url, err);
+    const errMsg = err.name === 'FetchError' ? err.name : err.toString();
+    console.warn('failed to run Lighthouse', url, errMsg);
 
     const update = {
-      lighthouseError: err.toString(),
+      lighthouseError: errMsg,
```
robdodson, line 67: is lighthouseError every logged? It seems like errMsg still does err.toString() in some instances so I'm not sure I understand how this fixes the issue?

robdodson: maybe a better question:

What was causing the issue in the first place? Are only FetchErrors displaying the key or do other kinds of errors?

devnook: LighthouseError is actually displayed in the UI. In case of FetchError, it means there was a problem in fetch from Pagespeed API, and the url (including key) of Pagespeed api is included in the error message. This is not the case for other errors which originate from Ligthouse service (and not pagespeed).

PR #5692, opened 2021-06-22
📝 Update Flexbox assessment

vegerot:
When first reading the first question, I thought it may have been a trick question.  This commit adds clarification that the Flexbox is using the initial flex values in an `ltr` language

### DIFF
--- a/src/site/content/en/learn/css/flexbox/conclusion.assess.yml
+++ b/src/site/content/en/learn/css/flexbox/conclusion.assess.yml
@@ -2,7 +2,14 @@ setLeader: Test your knowledge of flexbox
 height: unset
 tabLabel: question
 questions:
-  - type: multiple-choice
+  - stimulus: |
+      ```css
+      .container {
+        display: flex;
+        direction: ltr;
+      }
+      ```
+    type: multiple-choice
     cardinality: "1"
     correctAnswers: "0"
     stem: To vertically align with flexbox, use
@@ -11,7 +18,14 @@ questions:
         rationale: "Nice"
       - content: "justify keywords"
         rationale: "Sorry"
-  - type: multiple-choice
+  - stimulus: |
+      ```css
+      .container {
+        display: flex;
+        direction: ltr;
+      }
+      ```
+    type: multiple-choice
     cardinality: "1"
     correctAnswers: "1"
     stem: To horizonally align with flexbox, use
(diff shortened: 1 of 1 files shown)


src/site/content/en/learn/css/flexbox/flex.assess.yml
```diff
@@ -1,4 +1,4 @@
-setLeader: Test your knowledge of flexbox
+setLeader: Test your knowledge of flexbox. (Note: The following questions all assume `flex-direction: row`)
```
argyleink, line 1:
```suggestion
setLeader: Test your knowledge of flexbox: `display: flex`
```
what do you think of this? make it explicit that it's a default flex container in context?

vegerot: I like that.  It still opens the door to vagueness (but how many people other than me thought it was vague?).

To be the most specific, we could have a sub(-sub)-title that gives the full class definition

```
Test your knowledge of flexbox with the following example
```
```css
.container {
  display: flex;
}
```
image


But now I feel like I'm stretching a bit.  This may be over-explaining it to the reader, and it may be confusing to specify things like it's in an *ltr* language.

argyleink: I think you're making a stronger case, and I agree that the current quiz is making too many assumptions about the reader 🙂 

the quiz's support contextual code snippets as a stimulus: thoughts on this?

Kapture 2021-06-22 at 13 22 45

```yml
stimulus: |
  ```css
  .container {
    display: flex;
    direction: ltr;
  }
  ```
```

vegerot: 💯👌

I'll add that

vegerot: @argyleink I didn't test it locally, but I followed another [example of `stimulus`](https://github.com/GoogleChrome/web.dev/blob/068563f3e4eee82a8e62fa25b2b9ea049acb92d0/src/site/content/en/accessible/labels-and-text-alternatives/self-assessment.assess.yml#L12-L18) on the site, so 🤞

vegerot: Netlify-bot says the preview deployment failed.  Did I mess something up?

vegerot: @argyleink lookin' good :)

vegerot: Last nitpick, we need to add the same `stimulus` to the third question, since the question specifically mentions "height"

argyleink: > Last nitpick, we need to add the same `stimulus` to the third question, since the question specifically mentions "height"

nice, i'm down. let's change it to: "If you want content size used for child items" and let the height answer continue to be wrong?

vegerot: A few of the `rationale`s still mention `height`, so I'll put the `stimulus` there too

vegerot: Updated

argyleink (approved):

PR #5697, opened 2021-06-22
refactor(cards): break up cards into different purposeful macros, fixes #2717

### LINKED ISSUE #2717: Refactor cards component to be more flexible.

MichaelSolati:
Fixes #2717

5e77wy

- Remove BaseCard, PostCard, and AuthorCard shortcodes.
- Create 3 card macros, one for posts, one for authors, and one for tags.
- Update places where shortcodes were used to use relevant macros.


src/site/content/en/authors/index.njk
```diff
@@ -27,7 +28,7 @@ pagination:
   <section class="w-grid">
     <div class="w-grid__columns w-grid__columns--three" role="list">
       {% for element in paged.elements %}
-        {% AuthorCard element %}
+        {{ authorcard(element) }}
       {% endfor %}
```
robdodson, line 32: I'm curious if this would read better as:

```njk
{% for author in paged.elements %}
  {{ authorcard(author) }}
{% endfor %}
```

Since we're inside of authors/index.njk, it seems like there's enough context to use the word 'author' in the argument.

src/site/_includes/partials/paged.njk
```diff
@@ -1,7 +1,9 @@
+{% from 'macros/postcard.njk' import postcard with context %}
+
 <section class="w-grid">
   <div class="w-grid__columns w-grid__columns--three" role="list">
     {% for element in paged.elements %}
-      {% PostCard element %}
+      {{ postcard(element) }}
     {% endfor %}
```
robdodson, line 7: Is this always a post? If so I wonder if it would read better as:

```
{% for post in paged.elements %}
  {{ postcard(post) }}
{% endfor %}
```

src/site/content/en/tags/index.njk
```diff
@@ -24,5 +25,13 @@ pagination:
       <p class="w-page-header__copy">{{ description }}</p>
     </header>
   </div>
-  {% include 'partials/paged.njk' %}
+  <section class="w-grid">
+    <div class="w-grid__columns w-grid__columns--three" role="list">
+      {% for element in paged.elements %}
+        {{ tagcard(element) }}
+      {% endfor %}
```
robdodson, line 32: Similar point as above about doing `for tag in paged.element`

src/site/_includes/macros/authorcard.njk
```diff
@@ -0,0 +1,31 @@
+{% macro authorcard(collectionItem) %}
```
robdodson, line 1: See comment below about changing this argument from `collectionItem` to `post|tag|author` depending on which card macro we're using.

src/site/_includes/macros/authorcard.njk
```diff
@@ -0,0 +1,31 @@
...
+
+{% set url = '/authors/'+collectionItem.key %}
```
robdodson, line 3: It looks like the old shortcode was using `collectionItem.url` for urls. Is that not possible here?

src/site/_includes/macros/authorcard.njk
```diff
@@ -0,0 +1,31 @@
...
+{% set data = collectionItem.data %}
+{% set thumbnail = data.thumbnail or data.hero or null %}
+{% set alt = data.alt or '' %}
+
+<div class="w-card w-card-author" role="listitem">
+  <article class="w-card-base">
+    <div class="w-card-base__cover w-card-base__cover--with-image }}">
+      <a href="{{url}}" class="w-card-base__link" tabindex="-1" aria-hidden="true">
```
robdodson, line 11: Would you mind adding a comment here that say: "Because we are not adding descriptive alt text to this image, and because the link is duplicated when we link the author's name below, we chose to hide this link from screenreaders to make their navigation a bit more efficient."

src/site/_includes/macros/postcard.njk
```diff
@@ -0,0 +1,44 @@
...
+{% set thumbnail = data.thumbnail or data.hero or null %}
+{% set alt = data.alt or '' %}
+{% set isDraft = 'w-card--draft' if data.draft %}
+{% set authors = data.authors or [] %}
+
+<div class="w-card {{className}} {{isDraft}}" role="listitem">
+  <article class="w-card-base {{ 'w-card-base--featured' if featured }}">
+    {% if thumbnail %}
+      <div class="w-card-base__cover w-card-base__cover--with-image }}">
+        <a href="{{url}}" class="w-card-base__link" tabindex="-1" aria-hidden="true">
```
robdodson, line 14: Would you add a comment here saying that we intentionally hide the thumbnail image from screen readers because it appears out of context from the rest of the article content and we suspect it might be confusing. Since the link for the article is duplicated in the heading down below, we think this might be an OK tradeoff.

src/site/_includes/macros/tagcard.njk
```diff
@@ -0,0 +1,34 @@
...
+{% set url = '/tags/'+collectionItem.key %}
+{% set data = collectionItem.data %}
+{% set thumbnail = data.thumbnail or data.hero or null %}
+{% set alt = data.alt or '' %}
+
+<div class="w-card" role="listitem">
+  <article class="w-card-base">
+    {% if thumbnail %}
+      <div class="w-card-base__cover w-card-base__cover--with-image }}">
+        <a href="{{url}}" class="w-card-base__link" tabindex="-1" aria-hidden="true">
```
robdodson, line 12: Similar to the authorcard, can you add a comment explaining that we hide the image from screen readers here because we often don't provide descriptive alt text.

src/site/_includes/macros/postcard.njk
```diff
@@ -0,0 +1,51 @@
...
+{% set isDraft = 'w-card--draft' if data.draft %}
+{% set authors = data.authors or [] %}
+
+<div class="w-card {{className}} {{isDraft}}" role="listitem">
+  <article class="w-card-base {{ 'w-card-base--featured' if featured }} {{ 'w-card-base__with--cover' if thumbnail }}">
+    {% if thumbnail %}
+      <div class="w-card-base__cover">
+      {#
+        We intentionally hide the thumbnail image from screen readers because it appears out of context
+        from the rest of the article content and may confusing. Since the link for the article is duplicated
```
robdodson, line 23:
```suggestion
        from the rest of the article content and may be confusing. Since the link for the article is duplicated
```

src/site/_includes/macros/tagcard.njk
```diff
@@ -0,0 +1,41 @@
...
+{% set thumbnail = data.thumbnail or data.hero or null %}
+{% set alt = data.alt or '' %}
+
+<div class="w-card" role="listitem">
+  <article class="w-card-base {{ 'w-card-base__with--cover' if thumbnail }}">
+    {% if thumbnail %}
+      <div class="w-card-base__cover">
+        {#
+          We intentionally hide the thumbnail image from screen readers because it appears out of context
+          from the rest of the article content and may confusing. Since the link for the article is duplicated
```
robdodson, line 17:
```suggestion
          from the rest of the article content and may be confusing. Since the link for the article is duplicated
```

src/site/_includes/macros/tagschip.njk
```diff
@@ -0,0 +1,14 @@
+{#
+  @typedef {string[]} tags - An array of tag keys to display.
+  @property {string} [className] - A space separated list of classes to apply.
+#}
+{% macro tagschip(tags, className = '') %}
```
robdodson, line 5: Would you mind camel casing this and all other macros. I think we do that for other macros on the site like audioPlayer() and audioFab().

Also, could you rename this to tagChips instead of tagsChip. I think the UI element that we're rendering is called a chip, and we're rendering a set of them so—"tag chips".

robdodson (approved):
LGTM after doing the camel case rename

PR #5733, opened 2021-06-29
✏️ [Learn CSS] Fix typo in Border section

vegerot:
The attached CodePen (https://codepen.io/web-dot-dev/pen/poRraBp) uses `2px` instead of `1px`.


argyleink (approved):
excellent catch, thank you very much!

PR #5745, opened 2021-07-01
📝 [Learn CSS] Add Podcast link to conclusion

vegerot:


robdodson:
@argyleink are you cool with adding this?

argyleink:
hm yeah! at first i thought i wouldnt be cool with it, but after investigating and listening, it does make sense and fits well. curious what @una thinks?

vegerot:
Yeah.  The Conclusion "lesson" briefly describes what you've learned so far, and the "Season 1 summary" podcast episode briefly describes **each** episode in season 1.  So I think that listening to the summary episode is a great way to reinforce and remember everything learned so far.

rachelandrew:
@una @argyleink given we now have more than season 1 covered, I'm not sure this is an ideal addition. Can you confirm either way?

argyleink:
i still think it's nice to break it up and have a recap episode. but it's "nice to have", so i could go either way.

rachelandrew (approved):

PR #5758, opened 2021-07-05
fix typo in bullet list article

donroyco:
This fixes a small typo on [New CSS functional pseudo-class selectors :is() and :where()](https://web.dev/css-is-and-where/) page.

```
`:is()` takes the specificity of it's most specific selector.
```
»
```
`:is()` takes the specificity of its most specific selector.


rachelandrew (approved):
thank you!

PR #5770, opened 2021-07-07
 Add translation disclaimer to fallback pages

devnook:


devnook:
This is still causing a CLS, so I plan to fiddle with the design a bit.

MichaelSolati (approved):

PR #5806, opened 2021-07-15
Decrease margin on .w-stats container to get rid of CLS in the CLS article

### LINKED ISSUE #5766: CLS article has CLS. Oh the irony.
This article has some CLS on mobile.
https://web.dev/telegraph/

I think it's because there is some element causing horizontal scroll. When you scroll down on mobile, the appearance of the horizontal scroll bar causes the header to shift up.

tannerdolby:
Fixes #5766 

- Reduces margin for `.w-stats` container from `margin: 32px -32px` to `margin: 24px -24px`
- Note: Instead of jumping directly from 32px to 24px, a media query could be introduced to update the margin for mobile viewports as this is when things start to overflow.

### DIFF
--- a/src/styles/components/_stats.scss
+++ b/src/styles/components/_stats.scss
@@ -18,7 +18,7 @@
   display: flex;
   flex-direction: column;
   justify-content: space-around;
-  margin: 32px -32px;
+  margin: 24px -24px;
   padding: 64px 32px;
   text-align: center;
 


tannerdolby:
This should do the trick to fix CLS happening on this page! It's odd the CLS even happened in the first place, but indeed the root of the problem stemmed from the large margins on the `.w-stats` container.

robdodson (approved):

robdodson:
awesome, thank you @tannerdolby !

robdodson:
Just waiting on the tests to run and then i'll merge.

tannerdolby:
your welcome @robdodson !

robdodson:
ah I forgot that we have an issue where Percy tests don't run properly on forks, but I think that should be fine. Thank you again!

tannerdolby:
ah ok I see! Feel free to ping me if I need to push any changes to help the Percy tests pass/run properly. Your welcome!

PR #5809, opened 2021-07-15
Add i18n language directories scaffold

devnook:
Adds languages: Korean, Chinese, Japanese, Russian, Portuguese


robdodson (approved):
I think we can merge this to unblock the localization work (btw, should we remove Polish?)

But seeing all of the duplicated 11tydata files does have me a bit concerned that we should move more stuff into `_data`. I could see us changing the en/en.11tydata.js to reorganize some learning paths stuff, and forgetting to do that for other languages.

PR #5811, opened 2021-07-15
Add Global CSS and core utilities

andy-piccalilli:
This is first cut at global CSS—mainly aimed at HTML elements. Because of the already existing theme system, this CSS also supports light and dark modes. 

All the elements that have been styled can be seen at: https://deploy-preview-5811--web-dev-staging.netlify.app/design-system/temp-styleguide/

I've added a couple of utilities to make presenting these global styles easier. The rest of those will follow next! 

Edit: I'm also adding some core utilities to this because you only live once.


andy-piccalilli:
The focus ring utility uses box shadow to keep radius (we can switch it out when chromium honours radius) but also has a transparent outline for high contrast mode. 

Focus ring in windows high contrast mode

src/site/content/en/design-system/css-utilities.njk
```diff
@@ -1,12 +1,97 @@
 ---
 title: 'CSS Utilities'
+permalink: '/design-system/css-utilities/index.html'
 ---
 
 {% extends "design-system.njk" %}
 
+{% set content %}
```
andy-piccalilli, line 8: There's almost certainly a better way to do this, but for now, it'll get this bit of documentation where it needs to be.

robdodson: Not sure I understand the problem but could you have a separate css-utilities markdown file that contains this content and uses css-utilities.njk as its layout?

src/site/content/en/design-system/css-utilities.njk
```diff
@@ -1,12 +1,97 @@
...
+```
+
+## Focus ring
+
+A box-shadow powered focus ring that uses a transparent outline to provide
+focus for high contrast users. The box-shadow helps to support border-radius
+elements. 
+
+```html
+<button class="[ my-cool-button ] [ focus-ring ]">My cool button</button>
```
robdodson, line 45: I'm wondering if we want to do the square bracket thing for classes or not. We don't do anything like it on d.c.c. so I can imagine it being a little confusing for team members context switching between the two sites.

src/site/content/en/design-system/css-utilities.njk
```diff
@@ -1,12 +1,97 @@
...
+```
+
+## Wrapper 
+
+A horizontally centered wrapper that provides a consistent central column.
+This should be used in all contexts where the content needs to be in the
+center of the page.
+
+```html 
+<div class="wrapper">
```
robdodson, line 85: Random thing I noticed - none of these snippets seem to have syntax highlighting. Not a huge deal, but I was curious if you knew why that was?

robdodson (approved):

robdodson:
Sorry it took me a couple of days to get to this! Feel free to ping me on chat if you're waiting for a review. I'd like to make sure I'm unblocking you as quickly as possible, but sometimes I don't see all the GH notifications flying by.

src/site/content/en/design-system/css-utilities.njk
andy-piccalilli: Ah yes! I've created a `design-system-documentation.njk` layout because there'll be a few pages like this. This fixes the syntax highlighting too 🙌

src/site/content/en/design-system/css-utilities.njk
andy-piccalilli: Ah yeh I'll get rid of them.

src/site/content/en/design-system/css-utilities.njk
andy-piccalilli: Fixed with the above

andy-piccalilli:
No worries at all @robdodson. Most stuff won't be too dependent ongoing, but I'll definitely ping you for important ones

PR #5828, opened 2021-07-21
Update nginx.md

armanism24:
Added tutorial for HTTP to HTTPS redirection in NGINX because the old method is not working now. It occurs "ERR_TOO_MANY_REDIRECTS".

Fixes #SOME_ISSUE_NUMBER

- 
- 
-


mihajlija:
Hi @armanism24, thank you bringing this to our attention! I've checked the instructions on StackExchange, and though you were getting "ERR_TOO_MANY_REDIRECTS", it doesn't necessarily make the StackExchange thread wrong. There's a lot of variation in how nginx can be configured; different versions of nginx can also influence which instructions work. Since the StackExchange page includes several other working solutions, we'll keep that link for the time being.

PR #5865, opened 2021-07-22
feat(serverless): enable 11ty serverless for local development

MichaelSolati:
Fixes #SOME_ISSUE_NUMBER

- 
- 
-


src/build/output-permalink.js
```diff
@@ -37,23 +37,24 @@ const {defaultLocale} = require('../site/_data/site');
...
   }
 
   if (data.permalink) {
-    return data.permalink.replace(
+    permalink = data.permalink.replace(
       /^\/{{lang}}/,
       localization ? '/i18n/{{lang}}' : '',
     );
+  } else {
+    permalink = data.page.inputPath
```
samthor, line 53: What doest this do? The default permalink doesn't work and it needs to be provided?

src/site/_data/eleventyComputed.js
```diff
@@ -0,0 +1,17 @@
...
+    if (Object.keys(data).length === 0) {
+      return;
+    }
+    const build = outputPermalink(data);
+    const serverless = path.join(
+      '/serverless',
+      build.replace('index.html', ''),
+    );
+
+    return {build, serverless};
```
samthor, line 15: How does this work with serving this locally? Surely permalink is basically broken for normal development now?

PR #5872, opened 2021-07-23
Change `Flexbox` to `flexbox` 

kosamari:
This PR is a minor edit to older post for consistency. Change `Flexbox` to `flexbox` per spec language.


kosamari:
Note: the `master` word is part of URL that we can't change

kosamari:
ah, this was duplicate!

Issue #5875, opened 2021-07-25
Href values of the links at the "Table of Contents" section are not processed by i18n


shapkarin:
Hello. There are wrong anchors at `Table of Contents`. Looks like in attribute `href` data is not parsed in a right way and used just like a string. I guess It should be parsed by i18n with the provided key.
At this nunjucks template link href value is `not to call i18n` with the `topic.title` value: https://github.com/GoogleChrome/web.dev/blob/517699486a138a527d0edc6a3250ada425d4a7f9/src/site/_includes/path.njk#L56
Example of the `topic.title` value itself is:
https://github.com/GoogleChrome/web.dev/blob/068563f3e4eee82a8e62fa25b2b9ea049acb92d0/src/site/_data/paths/accessible.json#L9

**Steps to reproduce the behavior**
1. Go to https://web.dev/accessible
2. Click on `Understand the diverse needs of users`
3. Yours browser will get that URL fragment (hash) `i18n.paths.accessible.topics.understand_the_diverse_needs_of_users` and URL becomes https://web.dev/accessible/#i18n.paths.accessible.topics.understand_the_diverse_needs_of_users 
4. Page is not scrolled to the `Understand the diverse needs of users` topic

**Expected behavior**
1. Go to https://web.dev/accessible
2. Click on `Understand the diverse needs of users`
3. Yours browser will get that fragment `understand-the-diverse-needs-of-users` and URL will be https://web.dev/accessible/#understand-the-diverse-needs-of-users

**Screenshots**
After click to the TOC link:
<img width="1552" alt="Screenshot 2021-07-25 at 07 59 12" src="https://user-images.githubusercontent.com/1463086/126888264-f37c86ea-8c36-4873-a198-02583c209bc5.png">
How it should be after the click:
<img width="1552" alt="Screenshot 2021-07-25 at 07 58 53" src="https://user-images.githubusercontent.com/1463086/126888268-7b4b37f0-2ed0-44ec-8105-bf36f023f4c2.png">

PR #5878, opened 2021-07-25
Fix TOC links for the learn pages

shapkarin:
Related issue with the detailed bug description – #5875

### DIFF
--- a/src/site/_includes/path.njk
+++ b/src/site/_includes/path.njk
@@ -53,7 +53,7 @@ renderData:
       <ul class="w-icon-list">
         {% for topic in topics %}
           <li class="w-icon-list__item w-icon-list__item--check">
-            <a class="w-path-link" href="#{{ topic.title | slug }}">{{topic.title | i18n(locale)}}</a>
+            <a class="w-path-link" href="#{{ topic.title | i18n(locale) | slug }}">{{topic.title | i18n(locale)}}</a>
           </li>
 	      {% endfor %}
       </ul>


shapkarin:
@devnook, @robdodson maybe it will be nice to have a smooth scroll for such links? CSS property `scroll-behavior: smooth` can help with that. Or it can be achieved with JS `Element.scrollIntoView({ behavior: 'smooth' });`. Maybe to add that with an other PR?

Regarding to the current issue as you can check with the stage environment, bug with navigation is fixed and TOC links works well for any learn page. More info about the bug can be found at #5875 

#### Behavior at this PR's branch
- for example go to the https://deploy-preview-5878--web-dev-staging.netlify.app/accessible
- click on any link from the "Table of Contents" list
- you will be forwarded with scroll to the topic that you select at the previous step

#### How it acts at the `main` branch
- for example open https://web.dev/accessible
- click on any link from the TOC section
- URL now includes a key for the i18n translate
- there is no any kind of navigation

samthor:
Thanks for the fix. I'm happy to merge this.

re: smooth scrolling, it seems excessive to add JS for this. We could add the CSS property, but I'd rather defer to our designers.

shapkarin:
> Thanks for the fix. I'm happy to merge this.
> 
> re: smooth scrolling, it seems excessive to add JS for this. We could add the CSS property, but I'd rather defer to our designers.

OK. Let's defer to designers.

PR #5883, opened 2021-07-26
Correct URL component encoding example in URLPattern article.

jeremyroman:
A single character seems to have been lost along the way.


jeremyroman:
@jpmedley @jeffposnick This corrects a typo the URLPattern blog post (#5780).

jeffposnick (approved):
Thanks for catching that!

robdodson (approved):

PR #5889, opened 2021-07-27
refacor(firebase): migrate to v9 of firebase library, fixes #5884

### LINKED ISSUE #5884: Upgrade Firebase to take advantage of lighter version
We consistently get dinged by Lighthouse and PageSpeed for having a slow time to interactive. This is caused by our dependency on firebase.

image

My understanding is that the newest version is broken up into smaller bundles and we can only require the bits that we need. We should upgrade to see if this improves our TTI score.

MichaelSolati:
Fixes #5884

- Remove firebase loader script
- Use firebase v9 beta imports

I measured a reduction in JavaScript loaded (old method included injecting script tags, so I just added the size of all the scripts loaded for Firebase) from 161 KB to 67 KB.

Also the "Total Blocking Time" went from 20 ms to 0 ms.

Here is a [WebPageTest](https://webpagetest.org/result/210727_AiDcAF_670278a1ca7d610932bfc5562f6fecc8/?view=webvitals) of the site presently, before landing this PR, as a record of the old performance.


robdodson (approved):

robdodson:
looking good! ship it!

PR #5908, opened 2021-08-02
Theme switcher

andy-piccalilli:
This is the theme switcher that toggles dark/light mode, overriding the initial `prefers-color-scheme` setting that is honoured by default. The setting is stored in and loaded from local storage.

In `head.njk`, the little snippet of JavaScript applies the setting _before_ the CSS is loaded to prevent a flash of incorrect theme, when a user has a setting stored in local storage. 

In terms of UI on the toggle itself, I've just applied some really basic CSS so it doesn't look rubbish. We need to work out how the toggle will look, globally, so I'll tag in @jimoong for some thoughts on that. It's using `<button>` elements to toggle right now, but we could also lean into an `<input type="checkbox" role="switch" />` to support something more "toggle"-like. 

I added it to the `base.js` because we will need it on every page, ongoing. I doubt it'll have much of an impact on overall weight on pages using the now old CSS stuff. 

For now, I've just applied it to the temp styleguide: https://deploy-preview-5908--web-dev-staging.netlify.app/design-system/temp-styleguide/

### DIFF
--- a/src/lib/components/ThemeToggle/index.js
+++ b/src/lib/components/ThemeToggle/index.js
@@ -1,0 +1,93 @@
+class ThemeToggle extends HTMLElement {
+  constructor() {
+    super();
+
+    this.STORAGE_KEY = 'user-color-scheme';
+    this.COLOR_MODE_KEY = '--color-mode';
+  }
+
+  connectedCallback() {
+    // Allows --flow-space to work and we
+    // only want that to happen when it's
+    // wired up
+    this.style.display = 'block';
+
+    this.render();
+  }
+
+  getCSSCustomProp(propKey) {
+    let response = getComputedStyle(document.documentElement).getPropertyValue(
+      propKey,
+    );
+
+    // Tidy up the string if there’s something to work with
+    if (response.length) {
+      response = response.replace(/'|"/g, '').trim();
+    }
+
+    // Gorko will probably set this as a `var()` reference,
+    // we need to strip that stuff out
+    return response.replace('var(--color-').replace(')');
+  }
+
+  applySetting(passedSetting) {
+    // Attempts to load the setting from local storage
+    const currentSetting =
+      passedSetting || localStorage.getItem(this.STORAGE_KEY);
+
+    if (currentSetting) {
+      this.setStatus(currentSetting);
+      window.applyThemeSetting(currentSetting);
+    } else {
+      // Loads the value based on the CSS custom property value instead, with a fallback of light
+      const customPropValue =
+        this.getCSSCustomProp(this.COLOR_MODE_KEY) || 'light';
+      this.setStatus(customPropValue);
+      window.applyThemeSetting(customPropValue);
+    }
+  }
+
+  setStatus(currentSetting) {
+    // Loop the buttons and set aria-pressed (active) state depending on current theme
+    this.buttons.forEach((button) => {
+      button.setAttribute(
+        'aria-pressed',
+        currentSetting === button.getAttribute('data-theme') ? 'true' : 'false',
+      );
+    });
+  }
+
+  render() {
+    this.innerHTML = `
+      <div class="theme-toggle" aria-label="Change display theme">
+        <button class="button" data-theme="light">Light</button>
+        <button class="button" data-theme="dark">Dark</button>
+      </div>
+    `;
+
+    this.afterRender();
+  }
+
+  afterRender() {
+    this.buttons = this.querySelectorAll('[data-theme]');
+
+    // Loop each button to attach the toggle event
+    this.buttons.forEach((button) => {
+      button.addEventListener('click', (evt) => {
+        evt.preventDefault();
+
+        const setting = button.getAttribute('data-theme');
+        this.applySetting(setting);
+        localStorage.setItem(this.STORAGE_KEY, setting);
+      });
+    });
+
+    this.applySetting();
+  }
+}
+
+if ('customElements' in window) {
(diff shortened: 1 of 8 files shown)


andy-piccalilli:
A little supplementary screencast about UI @jimoong https://www.loom.com/share/8206a903bcb543b0a736fabacceaec65

I wouldn't let this block the PR being merged for now though. I reckon we apply the UI in another PR 👍

robdodson:
I personally really like the first example you showed where it had a clear label and a bit of iconography within the switch itself. I think having a label for something like this is important as it's still a newer feature and it prevents any ambiguity around what it does.

src/lib/components/ThemeToggle/index.js
```diff
@@ -0,0 +1,114 @@
...
+      );
+    });
+  }
+
+  render() {
+    this.innerHTML = `
+      <div class="theme-toggle" aria-label="Change display theme">
+        <button class="button" data-theme="light">Light</button>
+        <button class="button" data-theme="dark">Dark</button>
+      </div>
```
robdodson, line 86: What I've done in other places is to have this content in the light DOM, and the custom element adds the listeners when it upgrades. The upside is that you can style everything and avoid any layout shift. It does create the issue that technically you have buttons on screen that don't do anything if the JS fails to load, but I think that's an OK tradeoff for the improved layout shift metric in CWV.

src/lib/components/ThemeToggle/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+    this.STORAGE_KEY = 'user-color-scheme';
+    this.COLOR_MODE_KEY = '--color-mode';
+  }
+
+  connectedCallback() {
+    // Allows --flow-space to work and we
+    // only want that to happen when it's
+    // wired up
+    this.style.display = 'block';
```
robdodson, line 34: I left a comment down in the render function which I think is related to this. I was wondering if we can go ahead and have all of the html and CSS for the element stamped out already, and in connectedCallback we only do the afterRender() stuff.

robdodson (approved):
LGTM with one comment about maybe moving the innerHTML stuff into light DOM

jimoong:
I agree that we need the label and also like the idea of using icon to reinforce it. But not quite sure about using icons in the toggle itself. It looks cool and I'd like to have it :) but considering other UI controls in web.dev are somewhat 'pure' looking, feeling a bit hesitant with having something too special. 
I'm curious about where we put the control cause we don't have a proper surface to put the site controls. Same question we had for the language selector and we ended up putting it on the footer. We can do the same thing, but it's terrible for discovery, so I brought back the exploration that we had for the profile card and tried a quick mock. Using some basic controls.

Screen Shot 2021-08-03 at 10 46 03 AM

robdodson:
oh I like the idea of a settings icon for color mode and language!

robdodson:
btw @andy-piccalilli I was never able to get the theme switcher to do the flicker that you mentioned. But if it turns up I'm sure we can sort it out.

src/lib/components/ThemeToggle/index.js
andy-piccalilli: Yeh I'll switch this to use a switch-style pattern (how many times can I say switch) based on what we've discussed about UI and pass it as light DOM to avoid layout shift. 

How about this for a strategy to avoid CLS and avoid rendering a control that might not work:

1) We pass the input and label as light DOM but with `visibility: hidden` 
2) This means it won't show up for sighted users _and_ SR users 
3) We set `visibility: visible` when it's ready

It'll still render the element and assign it space in the content flow if we do that

andy-piccalilli:
Yeh really loving that idea @jimoong! I'll create us a generic `.switch` block that can account for that UI for us.

src/scss/blocks/_theme-toggle.scss
```diff
@@ -0,0 +1,20 @@
+/// THIS IS VERY MUCH A TEMPORARY
+/// BIT OF CSS UNTIL WE SLING SOME
+/// PROPER UI DESIGN ON THE THEME TOGGLE
+.theme-toggle {
```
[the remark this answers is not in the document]
andy-piccalilli: p sure we can just delete this now 🎉

src/lib/components/ThemeToggle/index.js
andy-piccalilli: Just realised the above is unnecessary because [we're doing this approach anyway](https://github.com/GoogleChrome/web.dev/pull/5908#issuecomment-892039762)

robdodson (approved):

soffieprincess02:
> This is the theme switcher that toggles dark/light mode, overriding the initial `prefers-color-scheme` setting that is honoured by default. The setting is stored in and loaded from local storage.
> 
> In `head.njk`, the little snippet of JavaScript applies the setting _before_ the CSS is loaded to prevent a flash of incorrect theme, when a user has a setting stored in local storage.
> 
> In terms of UI on the toggle itself, I've just applied some really basic CSS so it doesn't look rubbish. We need to work out how the toggle will look, globally, so I'll tag in @jimoong for some thoughts on that. It's using `<button>` elements to toggle right now, but we could also lean into an `<input type="checkbox" role="switch" />` to support something more "toggle"-like.
> 
> I added it to the `base.js` because we will need it on every page, ongoing. I doubt it'll have much of an impact on overall weight on pages using the now old CSS stuff.
> 
> For now, I've just applied it to the temp styleguide: https://deploy-preview-5908--web-dev-staging.netlify.app/design-system/temp-styleguide/

PR #5910, opened 2021-08-02
Switch to block level image.

robdodson:
I noticed this image looked a bit odd on the live article. I think the layout that floats an image on either side of the text doesn't work super great on our site.

image

This tweaks the article to use a block image instead. @jpmedley what do you think?

image


jpmedley:
We're waiting on a samples repo to be moved to merge [a new version](https://github.com/GoogleChrome/web.dev/pull/5545) of this section. Merging this would likely create a merge conflict.

samthor:
It seems like the other PR is merged now. @robdodson can you merge this?

robdodson:
yep, just need to fix the conflict

robdodson:
oh this actually looks correct now. I think they changed the layout when they did the section update.

image

PR #5930, opened 2021-08-05
Add patterns data structure (pseudo-collection)

devnook:
This adds a basic CodePatterns infrastructure to web.dev (without the pattern shortcode yet).


src/lib/patterns.js
```diff
@@ -0,0 +1,71 @@
...
+const matter = require('gray-matter');
+const md = require('markdown-it')();
+
+const stripDot = /^\./;
+const basePath = 'src/site/content/en/patterns';
+const files = glob.sync(path.join(basePath, '**', 'index.md'));
+
+const ids = new Set();
+
+const allPatterns = files
```
samthor, line 28: Can we make this build as part of calling the function (potentially caching it if necessary)? I think we struggle with stack traces a lot in our tooling already, and doing this at "load" is probably going to break people's brain in future.

src/lib/patterns.js
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+const glob = require('fast-glob');
+const fs = require('fs');
+const path = require('path');
+const matter = require('gray-matter');
+const md = require('markdown-it')();
+
+const stripDot = /^\./;
+const basePath = 'src/site/content/en/patterns';
```
samthor, line 23: I'd probably do something like:

```js
const basePath = path.join(__dirname, '../../src/site/content/en/patterns');
```

... in case this is run from somewhere not the top-level folder.

samthor: I'm also curious about plans for non-en, but I'm also happy to hear "none" for now, in that we have no plans to translate API docs either.

src/lib/patterns.js
```diff
@@ -0,0 +1,71 @@
...
+const files = glob.sync(path.join(basePath, '**', 'index.md'));
+
+const ids = new Set();
+
+const allPatterns = files
+  .map((file) => {
+    const id = path.relative(basePath, path.dirname(file));
```
samthor, line 30: This basically seems like we're checking that no two files have the same path. Is that right? Is this helpful? I feel like it'll never fire for any real file-system.

src/lib/patterns.js
```diff
@@ -0,0 +1,71 @@
...
+      out[basename] = {
+        content,
+        type,
+        name: basename,
+      };
+      return out;
+    }, {});
+
+    const content = md.render(fileContents.content);
+    const demo =
```
samthor, line 55: I think it's not obvious that "demo.html" is the default file being shown, or you can set the "demo" property in the data structure. I commented in the doc too, but a comment here will help.

src/lib/patterns.js
```diff
@@ -0,0 +1,71 @@
...
+  .map((file) => {
+    const id = path.relative(basePath, path.dirname(file));
+    if (ids.has(id)) {
+      throw new Error(`Duplicate pattern id found: ${id}`);
+    }
+    ids.add(id);
+
+    const fileContents = matter(fs.readFileSync(file, 'utf-8'));
+    if (fileContents.data.layout !== 'pattern') {
+      return;
```
samthor, line 38: Should this be a mistake?

src/site/_filters/patterns.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Finds a pattern by the pattern's id in the paaterns data object.
```
samthor, line 18: paaterns => patterns

src/site/_filters/patterns.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {string} id Pattern Id
+ * @param {Array} patterns Patterns data object
+ * @return {Object} Pattern object
+ */
+const getPattern = (id, patterns) => {
+  return patterns.find((item) => item.id === id);
```
samthor, line 24: I'm super nitpicking but this method implies that it should be stored in a dictionary, not a list. But we have O(1) things so ¯‍\‍_‍(‍ツ‍)‍_‍/‍¯

samthor (commented):
I know we're not at the `<iframe>` stage yet but I wonder if we should serve these cross-origin just for safety (e.g., "patterns.web.dev").

robdodson:
When we were working on Learn CSS I created an 11tydata.js file which would set the page permalinks to false if it was a production build so that things wouldn't go live until the feature was complete. I think we should do the same thing for `/en/patterns/`.

robdodson:
I just tested this, here are the two changes you'd need to add.

```js
// src/site/content/en/patterns/patterns.11tydata.js
module.exports = {
  eleventyComputed: {
    permalink: (data) => {
      if (process.env.ELEVENTY_ENV === 'prod') {
        return false;
      } else {
        return (data.page.filePathStem + '.html').replace(/^\/en\//, '/');
      }
    },
  },
};
```

```njkk
{# src/site/_includes/pattern.njk #}

---
layout: default
---

{% if page.url %}
  {% set patternId = page.url | getRelativePath('patterns') %}
  {% set pattern = patternId | getPattern(patterns) %}

  {{ content | safe }}
{% endif %}
```

src/lib/patterns.js
samthor: Ok, we talked about this. Can you call out that `fileContents.data.demo`, if set, is expected to be an external URL or something like that?

src/lib/patterns.js
devnook: None for now.

src/lib/patterns.js
devnook: As discussed, let's add caching to json separately.

src/lib/patterns.js
devnook: Right

src/lib/patterns.js
devnook: No, it just allows us to skip files that are not pattern files.

src/site/_filters/patterns.js
devnook: We want patterns to resemble eleventy collections, so it has to behave like a list. I think replicating it as a dictionary for quick access by key is not critical here.

src/site/_filters/patterns.js
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Finds a pattern by the patterns id in the paaterns data object.
+ * @param {string} id Pattern Id
+ * @param {Array} patterns Patterns data object
+ * @return {Object} Pattern object
```
MichaelSolati, line 21: Can the pattern object get a typescript interface to be used to declare what the object should look like?

src/site/_filters/patterns.js
MichaelSolati: Ignoring if it's critical or not, I don't know if having a collection as an array is that important.

1) I think 11ty does it only because the built in collections are just filtering all content based on tags, so that makes sense in a way.
2) 11ty pagination accepts objects also, not just arrays.
3) Collections such as shows, authors, and tags are objects (key/values), so we already have dict like collections.

I guess I'm just saying the whole array collection isn't really necessary and I don't think we need to resemble 11ty's built in collections.

src/site/_includes/pattern-suite.njk
```diff
@@ -0,0 +1,14 @@
+---
+layout: default
+---
+
+{{ content | safe }}
+
+<ul>
+{% for pattern in patterns %}
+  {% if pattern.suite == page.url %}
+    <li><a href="/patterns/{{pattern.id}}">{{ pattern.id }}</a></li>
```
MichaelSolati, line 10: Could a pattern object have its own url as a property? Not necessary, just wondering if it would make sense.

MichaelSolati (approved):
Have a few thoughts, though I'm not adamant about anything I've commented. I don't want to be a stick in the mud. So I'm happy with it landing as it is.

src/site/_includes/pattern-suite.njk
devnook: I think pattern id is a good approximation of a url. Also we collect patterns from files, so theorethically the do not have urls assigned to them yet (e.g. this depends on the eleventy build later on).

src/site/_filters/patterns.js
devnook: Ok, let's try to have it as an object.

src/site/_filters/patterns.js
devnook: Added in /lib/patterns.js

PR #5937, opened 2021-08-06
Basic design system UI

andy-piccalilli:
This is just a basic UI for the design system so it's no longer blocking the next step: patterns build. 

Right now, it's just using a sidebar composition, but ongoing, it'll be nice to implement the responsive burger-style system we already have going. 

here's a loom to show you it in action: https://www.loom.com/share/cd03477b1dc442e7a4ab50253a665295

Design system UI


src/site/_filters/md.js
```diff
@@ -16,7 +16,13 @@ module.exports = (content, opts) => {
     md.set(opts);
   }
 
-  return content.split('\n').length > 1
-    ? md.render(content)
-    : md.renderInline(content);
+  let inline = !content.includes('\n');
+
+  // If there's quite a bit of content, we want to make sure
```
robdodson, line 21: I need to double check if this breaks anything on the site. Our screenshot coverage with Percy isn't very comprehensive.

robdodson: reading through the code again, this _seems_ ok... let's go with it and tinker if something looks off.

robdodson (approved):

src/site/_filters/md.js
andy-piccalilli: Yeh I reckon it'll be fine. It's just catching cases where you get like a paragraph of text that accidentally gets no formatting

PR #5951, opened 2021-08-09
Add CodePattern shortcode

devnook:
This uses a hardcoded pattern data for now (blocked on #5930 )


package.json
```diff
@@ -126,6 +126,7 @@
     "node-fetch": "^2.6.1",
     "npm-run-all": "^4.1.5",
     "postcss": "^8.2.10",
+    "prism-js": "^1.0.0",
     "prismjs": "^1.24.1",
```
robdodson, line 130: I'm a bit confused as it looks like we already depend on `prismjs`

src/site/_includes/components/CodePattern.js
```diff
@@ -0,0 +1,66 @@
...
+    },
+    demo: '/patterns/example-pattern/demo.html',
+    suite: '/patterns/',
+  };
+
+  const assets = Object.values(pattern.assets)
+    .map((asset) => {
+      const content = Prism.highlight(
+        asset.content,
+        Prism.languages.html,
```
robdodson, line 33: should this be Prism.languages[asset.type] ? Possibly done in some kind of try/catch with a fallback to 'text' if there's an error?

src/site/_includes/components/CodePattern.js
```diff
@@ -0,0 +1,66 @@
...
+        'html',
+      );
+      return html`
+        <web-tab title="${asset.type}">
+          <pre>
+            <code class="language-${asset.type}">${content}</code>
```
robdodson, line 39: Can you try a multiline code example. My concern is that `pre` preserves whitespace, so it seems like the way this is formatted will lead to weird indentation.

src/site/content/en/handbook/web-dev-components/index.md
```diff
@@ -37,6 +37,16 @@ guidance about how to use them effectively.
 1. [Tooltips](#tooltips)
 1. [Video](#video)
 
+## Code Patterns
```
robdodson, line 40: This needs to be added to the table of contents above. The headings are alphabetized so this item should come before the Code Pen shortcode in the doc

src/site/_includes/components/CodePattern.js
robdodson: It looks like the pre element is including the whitespace, the first line of html is shifted to the right a bit.

image

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+.code-pattern {
+  width: 100%;
+
+  .code-pattern__content {
+    display: grid;
+    gap: 1rem; // sass-lint:disable-line no-misspelled-properties
+    grid-template-columns: 50% auto;
+    justify-items: stretch;
+    height: 300px;
+    margin-bottom: 1rem;
```
robdodson, line 19: you can remove this. Since you are using the `flow` utility on the parent, it's already pushing the demo button down.

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+    overflow: scroll;
```
robdodson, line 20: You can remove this. `overflow:scroll` creates visible gutters on macOS even if there is no overflow.

image

You can use `overflow: auto` to make them go away. But in this case I don't think you want this element to have any overflow, it seems like you'd probably want the actual demo and the tabs to have the overflow.

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+//
+// =============================================================================
+
+.code-pattern {
+  width: 100%;
+
+  .code-pattern__content {
+    display: grid;
+    gap: 1rem; // sass-lint:disable-line no-misspelled-properties
+    grid-template-columns: 50% auto;
```
robdodson, line 16:
```suggestion
    grid-template-columns: 1fr 1fr;
```

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+    justify-items: stretch;
```
robdodson, line 17: I don't think you need this? The tabs seem to stretch with the iframe if is height increases.

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+    height: 300px;
```
robdodson, line 18: Probably don't want this here. I think we may instead give the iframe an explicit size (and maybe make it configurable by the author somehow?) and let that define the overall height for the container.

src/styles/components/_code-pattern.scss
```diff
@@ -0,0 +1,37 @@
...
+    iframe,
+    web-tabs {
+      height: 100%;
+      width: 100%;
+    }
+  }
+
+  .w-button {
+    text-transform: none;
+  }
```
robdodson, line 36: I think we would leave it with the uppercase styling that other buttons have on the site. I think that may be different from Jiwoong's mock, but that might be because he forgot. I think newer Material buttons don't do the uppercase thing, but the old ones did and we should probably be consistent with the rest of our UI.

src/styles/components/_code-pattern.scss
devnook: It is causing a CLS if there's no height defined.

robdodson: right I meant you would give the iframe the explicit size, not the parent

robdodson: I guess this works too if you want to leave it like this. But we'll need to come up with a way for this to be configurable by the author. One option is for the shortcode to accept a size which it then applies to the element as an inline style

robdodson (approved):

PR #5964, opened 2021-08-11
Add placeholder learn/forms section

robdodson:

- Adds the learn/forms course
- Includes a feature flag in 11tydata.js so the course will not be published to production
- Changes the meta.yml format a bit so it doesn't contain a bunch of implementation details
- Fixes a bug where adding a second course was causing the course progress to be overwritten in localstorage
- Stubs out a bit of documentation explaining how to add a course. I'll fill this out in a follow up PR so that this one doesn't get any longer 😅


src/site/_data/i18n/courses.yml
```diff
@@ -4,4 +4,9 @@ learn_css:
...
   description:
     en: "Let’s gleefully breakdown complex aspects of CSS into digestible episodes covering everything from accessibility to z-index."
-    es: "Dividamos alegremente los aspectos complejos de CSS en episodios digeribles que abarquen todo, desde la accesibilidad hasta el z-índex."
\ No newline at end of file
+    es: "Dividamos alegremente los aspectos complejos de CSS en episodios digeribles que abarquen todo, desde la accesibilidad hasta el z-índex."
+learn_forms:
+  title:
+    en: Learn Forms!
+  description:
+    en: "..."
```
samthor, line 12: Is this a placeholder or literally supposed to be dots?

src/site/_data/courses/forms/meta.yml
```diff
@@ -0,0 +1,19 @@
...
+date: 2021-08-11
+url: /learn/forms/
+draft: true
+title: i18n.courses.learn_forms.title
+description: i18n.courses.learn_forms.description
+# This is the logo that appears in the navigation drawer and header.
+logo: '/images/courses/css/logo.svg'
+# This is used to provide the background for the navigation drawer and header.
+background: '/images/courses/css/background.svg'
+# This provides the cover image for the card on /learn/
```
samthor, line 16: This is kind of interesting - we have images that are clearly meant to be in local `` tags, but some which are meant to be served through imgix. This is probably fine bit a bit ambiguous.

src/site/content/en/learn/forms/example/index.md
```diff
@@ -0,0 +1,382 @@
+---
```
samthor, line 1: I'm not going to review this course. I presume it's a copy of something else.

src/site/content/en/learn/forms/index.md
```diff
@@ -0,0 +1,45 @@
...
+course from MDN to learn all about how to write markup and link your stylesheets.
+
+Here's what you'll learn:
+
+{% include 'partials/course-index.njk' %}
+
+So, are you ready to learn CSS? [Let's get started](/learn/css/box-model/).
+
+{% Aside %}
+This course is written and edited by CSS community experts: <a href="https://twitter.com/piccalilli_">Andy Bell</a>, <a href="https://twitter.com/rachelandrew">Rachel Andrew</a>, <a href="https://twitter.com/Una">Una Kravets</a>, and <a href="https://twitter.com/argyleink">Adam Argyle</a>. <a href="https://twitter.com/rob_dodson">Rob Dodson</a> and <a href="https://twitter.com/jiwoong">Jiwoong Lee</a> shepherded the UX & production build, along with help from <a href="https://twitter.com/devnook">Ewa Gasperowicz</a> and <a href="https://twitter.com/piccalilli_">Kayce Basques</a>.
```
samthor, line 44: nit: should this go to web.dev authors pages eventually? Or why is this here vs. all being listed in the front matter? @andy-piccalilli

src/site/content/en/learn/forms/forms.11tydata.js
```diff
@@ -0,0 +1,38 @@
...
+      thumbnail: (data) => {
+        // Use thumbnail defined in the frontmatter or in the meta.yml file.
+        return data.thumbnail || data?.courses?.[projectKey]?.meta?.thumbnail;
+      },
+      // Use this flag to prevent the course from showing up in production.
+      permalink: (data) => {
+        if (process.env.ELEVENTY_ENV === 'prod') {
+          return false;
+        }
+        return (data.page.filePathStem + '.html').replace(/^\/en\//, '/');
```
samthor, line 34: So this is basically a TODO right - we're not publishing this yet?

I'm a bit confused as to why the second non-prod case is needed. Is this just the "default" behavior that we don't get because there's an override here?

src/site/content/en/learn/forms/index.md
andy-piccalilli: If I remember rightly this got slotted into the CSS course post-launch. There's a credits section right at the end of the course too.

src/site/_includes/partials/header-course.njk
```diff
@@ -1,12 +1,18 @@
 {% from 'macros/icon.njk' import icon, svg with context %}
 
-<web-header class="header-course lg:display-none" role="navigation" style="--course-background: {{ courseData.meta.background }}">
+<web-header class="header-course lg:display-none" role="navigation" style="--course-background: url('{{ courseData.meta.background }}')">
```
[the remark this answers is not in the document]
robdodson: no, previously in the meta.yml [the author needed to write `background: url('foo.svg')`](https://github.com/GoogleChrome/web.dev/blob/2a734efa02de83c96f3e3940a711c340ce358596/src/site/_data/courses/css/meta.yml#L15). That felt like including too much implementation detail in the course configuration, so I changed it.

src/site/_data/i18n/courses.yml
robdodson: Placeholder. Would you like it to be changed?

src/site/_data/courses/forms/meta.yml
robdodson: Yeah I'm not in love with this. I think it's oddly consistent with the way the site works (all of the /learn/ cards have their svg images in /images) but...it's a bit weird. I didn't think this PR was the right place to change it up tho.

src/site/content/en/learn/forms/example/index.md
robdodson: yeah i stole this from learn css just to get them started

src/site/content/en/learn/forms/index.md
robdodson: oh I didn't notice this aside was on the overview page.

To answer your question Sam, because there are so many authors, we put them at the end of the course:
https://web.dev/learn/css/conclusion/#brought-to-you-by

I did this because otherwise we'd end up with a huge stack of authors on each page that would force down the first heading and look a bit odd.

src/site/content/en/learn/forms/forms.11tydata.js
robdodson: > So this is basically a TODO right - we're not publishing this yet?

Yes. This lets folks check in content without it actually going live.

> I'm a bit confused as to why the second non-prod case is needed. Is this just the "default" behavior that we don't get because there's an override here?

Yeah, if I recall correctly, I had to add this otherwise the page's permalink did not work correctly (I think it ended up including /en/). I had a comment for this the first time I used this trick: https://github.com/GoogleChrome/web.dev/pull/5304/files#diff-383d1becf08325fc7dd2a0672ea3de9dd434dabdff8325c74131f392975c1f20L23

In theory our output-permalink.js thing that Michael wrote _should_ be stripping the /en/. But something about the order that these things run in made that not happen. I'll include a comment so folks understand what this is doing.

src/lib/components/CourseLinks/index.js
```diff
@@ -43,6 +43,22 @@ class CourseLinks extends HTMLElement {
     slot.addEventListener('slotchange', this.onSlotChange);
   }
 
+  /**
+   * Get the user's overall course progress from localstorage.
+   * This returns an object where each string key represents a different course.
+   * Example: {css: {pages: ['intro', ...], percent: 10}, pwa: {...}}
+   * @returns {Object.<string, Course>}
```
samthor, line 50: nit: TS writes this like `{{[courseName: string]: Course}}`

src/lib/components/CourseLinks/index.js
```diff
@@ -43,6 +43,22 @@ class CourseLinks extends HTMLElement {
...
+   */
+  getProgress = () => {
+    let progress;
+    try {
+      progress = JSON.parse(localStorage['webdev_course_progress']);
```
samthor, line 55: Since this is now in try/catch, you can just `return` this line and `return {}` below. But whatever 🤷

src/lib/components/CourseLinks/index.js
```diff
@@ -53,44 +69,37 @@ class CourseLinks extends HTMLElement {
...
+    const progress = this.getProgress();
+    const course = progress[courseKey] || {...initialCourse};
     // Find every anchor that the user has already visited and set an attribute
     // on it so it renders a checkmark.
     children.forEach((child) => {
       if (
-        courseProgress.pages &&
-        courseProgress.pages.indexOf(child.getAttribute('href')) > -1
+        course.pages &&
+        course.pages.indexOf(child.getAttribute('href')) > -1
```
samthor, line 80: [96%+](https://caniuse.com/array-includes)

src/site/_data/courses/forms/meta.yml
samthor: Ack

src/site/_data/i18n/courses.yml
samthor: I don't mind. I guess it is obvious a placeholder but I worry that without a TODO it will be missed :)

src/site/content/en/learn/forms/index.md
samthor: Sure. I'm a bit confused because what you've linked to @robdodson feels structured and this part clearly is not. My point is more—if this is important it should be a front-matter list of contributors, which I feel like what was linked to definitely is.

src/site/content/en/learn/forms/forms.11tydata.js
samthor: Yeah, this feels gross. But it's fine for now.

samthor (approved):

src/site/content/en/learn/forms/index.md
robdodson: yeah, I'm not sure how this Aside was added. The thing I wrote uses the authors in the frontmatter and sends them to a macro. https://github.com/GoogleChrome/web.dev/blame/main/src/site/content/en/learn/css/conclusion/index.md

I can delete the aside from this PR so it doesn't encourage the Learn Forms authors to do the same thing.

src/lib/components/CourseLinks/index.js
robdodson: yeah I was trying to look up the TS syntax because I knew it was something like that. The TS JSDocs had this example
image

I'll switch it.

samthor:
Yeah, both are fine, but the new format is more modern and you give a hint
about what the 'key' is supposed to be.

src/lib/components/CourseLinks/index.js
robdodson: Hit a bug. Since the script is inlined into the page, nunjucks is interpreting the {{ }} as templating syntax. I'm going to leave it with the Object<> style

PR #5967, opened 2021-08-13
add includeRaw shortcode

samthor:
This follows-up from @robdodson 's PR #5964 and adds an `includeRaw` helper.

Very confusingly, all the previous includes that go outside "src/site/_includes" seem to have needed "../../.." and yet somehow still end up in "src/site" as the base path. I can't explain it, but I think Nunjucks' include function must do something different or special here.

So I've simplified that here.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -19,6 +19,7 @@ const pluginRss = require('@11ty/eleventy-plugin-rss');
 const pluginSyntaxHighlight = require('@11ty/eleventy-plugin-syntaxhighlight');
 const yaml = require('js-yaml');
 const fs = require('fs');
+const path = require('path');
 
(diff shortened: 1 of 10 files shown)


.eleventy.js
```diff
@@ -215,6 +216,10 @@ module.exports = function (config) {
   config.addShortcode('Tooltip', Tooltip);
   config.addShortcode('Video', Video);
   config.addShortcode('YouTube', YouTube);
+  config.addShortcode('includeRaw', (arg) => {
```
robdodson, line 219: do you mind moving this into it's own file? I know it's a small shortcode, but it'd be nice to be consistent with the other shortcodes.

robdodson (approved):

samthor:
Percy changes are noise from other changes.

PR #5970, opened 2021-08-13
@devnook Implement PatternSets

devnook:
This PR:
+ Adds pattern-sets and pattern-suites (pattern-suite is a top-level pattern set and gets it's own card in /patterns/)
+ Adds breadcrumbs macro (Breadcrumbs.js is too specific to posts, can't show deep urls structure)
+ Adds a bunch of patterns to demo the sets & suites (all have permalink=false in prod)

Out of scope for this PR (known outstanding issues):
+ CodePattern tabs css, esp on resize TBD
+ Pattern filter TBD


src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,130 @@
...
+  <div class="code-patterns-nav__filter">
+  <input type="text" placeholder="filter"/>
+  </div>
+</div>
+
+<section class="w-layout-container code-patterns--main">
+  <div>
+    {# Show patterns from current set (if set) or from the whole suite #}
+
+    {% if currentPatternSet %}
```
samthor, line 78: I'm really confused about where this template is used multiple times—seems like one for a whole "suite" and one for a "set"? I can't see it in the `.md` files.

src/site/_includes/macros/breadcrumbs.njk
```diff
@@ -0,0 +1,44 @@
+{#
+  @typedef {[string: string]} crumbs - The titles and urls of breadcrumbs to display.
```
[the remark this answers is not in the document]
devnook: I thought this is from typescript linter. Copied from other macros in /macros folder

src/site/_includes/pattern-set.njk
devnook: Suite is just a special case of a set (it is top level and gets a fancy card in /patterns/). if the page is rendered for a set and not whole suite, their link in menu gets highlighted.

src/styles/pages/_code-patterns.scss
```diff
@@ -0,0 +1,88 @@
...
+  @include bp(md) {
+    display: flex;
+    justify-content: space-between;
+  }
+
+  @include bp(lg) {
+    display: none;
+  }
+
+  .code-patterns-nav__filter {
```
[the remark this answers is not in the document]
devnook: so actually they are not unique on the page, and they look differently depending whether they are in --top or in --side navigation. I wonder what does BEM say about such case - will check.

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+        {% endif %}
+      {% endif %}
+    {% endfor %}
+  </nav>
+  <div class="code-patterns-nav__filter">
+    <input type="text" placeholder="filter"/>
+  </div>
+</div>
+
+<section class="w-layout-container code-patterns--main">
```
robdodson, line 76: Probably want to get rid of the top and bottom padding on this and replace it with just 64px of top margin.

robdodson: What is the --main modifier supposed to signify? Usually a modifier in BEM is for when you're changing the state of something (--disabled, for example).

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+  {% from 'macros/breadcrumbs.njk' import breadcrumbs with context %}
+  {% set crumbs = [
+    ['All patterns', '/patterns/'],
+    [patternSuite.title, '/patterns/' + patternSuite.id]
+  ] %}
+  {{ breadcrumbs(crumbs) }}
+</section>
+
+<header class="w-masthead-about">
+  <div class="w-masthead-about__container">
```
robdodson, line 34: Hm... it's not great that we're reusing this component (it's kind of coupled to the about page) but I understand why it's being used.

I'd suggest overriding it a bit and setting padding top/bottom to 0 and replacing them with margin top/bottom 64px.

Otherwise there will be more whitespace above the divider than below it.

src/site/_includes/components/CodePattern.js
```diff
@@ -16,18 +15,18 @@ module.exports = (patternId, height = 300) => {
...
           <pre><code class="language-${asset.type}">${content}</code></pre>
         </web-tab>
       `;
     })
     .join('\n');
 
-  return html`<div class="code-pattern flow flow-space-400">
+  return `<div class="code-pattern flow flow-space-400">
     <div class="code-pattern__content" style="height: ${height}px;">
       <div class="code-pattern__demo">
```
robdodson, line 28: These elements need to stack vertically on mobile.

src/styles/components/_code-pattern.scss
```diff
@@ -26,12 +26,14 @@
   pre {
     border: 0;
     background: 0;
+    white-space: pre-wrap;
```
robdodson, line 29: I don't think we want this to pre-wrap. I think we want it to scroll horizontally.

src/site/_includes/pattern.njk
```diff
@@ -1,5 +1,5 @@
 ---
-layout: patterns
+layout: pattern-page
 pageScripts:
```
robdodson, line 3: Any page that shows the tabs widget will need to inline it into the head, otherwise there's a big flash of unstyled content.

Kapture 2021-08-13 at 16 10 39

You can inline script into the head using the `inlineScript` yaml.
https://github.com/GoogleChrome/web.dev/blob/67c579c98bc1e21d1e2989ad459eed80aadc7e6a/src/site/_includes/partials/head.njk#L45-L53

The path needs to be relative to the _includes folder. Here's an example:
https://github.com/GoogleChrome/web.dev/blob/67c579c98bc1e21d1e2989ad459eed80aadc7e6a/src/site/_includes/course.njk#L5-L6

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+      {% if set.suite == patternSuite.id %}
+        {% if setId == currentPatternSet.id %}
+          <a class="selected" href="/patterns/{{set.id}}">{{set.title}}</a>
+        {% else %}
+          <a href="/patterns/{{set.id}}">{{set.title}}</a>
+        {% endif %}
+      {% endif %}
+    {% endfor %}
+  </nav>
+  <div class="code-patterns-nav__filter">
```
robdodson, line 71: I'd suggest commenting this element out until we're ready to enable it.

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+        {{ patternSuite.description }}
+      </p>
+    </div>
+    <div class="w-masthead-about__right">
+      {% Img
+          height=220,
+          width=330,
+          src=patternSuite.hero,
+          alt="Collection cover image",
+          loading="lazy",
```
robdodson, line 49: Since this is above the fold it should not be lazy loaded.
https://web.dev/lcp-lazy-loading/

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+    {% else %}
+
+      {% for patternSetId, patternSet in patterns.sets %}
+        {% if patternSet.suite == patternSuite.id %}
+          <h2> {{ patternSet.title }} </h2>
+          {{ patternSet.content | safe }}
+          {% for patternId, pattern in patterns.patterns %}
+            {% if pattern.set == patternSetId %}
+              <div class="flow">
+                <a href="/patterns/{{pattern.id}}"><h3 id="{{pattern.id}}">{{pattern.title}}</h3></a>
```
robdodson, line 103: I think you may want to move the anchor inside of the h3. Because the h3 is a block element the link itself is 800 - 960px wide and is very easy to click by accident.

src/site/_includes/pattern-page.njk
```diff
@@ -0,0 +1,7 @@
+---
```
robdodson, line 1: Could you get rid of this file and instead have pattern.njk extend post.njk?

src/site/_includes/pattern.njk
```diff
@@ -1,5 +1,5 @@
 ---
-layout: patterns
+layout: pattern-page
```
robdodson, line 2: It seems like we don't want the "Was this page helpful" widget.

I think we should actually just comment it out of the post template. Nobody uses it—Kayce said the data it sent back in analytics was too noisy to be useful.

src/styles/components/_code-pattern.scss
```diff
@@ -26,12 +26,14 @@
...
   }
 
   iframe {
     border: 0;
     height: 100%;
     width: 100%;
+    min-width: 300px;
```
robdodson, line 36: It seems like you shouldn't need this?

src/site/_includes/pattern-set.njk
robdodson: I'm also a bit confused between what's a suite and what's a set. It might be good to add a comment somewhere explaining the difference.

It seems like there used to be a suite in this example but it was deleted in this PR?

src/site/_includes/macros/breadcrumbs.njk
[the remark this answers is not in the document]
samthor: Yup, of course. It's just that typedef means something very different - it should be param in both places.

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+              </div>
+            {% endif %}
+          {% endfor %}
+        {% endif %}
+      {% endfor %}
+
+    {% endif %}
+  </div>
+
+  <section class="code-patterns-nav--side">
```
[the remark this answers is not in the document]
samthor: Yeah, that was my mistake, I didn't realise there was one in web.dev as well.

I wouldn't block @devnook on this - get it in and revisit the spy/sticky code?

src/site/_includes/pattern.njk
samthor: I appreciate that inlining works well (and we do CSP it and so on) but I _also_ do feel like `web-header` could have a sensible default set of styles here, since we know how big it will end up being.

src/site/_includes/macros/breadcrumbs.njk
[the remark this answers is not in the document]
devnook: Well I think the idea was that this comment describes what object the macro is outputting in html / styling, and was not intended as function jsdoc. In such case, @typedef is reasonable and more universal - it describes a type present in the macro (in this case passed as a param, but that's not always the case, it can come from context in other macros).

How about I make you both happy and add both :)

src/site/_includes/pattern-set.njk
devnook: All pages are disabled for now - permalink: false. I think it's better for development to keep it here.

src/site/_includes/components/CodePattern.js
devnook: I'd rather do it in a separate Pr, as this one is about suits&sets pages rather than CodePattern element.

devnook: Here's an issue to track: https://github.com/GoogleChrome/web.dev/issues/5979

src/site/_includes/macros/patterns-card.njk
```diff
@@ -0,0 +1,34 @@
+{#
```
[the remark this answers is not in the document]
devnook: I think it's an overkill and less readable.

src/site/_includes/pattern-page.njk
devnook: Do you mean I should introduce {% blocks %} to post.njk? It was made with includes in mind rather than extends.

src/site/_includes/pattern-set.njk
devnook: Are you sure? 64px top margin makes a lot of whitespace before the first pattern set title starts, pushing all content under the fold.

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,132 @@
...
+      {% endif %}
+    {% endfor %}
+  </nav>
+  <div class="code-patterns-nav__filter">
+    <input type="text" placeholder="filter"/>
+  </div>
+</div>
+
+<section class="w-layout-container code-patterns--main">
+  <div>
```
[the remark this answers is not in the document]
devnook: The inner div is needed to stack the side navigation. If I make it 800 px this page is inconsistent with the rest of the side. Is this by design?

src/site/_includes/pattern-set.njk
devnook: Yes, before there was no distinction between suits & sets and this Pr introduces that distinction.

src/site/_includes/pattern-set.njk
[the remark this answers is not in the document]
devnook: Right now the nav is just normal links that do reload the page, they do not scroll. If we want to have scrollspy/sticky behavior, I'd rather do it in a separate Pr as Sam suggested: https://github.com/GoogleChrome/web.dev/issues/5980

src/site/_includes/pattern.njk
devnook: Let's not mix Patterns and Post changes here - will submit separately.

src/styles/components/_code-pattern.scss
devnook: How to achieve it without setting the explicit width of the tabs (we'd like it to be 50% of the component)?

devnook:
PTAL

devnook (commented):
PTAL

src/site/_includes/pattern-set.njk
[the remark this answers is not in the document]
robdodson: The 800px comes from Jiwoong's figma design. I'm not sure what you mean when you say that it's inconsistent with the side?

src/site/_includes/pattern.njk
robdodson: I'm not sure I understand how it relates to `web-header`? You could try to have default styles but you'll have to make sure to test with a production build to make sure purgeCSS doesn't remove them. I think inlining the component on this page may be simpler.

src/site/_includes/pattern-page.njk
robdodson: no, sorry for the confusion. I thought pattern.njk could use post.njk as its layout (removing the need for pattern-page.njk). But I see now that post.njk does the weird flattened permalink thing, so we have to have pattern-page.njk if we want nested urls.

src/site/_includes/pattern.njk
robdodson: yep that's fine

src/site/_includes/components/CodePattern.js
```diff
@@ -1,7 +1,11 @@
-const {html} = require('common-tags');
 const Prism = require('prismjs');
-const patterns = require('../../_data/patterns')();
+const patterns = require('../../_data/patterns').patterns;
 
+/**
+ * A component to display code samples and a code demo side-by-side.
+ * @param {string} patternId Id of the Code Pattern to be displayed.
+ * @param {number?} height Optional desired height of the component.
```
robdodson, line 7:
```suggestion
 * @param {number} [height=300] Optional desired height of the component.
```

src/site/_includes/pattern-set.njk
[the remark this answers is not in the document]
robdodson: I think it probably is different from the rest of the site because the layout is unique

src/site/_includes/pattern-set.njk
robdodson: > Yes, before there was no distinction between suits & sets and this Pr introduces that distinction.

What's an example of a suite versus a set?

src/site/_includes/pattern-set.njk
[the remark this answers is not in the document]
robdodson: ok

src/site/_includes/pattern.njk
robdodson: I realized this won't work in this case because the Tabs element has a number of dependencies and we'd have to inline all of those as well, which seems like a lot...

So it seems like having default styles is probably our best bet. My hunch is they should do something like:

```
web-tabs:not(:defined) {
  // make all of the `pre` content inside of tabs invisible
}
```

src/site/_includes/pattern.njk
```diff
@@ -1,7 +1,9 @@
...
 pageScripts:
   - '/js/content.js'
+inlineScripts:
+   - '../../../lib/components/Tabs/index.js'
```
robdodson, line 6: It looks like this causes a runtime error because the inline script is trying to use module imports but we're inlining it using a regular script tag. Given how many dependencies the Tabs widget has, maybe inlining it isn't the right approach. It seems like we'll need to use placeholder styles to prevent CLS and hide the content until it's ready.

src/site/_includes/pattern-set.njk
robdodson: Another option might be to reduce the margin on the element above it (the masthead) so it's 32px so the two are even.

image

src/styles/components/_code-pattern.scss
robdodson: This is a weird quirk of grid. You have to define a minimum width for the columns.
https://css-tricks.com/preventing-a-grid-blowout/

```
grid-template-columns: minmax(0, 1fr) minmax(0, 1fr)
```

Then you'd remove the `white-space` property from the pre element and it should start displaying a scrollbar.

image

However, if there are multiple lines of code then we'll end up with both a horizontal and vertical scrollbars, which isn't great...

I noticed that CodePen wraps the text, but you can also drag to resize the container to read more.
image

I tried putting `position: relative; resize: both` on the `pre` element but it pushed around other elements on the page when I tried to resize it. My thinking is maybe we could leave pre-wrap if we can figure out a way to let the user resize the code? We may try asking Una to help with that CSS

devnook: We need to make demo resizeable so why not both sides.

devnook: Issue: https://github.com/GoogleChrome/web.dev/issues/5988

src/site/_includes/pattern.njk
devnook: TIL :defined !!

src/styles/components/_code-pattern.scss
devnook: If for some reason iframe does not load it will break the layout with width=0

src/site/_includes/pattern-set.njk
devnook: Suite is a top-level set, e.g. /patterns/PatternSuite/PatternSet/Pattern/index.md . It's special power is that it can contain other sets and that it gets a dedicated card in /patterns/ page.

devnook:
PTAL

robdodson (approved):

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,142 @@
...
+
+{% if patternSetId %}
+  {% set patternSet = patterns.sets[patternSetId] %}
+{% endif %}
+
+{% if patternSet.suite %}
+  {# This is a page for a PatternSet.
+    patternSuite variable is set to the parent of the current set. #}
+  {% set patternSuite = patterns.sets[patternSet.suite] %}
+  {% set currentPatternSet = patternSet %}
```
samthor, line 27: please duplicate this line below and set it to the empty string.

src/site/_includes/pattern-set.njk
```diff
@@ -0,0 +1,142 @@
...
+#}
+
+{% if page.url %}
+  {% set patternSetId = page.url | getRelativePath('/patterns/') %}
+{% else %}
+  {% set patternSetId = '' %}
+{% endif %}
+
+{% if patternSetId %}
+  {% set patternSet = patterns.sets[patternSetId] %}
```
samthor, line 20: duplicate this line in an `{% else %}` and set to `null`

src/styles/components/_code-pattern.scss
```diff
@@ -16,16 +16,27 @@
...
+    // grid-template-columns: minmax(0, 1fr) minmax(0, 1fr)
   }
 
   .code-pattern__assets,
   .code-pattern__demo {
     border: 1px solid $BORDER_COLOR;
   }
 
+  web-tabs:not(:defined) {
+    display: none;
```
samthor, line 28: Does this still cause a flash of unstyled content?

src/site/_includes/components/CodePattern.js
```diff
@@ -1,7 +1,11 @@
...
 const Prism = require('prismjs');
-const patterns = require('../../_data/patterns')();
+const patterns = require('../../_data/patterns').patterns;
 
+/**
+ * A component to display code samples and a code demo side-by-side.
+ * @param {string} patternId Id of the Code Pattern to be displayed.
+ * @param {number} [height=300] Optional desired height of the component.
+ */
 module.exports = (patternId, height = 300) => {
```
samthor, line 9: Just a drive-by, this 300 is what makes our unstyled content work well because we're not reflowing the page. I'm happy with this.

src/site/_includes/components/CodePattern.js
```diff
@@ -16,18 +20,18 @@ module.exports = (patternId, height = 300) => {
...
+      return `
         <web-tab title="${asset.type}" data-label="${asset.type}">
           <pre><code class="language-${asset.type}">${content}</code></pre>
         </web-tab>
       `;
     })
     .join('\n');
 
-  return html`<div class="code-pattern flow flow-space-400">
+  return `<div class="code-pattern flow flow-space-400">
```
samthor, line 31: why drop the html common-tags thing?
(To be clear, it seems to just - strip whitespace? Maybe we should use `safeHtml`?)
I don't need you to solve this now. Just wondering why.

samthor (commented):
I want to get this in. @devnook work with me when you come online and I'll approve.

samthor:
(Oh, it's already approved! Hooray.)

src/styles/components/_code-pattern.scss
devnook: No, this solves it (wrapper has a defined height and content is hidden).

src/site/_includes/components/CodePattern.js
devnook: Because they were removing line breaks and we need to preserve line breaks to show code as multiline strings rather than one big blob.

PR #5981, opened 2021-08-16
Remove feedback widget from post.njk

devnook:
See https://github.com/GoogleChrome/web.dev/pull/5970/files#r688825126

### DIFF
--- a/src/site/_includes/partials/post.njk
+++ b/src/site/_includes/partials/post.njk
@@ -116,7 +116,7 @@
     {% if codelabs %}
       {% CodelabsCallout codelabs, lang %}
     {% endif %}
-    {# TODO: Consider removing the feedback widget as noisy and unhelpful. #}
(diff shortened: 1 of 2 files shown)


src/site/_includes/partials/post.njk
```diff
@@ -116,7 +116,8 @@
     {% if codelabs %}
       {% CodelabsCallout codelabs, lang %}
     {% endif %}
-    <web-feedback additional-questions="{{ feedback }}"></web-feedback>
+    {# TODO: Consider removing the feedback widget as noisy and unhelpful. #}
```
samthor, line 119: This is phrased with the wrong tense. We have removed it now - maybe something like:

{# TODO: We've removed the feedback widget for now as it's noisy and unhelpful. #}

samthor (commented):
You should also remove the import of `Feedback` from "src/lib/pages/content.js", since it's not used anywhere else.

samthor (approved):

PR #5994, opened 2021-08-17
Add image demo to codePattern

### LINKED ISSUE #5993: Allow to use an image as demo in CodePattern

devnook:
Fixes #5993


src/site/content/en/patterns/image-demo-example/index.md
```diff
@@ -0,0 +1,26 @@
...
+layout: pattern
+title: Image demo pattern
+description: A description for the example pattern
+date: 2021-08-10
+height: 400
+---
+
+It is possible to use an image (static or animated) as a demo in a CodePattern.
+
+Use an `{% raw %}{% Img %}{% endraw %}` shortcode wrapped in `<figure>` tag in the demo.md
```
jpmedley, line 11:
```suggestion
Use an `{% raw %}{% Img %}{% endraw %}` shortcode wrapped in a `<figure>` tag in the demo.md
```

jpmedley (changes requested):

jpmedley:
I did not review your code. You'll need to get an engineer to do that.

src/site/content/en/patterns/image-demo-example/assets/sample.html
```diff
@@ -0,0 +1,3 @@
+<figure>
```
samthor, line 1: This file seems unused because you load "body.html".

src/site/content/en/patterns/image-demo-example/demo.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: demo
+patternId: image-demo-example
+---
+
+<figure >
```
samthor, line 6: I figured you might eventually add this (and it works with `{{ content | safe }}`) :)

samthor (commented):
I just want to clarify "sample.html".

src/site/content/en/patterns/image-demo-example/assets/sample.html
devnook: This is a sample that gets displayed in one of the assets tab. It is independent of the image demo.

src/site/content/en/patterns/image-demo-example/demo.md
devnook: Yeah :)

src/site/content/en/patterns/image-demo-example/assets/sample.html
samthor: Right, but it seems like it's purpose is to match the content that is in the Markdown, this is the same demo. Is that correct? Should it just be body.html?

devnook: Not really, they should be independent. There are 4 types of demos:

Allow demo to be built by:
+ Providing html, js and css snippet
+ Providing a complete html page
+ Providing an external, frameable demo url
+ Providing a static image or gif

See here:
https://docs.google.com/document/d/1jpTAHkuwg4XDJ2pvHKNuRxywTk6-dCgBgdriMqPteAY/edit?usp=sharing

Only in the 1st case the body.html is identical with the .../demo.html url content.

The purpose of image demo is that you provide an image illustrating the concept described in the pattern, e.g. a screencast of devtools to illustrate assets loading.

src/site/content/en/patterns/image-demo-example/demo.md
samthor: This doesn't seem to render, so I'm confused:
https://deploy-preview-5994--web-dev-staging.netlify.app/patterns/image-demo-example/demo

(Whereas sample.html is included in the output; although I appreciate that might be sort of unintentional as it is inlined.)

samthor (commented):
I'd be happy to jump on a VC to help understand this.

src/site/content/en/patterns/image-demo-example/demo.md
devnook: Ah good catch, it is because an unclean merge from main in the meantime. Let me fix it.

devnook:
I updated the samples so that maybe it's clearer what is happening. If its still confusing, let's vc.

src/site/_includes/demo.njk
```diff
@@ -6,13 +6,25 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>{{ title }}</title>
+  <style>
+    html, body {
+      height: 100%;
+    }
+    figure {
```
samthor, line 13: I appreciate the sensible defaults here. But in this literal demo, the image actually ends up being too big for the patterns preview thing.

Can you add CSS for e.g. `figure > img` so it has `max-width: 100%` and `object-fit: contain`? (There might be more to make it fit in the center, that's just off the top of my head).

samthor (approved):

PR #5990, opened 2021-08-17
Add customizable codePattern height + resizeable

devnook:
Fixes #5988

There are 3 ways to define height of the element:

1. Add height param in CodePattern shortcode include : {% CodePattern patternId HEIGHT %}
2. Add height in the Pattern frontmatter
3. Default height calculated from the number of lines of the longest code snippet

The precedence is 1 > 2 > 3.

Th iframe is now also resizeable (see handle in bottom right corner of the iframe).

PR #5998, opened 2021-08-17
Make code snippet scrollable; don't overflow container

khempenius:


src/lib/components/Tabs/_styles.scss
```diff
@@ -88,6 +88,7 @@ web-tabs {
     display: flex;
     flex: 1;
     flex-direction: column;
+    overflow-y: scroll;
```
samthor, line 91: Do you want it to always be scrollable? Seems like `auto` would show the scrollbar when it's needed.

khempenius:
Closing - this was fixed by https://github.com/GoogleChrome/web.dev/pull/5990.

PR #6001, opened 2021-08-18
Improve aria and i18n in BrowserCompat

devnook:


tomayac:
I just ran the [preview URL](https://deploy-preview-6001--web-dev-staging.netlify.app/handbook/web-dev-components/#browsercompat) through [Assistive Labs](https://assistivlabs.com/session?claim_id=5twemganwemVK7SMLsRC&keyboard=1) using the NVDA screen reader, and I think this is not working yet. See the speech log:

Screen Shot 2021-08-18 at 16 21 55

It says "times" for not supported, which is, I believe, due to the fact that the `span` has text content. I am no expert on this, we should definitely consult @robdodson.

robdodson:
Is the CSS messed up? In the previous PR it looked like the browser icons were much larger and the compat data overlapped with them. https://github.com/GoogleChrome/web.dev/pull/5665#issuecomment-880251325

image

But in the preview it seems like they're small with a lot of whitespace between them.
https://deploy-preview-6001--web-dev-staging.netlify.app/handbook/web-dev-components/#browsercompat

image

robdodson:
Often times an aria label won't do anything on an element if it doesn't have a `role`. I think an easier way of solving this will be to put the text into the span for the browser, and make the rest of the stuff aria-hidden. We can use the visually-hidden class (https://github.com/GoogleChrome/web.dev/blob/main/src/styles/utilities/_visually-hidden.scss) so the text will not be visible on screen, but will be accessible to screen readers. It would look something like this:

```
<div class="browser-compat__icon browser-compat--chrome">
  <span class="w-visually-hidden">Chrome 74, Supported</span>
</div>
<span class="browser-compat__version browser-compat--true" aria-hidden="true">74 </span>
```

jpmedley:
This looks great. Can this be an npm package so that other Google properties can use it? 

Should "Source" link to the #browser_compatibility anchor? You might be worried that the anchor might not be present on every page. A PR test on MDN will enforce this. If it is ever missing, the link just goes to the top of the MDN page.

This looks like it will handle interface members, so `api.XRSession.depthDataFormat` should work. Is this correct?

There's a few things you should know about `mdn_url`. First, sometimes it's missing. I have a script that constructs it when it's missing. [here](https://github.com/jpmedley/mdn-helper/blob/main/bcd.js#L124) and [here](https://github.com/jpmedley/mdn-helper/blob/main/bcd.js#L20). Sometimes it's present before the page is. Not sure how you want to handle that.

devnook:
Re:CSS messed up - no, this is the new design from Jiwoong:) We asked him to make it a bit smaller and easier to inline. You can see it in the patterns figma file.

src/site/_includes/components/BrowserCompat.js
```diff
@@ -14,43 +14,53 @@
...
     const sourceLink = source
       ? `<span class="browser-compat__link">
-        <a href="${source}" target="_blank">Source</a>
+        <a href="${source}" target="_blank">${sourceLabel}</a>
       </span>`
       : '';
+    const supportLabel = i18n(`i18n.browser_compat.browser_support`, locale);
     return `<div class="browser-compat">
-      <span>Browser support:</span>
+      <span>${supportLabel}</span>
```
samthor, line 56: add the : back after the label

samthor (approved):
Fix nits and 👍

devnook:
Re: missing mdn_url -> if the link is missing, we ignore it for now. If this proves to be a common problem, I'll be happy to use your scripts to improve the coverage @jpmedley

PR #6002, opened 2021-08-18
Move example patterns to one set

devnook:
+ Move all examples to one directory
+ Set draft: true for example patterns
+ Filter out drafts in /patterns/ page


src/site/_includes/patterns.njk
```diff
@@ -18,8 +18,11 @@ layout: default
   <section class="w-grid">
     <div class="w-grid__columns w-grid__columns--three" role="list">
       {% for setId, set in patterns.sets %}
-        {% if not set.suite %}
-          {{ PatternSuiteCard(set) }}
+        {% if site.env == 'prod' and set.draft %}
```
samthor, line 21: Can you rewrite this a bit? Maybe:

```njk
{% if (site.env == 'prod' and set.drafft) or set.suite %}
  {# Don't display drafts cards in prod, and don't display ??? #}
{% else %}
  {{ PatternSuiteCard(set) }}
{% endif %}
```

src/site/content/en/patterns/example-set/index.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: pattern-set
+title: Example PatternSet
+description: Example set of patterns
+hero: image/tcFciHGuF3MxnTr1y5ue01OGLBn2/dYAbN2LdttJ1qYk1gssh.svg
+draft: true
```
samthor, line 6: Could you instead create "example-set.11tydata.js" in this folder which sets `draft: true` for everything? Or would you prefer it set on invididual files in case they're copied and pasted for new demos?

devnook: We need this to be in frontmatter and not in 11tydata file, because it is used by parsing directly the frontmatter from the file, before any of 11ty magic kicks in (in /lib/patterns.js).

devnook:
PTAL

src/lib/patterns.js
```diff
@@ -46,6 +46,7 @@ const allPatterns = files.reduce((patterns, file) => {
       title: fileContents.data.title,
       description: fileContents.data.description,
       hero: fileContents.data.hero,
+      draft: fileContents.data.draft,
```
samthor, line 49: Could you add a comment above this section (const/if) saying something like "This only reads front-matter from files, and doesn't inherit any "11tydata.js" files".

samthor (approved):

PR #6004, opened 2021-08-18
@devnook Add patterns that do not have a parent PatternSet to the pattern-set page

devnook:
Some patterns might belong directly to the pattern suite (/patterns/PatternSuite/Pattern) instead of a PatternSet (/patterns/PatternSuite/PatternSet/Pattern). They are displayed under "other patterns" heading in the pattern suite page.

### DIFF
--- a/src/site/_includes/pattern-set.njk
+++ b/src/site/_includes/pattern-set.njk
@@ -124,11 +124,13 @@ pageScripts:
       {% endfor %}
 
       {# Patterns that belong directly to the pattern suite #}
+      {% set first = true %}
       {% for patternId, pattern in patterns.patterns %}
         {% if pattern.set == patternSuite.id %}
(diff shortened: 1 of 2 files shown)


src/site/_includes/pattern-set.njk
```diff
@@ -122,6 +123,20 @@ pageScripts:
         {% endif %}
       {% endfor %}
 
+      {# Patterns that belong directly to the pattern suite #}
+      {% for patternId, pattern in patterns.patterns %}
+        {% if pattern.set == patternSuite.id %}
+          {% if loop.first %}
+            <h2> Other patterns</h2>
```
samthor, line 130: You should make this an i18n string. I don't think there's any other literal English on this page.

samthor (approved):

PR #6011, opened 2021-08-19
Blog patch: css accent color to use latest browser support widget

argyleink:

- adds new browser compat shortcode `{% BrowserCompat 'css.properties.accent-color' %}`

Screen Shot 2021-08-19 at 10 59 37 AM

It's a little misleading though..! it says 91 supports it, and it does... behind a flag. 93+ it's in stable. @devnook @robdodson should I file a separate feature request to only indicate browser support if it's non-experimental? 

side note:
the pr includes commits from the original branch but properly shows the small mod to the 1 file. squash and merge makes that poo poo commit history a non issue yeah? or do you want me to make a separate branch so the commit is cleaner?


devnook:
@argyleink feel free to submit a feature request, I'm sure we'll iterate on this widget anyways. 

As for commits, they should get squashed on merge so no prob (I think).

robdodson:
yeah if something is behind a flag we probably need to show that it's not supported or in some sort of in-between state.

PR #6027, opened 2021-08-23
add App History API article

samthor:
This has been reviewed by @jpmedley and others in Docs. Importing here for publish.

DO NOT MERGE label is added because there's a demo video/gif that's TBD. Should be added within the next few hours.

### DIFF
--- a/src/site/content/en/blog/app-history-api/index.md
+++ b/src/site/content/en/blog/app-history-api/index.md
@@ -1,0 +1,392 @@
+---
+title: "Modern Client-Side Routing: The App History API"
+subhead: "Standardizing client-side routing through a brand new API which completely overhauls building single-page applications."
+authors:
+  - samthor
+date: 2021-08-25
+hero: image/QMjXarRXcMarxQddwrEdPvHVM242/nfRcF1VBdvRcsYI7RSIQ.jpg
+alt: "Sculpture adorning the General Post Office, Sydney, Australia"
+description: "Learn about the App History API, a new API which adds improved functionality to build single-page applications."
+tags:
+  - blog # blog is a required tag for the article to show up in the blog.
+  - capabilities
+---
+
+Single-page applications, or SPAs, are defined by a core feature: these are websites which dynamically rewrite their content as the user interacts with the site, instead of the default method of loading entirely new pages from the server.
+
+While SPAs have been able to bring you this feature via the History API (or in limited cases, by adjusting the site's #hash part), it's a [clunky API][clunky-history-api] developed long-before SPAs were the norm—and the web is crying out for a completely new approach.
+The App History API is a proposed API that completely overhauls this space, rather than trying to simply patch up History API's rough edges (e.g., [Scroll Restoration][scroll-restoration] patched the History API rather than trying to reinvent it).
+
+[clunky-history-api]: https://html5doctor.com/interview-with-ian-hickson-html-editor/#:~:text=My%20biggest%20mistake%E2%80%A6there%20are%20so%20many%20to%20choose%20from!%20pushState()%20is%20my%20favourite%20mistake
+[scroll-restoration]: https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration
+
+{% Aside %}
+
+The App History API is currently in development, but available in Chromium-based browsers behind the "Experimental Web Platform features" flag.
+[Check out a demo here][demo].
+
+{% endAside %}
+
+This post will describe the App History API at a high level.
+If you'd instead like to dive straight into the technical proposal, [check out the Draft Report in the WICG repository][wicg-report].
+
+## Example Usage
+
+When using the App History API, you'll start by adding a handler for the "navigate" event.
+This event is fundamentally _centralized_: it will fire for all types of navigations, whether because the user performed an action (such as clicking a link, submitting a form, or going back and forward) or when navigation is triggered programmatically (i.e., via your site's code).
+And, in most cases, it then lets your code override the browser's default behavior for that action.
+For SPAs, that likely means keeping the user on the same page and loading or changing the site's content.
+
+{% Aside %}
+
+This new event is very loosely analogous to the History API's "popstate" event, although way more powerful for you as a developer.
+Read on to find out more!
+
+{% endAside %}
+
+The "navigate" event contains information about the navigation, such as the target URL, and allows you to respond to the navigation in one centralized place.
+A basic implementation on example.com could look like this:
+
+```js
+appHistory.addEventListener("navigate", event => {
+  switch (e.destination.url) {
+    case 'https://example.com/':
+      event.transitionWhile(loadIndexPage());
+      break;
+    case 'https://example.com/cats':
+      event.transitionWhile(loadCatsPage());
+      break;
+  }
+});
+```
+
+In this example, we call `transitionWhile()` on the event with a promise generated from async helper functions.
+By calling this method with a response, the browser knows that your code needs to run to configure the next state of your site.
+This will create a transition object, which tracks the success or failure of the `Promise`—for example, it might reject because the user's browser is offline.
+
+You can handle the event in one of two ways:
+
+- by calling `transitionWhile()` (as described above) to handle the navigation
+- by calling `preventDefault()`, which can cancel the navigation completely
+
+Both these methods are usually allowed, but have cases where they're unable to be called.
+You can't handle navigations via `transitionWhile()` if the navigation is a cross-origin navigation: i.e., it's leaving your domain.
+And you can't cancel a navigation via `preventDefault()` if the user is pressing the Back and Forward buttons in their browser: you should not be able to trap your users on your site.
+
+Even if these methods aren't allowed, the "navigate" event is still _informative_, so your code could still, e.g., log an analytics event or perform cleanup.
+
+## Why another event handler?
+
+The "navigate" event centralizes handling URL changes inside a SPA.
+This is a difficult proposition using older APIs.
+If you've ever written the routing for your own SPA using the History API, you might have added code like this:
+
+```js
+function updatePage(event) {
+  event.preventDefault();  // we're handling this link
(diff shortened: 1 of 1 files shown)


samthor:
Ready for review now.

samthor:
Direct preview link here: https://deploy-preview-6027--web-dev-staging.netlify.app/app-history-api/

jpmedley:
FYI. The early review I do is high level and not meticulous. It's main goal is to find problems that authors might be disinclined to fix at the 11th hour (which is when PRs occur). Structural problems, for example.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+
+{% endAside %}
+
+Lastly, there's not yet consensus on programatically modifying or rearranging the list of entries the user has navigated through while using your site.
+This is [currently under discussion][bug-edit-entries], but one option could be to allow only deletions: either historic entries or "all future entries".
+The latter would allow temporary state, e.g., as a developer, I could:
+
+- ask the user a question by navigating to new URL or state
+- allow the user to complete their work (or go Back)
+- on completion, the App History API can remove that entry
```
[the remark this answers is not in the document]
jpmedley: Bullets should have parallel grammatical structure.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+
+This could be perfect for temporary modals or interstitals: the new URL is something that a user can use the Back gesture to leave from, but they then cannot accidentaly go Forward to open it again (because the entry has been removed), and is just not possible with the current History API.
```
jpmedley, line 350:
```suggestion
This could be perfect for temporary modals or interstitals: the new URL is something that a user can use the Back gesture to leave from, but they then cannot accidentaly go Forward to open it again (because the entry has been removed). This is just not possible with the current History API.
```

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+
+## Try the App History API
+
+You can try the App History API today in Chrome or Chromium-based browsers by enabling the "Experimental Web Platform features" flag.
```
jpmedley, line 354: What version? This future proofs this against flag changes, inauguration of an origin trial, and shipping.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+You can also [try out a demo][demo] by [Domenic Denicola][domenic].
+
+We're especially eager for feedback on issues labelled with ["feedback wanted"][feedback-wanted] on GitHub.
+You can also check out the repo and spec more generally at WICG/app-history, including filing new issues.
```
jpmedley, line 358: Please provide a complete link.

jpmedley (changes requested):
Thank you for submitting this article. I've found some things that need to be addressed. I've explained things where I think they're not clear. I will no doubt have questions. I'm here to answer them.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+
+// Navigate away, but the button will always work.
+await appHistory.navigate('/another_url');
+```
+
+### State
+
+The App History API surfaces a notion of "state".
+This is similar but improved from `history.state` from the History API. In the App History API, you can call the `.getState()` method of the current entry (or any entry) to return a copy of its state.
+The returned copy is mutable, but changes to the returned object won't affect further calls to `.getState()`. For example:
```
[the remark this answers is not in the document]
samthor: I've rewritten this a bit by definining state and introducing how you set it first. PTAL.

samthor (commented):
Thank you for the thorough review, I think I've addressed all your comments now, and left any unclear points unresolved  for you to take another look at.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+You can handle the event in one of two ways:
+
+- by calling `transitionWhile()` (as described above) to handle the navigation
+- by calling `preventDefault()`, which can cancel the navigation completely
+
+Both these methods are usually allowed, but have cases where they're unable to be called.
+You can't handle navigations via `transitionWhile()` if the navigation is a cross-origin navigation: i.e., it's leaving your domain.
+And you can't cancel a navigation via `preventDefault()` if the user is pressing the Back and Forward buttons in their browser: you should not be able to trap your users on your site.
+
+Even if these methods aren't allowed, the "navigate" event is still _informative_, so your code could still, e.g., log an analytics event or perform cleanup.
```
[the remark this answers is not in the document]
jpmedley: I'm rereading the line again. Do you mean that the occurrence of the event is informative or that the data provided by the event is informative. If the ladder, I think you need to name it and make it clear that the even is what you're talking about.

Maybe this isn't the place to do it, but I was also keen to get the name of the event interface into the article. This event will eventually have its own page on MDN and it's name is what developers will use to look it up.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+As such, this API introduces a semantic concept that the browser understands: a SPA navigation is currently occurring, over time, changing the document from a previous URL and state to a new one.
+This has a number of potential benefits, including that of accessibility: browsers can surface the beginning, end, or potential failure of a navigation in a variety of ways.
+Chrome, for example, intends to [reactivate its native loading indicator][loading-crbug] during this time.
+
+### Transition Success and Failure
+
+After the "navigate" event completes normally, the URL being navigated to will take effect. This happens immediately, even if you've called `transitionWhile()`, but if you have called it with a `Promise`, one of two things will happen:
+
+- If that succeeds (or you did not call `transitionWhile()`), the App History API will fire "navigatesuccess".
+- If the `Promise` rejects, the API will fire "navigateerror".
```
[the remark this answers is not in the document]
jpmedley: I see what you did with `navigateerror`. I can't find `navigatesuccess`.

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+### Access All Entries
+
+The "current entry" is not all, though.
+The API also provides a way to access the entire list of entries that a user has navigated through while using your site via its `appHistory.entries()` call, which returns a snapshot array of entries.
+This could be used to, e.g., show different UI based on how the user navigated to a certain page, or just to look back at the previous URLs or their state.
+This is impossible with the current History API.
+
+Entries themselves support events that allow operations to be taken when the entry finishes loading, is navigated to or from, or if it's disposed of from the user's history (e.g., an entry forward of the current position can be disposed because the user clicked a link and cleared the existing stack).
```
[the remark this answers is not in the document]
jpmedley: Much better. I have some additional thoughts.

'Entries' is plural. The latter parts of the sentence need to also be plural. So: "are navigated to or from" and "if they're disposed of".

>or if it's disposed of

I just want to confirm this. 'It' refers to 'entries', right?

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -247,7 +258,7 @@ appHistory.updateCurrent({ state });
 
 The "current entry" is not all, though.
 The API also provides a way to access the entire list of entries that a user has navigated through while using your site via its `appHistory.entries()` call, which returns a snapshot array of entries.
-This could be used to, e.g., show different UI based on how the user navigated to a certain page, or just to look back at the previous URLs or their state.
+This could be used to, e.g., show a different UI based on how the user navigated to a certain page, or just to look back at the previous URLs or their states.
 This is impossible with the current History API.
 
 Entries themselves support events that allow operations to be taken when the entry finishes loading, is navigated to or from, or if it's disposed of from the user's history (e.g., an entry forward of the current position can be disposed because the user clicked a link and cleared the existing stack).
```
jpmedley:
```suggestion
Entries themselves support events that allow operations to be taken when the entry finishes loading, is navigated to or from, or if it's disposed of from the user's history. (E.g., an entry forward of the current position can be disposed because the user clicked a link and cleared the existing stack.)
```

src/site/content/en/blog/app-history-api/index.md
```diff
@@ -260,23 +271,25 @@ And while the App History API is particularly powerful even with a simple event
...
 
-The `.navigate()` method returns a `Promise`, so the invoker can wait until the transition is complete (or is rejected due to failure or being preempted by another navigation).
-It also has an optional options object, allowing you to `replace` the current URL, set a new immutable `state` (to be made available via `getState()`), and `info`.
+The `AppHistory.navigate()` method returns a `Promise`, so the invoker can wait until the transition is complete (or is rejected due to failure or being preempted by another navigation).
+It also has an optional options object, allowing you to `replace` the current URL, set a new immutable `state` (to be made available via `AppHistoryEntry.getState()`), and `AppHistoryNavigateEvent.info`.
 
-The `info` option is worth calling out.
+The `info` property is worth calling out.
 It allows you to pass transient information about this specific navigation event into the "navigate" handler.
 This could be useful to, for example, denote a particular animation that causes the next page to appear (the alternative might be to set a global variable or include it as part of the #hash. Both options are a bit awkward).
```
jpmedley, line 306:
```suggestion
This could be useful to, for example, denote a particular animation that causes the next page to appear. (The alternative might be to set a global variable or include it as part of the #hash. Both options are a bit awkward.)
```

src/site/content/en/blog/app-history-api/index.md
jpmedley: Don't mention Chrome 93 at all. Do mention Chrome 95. I don't recommend trying to remember to come back. Statements like your original are hard to keep track of, harder to find and are rarely fixed after the fact no matter how much writers intend to. The combination of "95" and "flag" points to the fact that this is limited and does so in a very specific way.

PR #6030, opened 2021-08-23
docs: tweak copy to correspond with image (makes it a little bit easier)

j9t:
Change helps create a mental connection between the written example and the image below ([“Grid lines” section](https://web.dev/learn/css/grid/)).

Signed-off-by: Jens Oliver Meiert <jens@meiert.com>


argyleink (approved):
nice, agree that matches better 👍🏻

PR #6058, opened 2021-08-30
Clarify drag and drop article handleDrop

### LINKED ISSUE #6051: Didn't call handleDrop at the end
```
function handleDrop(e) {
  e.stopPropagation();

  if (dragSrcEl !== this) {
    dragSrcEl.innerHTML = this.innerHTML;
    this.innerHTML = e.dataTransfer.getData('text/html');
  }

  return false;
}
```

**To Reproduce**
1. Go to bottom of this article
2. find 'handleDrop function'
3. 'As other functions are being used in event listener '
4. handleDrop function isn't being called anywhere and that's why code isn't working

# My code isn't working because I don't know where to call handleDrop function.

samthor:
Fixes #6051

### DIFF
--- a/src/site/content/en/blog/drag-and-drop/index.md
+++ b/src/site/content/en/blog/drag-and-drop/index.md
@@ -4,7 +4,7 @@ authors:
   - ericbidelman
   - rachelandrew
 date: 2010-09-30
-updated: 2020-07-29
+updated: 2021-08-30
 description: >
   The HTML5 Drag and Drop (DnD) API means that we can make
   almost any element on our page draggable. In this post we’ll explain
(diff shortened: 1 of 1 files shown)


rachelandrew (changes requested):
This has messed up the code highlighting in the main code section. If you compare the preview with the live article at https://web.dev/drag-and-drop/ I think the highlighted lines need adjusting.

The updated date should also be updated.

samthor:
I reworked this a bit and added a different section to call out adding thie handler.

src/site/content/en/blog/drag-and-drop/index.md
```diff
@@ -205,7 +206,7 @@ consider [throttling or debouncing your event listener](https://css-tricks.com/d
 ## Completing the drop
 
 To process the actual drop,
-add an event listener for the `drop` event.
+add an event listener for the `drop` event (we do this above, so you can fill it in).
```
rachelandrew, line 209:
```suggestion
add an event listener for the `drop` event.
```

rachelandrew: now you've changed the structure I think we don't need this.

rachelandrew (approved):
I think the "we do this above" comment is now redundant, otherwise lgtm.

PR #6069, opened 2021-08-31
build(presubmit): create a presubmit step, fixes #5793

### LINKED ISSUE #5793: Create a `presubmit` step
We do not always test to see if the site builds fully, so some PRs break the build and deploy.

Add a flag to run expensive tests/actions, such as building the entire site, before allowing it to merge.

MichaelSolati:
Fixes #5793

- add new workflow that runs a prod build
- add action that checks if there is a presubmit label and removes it if there is

### DIFF
--- a/.github/actions/presubmit/index.js
+++ b/.github/actions/presubmit/index.js
@@ -1,5 +1,5 @@
 /**
- * @file This action checks if the `presubmit` label has been added to a Pull Request.
+ * @file This action checks if the `$-presubmit` label has been added to a Pull Request.
  * If the label exists, then remove the label and allow the rest of the workflow the
  * action belongs to to continue. If not fail the action causing the workflow to fail.
  *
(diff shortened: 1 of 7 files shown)


.github/workflows/presubmit-worflow.yml
```diff
@@ -0,0 +1,27 @@
+name: Presubmit
```
samthor, line 1: this file is typoed, should be "presubmit-wor**k**flow"

.github/actions/presubmit/index.js
```diff
@@ -0,0 +1,32 @@
+const core = require('@actions/core');
+const github = require('@actions/github');
+
+const LABEL = 'presubmit';
```
samthor, line 4: Can you explain what this actually does? Just some comments would be nice—like I don't understand if the build passing adds the presubmit tag, or whether it needs to exist to trigger it, or ... ?

.github/pull_request_template.md
```diff
@@ -8,6 +8,8 @@ Fixes #SOME_ISSUE_NUMBER
...
 Changes proposed in this pull request:
 
-- 
-- 
-- 
+-
+-
+-
+
+When you're ready to submit your PR, dont forget to add the `presubmit` label.
```
samthor, line 15: dont => dont'

It sounds like what the presubmit action will do is then immediately _remove_ the presubmit label and just use it as a signal to start the process. Is that right?

Can we change the label to start with e.g., "#" or "+" so that it sorts before other labels?

samthor:
I found that "-" sorts before all other letters. Notably this means you can put a "-" on the left side of text, even if there's other symbols there. I created label [$-presubmit](https://github.com/GoogleChrome/web.dev/labels/%24-presubmit), which sorts first—the $ is helpful as "-" as leading prefix is probably confusing or implies negation.

Can you swap to that or something similar?

.github/pull_request_template.md
samthor: typo to fix still

samthor (approved):

PR #6082, opened 2021-09-01
build(cloudbuild): migrate deploy to cloud build

MichaelSolati:


README.md
```diff
@@ -87,19 +87,14 @@ of the next hour.
...
-- @GoogleChrome/web-dev-eng
-- @GoogleChrome/web-devrel
+- web.dev-eng
+- web.dev-owners
 
-1. Navigate to [the Deploy workflow in the Actions panel](https://github.com/GoogleChrome/web.dev/actions?query=workflow%3ADeploy).
-2. Click the **Run workflow** button. Make sure the branch says `main`, then click the green **Run workflow** button.
-
-![An expanded workflow popup with a green run workflow button inside of it.](https://user-images.githubusercontent.com/1066253/89584965-da6eb500-d7f1-11ea-8a43-d8b1abe2cd3b.png)
+1. Navigate to [the Cloud Build Triggers page](https://console.cloud.google.com/cloud-build/triggers?project=web-dev-production-1).
```
samthor, line 95: This is good but I want to note here—like we do for dcc, I think—that in practice because this will happen every ~hours, you need to stop or disable that task for manual builds to really make sense. Can you add a comment?

samthor (approved):

PR #6089, opened 2021-09-01
Let authors use markdown in course titles and descriptions.

robdodson:
Previously you could not use markdown backticks in the titles or descriptions for course content. This already works for regular blog posts, but it didn't for courses.

This also adds an `.unstyled-code` class which can be used as an exception to disable most of our code styles.

Before:
image

After:
image

### DIFF
--- a/src/site/_includes/course.njk
+++ b/src/site/_includes/course.njk
@@ -55,8 +55,8 @@ inlineScripts:
             <div class="course-index font-mono">{{ loop.index0 | padStart(3, '0') }}</div>
           {% endif %}
         {% endfor %}
-        <h1 class="course-headline">{{ title }}</h1>
-        <p class="course-description">{{ description }}</p>
+        <h1 class="course-headline">{{ title | md }}</h1>
+        <p class="course-description">{{ description | md }}</p>
         {% include 'partials/toc-inner.njk' %}
         <div class="course-copy flow">
           {% if audio %}
--- a/src/site/_includes/course.njk
+++ b/src/site/_includes/course.njk
@@ -55,8 +55,12 @@ inlineScripts:
             <div class="course-index font-mono">{{ loop.index0 | padStart(3, '0') }}</div>
           {% endif %}
         {% endfor %}
-        <h1 class="course-headline">{{ title | md }}</h1>
-        <p class="course-description">{{ description | md }}</p>
+        <h1 class="course-headline">{{ title | md | safe }}</h1>
+        <p class="course-description">
+          {# If authors use backticks in their description markdown we don't #}
+          {# want the code to have any border or background color. #}
+          <span class="unstyled-code">{{ description | md | safe }}</span>
+        </p>
         {% include 'partials/toc-inner.njk' %}
         <div class="course-copy flow">
           {% if audio %}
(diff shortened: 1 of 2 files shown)


src/site/_includes/course.njk
```diff
@@ -55,8 +55,12 @@ inlineScripts:
             <div class="course-index font-mono">{{ loop.index0 | padStart(3, '0') }}</div>
           {% endif %}
         {% endfor %}
-        <h1 class="course-headline">{{ title }}</h1>
-        <p class="course-description">{{ description }}</p>
+        <h1 class="course-headline">{{ title | md | safe }}</h1>
```
samthor, line 58: Is this disingenious because it doesn't work for the actual `<title></title>`, therefore we shouldn't do this here?

src/styles/generic/_code.scss
```diff
@@ -80,16 +80,21 @@ pre code {
...
 h4 code,
 h5 code,
-h6 code {
+h6 code,
+.unstyled-code code {
   background: transparent;
   border: 0;
   color: inherit;
   font-size: inherit;
+  font-style: inherit;
```
samthor, line 97: at what point is this just `font: inherit`?

samthor (approved):

src/styles/generic/_code.scss
robdodson: I think the difference is we don't want it to use the same font face (we want to use the mono one). I'm not sure about the specificity with shorthands but I'm happy to try `font: inherit` to see if it still gives us the result we want.

src/site/_includes/course.njk
robdodson: I may be misunderstanding but I think we get the results that we want.

Using this PR, if I have the following frontmatter:

```yaml
title: Learn `<ol>` CSS
authors:
  - andybell
description: An evergreen `<ul>` CSS course and reference to level up your web styling expertise.
date: 2021-05-04
tags:
  - css
  - layout
```

I end up with:

```html
<title>Learn <ol> CSS</title>
<meta name="description" content="An evergreen <ul> CSS course and reference to level up your web styling expertise." />
```

And the page renders as expected:

image

robdodson: Although it's maybe not great to have an actual piece of html inside of the title because you could in theory wrote `</title>` and the page would blow up

src/styles/generic/_code.scss
robdodson: Follow up: I tried `font: inherit` but that made it stop using the mono typeface

src/site/_includes/course.njk
robdodson: @samthor what would you want to appear in `<title>`? We could look into having Meta.js html escape stuff?

robdodson: ok I tested it and I think it _may_ be ok.

Here's what I got when I did this:

image

```
<h1 class="course-headline">Learn CSS <code>&lt;script&gt;alert('hello');&lt;/script&gt;</code></h1>
```

The `{{ title | md | safe }}` thing is what we already do for blog posts and (so far at least) it hasn't created any issues.

PR #6090, opened 2021-09-01
Fix broken indentation that is messing up translation PRs

robdodson:
The indentation on the code block is causing our translation PRs to double wrap things in backticks.


robdodson:
I'm just going to merge this as it's a trivial change

PR #6091, opened 2021-09-02
fix(write-version): use git to get commit sha

MichaelSolati:
Forgot to update the write version to get the commit sha from git rather than the github env variable.


samthor (approved):
Yeah, sorry I missed this in review. 👍

PR #6092, opened 2021-09-02
Allow for nested content in course sidebar

### LINKED ISSUE #6024: Add support for headings in the navigation for courses.
https://web.dev/learn/css/ does not have any section headings in the left nav, but I've seen a number of upcoming outlines that will need it. 

We would need to add it to https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/partials/navigation-drawer-course.njk and we would likely need to modify https://github.com/GoogleChrome/web.dev/blob/main/src/site/_filters/navigation.js to support it.

samthor:
Fixes #6024.

This also fixes the types a bit—the next/prev entries were never drawn from 11ty. Also, the "NEW" badge was never appearing because it wasn't part of the data cloned from the source item.

Screen Shot 2021-09-02 at 15 17 05

I've made the headings fairly bland text, and added extra padding to nested entries. This doesn't effect the `<courses-links>` element, which looks at its direct children, because each nested entry is still attached to the root (it's just padded more). It's also possible for a URL entry to have section children.


src/site/_includes/partials/course-index.njk
```diff
@@ -1,9 +1,14 @@
 {%- set courseData = courses[projectKey] -%}
 {%- set pageNavigation = collections.all | navigation(courseData.toc) -%}
-{%- for item in pageNavigation -%}
+{%- for item in pageNavigation.list -%}
```
robdodson, line 3: I think the card on the /learn/ page uses the number of items in pageNavigation to indicate how many resources are in the course (note how in the image below it just says ___ resources). So we may need to update that card to count the number of items in pageNavigation.list, or use some other approach to get the total number of  items.

image

src/site/_data/courses/css/toc.yml
```diff
@@ -7,8 +7,12 @@
 - url: /learn/css/color
 - url: /learn/css/sizing
 - url: /learn/css/layout
-- url: /learn/css/flexbox
+- title: i18n.header.open_menu
+  sections:
+  - url: /learn/css/flexbox
+- title: i18n.header.about
 - url: /learn/css/grid
+  title: i18n.header.live
```
robdodson, line 15: Just a note to clean this up before merging

robdodson (approved):

src/site/_includes/partials/course-index.njk
robdodson: ah I think this is also why percy is failing so when that's fixed that test should pass and unlock the PR

samthor: Fixed, I missed this one.

PR #6098, opened 2021-09-02
[DO NOT MERGE] Update WebTransport OT article with release plans

jeffposnick:
Updates to the WebTransport origin trial article to include info about the current release plans.

I'm marking this as DO_NOT_MERGE for the time being, while I get sign-off from the relevant teams that the information is accurate. At that point I'll remove the label.

CC: @yutakahirano

### DIFF
--- a/src/site/content/en/blog/webtransport/index.md
+++ b/src/site/content/en/blog/webtransport/index.md
@@ -21,15 +21,14 @@ feedback:
 ---
 
 {% Aside 'caution' %}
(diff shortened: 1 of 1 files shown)


jpmedley (approved):

jeffposnick:
The OT was restarted, so no updates to the text of this article need to be made at this time.

When WebTransport graduates out of OT and ships by default we'll need to make a different set of updates.

PR #6103, opened 2021-09-03
maybe deploy script and split

samthor:
Splits Cloud Build into two parts. One which checks the current version and might set off another task, and one that actually does the deploy.

Alternatives considered:

- one single task with conditional steps (not really supported, needs lots of bash): fine, but unreadable
- detecting cancels (hard, they're logged as failures)

Once this is ready to go, I will flip the hourly trigger over on Cloud Console, as it currently points to "deploy". We'll need to point it to this "maybe-deploy" trigger.

If either tasks fails, we'll be notified, because we're still watching for errors. That process doesn't need to change.


MichaelSolati (approved):
LGTM

Though 1) the cron on GCP will need to be updated to trigger the maybe deploy. 2) I don't like the "maybe deploy" name... Maybe it could be "deploy check" or "release check?"

samthor:
Renamed back to "version-check" as per discussion in Chat.

PR #6106, opened 2021-09-05
Fix Url in Does not have a <meta name="viewport"> tag with width or i…

digi-booster:
…nitial-scale article

Fixes #SOME_ISSUE_NUMBER

- 
- 
-


digi-booster:
@googlebot I signed it!

PR #6145, opened 2021-09-06
fixing url in Improved Next.js and Gatsby page load performance with granular chunking blog

digi-booster:
Fixes #SOME_ISSUE_NUMBER

- 
- 
-


samthor (commented):
I'm not entirely sure whether this should link to a specific revision or 'main'.

I'd prefer 'main' for now, because we know if it's broken again in future.

PR #6176, opened 2021-09-09
Fix grid bug in CH93

devnook:


src/styles/components/_code-pattern.scss
```diff
@@ -19,12 +19,16 @@
     grid-template-columns: minmax(0, 1fr);
 
     @include bp(md) {
-      grid-template-columns: minmax(0, 1fr) minmax(0, 1fr);
+      grid-template-columns: 1fr 1fr;
```
una, line 22: Can we keep the old code commented to revert in M94? And open an issue to revert?

una (approved):
Approving the hotfix,  but would like to keep comments and open an issue as a reminder to revert this

PR #6185, opened 2021-09-10
Remove lint rules

devnook:


robdodson (approved):

samthor:
For context this is to support i18n where we don't get be so restrictive.

PR #6198, opened 2021-09-13
feat: Copying a heading copies the link to the clipboard

tomayac:
This is the successor to #4570.

This now uses the `web-snackbar`.

### DIFF
--- a/src/lib/copy.js
+++ b/src/lib/copy.js
@@ -1,0 +1,21 @@
+if ('clipboard' in navigator && 'writeText' in navigator.clipboard) {
+  document.addEventListener('click', async (event) => {
+    if (
+      // @ts-ignore
+      !event.target.classList ||
+      // @ts-ignore
+      !event.target.classList.contains('w-headline-link')
+    ) {
+      return;
+    }
+    event.preventDefault();
+    try {
+      await navigator.clipboard.writeText(event.target.href);
+      const snackbar = document.querySelector('web-snackbar');
+      snackbar.setAttribute('type', 'copied');
+      snackbar.setAttribute('open', 'open');
+    } catch (err) {
+      console.warn('Failed to copy: ', err);
+    }
+  });
+}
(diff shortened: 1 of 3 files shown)


src/lib/copy.js
```diff
@@ -0,0 +1,21 @@
+if ('clipboard' in navigator && 'writeText' in navigator.clipboard) {
+  document.addEventListener('click', async (event) => {
+    if (
```
samthor, line 3: To avoid the typing errors, this should be:

```js
if (!(event.target instanceof Element) || !event.target.classList.contains('w-headline-link')) {
  return;
}
```

src/lib/copy.js
```diff
@@ -0,0 +1,21 @@
...
+      // @ts-ignore
+      !event.target.classList ||
+      // @ts-ignore
+      !event.target.classList.contains('w-headline-link')
```
tomayac, line 7:
```suggestion      
      !(event.target instanceof Element) ||
      !event.target.classList.contains('w-headline-link')
```

src/lib/copy.js
tomayac: Good idea. Done.

samthor (commented):
In general I think using the snackbar is good. However the code is a bit awkward (to be fair there's a few timeouts already in this file).

Also technically it's possible to click a link and hide the cookie notice without saying "ok". Let me have a think.

tomayac:
> In general I think using the snackbar is good. However the code is a bit awkward (to be fair there's a few timeouts already in this file).

As a random data point, I noticed that the snackbar equivalent on d.c.c. works entirely different. To be quite honest, I think the current web.dev snackbar should be uniquely used for cookie notices, and a new toast component should be added on both d.c.c. and web.dev for use cases like this.
 
> Also technically it's possible to click a link and hide the cookie notice without saying "ok". Let me have a think.

Is this GDPR compliant?

PR #6207, opened 2021-09-14
Enable no-smart-quotes linter for code samples

devnook:


src/site/content/en/handbook/markup-code/index.md
```diff
@@ -100,9 +100,9 @@ Indent using two spaces.
 Always use straight quotes. (Readers often copy code directly from the site, and smart quotes can break code.)
 
 {% Compare 'worse' %}
-`const imagemin = require(‘imagemin’);`
+<code>const imagemin = require(‘imagemin’);</code>
```
robdodson, line 103: I'm surprised you needed to remove the backticks in this scenario. It looks like Compare adds a newline so the markdown parser should kick in (https://github.com/GoogleChrome/web.dev/blob/c36a2b0f605b7a8ba8dc72dd7385701ab2069bd9/src/site/_includes/components/Compare.js#L53). Was it not doing that?

robdodson (approved):
LGTM, one question

src/site/content/en/handbook/markup-code/index.md
devnook: It was, but it also directly violates our rule that we do not allow smart quotes in code samples in .md files...

devnook:
PS. This rule checks the markdown code blocks (```xxx``` and `xxx`) but not the <code> tags content directly.

PR #6212, opened 2021-09-14
test(presubmit): move some other workflow steps to presubmit workflow, fixes #4612

### LINKED ISSUE #4612: Integration tests are slow because they optimize images and do a full site build.
I think we can speed these up by doing a limited site build (similar to what we do when Percy runs) and mostly ignore content images.

MichaelSolati:

- Move all JS tests and JS/SCSS linting from `Lint and test` workflow to `Presubmit` workflow 
- Run Lighthouse CI on `Presubmit` workflow
- Remove `pull_request` trigger from `Lighthouse CI` workflow


.github/workflows/lint-and-test-workflow.yml
```diff
@@ -36,16 +36,6 @@ jobs:
           filters: |
```
robdodson, line 36: should we rename this workflow to `lint-markdown`?

.github/workflows/presubmit-workflow.yml
```diff
@@ -25,3 +25,50 @@ jobs:
...
+      # Checks to see if any files in the PR match one of the listed file types.
+      # We can use this filter to decide whether or not to run linters or tests.
+      # You can check if a file with a listed file type is in the PR by doing:
+      # if: ${{ steps.filter.outputs.md == 'true' }}
+      # This will return true if there's a markdown file that was changed
+      # in the PR.
+      - uses: dorny/paths-filter@v2.2.1
+        id: filter
+        with:
+          filters: |
```
robdodson, line 38: Since this only runs when we do presubmit, maybe we remove all the file path filtering stuff and just do lint and test?

test/integration/build-test.js
```diff
@@ -12,14 +12,20 @@ describe('Build test', function () {
...
-    } catch (err) {
-      assert.fail(err);
+    if (fs.existsSync('./dist')) {
+      console.log(
+        '`dist` folder already exists, build completed. Starting tests.',
+      );
+    } else {
+      try {
+        // This copies everything except for images because gulp will try to
+        // optimize those during a prod build which slows things right down.
```
robdodson, line 22: I think this comment is outdated and can be removed

robdodson (approved):
Some small comments but overall looks good to me. Deferring to @samthor for the final sign off.

.github/workflows/lighthouse-ci-workflow.yml
```diff
@@ -1,19 +1,9 @@
 name: Lighthouse CI
 
-# By default, runs when a pull_request's activity type is opened, synchronize,
-# or reopened and on main so comparisons to ancestors can be made.
 on:
   push:
     branches:
```
samthor, line 5: This is actually interesting; what value does this action have at all right now? Like, I feel like doing this _again_ for every PR after it's merged is just burning CPU.

samthor (commented):
Moving Lighthouse? Yes, totally fine to do in a presubmit.

The linter checks? Ehh, I'm not sure. My concern is that, they're so cheap, and authors who are editing in GitHub directly can see feedback straight away. (This is not hugely different from the external review bot, which is also running every commit.)

.github/workflows/lighthouse-ci-workflow.yml
MichaelSolati: I did agree except for the fact "on main so comparisons to ancestors can be made."

I'm assuming we want a log/history of commits to main.

MichaelSolati:
> Moving Lighthouse? Yes, totally fine to do in a presubmit.
> 
> The linter checks? Ehh, I'm not sure. My concern is that, they're so cheap, and authors who are editing in GitHub directly can see feedback straight away. (This is not hugely different from the external review bot, which is also running every commit.)

Most linters are cheap, but the JS tests are. Specifically the [integration test that runs a full prod build](https://github.com/GoogleChrome/web.dev/blob/1041c85ac8172a650926a46e3b1e3ce6924e7567/test/integration/build-test.js#L18).

@samthor should we keep the lints as its own thing and move the js test into this?

.github/workflows/lighthouse-ci-workflow.yml
samthor: Yes, but... we don't really do this right now. Like, do we even get an email if this fails after merge?

samthor (commented):
Yes. And the test isn't a linter (even though perhaps it was in the wrong file). I mean like `npm run lint:js` is cheap. So yeah, I'd keep the cheap lint steps, but move the full `npm run test` to the presubmit.

PR #6218, opened 2021-09-15
Render component helper

andy-piccalilli:
This little helper allows us to render design system components like this: 
```{{ design.components.render('button', { text: 'I am a cool button' }) | safe }}```

You can render the pattern verbatim like this too: 
```{{ design.components.render('button') | safe }}```

### CODE REVIEW

File: src/scss/blocks/_toggle-switch.scss [hunk 1]
```diff
@@ -14,7 +14,8 @@ $toggle-switch-decor-ratio: 0.8;
 
 .toggle-switch__input {
   opacity: 0;
-  position: absolute;
+  // Yup, the global input styles have some pretty high specificity
+  position: absolute !important;
```
andy-piccalilli, line 18: A little bit of a CSS tip: I would add `!important` *all year* rather than trying to create a higher specificity selector. Readability FTW

src/site/_data/design/components.js
```diff
@@ -284,4 +286,38 @@ module.exports = {
...
     return response;
   },
+
+  // Takes a handle, tries to match that with a component or variant then
+  // applies passed data if available
+  render(handle, data = null) {
+    const item = this.previews.find((item) => item.handle === handle);
+
+    if (!item) {
+      console.log(error(`Component, “${handle}” couldn’t be found`));
```
samthor, line 296: no smart quotes

src/site/_includes/macros/design-system-component.njk
```diff
@@ -1,29 +1,37 @@
...
-      <web-copy-code>
-        {% highlight "html" %}{{ markup | safe }}{% endhighlight %}
-      </web-copy-code>
-      {% endif %}
+  <div class="gap-top-size-2">
+    <details>
+      <summary>Code</summary>
+      <div class="flow">
+        {% if markup %}      
+        <p class="text-size-3">Source (Nunjucks)</p>
```
samthor, line 20: nit: indent the HTML here and below on line 27

src/site/content/en/design-system/render-component-test.njk
```diff
@@ -0,0 +1,24 @@
+---
+permalink: '/design-system/render-component-test/index.html'
+title: 'Test'
+button: 
+  text: 'Override data'
+aside:
+  content: 'Some **markdown** in here'
```
samthor, line 7: At some point we need to care about i18n for these. But that is not today I suspect.

src/site/content/en/design-system/render-component-test.njk
```diff
@@ -0,0 +1,24 @@
...
+{% block inner %}
+
+{{ design.components.render('button') | safe }}
+{{ design.components.render('button-primary') | safe }}
+{{ design.components.render('button-primary', button) | safe }}
+
+{{ design.components.render('aside') | safe }}
+{{ design.components.render('aside-warning', aside) | safe }}
+
+{# Should throw a console error #}
```
samthor, line 21: Throw implies the site fails to build, and clearly it builds fine. I'm in two minds, I actually wouldn't mind this to crash, so in this case could you just comment this out and say "this would have crashed"?

src/site/_data/design/components.js
```diff
@@ -284,4 +286,38 @@ module.exports = {
...
+      return '';
+    }
+
+    let renderData = {};
```
samthor, line 300: I think you could simplify this. Yes, using the spread syntax when we only have one object is a copy rather than direct use, but I think all the conditionals here just make it harder to read. Remember spread on `null` or `undefined` is totally valid.

```js
// Passed data overwrites any context data.
const renderData = {...item.data.context, ...data};
return nunjucksEnv.renderString(item.markup, { data: renderData });
```

samthor (approved):

src/site/content/en/design-system/render-component-test.njk
andy-piccalilli: Yeh for sure. We can pass the i18n stuff in just like we do through the templates etc

src/site/content/en/design-system/render-component-test.njk
andy-piccalilli: I'll just delete this page once this PR is approved anyway. It's a temporary test

src/site/_data/design/components.js
andy-piccalilli: Very tidy!

andy-piccalilli:
Sweet, I'll merge this one and prototype shortcode context stuff in a different PR,  FYI @robdodson

PR #6220, opened 2021-09-15
fix(LHCI): add missing step for GitHub App setup

patrickhulce:

- Adds a missing step in the LHCI GitHub App instructions where you need to expose it to the environment.
- Updates the LHCI version to 0.8.x

### DIFF
--- a/src/site/content/en/blog/lighthouse-ci/index.md
+++ b/src/site/content/en/blog/lighthouse-ci/index.md
@@ -390,16 +390,16 @@ to display these results on each pull request.
         runs-on: ubuntu-latest
         steps:
           - uses: actions/checkout@v1
-          - name: Use Node.js 10.x
+          - name: Use Node.js 12.x
             uses: actions/setup-node@v1
             with:
-              node-version: 10.x
+              node-version: 12.x
           - name: npm install
             run: |
               npm install
           - name: run Lighthouse CI
             run: |
-              npm install -g @lhci/cli@0.3.x
+              npm install -g @lhci/cli@0.8.x
               lhci autorun --upload.target=temporary-public-storage || echo "LHCI failed!"
     ```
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/lighthouse-ci/index.md
```diff
@@ -472,8 +472,21 @@ The steps below explain how to set up a status check for Lighthouse CI.
...
+7.  Expose the `LHCI_GITHUB_APP_TOKEN` secret to the environment of LHCI in 
+    GitHub Actions.
+  
+    ```yaml
+          - name: run Lighthouse CI
+            env:
+              LHCI_GITHUB_APP_TOKEN: ${{ secrets.LHCI_GITHUB_APP_TOKEN }}
+            run: |
+              npm install -g @lhci/cli@0.8.x
+              lhci autorun --upload.target=temporary-public-storage || echo "LHCI failed!"
```
OliverJAsh, line 485: I think we also need to document the fact that this is required in the checkout action?

```
with:
          ref: ${{ github.event.pull_request.head.sha }}
```

https://github.com/GoogleChrome/lighthouse-ci/issues/172#issuecomment-571928200

patrickhulce: that's when actions are run on pull requests only, this example sets up running LHCI and all pushes, which should not have that same issue (in addition to not having an `github.event.pull_request` to be able to reference)

OliverJAsh: > in addition to not having an `github.event.pull_request` to be able to reference

We worked around this like so:

```
ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
```

> this example sets up running LHCI on all pushes, which should not have that same issue

FWIW this is what we're doing (running LHCI on all pushes, `master` + PRs) and this workaround was required for us.

patrickhulce: > and this workaround was required for us.

Required for just the PR portion, correct? I have not seen a scenario where running on pushes to all branches requires this explicit checkout ref.

OliverJAsh: Yes you're right! My bad.

PR #6226, opened 2021-09-16
WIP: prototype rendering new design system components without disturbing older stuff

andy-piccalilli:
I recorded a screencast of this admittedly, questionable approach: https://www.loom.com/share/a5f832e2f84948d4a81b7ee86c6d1b1e

If we go down this route, the main problem will be how do we handle nested shortcodes. For example, how do we handle a `{% Img %}` being passed into an `{% Aside %}`? The markdown is processed using [our filter](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_filters/md.js).

Other than that and the frightening globals, this could have legs—especially as it is very temporary. 

Let me know ur thoughts and please, feel free to play around on this branch too! 

Example page that uses the new design system and an existing shortcode: https://github.com/GoogleChrome/web.dev/blob/prototype-shortcode-context/src/site/content/en/design-system/css-methodology.md

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -157,14 +157,12 @@ module.exports = function (config) {
   // CSS_ORIGIN set to 'next'. This allows shortcodes to determine if we
   // are in a design system context or a legacy context
   config.addCollection('designSystemGlobals', (collection) => {
-    global.__designSystemPaths = [
-      ...new Set(
-        collection
-          .getAll()
-          .filter(({data}) => data.CSS_ORIGIN === 'next')
-          .map(({filePathStem}) => filePathStem),
-      ),
-    ];
+    global.__designSystemPaths = new Set(
+      collection
+        .getAll()
+        .filter(({data}) => data.CSS_ORIGIN === 'next')
+        .map(({filePathStem}) => filePathStem),
+    );
 
     return global.__designSystemPaths;
   });
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -166,8 +166,6 @@ module.exports = function (config) {
       ),
     ];
 
-    global.__designSystemComponents = require('./src/site/_data/design/components');
-
     return global.__designSystemPaths;
   });
 
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -153,6 +153,24 @@ module.exports = function (config) {
     return memoize(collection.getAll());
   });
 
+  // Filters through all collection items and finds content that has
+  // CSS_ORIGIN set to 'next'. This allows shortcodes to determine if we
+  // are in a design system context or a legacy context
+  config.addCollection('designSystemGlobals', (collection) => {
+    global.__designSystemPaths = [
+      ...new Set(
+        collection
+          .getAll()
+          .filter(({data}) => data.CSS_ORIGIN === 'next')
+          .map(({filePathStem}) => filePathStem),
+      ),
+    ];
+
+    global.__designSystemComponents = require('./src/site/_data/design/components');
+
+    return global.__designSystemPaths;
+  });
+
   // ----------------------------------------------------------------------------
   // FILTERS
   // ----------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


src/site/_includes/components/Aside.js
```diff
@@ -22,6 +25,33 @@ function Aside(content, type = 'note') {
   const locale = getLocaleFromPath(this.page && this.page.filePathStem);
 
   let prefix;
+
+  // Renders the new aside
+  if (isDesignSystemContext(this.page.filePathStem)) {
```
samthor, line 30: If I have the page, can't I just read its data here to find this out? (i.e., what else is available on `this`? 11ty is bad at types, time to `log`). I feel like the collection backing `isDesignSystemContext` is adding a weird layer of abstraction we don't need.

.eleventy.js
```diff
@@ -153,6 +153,24 @@ module.exports = function (config) {
...
+    global.__designSystemPaths = [
+      ...new Set(
+        collection
+          .getAll()
+          .filter(({data}) => data.CSS_ORIGIN === 'next')
+          .map(({filePathStem}) => filePathStem),
+      ),
+    ];
+
+    global.__designSystemComponents = require('./src/site/_data/design/components');
```
samthor, line 169: Just make your code require this when it needs it. Unless I'm missing something and this is trying to work around 11ty's reload semantics. In which case this should be commented anyway 🤷

src/site/_includes/components/Aside.js
```diff
@@ -22,6 +25,33 @@ function Aside(content, type = 'note') {
...
+    switch (type) {
```
samthor, line 31: Some quick comments about why there's special-cases here would be good. e.g., "note" will end up with `title` being `undefined`.

src/site/_includes/components/Aside.js
andy-piccalilli: Nah 11ty only gives [page variables](https://www.11ty.dev/docs/data-eleventy-supplied/#page-variable-contents). It's a weird layer: you're right. It just helps us get over the line then it'll get yoinked out again

.eleventy.js
andy-piccalilli: This is going to go now. It was just a test-case that was janky anyway.

src/site/_includes/components/Aside.js
andy-piccalilli: This is getting deleted out but yep: noted (pun intended)

andy-piccalilli:
I've cleaned out all the stuff I was prototyping with now @samthor. All I need ongoing is the `isDesignSystemContext` and I can get cracking rolling stuff out.

src/site/_data/design/components.js
```diff
@@ -298,7 +299,12 @@ module.exports = {
     }
 
     // Passed data overwrites any context data.
+    // Regex removes whitespace and minifies: https://stackoverflow.com/a/59357436
```
samthor, line 302: Wait—we literally have a minify step in our build. Is this important earlier than that?

.eleventy.js
```diff
@@ -153,6 +153,22 @@ module.exports = function (config) {
     return memoize(collection.getAll());
   });
 
+  // Filters through all collection items and finds content that has
+  // CSS_ORIGIN set to 'next'. This allows shortcodes to determine if we
+  // are in a design system context or a legacy context
+  config.addCollection('designSystemGlobals', (collection) => {
+    global.__designSystemPaths = [
```
samthor, line 160: I think you should actually keep this as a `Set`, and use `.has()` in the helper.

samthor (approved):
I don't like the global but it's fine for now.

src/site/_data/design/components.js
andy-piccalilli: Ah, good spot because we actually don't need this now. It was playing up in a shortcode context but that won't be a problem now. I'll remove 👍

andy-piccalilli:
Yep I do not like it either, but it'll get us over the last hurdle

robdodson (approved):

PR #6228, opened 2021-09-16
Add translations for i18n strings

ilyaspiridonov:
This PR adds translations for strings under _data/i18n (Spanish, Portuguese, Russian, Korean, Japanese and Chinese).
Translated by Alconost.


src/site/_data/i18n/authors.yml
```diff
@@ -1,1116 +1,3347 @@
...
     en: Sam Thorogood
+    es: Sam Thorogood
+    ja: Sam Thorogood
+    ko: Sam Thorogood
+    pt: Sam Thorogood
+    ru: Sam Thorogood
+    zh: Sam Thorogood
   description:
     en: ¯‍\‍_‍(‍ツ‍)‍_‍/‍¯
+    es: '¯‍\‍_‍(‍ツ‍)‍_‍/‍¯'
```
samthor, line 678: Technically correct, the best kind of correct.

src/site/_data/i18n/authors.yml
```diff
@@ -1,1116 +1,3347 @@
...
+    zh: Orange Polska S.A. 销售平台开发与维护总监
 
 jiwoong:
   title:
     en: Jiwoong Lee
+    es: Jiwoong Lee
+    ja: Jiwoong Lee
+    ko: Jiwoong Lee
+    pt: Jiwoong Lee
+    ru: Чжи Ун Ли
```
samthor, line 2978: Most names don't seem to be translated but there are some cases like this, which are sort of oddly inserted in the set.

I presume @jimoong has suggested text for `ko`, which hasn't benn done.

Mostly I question the validity of, for nearly every author, just copying the English text to the other languages. I'd rather _not_ do this and fill in the gaps where we need to.

src/site/_data/i18n/browser_compat.yml
```diff
@@ -1,11 +1,35 @@
 browser_support:
   en: Browser support
```
jimoong, line 2: @samthor 
What's the context of this text?

samthor: I think this is @devnook 's "compatibility widget".

I suspect your (reasonable) concern is that the translations may have been done without any serious context on each string. If it's not high-quality we may need to find a way to feed context to translators, perhaps through a "_description" key here (i.e., adjacent to but not a language code).

src/site/_data/i18n/authors.yml
jimoong: ko: 이지웅

src/site/_data/i18n/authors.yml
```diff
@@ -1,1116 +1,3347 @@
...
+    zh: Jiwoong Lee
```
jimoong, line 2979: zh: 李智雄

src/site/_data/i18n/browser_compat.yml
jimoong: yeah, translation here looks correct and direct, but it may be wrong in different context. In that context, I'd suggest 
"지원되는 브라우저" and I guess this translation is more natural in most context.

src/site/_data/i18n/common.yml
```diff
@@ -1,55 +1,143 @@
...
+  ko: 핵심 용어
+  pt: Termo chave
+  ru: Ключевой термин
+  zh: 重要词汇
 
 gotchas:
   en: Gotchas
   es: Gotchas
+  ja: 問題の原因
+  ko: 갓차
```
jimoong, line 68: This is definitely wrong :) It's just transliterated. Could you explain what exactly the 'Gotcha' means or the usecases?

jimoong:
I briefly looked through and found some issues in Korean translation. Will
take a look more later :)



*Jiwoong Lee**Interaction Designer, Google*

samthor:
I can explain this but I don't want to play whack-a-mole with bad
translations. I think we need to write descriptions and send this all back
for another pass.

devnook:
@samthor @jimoong The context was provided via this doc: https://docs.google.com/document/d/1YF-eLtWwGRPYQReeaPOzUvhoKwBoLLpFv9RM2Lv30fc/edit?usp=sharing

src/site/_data/i18n/common.yml
devnook: I guess it's because this is a very "en" specific word if not "slangish". I'd definitely have problem to translate it to Polish.

src/site/_data/i18n/common.yml
```diff
@@ -1,55 +1,143 @@
...
+  pt: Pegadinhas
+  ru: Обратите внимание
+  zh: 问题
 
 do:
   en: Do
   es: Hacer
```
devnook, line 75: Do and Don't are like a pair, so should be in the same verb form. I think Hacer is an infinitive instead of 2st person form (No hagas).

src/site/_data/i18n/authors.yml
ilyaspiridonov: @samthor 

> Most names don't seem to be translated but there are some cases like this, which are sort of oddly inserted in the set.
> 
> I presume @jimoong has suggested text for `ko`, which hasn't been done here.
> 
> Mostly I question the validity of, for nearly every author, just copying the English text to the other languages. I'd rather _not_ do this and fill in the gaps where we need to.

@samthor Thank you. Do you mean we should only leave the EN version (unless other versions are provided by the Authors themselves)? In this case, if the Description is translated, and the name is not translated (i.e. only the en: key is present - will the page fall back to English version (and display the Description in English?) 
We're not copying over the English version into other languages manually - this is how translations tools usually work: if something is not translated, the English version is used, but the key for the target language is still generated by the reverse parser in case there is no fallback to English. 
But we don't  mind post-editing manually and removing the unused keys. Let me know what you think. 
CC : @devnook

src/site/_data/i18n/authors.yml
ilyaspiridonov: @jimoong Thank you, I have updated zh and ko versions

src/site/_data/i18n/browser_compat.yml
ilyaspiridonov: > I think this is @devnook 's "compatibility widget".
> 
> I suspect your (reasonable) concern is that the translations may have been done without any serious context on each string. If it's not high-quality we may need to find a way to feed context to translators, perhaps through a "_description" key here (i.e., adjacent to but not a language code).

@samthor While @devnook helped a lot with providing additional context for our translators, the main issue here is that with this particular case it's often impossible to run Linguistic QA on a localized build, as these strings seem to be rendered only for the pages/articles that have been translated (while many pages are in English only).  Of course, going forward, any additional description/context would help.

src/site/_data/i18n/common.yml
ilyaspiridonov: @devnook What would seem like the closets purely English synonym here?

devnook (approved):

PR #6257, opened 2021-09-21
test(presubmit): build a staging version of site to speed up HTML and CSS transforms

MichaelSolati:

- Change build command in presubmit to staging environment
- Update `purify-css` and `minify-html` to run on LHCI urls when in staging
-

### DIFF
--- a/.github/workflows/presubmit-workflow.yml
+++ b/.github/workflows/presubmit-workflow.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Run a production build of site
         run: npm run build
         env:
-          ELEVENTY_ENV: prod
+          ELEVENTY_ENV: staging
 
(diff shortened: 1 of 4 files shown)


samthor (commented):
This stops inlining CSS, but will the `<link rel="stylesheet"... Removal still happen - are the untested pages going to work? Do we care?

.github/workflows/presubmit-workflow.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Run a production build of site
```
samthor, line 26: comment is out of date

samthor (approved):
Actually, this is fine. I see the transform is happening on the HTML, not the CSS.

PR #6259, opened 2021-09-22
Add translated-to widget to posts

devnook:
Similar to d.c.c. one:
Screenshot 2021-09-22 at 16 44 24

### DIFF
--- a/src/site/_filters/urls.js
+++ b/src/site/_filters/urls.js
@@ -14,18 +14,22 @@
  * limitations under the License.
  */
 const path = require('path');
+const {supportedLocales} = require('../../../shared/locale');
+const {findByUrl} = require('./find-by-url');
+
+const i18nRegex = /i18n\/\w+\//;
 
 /**
  * Removes any i18n data from a URL to determine the default URL.
- *
  * @param {string} url URL with possible i18n folder and locale in prefix.
  * @returns {string} URL without any i18n information.
  */
-const getDefaultUrl = (url) =>
-  url.replace(
-    /^\/i18n\/([a-z]{2}|[A-Z]{2}|ALL)(_([a-z]{2}|[A-Z]{2}|ALL))?\//,
-    '/',
-  );
+const getDefaultUrl = (url) => {
+  if (typeof url !== 'string') {
+    return url; // shows up for `permalink: false`
+  }
+  return url.startsWith('/i18n/') ? url.replace(i18nRegex, '') : url;
+};
 
 /**
  * Extracts relative path from a url.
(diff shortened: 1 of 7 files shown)


src/site/_includes/components/LanguageList.js
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const {getTranslatedUrls} = require('../../_filters/urls');
+const {i18n} = require('../../_filters/i18n');
+const {languageNames} = require('../../../../shared/locale');
+
+module.exports = (url, lang) => {
+  const langhrefs = getTranslatedUrls(url)
+    .filter((langhref) => langhref[0] !== lang)
```
samthor, line 22: I'm not sure what's going on here but e.g. on a [non-English page](https://deploy-preview-6259--web-dev-staging.netlify.app/i18n/zh/vitals/), we don't show English in the list.

This is confusing because it _is_ being rendered in the meta tags.

devnook: We don't have a mechanism to force an English version of the page atm. Link web.dev.i18n/en/foo does not exist, and web.dev/foo will serve whichever language the browser is set to, defaulting to English. In meta tags it does not matter, because robots usually do not have language preference. 

This is I was reluctant to link to English, as there is no way to reliably point to "en" version of the page. This said, I think the widger do add value by making the other translations more discoverable, so I opted to still send a PR with this solution.

src/site/_filters/urls.js
```diff
@@ -14,18 +14,22 @@
...
-const getDefaultUrl = (url) =>
-  url.replace(
-    /^\/i18n\/([a-z]{2}|[A-Z]{2}|ALL)(_([a-z]{2}|[A-Z]{2}|ALL))?\//,
-    '/',
-  );
+const getDefaultUrl = (url) => {
+  if (typeof url !== 'string') {
+    return url; // shows up for `permalink: false`
+  }
+  return url.startsWith('/i18n/') ? url.replace(i18nRegex, '') : url;
```
MichaelSolati, line 31: Do we care if it's not a valid ISO code?

src/site/_includes/components/LanguageList.js
MichaelSolati: `web.dev/i18n/en/foo` does exist, but only in prodution. There's a gulp (I think) task that copies over all of the content not in the `i18n` folder to an en folder. This was because of when you have english as the primary language of the browser and spanish as the secondary firebase would try to serve spanish first.

MichaelSolati (approved):
Before merging take a look at what I said about the `en/foo`. If you don't think that makes a difference to anything then this PR is a LGTM.

src/site/_includes/components/LanguageList.js
devnook: Ah sweet! I totally forgot about this quirk. Thanks @MichaelSolati !

src/site/_filters/urls.js
devnook: This is more firebase-specific language combination code than iso code. I'll add a comment.

PR #6270, opened 2021-09-23
Adds linter rule for MDN links with locale, fixes all non-conforming URLs.

### LINKED ISSUE #6133: Don't link to en-US on MDN
As part of i18n, we should stop linking to en-US pages on MDN. Linking to the article without this seems to pick the best locale.

To be fair we do this... ~1351 times right now. We should script it.

malchata:
Fixes #6133

- Adds Markdown linting to detect MDN urls with locale info in them (e.g., en-US).
- Fixes all articles which contain URLs that violated the new linting rule.


tools/linting/bad-urls.js
```diff
@@ -21,6 +21,9 @@ const siteData = require('../../src/site/_data/site');
 
 module.exports = rule('remark-lint:bad-urls', checkURL);
 
+// Based on this Stack Overflow answer: https://bit.ly/3ELKYq3
+const locale = /^\/[a-z]{2,3}-[a-z0-9]{2,4}(-[a-z]{2})?/i;
+
```
devnook, line 26: I think ^ enforces the beginning of the string, so this regex will catch only relative urls starting with /locale/

malchata: `^` does enforce the beginning of the string, but that works here since `parsed.pathname` doesn't include the URL, just the path.

devnook (approved):
Thanks Jeremy!

Just one question, other lgtm.

PR #6274, opened 2021-09-24
Add i18n strings to toc

devnook:


src/component-library/web-table-of-contents/web-table-of-contents.njk
```diff
@@ -1,4 +1,4 @@
-<web-table-of-contents opened>
+<web-table-of-contents opened i18n="{{ i18n.toc | dump }}">
```
MichaelSolati, line 1: We may not need the entire toc file if were just getting the title. I wonder if we shouldn't do this.

devnook: I think i wanted to be universal here and always just pass all the translation strings the component might need. The i18n.toc object looks like this right now: https://github.com/GoogleChrome/web.dev/blob/main/src/site/_data/i18n/toc.yml so its just one property.
But maybe it is premature optimisation and we should pass only title, and add more i18n props if needed over time.

samthor:
My 2c on this generally: I like that we're only including a single YAML file, and one key versus two or even a few - it's so small that it's not worth worrying about.

However, I'd probably do this by adding a loader into our JS builder which lets us import ".yaml" as JSON objects. This is not significantly different from how we're allowed to import stylesheets, in that it's a custom loader that does _something_. 

The advantage here is that strings like this get included once in JS and not in every document, although you could argue again that this is really small.

Also, doesn't need an answer now, but I don't quite know how either approach works if this code gets moved to our infra repro. Probably the same, except that we either will have to:
(a) have to pass a i18n file down from the client repros (as this PR is now)
(b) have to support the ".yaml" loading behavior, and embed the YAML inside the infra repro
(c) just dump small strings like this into JSON directly inside the infra repro.

src/component-library/web-table-of-contents/web-table-of-contents.njk
samthor: This line doesn't need to be here yet right, and it still dumps the whole TOC?

It just seems out of place with the other usage right now.

src/lib/components/TableOfContents/index.js
```diff
@@ -27,6 +27,7 @@ class TableOfContents extends BaseStateElement {
   static get properties() {
     return {
       opened: {type: Boolean, reflect: true},
+      i18n: {type: Object},
```
samthor, line 30: Is it worth documenting on each class what properties we expect here? This might become useful if we move this to infra repro.

src/lib/components/TableOfContents/index.js
```diff
@@ -88,8 +91,9 @@ class TableOfContents extends BaseStateElement {
     this.observer.disconnect();
   }
 
-  onStateChanged({isTocOpened}) {
+  onStateChanged({isTocOpened, currentLanguage}) {
     this.opened = isTocOpened;
+    this.locale = currentLanguage;
```
samthor, line 96: This is fine for now but since we are no longer a SPA, and also language changes always caused a full reload, this could just be static.

src/site/_filters/i18n.js
```diff
@@ -60,6 +60,24 @@ const walk = (dir) => {
...
+/**
+ * Flattens the i18n yaml object to contain only translations from given locale.
+ * @param {Object} obj Object with i18n translations (from _data/i18n).
+ * @param {string} locale A locale prefix (example: 'en', 'pl')
+ * @return {Object|string}
+ */
+const getTranslation = function (obj, locale) {
+  const out = {};
+  for (const prop in obj) {
+    if (typeof obj[prop] === 'string') {
```
samthor, line 72: I think this is a bit confusing but I get the intention. It's worth documenting.

- If this object has any keys which point to a string, assume it's a leaf node that has translations.
- Otherwise it's probably an object which nests things, so convert that.

There's a weird corner case where we accidentally set some strings and some objects. I've seen authors do this because they were confused about yaml indentation. But I think this is rare enough not to validate again here.

src/site/_filters/i18n.js
```diff
@@ -60,6 +60,24 @@ const walk = (dir) => {
...
+      return obj[locale] ?? obj[defaultLocale];
+    } else {
+      out[prop] = getTranslation(obj[prop], locale);
+    }
+  }
+  return out;
+};
+
 /**
  * Looks for the i18n string that matches the path and locale.
```
samthor, line 82: This method now returns either a group of messages or an individual string.

Is it worth making two methods since the expected results are so different - we can throw in either case?

PR #6285, opened 2021-09-26
added singleQuote configuration to prettierrc.js

akshitarora921:


akshitarora921:
@googlebot I signed it!

samthor:
This does nothing. We already set this from the GTS config.

Issue #5323, opened 2021-05-11
CTA for opening side nav in Courses at top of modules


una:
Learn CSS modules can be like `001 / 023` as a button that opens the side nav

robdodson:
It's been a while since we discussed this so my memory is a bit fuzzy. @una is the idea ath the `000` that appears at the top of the course heading would be clickable, and clicking it would trigger the side navigation to open?

![image](https://user-images.githubusercontent.com/1066253/120719512-42784100-c47f-11eb-90c3-87bfa0af8db4.png)

PR #6299, opened 2021-09-28
add counter to courses

devnook:
Partially addresses #5323

### DIFF
--- a/src/site/_filters/navigation.js
+++ b/src/site/_filters/navigation.js
@@ -87,8 +87,10 @@ module.exports = function navigation(collection, toc) {
   // next/previous links and anything else.
   // The list only contains items with actual URLs.
   const list = [...map.values()];
+  const counterAll = list.length.toString().padStart(3, '0');
   list.forEach((item, idx) => {
     item.counter = idx.toString().padStart(3, '0');
+    item.counterAll = counterAll;
     item.prev = list[idx - 1] || null;
     item.next = list[idx + 1] || null;
   });
--- a/src/site/_includes/course.njk
+++ b/src/site/_includes/course.njk
@@ -52,7 +52,7 @@ inlineScripts:
       <main id="main">
         {% for item in pageNavigation.list %}
           {% if item.url === page.url %}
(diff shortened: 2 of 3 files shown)


robdodson:
I wonder if this is a bit confusing.

If you go here https://deploy-preview-6299--web-dev-staging.netlify.app/learn/css/conclusion/ it shows 024 / 025 which makes it seem like there should be one more page. But it may also be confusing to be on the first page (which is 000) and have it say 001 / 025.

If I recall correctly, Una's original issue was created because she wanted clicking on the course number to open the side nav—her primary concern was that the side nav may be hard to discover on smaller screens. So I'm not sure if showing 001 / 025 is as big of a priority as making the course number clickable. @una does that sound right? If that's the case then maybe we'd leave it as just `000` but clicking it would open the side nav?

IMO this may be more than necessary, the courses have been launched for a while and I don't think anyone has filed issues with the current design being hard to navigate on mobile.

MichaelSolati (approved):

MichaelSolati:
> I wonder if this is a bit confusing.
> 
> If you go here https://deploy-preview-6299--web-dev-staging.netlify.app/learn/css/conclusion/ it shows 024 / 025 which makes it seem like there should be one more page. But it may also be confusing to be on the first page (which is 000) and have it say 001 / 025.
> 
> If I recall correctly, Una's original issue was created because she wanted clicking on the course number to open the side nav—her primary concern was that the side nav may be hard to discover on smaller screens. So I'm not sure if showing 001 / 025 is as big of a priority as making the course number clickable. @una does that sound right? If that's the case then maybe we'd leave it as just `000` but clicking it would open the side nav?
> 
> IMO this may be more than necessary, the courses have been launched for a while and I don't think anyone has filed issues with the current design being hard to navigate on mobile.

I did not notice the number issue, I'd say that is important.

devnook:
AFAIK the current design does not allow to open the drawer easily on click. We would need to make the parent of the drawer aware that the opening action happened - similar as we do in ToC on articles. It's doable, but requires a biggish rewrite of courses code, so I'd like to know how important it is before jumping into it - @una can you advise?

PR #6316, opened 2021-09-30
fix(presubmit): cancel instead of fail if not presubmit

MichaelSolati:


robdodson (approved):

.github/actions/presubmit/index.js
```diff
@@ -36,11 +36,13 @@ async function run() {
       });
       core.info(`Succeeded to remove label: ${LABEL}`);
     } catch (e) {
-      core.notice(`Failed to remove label: ${LABEL}`);
+      core.warning(`Failed to remove label: ${LABEL}`);
```
samthor, line 39: This should still be an error, because it's been a failure in GitHub's API.

samthor (commented):
- I didn't know that actions can support their own "package.json" files. That's cool.
- Can you update the description a bit? It seems as if this does two things:

a) makes the presubmit task skip all its steps (and "pass") if it's not presubmit time
b) adds an additional task, which cancels (instead of fails), to check presubmit is present

Is this now race-y? Will the fact that "presubmit" have already passed, and the addtional task passes _then_ causes "presubmit" to be reevaluated - is there a moment there where everything is green but the presubmits haven't run?

I just wonder if this is simpler if we just have a single task that cancels, which itself is a cool property to display: here, we're basically saying "presubmit succeeds", which is bit misleading.

.github/actions/presubmit/index.js
MichaelSolati: An error will fail the overall workflow, no? Just cause we can't remove a tag doesn't mean that the overall status should be a fail if the other tests pass successfully.

robdodson (approved):

PR #6330, opened 2021-10-01
Add youtube playlist as a shortcode

### LINKED ISSUE #6242: Add ability to embed YouTube playlists
https://www.npmjs.com/package/eleventy-plugin-youtube-embed says playlists aren't supported, but I'm wondering if there's a workaround.

I've tried adding <iframe> code, but that doesn't seem to work too well.

Sample playlist: https://www.youtube.com/playlist?list=PLNYkxOF6rcICntazGfSVKSj5EwuR9w5Nv

devnook:
Fixes #6242

### DIFF
--- a/src/site/_includes/components/YouTubePlaylist.js
+++ b/src/site/_includes/components/YouTubePlaylist.js
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Iframe embed of a Youtube playlist
+ * @param {string} id Id of the playlist from Youtube
+ * @param {{src?: string; allow?: string; style?: string; title?: string;}} options
+ * @return string
+ */
+module.exports = (id, options) => {
+  if (!id) {
+    return;
+  }
+  const iframeStyle = [
+    'height: 100%;',
+    'left: 50%;',
+    'position: absolute;',
+    'top: 50%;',
+    'transform: translate(-50%, -50%);',
+    'border: 0;',
+    'width: 100%',
+  ];
+  let iframeProps = {
+    allow:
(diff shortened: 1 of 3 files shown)


src/site/_includes/components/YouTubePlaylist.js
```diff
@@ -0,0 +1,62 @@
...
+    title: 'IFrame content',
+  };
+  if (options && options.constructor === {}.constructor) {
+    iframeProps = {...iframeProps, ...options};
+  }
+  const {allow, src, style, title} = iframeProps;
+  return `
+    <div class="youtube">
+      <iframe
+        src="https://www.youtube.com/embed/videoseries?list=PLNYkxOF6rcICntazGfSVKSj5EwuR9w5Nv"
```
MichaelSolati, line 51: Should swap the list id to the `id` variable.

src/site/_includes/components/YouTubePlaylist.js
```diff
@@ -0,0 +1,62 @@
...
+        allow="${allow}"
+        loading="lazy"
+        src="${src}"
```
MichaelSolati, line 54: Why do we have two `src`?

src/site/_includes/components/YouTubePlaylist.js
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Iframe embed of a Youtube playlist
+ * @param {string} id Id of the playlist from Youtube
+ * @param {{src?: string; allow?: string; style?: string; title?: string;}} options
```
MichaelSolati, line 20: Probably don't need `src` since we're already getting the playlist ID.

MichaelSolati (changes requested):
This could/should probably be moved to infra so we can also use this in dcc. But most importantly is the comments on the YouTubePlaylist.js

src/site/_includes/components/YouTubePlaylist.js
```diff
@@ -0,0 +1,62 @@
...
+  ];
+  let iframeProps = {
+    allow:
+      'accelerometer; autoplay; clipboard-write; ' +
+      'encrypted-media; gyroscope; picture-in-picture',
+    src: `https://www.youtube.com/embed/videoseries?list=${id}`,
+    style: iframeStyle.join(''),
+    title: 'IFrame content',
+  };
+  if (options && options.constructor === {}.constructor) {
```
samthor, line 44: Do you expect callers to be malicious here?

I'd just write the above as

```js
const iframeProps = {
  stuff: "whatever",
  ...options,
};
```

... remembering that if options are null/undefined, then nothing happens.

PR #6340, opened 2021-10-05
Design system rollout on courses

andy-piccalilli:
This is a **very** large PR—apologies. It was necessary though because this is the first—and rather large—rollout of the new design system. 

Most things have been a doddle to work with, but existing web components have been _testing_, in places. 

Here's a quick tour of some of those limitations: https://www.loom.com/share/c2e20173703c4509ae66d2dfc2b9e6db

I definitely need fresh eyes on this now: I've been looking at it too long. 

If you want to see the dark mode stuff: run it locally, [pop over to the design system section (local link)](http://localhost:8080/design-system/) and toggle the dark mode on. 

A concern I have now is that the typography is noticeably much larger in the new system than the old system. Maybe a short term fix would be to [reduce the text sizing scale](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_data/design/tokens.json#L277) until we have a full rollout. Interested to get @jimoong’s thoughts on that, especially. 

Other than all that: we have covered _a lot_ of ground with this PR so future rollout PRs should be much smaller :)


robdodson:
As you mentioned in the video, the self assessment widget is still a bit off. On desktop it isn't so bad (side-by-side comparison):

image

But on mobile it's broken because the current widget tries to be very clever and have you open a modal to interact with it.

**This PR**:
image

**Current site**
image

If you go into lib/components/Assessment/index.js and lib/components/AssessmentQuestion/index.js, you should be able to [delete the lines where they import their SCSS](https://github.com/GoogleChrome/web.dev/blob/2fcdedb8965bc6eed0c7a5a3d5f4eae8c36cb717/src/lib/components/Assessment/index.js#L20). You could maybe try moving their existing SCSS over to `scss/web-components` to see if they still look mostly OK?

andy-piccalilli:
I think I'm going to have to monkey patch the assessment at this point, @robdodson because it has quite a bit of coverage outside of the courses section. I think once we tackle the general articles, we'll be able to give it a full refactor where I'll actually build a proper component for it in the design system.

robdodson:
Overall looks good. A couple of points:

- There's a scss lint error that still needs to be fixed
- Is it possible to toss in a style to give the assessment a white background? It looks a bit odd being all grey.
image

If not, that's ok, but wanted to flag it. I'll go ahead and approve so you can land this, but be sure to fix the linter error before merging.

PR #6332, opened 2021-10-01
feat(lint): add a linter for tags, fixes #2240

MichaelSolati:
Fixes #2240

Changes proposed in this pull request:

- Create action to lint tags
-
-

When you're ready to submit your PR, don't forget to add the `$-presubmit` label.

PR #6346, opened 2021-10-06
Future-proof the reference to Chrome 93.

jpmedley:

Articles like this are around long after the release of Chrome versions they mention. This article in particular, published yesterday, says "when Chrome 93 is released". Chrome 93 was released 36 days ago. This PR changes the phrase to use the words, "In Chrome 93 or later."

### DIFF
--- a/src/site/content/en/blog/sanitizer/index.md
+++ b/src/site/content/en/blog/sanitizer/index.md
@@ -8,6 +8,7 @@ description: |
 authors:
   - jackjey
 date: 2021-10-06
+updated: 2021-10-07
 hero: image/PV7xjXdOKHP8LWt9XhstsToJeK82/R2KGdIHv2lBRDS9vs8q1.jpg
 alt: Sanitizing a desk with Sanitizer
 tags:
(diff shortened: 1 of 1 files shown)


jpmedley:
@MichaelSolati The script is failing this because of the missing presubmit label. But it was a script that removed it. Please help?

MichaelSolati:
Odd, ok. So it looks like the workflow ran a few times. I think what happened is that you added the tag and it ran, removed the tag, and [passed](https://github.com/GoogleChrome/web.dev/actions/runs/1312343400). But the CLA bot added the CLA tag (which triggers the script to run again, it looks for when tags are added) and at that point the the presubmit tag had been removed by the first execution of the presubmit workflow so it failed the second time.

I think going forward wait until after the CLA stuff is done and be aware other tags being added will trigger the test. I'm presently working on some improvements to the script so I'll see if I can prevent something like this from happening again.

Also, it looks like the tag linter found an issue with the use of the tag `chrome93`, tag linting got [added yesterday](https://github.com/GoogleChrome/web.dev/pull/6332). I do believe we could add a general `chrome` tag to our list of [supported tags](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_data/i18n/tags.yml) and you could use that newly created tag instead. @robdodson or @devnook how do y'all feel about adding a chrome tag to the site?

jpmedley:
@MichaelSolati I've been waiting on the presubmit for nearly 24 hours now. That seems a little long. Please help?

MichaelSolati:
@jpmedley is it ok if I just merge it?

Issue #6352, opened 2021-10-07
`BrowserCompat` widget is broken when used inside `Details` clause.


agektmr:
**Describe the bug**
When a `BrowserCompat` widget is inserted within a `Details` clause, it terminates the `Details` section and doesn't display the widget as well.

**To Reproduce**
1. Insert `{% BrowserCompat 'api.BackgroundFetchEvent' %}` to https://github.com/GoogleChrome/web.dev/blob/main/src/site/content/en/secure/security-headers/index.md?plain=1#L260
2. Render the page locally
3. Expland the `Details` section by clicking on "How to use CSP" and look for "Supported browsers". The section terminates the rest of `Details` section and the widget is not displayed.

**Expected behavior**
The widget is displayed and the rest of `Details` will be displayed.

**Screenshots**
<img width="905" alt="Screen Shot 2021-10-07 at 16 45 32" src="https://user-images.githubusercontent.com/218144/136341662-169f2e32-4016-4248-95a8-3e6feb37c6c4.png">

PR #6355, opened 2021-10-07
Add warning about async shortcodes

devnook:
Related to #6352


samthor:
Can I point out that the BrowserCompat shortcode is only async because the 11ty-cache-assets plugin requires you to be async, even though all we're doing is caching a file to-disk, which _could actually be done_ with `fs.writeFileSync`.

So there's two issues here.

1. Yes, it's incredibly awkward that an async shortcode works like this - it seems implausible that Nunjucks can't deal with it?
2. Having said that, do we need to cache this at all? The transform looks pretty fast honestly (flattening the content).

PR #6371, opened 2021-10-08
Add Cloudflare ASX to the SXG tooling list.

twifkak:
It is out of waitlist, and now generally available.

/cc @khempenius


twifkak:
Thanks all. Checks pass; please merge when you can.

PR #6382, opened 2021-10-11
Updates to handbook quick start

rachelandrew:
This PR makes the Quick Start guide generic to both sites and adds the content about writing blog posts.

At the bottom of the Quick Start Guide is a section about organizing content. It doesn't really seem as if that is the best place for that as it's only relevant to web.dev. So perhaps as part of this I move that elsewhere and just link to the information for the two sites?

cc: @mihajlija @malchata @jpmedley @robdodson please take a look, make any edits, or make any suggestions for stuff I should add to this PR.


rachelandrew:
> Is it worth updating [the DCC handbook](https://developer.chrome.com/docs/handbook/) with a link to this content writing guide? That handbook is much more about specific engineering tasks, rather than the general steps to writing articles.

yeah, that's essentially the next step along with pointing dcc contribs to the newly generic quickstart guide, but I'll do that once this is merged.

PR #6424, opened 2021-10-13
Implement measure tool with PSI instead of lighthouse-keeper

### LINKED ISSUE #6385: Reimplement measure page to use PSI

devnook:
This PR adds an alternative (unpublished) url for measure powered by PSI at /measure-psi.  
The page remains largely unstyled yet, as I wait for @andy-piccalilli to implement header and page structure in next.css first (see #6419). I still think it will be nice to check in the js parts of the logic now, and style the page later.

Fixes #6385


samthor:
This works really well. Two minor issues which are really styling things:

* the "PWA" site tick doesn't go anywhere

* when scrolling down, this tries to be `position: fixed` but it's under the header and doesn't have a background

Screen Shot 2021-10-14 at 14 43 47

devnook:
This is ready for a second review. 

I removed conditional logic around user saved urls. Let's keep sign-in and fb as is, and consider removing it separately in another PR (if we decide so). The goal of this PR is to get /measure-psi page render correctly with the new design, new api call and new lighthouse report embed. 

The copy on this page is still to be updated. Once we paste in new copy and test it thoroughly, we can swap /measure and /measure-psi (in some point in future).

PR #6525, opened 2021-10-25
feat(carousel): implement carousel widget as a new component, fixes #6390

### LINKED ISSUE #6390: Implement Carousel widget as a new component

MichaelSolati:
Fixes #6390

- Add HTML for carousel widget
- Add custom element wrapper for carousel widget to give it JS functionality
-


src/site/content/en/design-system/css-compositions.njk
```diff
@@ -126,6 +126,79 @@ configurable grid item sizes.</p>
...
+    <tr>
+      <th>Property</th>
+      <th>Default value</th>
+      <th>Description</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td><code>--reel-item-width</code></td>
+      <td><code>33.333%</code> (one third_</td>
```
samthor, line 148: nit: _ should be )

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+    );
+
+    this._carouselTrack.addEventListener('scroll', this._scroll);
+
+    this.moveSlide(0, false);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    if (this._previousButton)
```
samthor, line 82: This becomes much easier to read if you use optional chaining, which does have 90%+ support (https://caniuse.com/mdn-javascript_operators_optional_chaining) even if we don't transpile it.

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+      this._previousButton.removeEventListener('click', this.previous);
+    if (this._nextButton)
+      this._nextButton.removeEventListener('click', this.next);
+    if (this._timeout) clearTimeout(this._timeout);
+    this._items.forEach((e) => e.parentElement.removeChild(e));
```
samthor, line 87: nit: `e.remove()` is much easier

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+    this.moveSlide(0, false);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    if (this._previousButton)
+      this._previousButton.removeEventListener('click', this.previous);
+    if (this._nextButton)
+      this._nextButton.removeEventListener('click', this.next);
+    if (this._timeout) clearTimeout(this._timeout);
```
samthor, line 86: You don't need to check this. You can always safely clear a timeout. This should also be `window.clearTimeout`—I think right now you're typing it as `NodeJS.Timeout`, and ... this isn't Node. Using `window.setTimeout` means it's a `number`.

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+    const elementCount = this._items.length;
+    let newIndex = this._index + increment;
+
+    if (newIndex < 0) {
+      newIndex = elementCount - 1;
+    } else if (newIndex >= elementCount) {
+      newIndex = 0;
+    }
+
+    if (newIndex === this._index) return;
```
samthor, line 102: nit: always `{}`

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+    if (this._timeout) clearTimeout(this._timeout);
```
samthor, line 103: Again, you don't need the conditional. You can just `window.clearTimeout(this._timeout)`.

(Also, you never set it back to zero or undefined _anyway_.)

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+
+    this._items.forEach((e, i) => e.classList.toggle('active', i === newIndex));
+    this._index = newIndex;
+    this._active = this._items[this._index];
```
samthor, line 107: Remove `this._active`, you don't use it for anything.

Just replace ths section with:

```js
const active = this._items[this._index];
if (this.interval === 0) {
  active.focus({preventScroll: true});
} else {
  this._timeout = window.setTimeout(this.next, this.interval);
}
if (scroll) {
  this._carouselTrack.scrollTo(..);
}
```

(You can drop the `_setTimeout` helper, it's only used here)

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+      this._previousButton.removeEventListener('click', this.previous);
+    if (this._nextButton)
+      this._nextButton.removeEventListener('click', this.next);
+    if (this._timeout) clearTimeout(this._timeout);
+    this._items.forEach((e) => e.parentElement.removeChild(e));
+    if (this._carouselTrack)
+      this._carouselTrack.removeEventListener('scroll', this._scroll);
+  }
+
+  moveSlide(increment, scroll = true) {
```
samthor, line 92: document types

```js
/**
 * @param {number} increment How many items to move by
 * @param {boolean} scroll Whether to move the user's viewport.
 */
```

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+
+  next() {
+    this.moveSlide(1);
+  }
+
+  previous() {
+    this.moveSlide(-1);
+  }
+
+  _scroll() {
```
samthor, line 122: Could you change this name to `_onScroll` and give it a comment? It took me a minute to work out that this is trying to, on a user's scroll event, determine the best element to make active.

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {BaseElement} from '../BaseElement';
+
+class Carousel extends BaseElement {
+  static get properties() {
+    return {
+      interval: {type: Number},
```
samthor, line 22: This makes the carousel automagically move through items, right? Can you document this?
Is it used or just for fun now—if so, can we drop the code?

src/lib/components/Carousel/index.js
```diff
@@ -0,0 +1,146 @@
...
+    ]);
+
+    this._items.forEach((v, i) => v.setAttribute('data-index', '' + i));
+
+    this._previousButton = this.querySelector('button[data-direction="prev"]');
+    this._previousButton.addEventListener('click', this.previous);
+    this._nextButton = this.querySelector('button[data-direction="next"]');
+    this._nextButton.addEventListener('click', this.next);
+
+    this._items.forEach((e, i) =>
```
samthor, line 71: You don't need to change this now, but remember that `focusin` bubbles. You could actually just add the listener _once_, then work out which item fired it in the handler.

samthor (approved):
Just approving to unblock. Please address comments :)

andy-piccalilli:
Hey @MichaelSolati, I've given this another visual pass and it looks really decent now, nice work! 

I've added an extra wrapper to child elements to assist with a little CSS trick to prevent the weird overflow behaviour on cards in their hover state. If and when this is ready to roll out to the [course cards](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/partials/course-cards-next.njk), just make sure you make a note of [that markup change](https://github.com/GoogleChrome/web.dev/pull/6525/commits/92d6ced7f339707c67755f5c5790af6b6b22dfd9#diff-16a187ffcb5281eb5b7e14c1fd033795c938431afe227a38731ec3cdead28de3R9).

src/lib/components/Carousel/index.js
samthor: Not fixed? You can always clear timeout.

src/lib/components/Carousel/index.js
```diff
@@ -119,7 +128,12 @@ class Carousel extends BaseElement {
     this.moveSlide(-1);
   }
 
-  _scroll() {
+  /**
+   * Event listener function that determines which element a user has scrolled to.
+   *
+   * @returns {void}
```
samthor, line 134: this line isn't needed, it's implied 🤷

PR #6574, opened 2021-10-28
fix(LearnCollection): fix when no id is selected, fixes #6555

### LINKED ISSUE #6555: Collections > All filter hides all cards
**To Reproduce**
1. Go to https://web.dev/learn/
2. Click on Collections > All button
3. All collection cards disappear.

MichaelSolati:
Fixes #6555

- Ensure boolean is used rather than `undefined`
- Add tests.
-


MichaelSolati:
@devnook tests have been added. It was weird cause querying the shadow DOM of the custom element wouldn't work. It was empty. But I found this... `await filter.updateComplete;`  I guess the test was running before the first life cycle, so it just would be empty inside. This was driving me crazy for a while cause I thought that the array I was passing in was the issue, then I thought that I couldn't query the shadow DOM. Eventually I found the `updateComplete` on StackOverflow!

Anyway we should be good to go!

src/lib/components/LearnFilter/index.js
```diff
@@ -9,16 +9,19 @@ class LearnFilter extends BaseElement {
...
   }
 
   constructor() {
     super();
-    /** @type {{id?: string, title: string}[]} */
-    this.filters = [];
     /** @type {string|undefined} */
     this.active = undefined;
+    /** @type {string} */
+    this.all = 'All';
```
samthor, line 22: maybe comment that this is purely a fallback for i18n

src/lib/components/LearnFilter/index.js
```diff
@@ -33,11 +36,13 @@ class LearnFilter extends BaseElement {
     }
 
     for (const child of learningPathsElement.children) {
-      child.classList.toggle('hidden', id && id !== child.id);
+      child.classList.toggle('hidden', !!id && id !== child.id);
```
samthor, line 39: Ah, this freaking logic.

I think this change is a no-op, since `!!id` should always be the same truthiness or falsiness.

If you want to be very readable I'd wrap it in `Boolean(id)` though.

test/unit/src/lib/components/LearnFilter/index.js
```diff
@@ -0,0 +1,102 @@
...
+      allButton.click();
+
+      let notHidden = 0;
+
+      for (const child of learningPathsElement.children) {
+        notHidden += +!child.classList.contains('hidden');
+      }
+      assert(notHidden === collectionIds.length, 'no element should be hidden');
+    } finally {
+      filter.remove();
```
samthor, line 77: Can this go into an `after()` call?

samthor (approved):

test/unit/src/lib/components/LearnFilter/index.js
MichaelSolati: The elements don't live outside of the function so I don't think the `after` call will know about them.  ¯\_(ツ)_/¯

PR #6600, opened 2021-11-01
Design update snags

### LINKED ISSUE #6577: Top card gets squished in the new homepage at viewport ~330px
Top card gets squished in the new homepage at viewport ~330px:
Screenshot 2021-10-29 at 11 58 15

### LINKED ISSUE #6596: Stats and columns component are not rendering in the Learn courses
We tried to use the stats and the column components, and they don't seem to be rendering with the new CSS (within courses at least). I don't currently have a live example of this, as I'm working locally. I can update here once this PR preview is up though if it helps.

andy-piccalilli:
Fixes #6577

Also applies some typsetting tweaks discussed between @jimoong and I.

### DIFF
--- a/src/scss/utilities/_wrapper.scss
+++ b/src/scss/utilities/_wrapper.scss
@@ -4,8 +4,8 @@
 .wrapper {
   margin-left: auto;
   margin-right: auto;
-  width: 80rem; // For no clamp support
-  width: clamp(16rem, 95vw, 80rem);
+  width: 85rem; // For no clamp support
+  width: clamp(16rem, 95vw, 85rem);
   max-width: 100%;
   margin-left: auto;
   margin-right: auto;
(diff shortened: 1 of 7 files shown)


src/scss/utilities/_wrapper.scss
```diff
@@ -4,8 +4,8 @@
 .wrapper {
   margin-left: auto;
   margin-right: auto;
-  width: 80rem; // For no clamp support
-  width: clamp(16rem, 95vw, 80rem);
+  width: 85rem; // For no clamp support
```
devnook, line 7: Do we go wider wrapper again?

devnook (approved):

src/scss/utilities/_wrapper.scss
andy-piccalilli: Yeh we bumped it up a touch to allow for the better typesetting in the hero etc

PR #6614, opened 2021-11-01
refactor(course-cards-next): use carousel

MichaelSolati:

- Update course cards partial to use carousel
- Carousel tracks active card when tabbing and using arrow keys.
- Carousel scrolls when using forward/back buttons.

### DIFF
--- a/src/lib/components/Carousel/index.js
+++ b/src/lib/components/Carousel/index.js
@@ -44,6 +44,7 @@ class Carousel extends BaseElement {
 
     this.next = this.next.bind(this);
     this.previous = this.previous.bind(this);
+    this._onKeyup = this._onKeyup.bind(this);
     this._onScroll = this._onScroll.bind(this);
   }
 
@@ -60,8 +61,6 @@ class Carousel extends BaseElement {
       ...this._carouselTrack.children,
     ]);
 
-    this._items.forEach((v, i) => v.setAttribute('data-index', '' + i));
-
     this._previousButton = this.querySelector('button[data-direction="prev"]');
     this._previousButton.addEventListener('click', this.previous);
     this._nextButton = this.querySelector('button[data-direction="next"]');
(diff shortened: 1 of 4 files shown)


samthor:
This is fine but the new homepage doesn't actually load the "Home" entrypoint.

I think you want to add this to "homepage-next.njk":

```
pageScripts:
  - '/js/home.js'
```

(which matches "homepage.njk")

Approved but please do this.

samthor (approved):

devnook:
I think the keyboard interaction is broken, e.g:

1. Hit tab to focus the "<" button
2. Hit enter to activate it
3. Carousel scrolls right (>) and 2nd items gets focus.

or:

1. Hit tab to focus the "<" button
2. Hit tab to enter into carousel. 1st item gets focused
3. Hit right arrow. 2nd item gets focused
4. Hit right arrow again. Nothing happens.
5. You can still go left and right between first and second item, but cannot get to 3rd and 4th.

src/site/_includes/partials/course-cards-next.njk
```diff
@@ -1,27 +1,37 @@
...
-      </a>
-    {% endif %}
-  {% endfor %}
-</nav>
+<web-carousel>
+  <div class="carousel">
+    <button class="icon-button" data-direction="prev" aria-label="back">
+      {% include "icons/carat-back.svg" %}
+    </button>
+    <div class="carousel__track reel" data-scroll="snap" aria-label="available courses">
```
devnook, line 8: please add back the scrollbar class to that scrollbar is visible when idle.

src/lib/components/Carousel/index.js
```diff
@@ -60,90 +49,98 @@ class Carousel extends BaseElement {
...
     this._items.forEach((e, i) =>
-      e.addEventListener('focusin', () => this.moveSlide(i - this._index)),
+      e.addEventListener('focusin', () => this._moveSlide(i - this._index)),
     );
+    this._carouselTrack.addEventListener('keyup', this._onKeyup);
 
-    this._carouselTrack.addEventListener('scroll', this._onScroll);
-
-    this.moveSlide(0, false);
+    this._moveSlide(0);
```
devnook, line 63: Why this? It gives the focus on first card after render, even if no user action indicated they indicated to focus that card. 

It gives an additional side effect, that if they load the page, dont even scroll to the carousel, and hit accidentally enter, they will navigate to a random page.

devnook (changes requested):
Apart from the comment left, lgtm. I would love to see a test for this though at some point.

src/lib/components/Carousel/index.js
MichaelSolati: This is a total remnant of the way it used to work, it definitely shouldn't be here now.

MichaelSolati:
@devnook PTAL

devnook (approved):

PR #6624, opened 2021-11-02
Make good TBT thresholds consistent

### LINKED ISSUE #6623: content: inconsistent TBT threshold
In https://web.dev/tbt/#what-is-a-good-tbt-score it says 300 ms is the good TBT threshold.
In https://web.dev/lighthouse-total-blocking-time/#how-lighthouse-determines-your-tbt-score it says the threshold is 200 ms.

200 ms is the correct threshold as of Lighthouse 8 per @brendankenny, so /tbt should be updated.

rviscomi:
Fixes #6623

The good TBT threshold is 200 ms


brendankenny (approved):
Nice catch!

PR #6643, opened 2021-11-02
Remove flex none from lh buttons

devnook:


src/scss/next.scss
```diff
@@ -394,6 +394,7 @@ details summary {
   font-size: get-size('size-2');
   list-style: none;
   cursor: pointer;
+  padding-right: 28px;  // Prevent overlap with ::before element
```
andy-piccalilli, line 397: Can you try `get-space('size-1')` on this please?

andy-piccalilli (approved):

src/scss/next.scss
devnook: Should I then also use this on ::before height&width?

PR #6719, opened 2021-11-06
Added in a missing not so obvious step to the README

gssakash:

- I've added in one additional step to the README under the "Building the site" section which is not obvious to a lot of beginners. 
- I'm quite experienced with Git but I still ran all of the other commands without performing the above step. 
- I basically had to redo the steps again once I changed my current working directory.

Let me know I can make any improvements or if I've made any mistakes.


gssakash:
@googlebot I signed it!

PR #6731, opened 2021-11-08
fix(Snackbar): use new button classes for snackbar, fixes #6705

### LINKED ISSUE #6705: Web snackbar looks wonky
I don't think the snackbar received an updated look with the new design system and looks off:

Before:
image

After:
image

It's small, but should be looked at.

MichaelSolati:
Fixes #6705

- Add clear button class.
- Add clear button class/attribute to snackbar buttons.

Before:
image

After:
image

This'll obviously need some love and attention if/when we remove the component level styling and the old SCSS.

### DIFF
--- a/src/styles/tools/_mixins.scss
+++ b/src/styles/tools/_mixins.scss
@@ -121,7 +121,7 @@
   justify-content: center;
   letter-spacing: 1px;
   outline: 0;
-  padding: 0 16px;
+  padding: 0 16px !important;
   position: relative;
   text-decoration: none;
   text-transform: uppercase;
--- a/src/styles/tools/_mixins.scss
+++ b/src/styles/tools/_mixins.scss
@@ -121,7 +121,7 @@
   justify-content: center;
   letter-spacing: 1px;
   outline: 0;
-  padding: 0 16px !important;
+  padding: 0 16px !important; // `.button` from new design system removes padding, this keeps it.
   position: relative;
   text-decoration: none;
   text-transform: uppercase;
--- a/src/styles/tools/_mixins.scss
+++ b/src/styles/tools/_mixins.scss
@@ -121,7 +121,7 @@
   justify-content: center;
   letter-spacing: 1px;
   outline: 0;
-  padding: 0 16px !important; // `.button` from new design system removes padding, this keeps it.
+  padding: 0 16px;
   position: relative;
   text-decoration: none;
   text-transform: uppercase;
--- a/src/component-library/button/button.json
+++ b/src/component-library/button/button.json
@@ -28,6 +28,14 @@
       }
     },
     {
+      "name": "clear",
+      "title": "Clear",
+      "context": {
+        "text": "A clear button",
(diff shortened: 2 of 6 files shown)


src/styles/tools/_mixins.scss
```diff
@@ -121,7 +121,7 @@
   justify-content: center;
   letter-spacing: 1px;
   outline: 0;
-  padding: 0 16px;
+  padding: 0 16px !important;
```
devnook, line 124: Why important in a mixin? There must be a very strong reason to do so.

andy-piccalilli: To add to this, I always make sure to add an explainer comment for `!important` (because there are reasons to use it afterall) to help people understand its purpose when they discover it

MichaelSolati: Yes! Uhm, so adding the `button` class used by the new design system removes the padding from `w-button` in parts of the site that doesn't use the new design.

MichaelSolati: I'll add a comment explaining that now.

andy-piccalilli (approved):
LGTM

I've just pushed up a specificity rule for the buttons which will get sorted once and for all in #6747 

This has also made me open #6746 because it's becoming rather problematic

src/styles/tools/_mixins.scss
devnook: If this is a temporary fix to resolve the old and the new specificity issues it should all go into _rollout.scss, so that we know to delete it when rollout is complete.

devnook (changes requested):
Can you please rebase with recent Andy's changes (mentioned in his comment) and check if it still works as intended?

MichaelSolati:
@devnook it looks like Andy hasn't started reworking the snackbar, but we discovered an unused class causing the padding to be removed. So we yanked that and were able to remove the `!important`, so it should be gtg.

MichaelSolati:
@devnook PTAL

devnook:
Screenshot 2021-11-17 at 14 33 26

This is how it looks like atm. Should be more like in your screnshot i think?

MichaelSolati:
> Screenshot 2021-11-17 at 14 33 26
> 
> This is how it looks like atm. Should be more like in your screnshot i think?

That's how it should be when `isStacked` is set to `false` in `SnackbarContainer`. However if `this.type === 'cookies'` then `isStacked` is `true`. I pushed up a temporary commit (it will need to be removed before merging this PR) so you can see the snackbar now.

image
https://deploy-preview-6731--web-dev-staging.netlify.app/

image
https://deploy-preview-6731--web-dev-staging.netlify.app/mailru-cwv/

And here are the same two when `isStacked` is `false`.

image

image

MichaelSolati:
@devnook PTAL

devnook:
It looks much better when not stacked. why do we stack it then?

devnook (approved):

PR #6732, opened 2021-11-09
[File Handling article] Remove OT info

tomayac:

- Remove OT info.


jpmedley (approved):
So still not going out in 97? :(

tomayac:
> So still not going out in 97? :(

It _might_ still happen: https://bugs.chromium.org/p/chromium/issues/detail?id=829689#c135

PR #6750, opened 2021-11-09
fix(algolia): slim down the payload for algolia indexing while speeding up the json build, fixes #5274, fixes #6435

### LINKED ISSUE #5274: Algolia Language fallbacks should not be combined.
**Describe the bug**
See https://github.com/GoogleChrome/web.dev/pull/5273

We currently put the english contents into the spanish article as a fallback search result. But this means the spanish record is too big for algolia and we have to truncate it.

### LINKED ISSUE #6435: Indexing is failing due to size issues
```
Starting Step #4 - "Algolia index site"
Step #4 - "Algolia index site": Already have image: node:14
Step #4 - "Algolia index site": 
Step #4 - "Algolia index site": > web.dev@3.0.0 algolia /workspace
Step #4 - "Algolia index site": > node algolia.js
Step #4 - "Algolia index site": 
Step #4 - "Algolia index site": Indexing 1433 articles amongst 1 chunk(s).
Step #4 - "Algolia index site": {
Step #4 - "Algolia index site":   name: 'ApiError',
Step #4 - "Algolia index site":   message: 'Record at the position 379 objectID=5af46e585f15f89dace00eb8294f433c is too big size=24401/10000 bytes. Please have a look at https://www.algolia.com/doc/guides/sending-and-managing-data/prepare-your-data/in-depth/index-and-records-size-and-usage-limitations/#record-size-limits',
Step #4 - "Algolia index site":   status: 400,
Step #4 - "Algolia index site":   transporterStackTrace: [
Step #4 - "Algolia index site":     {
Step #4 - "Algolia index site":       request: [Object],
Step #4 - "Algolia index site":       response: [Object],
Step #4 - "Algolia index site":       host: [Object],
Step #4 - "Algolia index site":       triesLeft: 3
Step #4 - "Algolia index site":     }
Step #4 - "Algolia index site":   ]
Step #4 - "Algolia index site": }
Step #4 - "Algolia index site": (node:19) UnhandledPromiseRejectionWarning: #<Object>
Step #4 - "Algolia index site": (Use `node --trace-warnings ...` to show where the warning was created)
Step #4 - "Algolia index site": (node:19) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
Step #4 - "Algolia index site": (node:19) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

And we should be killing the process, but an unhandled rejection isn't failing the build. Ah, Node 14.

MichaelSolati:
Fixes #5274, fixes #6435

- Delete `src/site/_collections/pages.js`.
- Turn some functions/logic from `src/site/_collections/pages.js` into 11ty filters.
- Remove english fallback from entries.
- Move parsing logic from `src/site/_collections/pages.js` to the template `src/site/content/pages.njk`.
- Add `locales` to `pages` data in `algolia.js`.

I believe that parsing the JSON data in the `src/site/content/pages.njk` rather than in `src/site/_collections/pages.js` should be significantly faster. While enabling the JSON to be built in dev mode before it would take close to 4.5 minutes, using the njk template it takes .75 minutes.

This method is almost 4 minutes faster.


algolia.js
```diff
@@ -93,11 +88,25 @@ async function index() {
 
   const raw = fs.readFileSync('dist/pages.json', 'utf-8');
   /** @type {AlgoliaItem[]} */
-  const algoliaData = JSON.parse(raw).map((e) => {
-    // Set date of when object is being added to algolia.
-    e.indexedOn = indexedOn.getTime();
-    return e;
-  });
+  const algoliaData = JSON.parse(raw)
+    .map((/** @type {AlgoliaItem} */ e) => {
```
devnook, line 92: "e" is usually used for "event". What does it stand for here? Can you use full name instead?

algolia.js
```diff
@@ -93,11 +88,25 @@ async function index() {
...
+      addLocaleToUrl(e.locale, e.url);
+      // Set date of when object is being added to algolia.
+      e.indexedOn = indexedOn.getTime();
+      return e;
+    })
+    .map((/** @type {AlgoliaItem} */ e) => {
```
devnook, line 98: Optional: Having 2 map functions operating on the same array seems usually suspicious (raises question of "why double loop?"). You can improve readability by adding a comment, or by splitting it into 2 statements:

const posts = JSON.parse(raw);
const urlToLocale = posts.reduce(...);   // Here you build up your locale dict (equivalent of 1st loop)
const algoliaData = posts.map(...);   // (2nd loop)

This also lets you avoid having names like "urlToLocale" and "addLocaleToUrl" which are confusing what you add to whom.

algolia.js
```diff
@@ -93,11 +88,25 @@ async function index() {
...
+      if (e.locale === defaultLocale) {
+        const locales = urlToLocale[e.url];
+        e.locales = [
+          e.locale,
+          ...supportedLocales.filter((i) => locales.indexOf(i) === -1),
```
devnook, line 103: Use isSupportedLocale instead. You can also do this check when you build urlToLocales map instead of post-processing here.

devnook (changes requested):
Reviewed

algolia.js
```diff
@@ -83,8 +69,17 @@ const chunkAlgolia = (arr) => {
   return chunked;
 };
 
-async function index() {
+(async () => {
```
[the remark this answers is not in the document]
MichaelSolati: No good reason for this, at least not one that I can remember. I'll revert this.

src/site/_filters/live-pages.js
```diff
@@ -0,0 +1,41 @@
...
+ * @return {EleventyCollectionItem[]} The posts that should go be shown.
+ */
+function livePages(posts) {
+  return posts.filter(
+    (post) =>
+      livePosts(post) &&
+      post.url &&
+      post.data.title &&
+      !(
+        post.data.disable_algolia === true ||
```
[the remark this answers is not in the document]
MichaelSolati: Now that we're using just `noindex` we will still need to do 2 variations.. `!(post.data.noindex === true || post.data.noindex === 'true')`. `noindex` can end up being a string due to how `eleventyComputed` will parse the value, so we need to check if it is `'true'` or `'false'`. But two cases are simpler than three, so hopefully we're ok here.

src/site/_filters/live-pages.js
```diff
@@ -0,0 +1,41 @@
...
+const {livePosts} = require('./live-posts');
+
+/**
+ * @param {EleventyCollectionItem[]} posts An array of eleventy post object.
+ * @return {EleventyCollectionItem[]} The posts that should go be shown.
+ */
+function livePages(posts) {
+  return posts.filter(
+    (post) =>
+      livePosts(post) &&
```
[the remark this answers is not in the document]
MichaelSolati: `livePosts` is designed for checking one item instead of an array of items, so I don't think I can use it like this. Unless I created a new filter for `livePosts` that handled an array of posts... (Honestly, `livePosts` sounds like it should be called `livePost`).

types/utils/algolia.d.ts
```diff
@@ -20,6 +20,7 @@ declare global {
      * Date of index, only available after indexed. Basically can only be used by search.
      */
     indexedOn?: number;
+    locales: string[];
```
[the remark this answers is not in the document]
MichaelSolati: This is added in `algolia.json` before updating the alogia entries.

algolia.js
MichaelSolati: So this filter removes the locales that do exist as urls from the `supportedLocales` array. Where as `isSupportedLocale` ensures that the locale does exist. They kind of do the opposite thing. So I don't believe I can do what you're suggesting.

MichaelSolati:
@devnook PTAL, there is one hold up... https://github.com/GoogleChrome/web.dev/pull/6804 needs to land first then i can change the algolia indexable filter to use the new `livePosts`

types/eleventy/collection-item.d.ts
```diff
@@ -95,7 +91,7 @@ declare global {
       /**
        * Tells search engines not to index page (this includes algolia)
        */
-      noindex?: boolean;
+      noindex?: boolean|string;
```
devnook, line 94: Why would we allow string here?

src/site/content/en/newsletter/index.njk
```diff
@@ -5,6 +5,8 @@ permalink: false
 description: Stay up to date with the latest about web platform.
 date: 2020-03-13
 draft: true
+eleventyComputed:
+  noindex: "{{ paged.index > 0 }}"
```
devnook, line 9: Why not boolean here?

src/site/content/pages.njk
```diff
@@ -2,4 +2,15 @@
...
-{{ collections.pages | dump(2) | safe }}
+{% if site.env === 'prod' %}
+  {% set posts = collections.all | livePosts | algoliaIndexable %}
+{% else %}
+  {% set posts = [] %}
+{% endif %}
+
+[
+  {% for post in posts %}
+    {{ post | algoliaItem | stringify | safe }}
```
devnook, line 13: There is in-built filter "dump". Maybe you can use it here instead of custom stringify?

src/site/_filters/algolia-item.js
```diff
@@ -0,0 +1,46 @@
...
+const striptags = require('striptags');
+
+const md5 = require('./md5');
+const {getDefaultUrl} = require('./urls');
+const {defaultLocale} = require('../_data/site');
+const {generateImgixSrc} = require('../_includes/components/Img');
+
+/**
+ * @param {EleventyCollectionItem} post
+ * @return {AlgoliaItem}
```
devnook, line 25: Sweet!

devnook (commented):
Reviewed

src/site/content/en/newsletter/index.njk
MichaelSolati: I can't get the `noindex` value be a `boolean` when computed by the njk template. I can only get it to be processed as a `string`. If I did it in a JS file I could do a `boolean`, but because we're extending a NJK template we can't also do a 11tydata file.

MichaelSolati:
@devnook PTAL

devnook (approved):

PR #6785, opened 2021-11-15
Remove the instruction to enable the bfcache experiment

philipwalton:
This PR updates the DevTools section of the bfcache post to indicate that developers no longer have to enable the experiment to test the feature (it's now on by default).


malchata (approved):
Short review to be sure, but this LGTM.

PR #6786, opened 2021-11-15
Add Learn CSS Backgrounds module

lozandier:
Fixes #5728

Changes proposed in this pull request:

Add lozandier to list of web.dev authors
Add backgrounds course module inside the folder of existing LearnCSS course modules
Add the backgrounds guide in the table of contents manifest file for the LearnCSS course

PR #6794, opened 2021-11-16
Learn CSS backgrounds

rachelandrew:
This was based on #6786 which had a bunch of problems causing it not to build, I've managed to build this locally.


una:
Cleanup in #6802

PR #6806, opened 2021-11-16
/measure: adopt lighthouse renderer

paulirish:
fixes #6753 

also driveby to disable PWA category

Issue #6770, opened 2021-11-12
New web.dev/measure doesn't allow exporting reports (HTML and JSON)


isAdrisal:
**Describe the bug**
Before the recent update to web.dev/measure, it was possible to save Lighthouse reports as HTML or JSON. This feature has been lost in the new design.

**To Reproduce**
1. Go to web.dev/measure.
2. Run an audit of a valid URL.
3. There is no mechanism to save or share the test result.

**Expected behavior**
It should be possible to save Lighthouse reports — not being able to save or otherwise share a Lighthouse report makes the tool far less useful for sharing Lighthouse results in a team or client.

**Screenshots**

devnook:
We include the "View report" link, which allows to export. I'm marking it as fixed.

PR #6807, opened 2021-11-16
/measure: include topbar for sharing/exporting

paulirish:
(this is on top of PR #6806) 

i saw  #6770  where folks wanted a sharing mechanism. I'm not 100% sure the viewer/save flow answers their requests.. but.. perhaps it does?

image


devnook:
Thanks Paul! I don't want to include the topbar in the /measure though, as it does not fit in the overall page design. I would like to get it implemented within current style and design of the page.

PR #6814, opened 2021-11-17
Update UrlChooser component to prefer https as default

### LINKED ISSUE #6813: Update UrlChooser to prepend https as default when missing
Hey folks, when I ran `johnapostol.com` through the `/measure` tool, I noticed that it prepended `http` instead of `https`. Honestly, I didn't notice until I was running through the diagnostics and saw the redirect.

image

I'm inclined to think `https` should be the default in 2021, especially since Chrome defaults to `https` in the address bar nowadays. This issue comes with the assumption that the vast majority of users want/expect `https` instead of `http`. Even if users expected `http`, might this be a tiny signal for them to consider a change?

Also of note, if I _had_ noticed the prepend earlier, I would have had to refresh the page in order to correct it.

Prepend `https` instead of `http` if it's not clearly added to the UrlChooser.

Teach security-conscious users to prepend `https` manually

**Additional context**
_As an aside, I ran my employer's page through the tool and it highlighted the multiple redirects we happen to have to get people to our homepage if they follow a bookmark or link with `http` instead of `https`. This is something difficult to notice that we've quietly accepted, I guess. At least I'm more aware of it again._

johnapost:
Fixes #6813 
See that issue for full context.

- `http` -> `https`, hopefully not too earth-shattering 🙇

### DIFF
--- a/src/lib/components/UrlChooser/index.js
+++ b/src/lib/components/UrlChooser/index.js
@@ -178,7 +178,7 @@ class UrlChooser extends BaseElement {
       return;
     }
     if (!url.startsWith('https://') && !url.startsWith('http://')) {
-      url = `http://${url}`;
+      url = `https://${url}`;
     }
     if (url !== this._urlInput.value && this._urlInput) {
       this._urlInput.value = url;


MichaelSolati:
Preferring HTTPs over HTTP? Oof, are you trying to flame war here? I'm kidding of course, I think this is perfectly reasonable. Let me just run it by the team first.

MichaelSolati (approved):
And we're shipping it! Thank you for the contribution.

PR #6828, opened 2021-11-18
Change the word "dictionary" to "object"

### LINKED ISSUE #6827: content: Change the word "dictionary" to "object"

https://web.dev/isinputpending/

Web developers are more familiar with the term "object" rather than "dictionary", that's why I think we should use the word "object" here instead. MDN and other docs for web developres use the term "object" rather than "dictionary" too.

### What's the deadline?

There's no deadline.

#### Is it a hard deadline?

No.

Twaha-Rahman:
Fixes #6827 

- Change the word "dictionary" with the word "object" in the article.


jeffposnick:
Seems like a reasonable change. Assigning to some folks from the tech writing side to confirm whether that matches our overall style.

jpmedley (approved):

PR #6851, opened 2021-11-19
Remove value judgement from the recommendation

oliviertassinari:
For context, I'm maintaining [a guide](https://www.notion.so/mui-org/Technical-documentation-7e55b517ac2e489a9ddb6d0f6dd765de) with all the learnings we had at MUI to write great documentation. I have linked this section, but I don't feel great about people that would make this mistake and be categories as "less experienced".


mihajlija:
@oliviertassinari Thanks for pointing this out, I agree completely! You may find this useful for writing tips as well https://developers.google.com/style/accessibility?hl=en

mihajlija (approved):

PR #6861, opened 2021-11-22
Remove JS Self Profiling from translations in cross-origin isolation guide article

agektmr:
This patch removes "JS Self-Profiling API" from outdated translations in cross-origin isolation guide article


agektmr:
@mihajlija I know you don't know all languages, but can you approve anyway?

mihajlija (approved):

PR #6923, opened 2021-11-30
Update bug info and code example for the responsiveness post

philipwalton:
This PR updates the responsiveness post to indicate that some of the issues previously affecting the metric candidates have been fixed in Chrome 98.

It also address one small issue with the code example that I noticed while testing the metrics on Chrome 98.


malchata (approved):
My understanding of this is comprehensive, but it scans and looks good to me.

PR #6929, opened 2021-12-01
Deprecate web.dev/live and redirect to web.dev/shows

devnook:


MichaelSolati (approved):
Beautiful, I love it. LGTM

PR #6934, opened 2021-12-01
Convert posts that used w-stats to new layout

devnook:


src/site/content/en/blog/how-mercadolibre-optimized-web-vitals/index.md
```diff
@@ -66,9 +66,9 @@ Experience Report](https://developers.google.com/web/tools/chrome-user-experienc
 product detail pages had a poor FID. Based on that information, they decided to focus their efforts
 on improving the interactivity for product pages in the site.
 
-<figure class="w-figure">
+<figure>
   {% Img src="image/tcFciHGuF3MxnTr1y5ue01OGLBn2/gg8ohXTbFgr6Msacklt0.png", alt="Mobile and Desktop versions of a Mercado Libre product detail page.", width="800", height="346", class="w-screenshot" %}
```
MichaelSolati, line 70: Do we need/still-use the `.w-screenshot` class on the pages with the `post` layout.

MichaelSolati (approved):
One small comment/q about the use of `.w-screenshot`.

src/site/content/en/blog/how-mercadolibre-optimized-web-vitals/index.md
devnook: We probably don't but I'd rather change it when we rewrite the shortcode than here, so that we can make sure it has no side effects.

PR #6952, opened 2021-12-02
test(algolia): add tests for algolia, fixes #2611

### LINKED ISSUE #2611: Need tests for index-algolia script.
The `index-algolia` script runs during deploys. Currently we don't have any tests for it and typically the only way that we know it's failed is when our CI breaks.

I think we should make it so index-algolia can be run in a pre-flight mode to verify it outputs what we expect. Then we should write tests to assert that it is working as intended.

MichaelSolati:
Fixes #2611

- Add algolia integration test
- In above test check each generated item from `src/site/content/pages.njk` and `algolia.js`
- Export `maxItemSizeInBytes` and `index` function from `algolia.js` to use `maxItemSizeInBytes` in algolia test
- Instead of `index` function being an "IIFE" we execute in in the `npm run algolia` command, that way we can import  `maxItemSizeInBytes` in the algolia integration test without running the script


algolia.js
```diff
@@ -105,49 +99,58 @@ async function index() {
...
-  const postsCount = algoliaData.length;
-
-  // @ts-ignore
-  const client = algoliasearch(
-    process.env.ALGOLIA_APP_ID,
-    process.env.ALGOLIA_API_KEY,
-  );
-  const index = client.initIndex('prod_web_dev');
+  // This exports the parsed data for a test
+  if (process.env.ELEVENTY_ENV === 'test') {
```
devnook, line 106: This means that this script does entirely different thing on test and on real use. This renders testing completely obsolete.

devnook (changes requested):
Please rewrite the algolia.js so that it does not include special logic for TEST environment.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "web.dev",
   "version": "3.0.0",
   "scripts": {
-    "algolia": "node algolia.js",
+    "algolia": "node -e \"require('./algolia.js').index()\"",
```
devnook, line 5: Is this because you changed algolia to a module? 
Then create another script, called e.g runAlgolia, and keep index logic there rather than inline in package.json.

test/unit/algolia.js
```diff
@@ -0,0 +1,76 @@
...
+      expect(chunkedDummyItems.length).to.be.above(1);
+    });
+
+    it('`chunkAlgolia` returns one array if less than `maxItemSizeInBytes`', function () {
+      const dummyItems = [createAlgoliaItem(maxChunkSizeInBytes - 2)];
+      const chunkedDummyItems = chunkAlgolia(dummyItems);
+      expect(chunkedDummyItems.length).to.be.equal(1);
+    });
+
+    it('`chunkAlgolia` throws error if an element is equal to `maxItemSizeInBytes`', function () {
```
devnook, line 71: Add test to when it is bigger than maxItemSizeInBytes

test/unit/algolia.js
```diff
@@ -0,0 +1,76 @@
...
+    throw new Error('Please only input even numbers into this function');
+  }
+  length = length / 2;
+  const content = Array.from({length}, () => 'a').join('');
+  return /** @type {AlgoliaItem} */ ({content});
+};
+
+describe('algolia', function () {
+  describe('trimBytes', function () {
+    it('`trimBytes` trims content if bytes twice `maxItemSizeInBytes`', function () {
```
devnook, line 27: Remove 'trimBytes' for better readability (it is stated in line above)

test/unit/algolia/index.js
```diff
@@ -0,0 +1,81 @@
...
+
+const {
+  chunkAlgolia,
+  maxChunkSizeInBytes,
+  maxItemSizeInBytes,
+  trimBytes,
+} = require('../../../algolia');
+
+/**
+ * Very simple to mock dummy object, only accepts even byte sizes.
```
philipwalton, line 12: Replace "dummy" with "temp" or "placeholder" to avoid the ableist usage. See: https://developers.google.com/style/word-list#dummy-variable

test/unit/algolia/index.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+  length = length / 2;
+  const content = Array.from({length}, () => 'a').join('');
+  return /** @type {AlgoliaItem} */ ({content});
+};
+
+describe('algolia', function () {
+  describe('trimBytes', function () {
+    it('trims content if bytes twice `maxItemSizeInBytes`', function () {
+      const dummyItemSize = maxItemSizeInBytes * 2;
```
philipwalton, line 28: Same comment as above w.r.t "dummy" (here and throughout this file).

test/unit/src/site/_filters/algolia-item.js
```diff
@@ -0,0 +1,140 @@
...
+      title: 'Hello world',
+    },
+    templateContent: 'Hello world',
+    url: '/hello-world/',
+  });
+
+describe('algoliaItem', function () {
+  it('returns object with all required fields', function () {
+    const post = createPost();
+    const dummyItem = algoliaItem(post);
```
philipwalton, line 22: Same comment as above w.r.t "dummy" (here and throughout this file).

algolia.js
```diff
@@ -55,15 +57,21 @@ const chunkAlgolia = (arr) => {
...
-    if (tempSizeInBytes + currentSizeInBytes < maxChunkSizeInBytes) {
-      temp.push(current);
+    const currentSizeInBytes = byteof(arrItem);
+
+    if (currentSizeInBytes >= maxChunkSizeInBytes) {
+      throw new Error(
+        `${path.join(
+          arrItem.locale || '',
+          arrItem.url || '',
+        )} is >= than 10 MB`,
```
philipwalton, line 67: Nit: it might be worth creating a `maxChunkSizeInMiB` (or something like that) and using that to generate the error message so it stays in sync with the logic.

philipwalton (approved):
I'm not familiar with this code, but after an initial scan, the logic seems fine. I made a few small suggestions about the variable names, but other than that, LGTM.

MichaelSolati:
Thank you for the review @philipwalton, I'm using the term `mock` instead, which seems to be fine according to the word list (which was hella helpful)

PR #6978, opened 2021-12-06
Fix line breaks in I/O article.

jpmedley:
To understand the problem, look at this image.

image


rachelandrew (approved):

tomayac (approved):
LGTM

Thanks for the fix. Weird, why would this happen? We have been using this syntax in other articles and never had hard line breaks.

PR #6980, opened 2021-12-06
fix(codelabs): fix glitch portion of codelabs, fixes #6979, fixes #4790

### LINKED ISSUE #4790: Percy tests failing on the codelab page
It looks like the `PERCY=true` flag isn't hiding the glitch iframe on the codelab page so it's breaking.

### LINKED ISSUE #6979: Codelabs too small
**Describe the bug**
Embedded glitch codelabs are appearing too small. See => https://web.dev/codelab-make-installable/ (reported by @jpmedley )

**To Reproduce**
1. Go to any codelab page https://web.dev/codelab-make-installable/

**Expected behavior**
The glitch should take up about half the page.

**Screenshots**
image

MichaelSolati:
Fixes #6979, fixes #4790

- Re-add styling to get codelabs to take up right size.
- Add percy mode to codelab so we can run percy test, that way we can catch if the page breaks
-


src/site/_includes/codelab.njk
```diff
@@ -12,11 +12,11 @@ pageScripts:
     {# Use first found path as a default path. #}
     {% set path = paths[postPaths[0]] %}
   {% endif %}
-  <web-codelab class="web-codelab" glitch="{{ glitch }}" {% if glitch_path %}path="{{ glitch_path }}"{% endif %} {% if site.env === 'test' %}testmode{% endif %}>
+  <web-codelab class="web-codelab" glitch="{{ glitch }}" {% if glitch_path %}path="{{ glitch_path }}"{% endif %} {% if site.percy === 'true' %}percymode{% endif %}>
```
devnook, line 15: Instead of adding whole logic of percymode to the component, you can simply not render it here if site.percy === 'true'. It is not the omponent role to understand the test environment limitations.

src/lib/components/Codelab/index.js
```diff
@@ -95,7 +106,7 @@ class Codelab extends BaseElement {
         ? `This Glitch isn't loaded in a test environment`
         : `This Glitch isn't available on small screens`;
       return html`
-        <div class="w-sizer ${isTest ? 'w-test' : ''}">
+        <div class="${isTest ? 'w-test' : ''}">
```
devnook, line 109: w-test class does not exist anymore

src/lib/components/Codelab/index.js
```diff
@@ -85,6 +88,14 @@ class Codelab extends BaseElement {
       return html``;
     }
     const isTest = env === 'test';
```
devnook, line 90: In what situation this is true? Do we have a test for this component? Is there another scenatio where env === test that I just don't recall?

devnook (changes requested):
Reviewed

src/lib/components/Codelab/index.js
```diff
@@ -84,21 +86,14 @@ class Codelab extends BaseElement {
...
-    // If this is a test, always show the warning. Percy snapshots our DOM at a
-    // low resolution before resizing it, so we can't rely on _isDesktop being
-    // different for smaller or larger tests. The `w-test` ensures we test the
-    // sticky behavior of this element.
-    if (!this._isDesktop || isTest) {
-      const message = isTest
-        ? `This Glitch isn't loaded in a test environment`
-        : `This Glitch isn't available on small screens`;
+
+    if (!this._isDesktop) {
```
devnook, line 90: Is the warning written in lines 89-92 not relevant anymore?

devnook (approved):

src/lib/components/Codelab/index.js
MichaelSolati: I think so, it's to hide the codelab when on mobile (as the viewport is too small)

PR #6984, opened 2021-12-07
[i18n] Localize string in `Instruction.js`

### LINKED ISSUE #6982: Localize `Instruction.js`
In blog post, we use `{% Instruction 'devtools-network', 'ol' %}` to print common instruction on screen. However, the text are not translated. Currently blocking PR #6946.  

The build handle i18n for the text.

Manually translate the word without using the `Instruction.js` component, but not favorable.

chybie:
Fixes #6982 

Add an optional `locale` param in the `Instruction` shortcode. Can use that like below in any article, especially translated one.

```js
{% Instruction 'devtools-network', 'li', locale %}
```

Preferably able to access to `@this {EleventyPage}` object like the `components/Aside.js`, then we can eliminate this additional `locale` param. However, I couldn't get that work.

### DIFF
--- a/src/site/_includes/components/Instruction.js
+++ b/src/site/_includes/components/Instruction.js
@@ -15,6 +15,8 @@
  */
 
 const {html} = require('common-tags');
+const {i18n} = require('../../_filters/i18n');
+const {defaultLocale} = require('../../../../shared/locale');
 const capitalize = require('../../_filters/capitalize');
 
 /**
@@ -25,7 +27,7 @@ const capitalize = require('../../_filters/capitalize');
  * @param {string} listStyle The list style to use. Defaults to 'ul'.
  * @return {string} A list of instructions.
  */
-module.exports = (type, listStyle = 'ul') => {
+module.exports = (type, listStyle = 'ul', locale = defaultLocale) => {
   let instruction;
   let substitution;
   let bullet;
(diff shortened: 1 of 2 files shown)


src/site/_includes/components/Instruction.js
```diff
@@ -25,7 +27,7 @@ const capitalize = require('../../_filters/capitalize');
  * @param {string} listStyle The list style to use. Defaults to 'ul'.
  * @return {string} A list of instructions.
  */
-module.exports = (type, listStyle = 'ul') => {
+module.exports = (type, listStyle = 'ul', locale = defaultLocale) => {
```
devnook, line 30: Let's avoid changing the way this function is called (eg by adding new parameters). You can use getLocaleFromPath to get locale instead, as shown here: https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/components/Aside.js#L35

Please note that for this to work you need to change the notation of the component definition from arrow  function for "function() {...}"

devnook:
Fantastic, thanks a lot @chybie ! I left one comment everything else looks great.

src/site/_includes/components/Instruction.js
chybie: Oh!!! That's the trick. Thanks. Will update it now.

chybie:
Updated. PTAL.

chybie:
@devnook please help to approve the Percy web changes as well

PR #7001, opened 2021-12-09
fix some errors in Chinese translation

wingtao:
fix some errors about markdown format in Chinese translation

### DIFF
--- a/src/site/content/zh/metrics/user-centric-performance-metrics/index.md
+++ b/src/site/content/zh/metrics/user-centric-performance-metrics/index.md
@@ -55,8 +55,8 @@ tags:
 
 性能指标一般通过以下两种方式来进行测量：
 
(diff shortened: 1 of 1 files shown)


chybie:
A Chinese reader here. Why is the `: ` need to be part of the bold in Chinese?

wingtao:
> A Chinese reader here. Why is the `: ` need to be part of the bold in Chinese?

I just want change the bold format to display correctly, because it's seem not right now, like this
image

as for as Why is the `:` need to be part of the bold in Chinese, I think it wanted to be as same as the English version.

rachelandrew:
@wingtao this will need the CLA signed for us to consider merging it.

wingtao:
> @wingtao this will need the CLA signed for us to consider merging it.

ok, I have signed the CLA now, maybe need a rescan

PR #7050, opened 2021-12-15
Remove newlines in subhead from launch handler post

jeffposnick:
There are [a few ways](https://yaml-multiline.info/) to do multi-line strings in YAML; the use of `|` as a prefix will add in newlines to match the newlines in the YAML, while using `>` will translate the newlines into spaces. In the case of this post, the spaces approach is more appropriate:

Before:
Screen Shot 2021-12-15 at 11 47 38 AM

After:
Screen Shot 2021-12-15 at 11 57 20 AM

### DIFF
--- a/src/site/content/en/blog/launch-handler/index.md
+++ b/src/site/content/en/blog/launch-handler/index.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: Control how your app is launched
-subhead: |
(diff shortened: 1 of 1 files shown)


tomayac (approved):
LGTM, thank you.

rachelandrew (approved):

Issue #6571, opened 2021-10-28
Learn course's Test panel shows 'incorrect' even when answer is correct


PaulKinlan:
**Describe the bug**
The tests on the learn page show all the inccorrect answers even when you've selected the correct one.

**To Reproduce**
Go to the learn course for forms, click on the first topic and then take the test... It says Correct and shows "Incorrect" answers too.

**Expected behavior**
I'd expect it to just say "Correct" or "Incorrect" on the option that was chosen so that I can pick again if needed.

**Screenshots**

![Screenshot 2021-10-28 21 49 41](https://user-images.githubusercontent.com/45510/139333700-19e1aa9d-b38e-474e-8ed5-32c361cf516c.png)

rachelandrew:
@devnook I'm not sure who is best to look at this but they are behaving pretty strangely having just gone through the forms ones to check them.

PR #7051, opened 2021-12-15
Self assesment styles

### LINKED ISSUE #6850: Rebuild the CSS for Assessment modules
It needs simplifying and moving into `src/scss/web-components`.

Current assessment module

hankchizljaw:
Fixes #6850


MichaelSolati (approved):
I think this will require more tweaks, not necessarily now, but in general. Like there are some interesting implementations of the assessment component :

Heres before a submission
image

And then when you submit...

image

I'm thinking this should be hidden until the user clicks it... Basically set `src/site/content/en/accessible/control-focus-with-tabindex/index.md`'s layout to `post` to see this.

Also it breaks the styling of assessments on pages on the old design system

https://deploy-preview-7051--web-dev-staging.netlify.app/control-focus-with-tabindex/

hankchizljaw:
Ah good catch on that element. It's yet another part of the self assessment that isn't easy to find 🙃

This is also again, blocked by old articles not using the new system

hankchizljaw:
Upon looking at that quiz on live, it seems sort of useless to me, but I'll ship a quick update to the display logic styles

devnook (approved):

devnook:
Fixes #6571

PR #7052, opened 2021-12-15
refactor(codelabs): update to use `default-next` as base

MichaelSolati:

- Remove or update all `w-*` CSS classes
- Update codelab web component to new breakpoint and new warning CSS
- Update and simplify `web-codelab` SCSS
- Set codelab.njk to `layout: default-next` and remove unnecessary wrapper and `legacy-rollout` classes.
- Delete old `src/styles/pages/_codelab.scss`

I've checked the pages and they look roughly the same with these changes, nothing is breaking, and anything that didn't look right I fixed.


src/lib/components/Codelab/index.js
```diff
@@ -33,7 +33,7 @@ class Codelab extends BaseElement {
     // _isDesktop has no default value as it's only correctly set between connected/disconnected
     // callbacks via the MediaQueryList's listener.
 
-    this._mql = window.matchMedia('(min-width: 865px)');
+    this._mql = window.matchMedia('(min-width: 800px)');
```
devnook, line 36: I think this new breakpoint is too small. at 830px codelab looks like this:
Screenshot 2021-12-20 at 15 43 32

devnook (commented):
Generally looks good, it might be worthwhile to adjust the breakpoint though.

MichaelSolati:
@devnook breakpoints should be good. If you could PTAL at.

src/lib/components/Codelab/index.js
MichaelSolati: I kept the 800px bp, but changed the size of the columns so it should look better at the smaller viewport

image

philipwalton: If you don't have the sidebar closed (which I didn't by default), it looks like this, which is not ideal because the column collapse button is covered by the "Remix to Edit" banner:

Screen Shot 2022-01-07 at 1 07 09 PM

src/site/_includes/codelab.njk
```diff
@@ -1,98 +1,91 @@
 ---
-layout: default
+layout: default-next
 permalink: "/{{lang}}/{{page.fileSlug}}/index.html"
 show_banner: false
 pageScripts:
   - '/js/content.js'
 ---
+{% if path === undefined  %}
+  {% set postPaths = page | getPaths %}
```
philipwalton, line 9: I know this was in the previous code, but I just want to confirm that the single bar is correct here. My understanding is nunjucks uses single bars as filters, but this looks like an "or" expression to me.

src/site/_includes/codelab.njk
```diff
@@ -1,98 +1,91 @@
...
-          </div>
-        {% endif %}
-      </article>
-    </div>
-  </web-codelab>
-</div>
+      <div class="text-size-0 color-mid-text">
+        <span>
+          {% if updated %}
+          {{ 'i18n.post.last_updated' | i18n(locale) }}: <time>{{ updated | prettyDate }}</time>
```
philipwalton, line 60: Nit: I think this line (and 62) should be indented.

philipwalton (approved):
LGTM (modulo a few comments to check)

src/site/_includes/codelab.njk
MichaelSolati: It should be applied as a filter. I haven't tested this use specifically but I have used it before in a similar manner.

src/site/_includes/codelab.njk
MichaelSolati: Totally agree, I'll fix this now.

PR #7106, opened 2021-12-25
Fix meta viewport tag content

### LINKED ISSUE #7105: Meta viewport tag has incorrect "value" attribute instead of "content"
**Describe the bug**
There are two code samples that include a meta viewport tag without a "content" value which means they do not work

**To Reproduce**
1. Go to https://web.dev/learn/design/intro/#a-meta-element-for-viewport
2. Copy the code sample `<meta name="viewport" value="width=device-width, initial-scale=1">` and use in a web page
3. The web page is not responsive
---
1. Go to https://web.dev/learn/design/media-queries/
2. Scroll to the "Check your understanding" section
3. The correct answer for one question has a code sample that doesn't work: `<meta name="viewport" value="width=device-width, initial-scale=1">`

**Expected behavior**
The sample should use the "content" attribute:

```
<meta name="viewport" content="width=device-width, initial-scale=1">
```

**Screenshots**

ricmatsui:
Fixes #7105

- Fix meta viewport code samples to use the correct attribute: "content"
- https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#viewport_basics
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attributes

### DIFF
--- a/src/site/content/en/learn/design/intro/index.md
+++ b/src/site/content/en/learn/design/intro/index.md
@@ -214,7 +214,7 @@ If you're using responsive design you need to tell the browser not to do that sc
 You can do that with a `meta` element in the `head` of the web page:
 
 ```html
-<meta name="viewport" value="width=device-width, initial-scale=1">
+<meta name="viewport" content="width=device-width, initial-scale=1">
 ```
 
 There are two values, separated by commas. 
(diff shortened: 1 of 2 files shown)


malchata (approved):
LGTM!

PR #7115, opened 2021-12-31
Fix environment variable usage on screen-configurations page

zaguiini:

The https://web.dev/learn/design/screen-configurations/ page uses the environment variable in a wrong manner. [According to the syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/env()#syntax), it requires a variable and a fallback value. However, we were seeing `env(value fallback fallback)` which makes the browser unhappy about it.

### DIFF
--- a/src/site/content/en/learn/design/screen-configurations/index.md
+++ b/src/site/content/en/learn/design/screen-configurations/index.md
@@ -132,7 +132,7 @@ For dual screens with a vertical hinge, set the first column to be the width of 
 ```css
 @media (horizontal-viewport-segments: 2) and (vertical-viewport-segments: 1) {
   main article {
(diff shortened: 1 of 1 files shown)


zaguiini:
CLA accepted.

rachelandrew:
@zaguiini I believe the text is correct as per spec when talking about the viewport segments. See: https://drafts.csswg.org/css-env/#viewport-segments 

cc: @adactio

zaguiini:
That's absolutely right, @rachelandrew! I totally missed it 🙈 

Thanks for the heads up. How about we introduce a little note in the article saying that the syntax is actually correct and a bit different from other `var` or `env` usages?

rachelandrew:
Closing as the change wasn't right as explained in comments. Feel free to open a PR with a suggested note.

PR #7122, opened 2022-01-04
Adds internal link lint rule

malchata:

- Adds a linter rule for web.dev links to ensure they remain relative instead of absolute (e.g., `/about/` instead of `https://web.dev/about/`), as this is a common fix we make in PR reviews.
- Fixes all links which cause this new linter rule to throw.


rachelandrew (approved):
LGTM, and is an excellent addition, thank you!

PR #7129, opened 2022-01-06
fix: remove dead link

coliff:

Lighthouse no longer checks for links with `target="_blank"` having the `rel="noopener"` attribute since `rel="noopener"` is implied already in all current evergreen browsers.
(maybe the content on this page needs further updates to note that?)


malchata (approved):
LGTM.

PR #7151, opened 2022-01-11
Porting USB applications to the web. Part 1: libusb

RReverser:
Part 1 of #6905.


RReverser:
Hi @jpmedley as per request, creating a PR for copy-editing here. I still need to add a hero image, but otherwise should be good for review.

jpmedley:
@RReverser Before I start this, are all the other reviews done? Is your team and other interested parties finished commenting on this?

RReverser:
Yes, go ahead.

FWIW "other interested parties" is pretty wide because lots of people expressed interest in this topic, but I'm not going to include all of them as reviewers too :)

jpmedley:
It's broad because I don't know everyone you want for review. Sometimes I get, "I didn't think you meant so-and-so." The concern is not about _who_. It's about _when_. It's not an effective use of my time to edit a paragraph if someone's going to come along and rewrite it.

RReverser:
That's okay. As I said, it's ready for review.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+};
+```
+
+Looking through the properties, we can see that the struct includes the backend name, a set of its capabilities, handlers for various low-level USB operations in form of function pointers, and, finally, sizes to allocate for storing private device-/context-/transfer-level data.
+
+The private data fields are useful at least for storing OS handles to all those things, as without handles we don't know which item any given operation applies to. In the web implementation, the OS handles would be the underlying WebUSB JavaScript objects. The natural way to represent and store them in Emscripten would be via the [`emscripten::val`](https://emscripten.org/docs/api_reference/val.h.html#_CPPv4N10emscripten10emscripten3valE) class, which is provided as part of [Embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html) (Emscripten's bindings system).
+
+Most of the backends in the folder are implemented in C, but a few are implemented in C++ as well. Embind only works with C++, so the choice was made for me and I've added `libusb/libusb/os/emscripten_webusb.cpp` with the required structure and with `sizeof(val)` for the private data fields:
```
jpmedley, line 123:
```suggestion
Most of the backends in the folder are implemented in C, but a few are implemented in C++. Embind only works with C++, so the choice was made for me and I've added `libusb/libusb/os/emscripten_webusb.cpp` with the required structure and with `sizeof(val)` for the private data fields:
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+    .caps = LIBUSB_CAP_HAS_CAPABILITY,
+    // …handlers - function pointers to implementations above
+    .device_priv_size = sizeof(val),
+    .transfer_priv_size = sizeof(val),
+};
+```
+
+### Storing WebUSB objects as device handles
+
+libusb provides ready-to-use pointers to the allocated area for private data. To work with those pointers as `val` instances, I've added small helpers that can construct them in-place, retrieve them as references, and move values out:
```
jpmedley, line 146:
```suggestion
libusb provides ready-to-use pointers to the allocated area for private data. To work with those pointers as `val` instances, I've added small helpers that construct them in-place, retrieve them as references, and move values out:
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+  WebUsbTransferPtr(usbi_transfer *itransfer)
+      : ValPtr(static_cast<val *>(usbi_get_transfer_priv(itransfer))) {}
+};
+```
+
+### Async web APIs in synchronous C contexts
+
+Now I've needed a way to handle async WebUSB APIs where libusb expects synchronous operations. For this, I could use Asyncify, or, more specifically, its Embind integration via [`val::await()`](https://emscripten.org/docs/api_reference/val.h.html#_CPPv4NK10emscripten10emscripten3val5awaitEv).
+
+I also wanted to correctly handle WebUSB errors and convert them into libusb error codes, but Embind currently doesn't have any way to handle JavaScript exceptions or `Promise` rejections from the C++ side. This problem can be worked around by catching a rejection on the JavaScript side and converting the result into an `{ error, value }` object that can be now safely parsed from the C++ side. This was accomplished by combination of [`EM_JS`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.EM_JS) macro and [`Emval.to{Handle, Value}`](https://emscripten.org/docs/api_reference/val.h.html#_CPPv4NK10emscripten10emscripten3val9as_handleEv) APIs:
```
jpmedley, line 186:
```suggestion
I also wanted to correctly handle WebUSB errors and convert them into libusb error codes, but Embind currently doesn't have any way to handle JavaScript exceptions or `Promise` rejections from the C++ side. This problem can be worked around by catching a rejection on the JavaScript side and converting the result into an `{ error, value }` object that can be now safely parsed from the C++ side. I did this with a combination of the [`EM_JS`](https://emscripten.org/docs/api_reference/emscripten.h.html#c.EM_JS) macro and [`Emval.to{Handle, Value}`](https://emscripten.org/docs/api_reference/val.h.html#_CPPv4NK10emscripten10emscripten3val9as_handleEv) APIs:
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+  static promise_result await(val &&promise) {
+    promise = em_promise_catch(std::move(promise));
+    return {promise.await()};
+  }
+};
+```
+
+Now I could use `promise_result::await` on any `Promise` returned from WebUSB operations and inspect its `error` and `value` fields separately.
+
+For example, retrieving a `val` representing a [`USBDevice`](https://developer.mozilla.org/docs/Web/API/USBDevice) from `libusb_device_handle`, calling its [`.open()`](https://developer.mozilla.org/docs/Web/API/USBDevice/open) method, awaiting its result and returning an error code as a libusb status code looks like this:
```
jpmedley, line 252:
```suggestion
For example, retrieving a `val` representing a [`USBDevice`](https://developer.mozilla.org/docs/Web/API/USBDevice) from `libusb_device_handle`, calling its [`open()`](https://developer.mozilla.org/docs/Web/API/USBDevice/open) method, awaiting its result, and returning an error code as a libusb status code looks like this:
```

jpmedley: We don't include the period for standalone method references. We do use a comma before the 'and' for items in a list (Oxford comma).

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+```cpp
+int em_open(libusb_device_handle *handle) {
+  auto web_usb_device = get_web_usb_device(handle->dev);
+  return promise_result::await(web_usb_device.call<val>("open")).error;
+}
+```
+
+### Device enumeration
+
+Of course, before any device can be opened, libusb needs to retrieve a list of available devices. The backend must implement this operation via a `get_device_list` handler.
```
jpmedley, line 263:
```suggestion
Of course, before I can open any device, libusb needs to retrieve a list of available devices. The backend must implement this operation via a `get_device_list` handler.
```

jpmedley: A certain amount of passive voice is OK, but general, active voice is better.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+The difficulty is that, unlike on other platforms, there is no way to enumerate all the connected USB devices on the web for security reasons. Instead, the flow is split into two parts. First, the web application can request devices with specific properties via [`navigator.usb.requestDevice()`](https://developer.mozilla.org/docs/Web/API/USB/requestDevice) and the user will manually choose which device they want to expose or they can reject the permission prompt. Then, you can list the already approved and connected devices via [`navigator.usb.getDevices()`](https://developer.mozilla.org/docs/Web/API/USB/getDevices).
```
jpmedley, line 265:
```suggestion
The difficulty is that, unlike on other platforms, there is no way to enumerate all the connected USB devices on the web for security reasons. Instead, the flow is split into two parts. First, the web application requests devices with specific properties via [`navigator.usb.requestDevice()`](https://developer.mozilla.org/docs/Web/API/USB/requestDevice) and the user will manually choose which device they want to expose or they can reject the permission prompt. Then, you can list the already approved and connected devices via [`navigator.usb.getDevices()`](https://developer.mozilla.org/docs/Web/API/USB/getDevices).
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+  for (uint8_t i = 0; i < devices_num; i++) {
+    auto web_usb_device = web_usb_devices[i];
+    // …
+    *devs = discovered_devs_append(*devs, dev);
+  }
+  return LIBUSB_SUCCESS;
+}
+```
+
+Most of the backend code uses `val` and `promise_result` in a similar way as already shown above. There were few more interesting hacks I've had to do to convert properties exposed by WebUSB to those expected by libusb, both in this method and in others handling data transfer, but those implementation details are less important for the purposes of this article. Make sure to check the code and the comments if you're interested.
```
jpmedley, line 297:
```suggestion
Most of the backend code uses `val` and `promise_result` in a similar way to what's shown above. There were few more interesting hacks I've had to do to convert properties exposed by WebUSB to those expected by libusb, both in this method and in others handling data transfer, but those implementation details are less important for the purposes of this article. Make sure to check the code and the comments if you're interested.
```

jpmedley: I can't work what the middle sentence is saying. Please try something else. This problem usually comes from trying to say too much in a single sentence. Try breaking it up.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+## Porting event loops to the web
+
+One more piece of `libusb` that I do want to stop on is event handling. As described in the previous article, most APIs in system languages like C are synchronous, and event handling is no exception. It's usually implemented via an infinite loop that "polls" (tries to read data or blocks execution until some data is available) from a set of external I/O sources, and, when at least one of those responds, passes that as an event to the corresponding handler. Once the handler has finished its job, the control returns to the loop, and it pauses for another poll.
```
jpmedley, line 301:
```suggestion
One more piece of `libusb` that I want to discuss is event handling. As described in the previous article, most APIs in system languages like C are synchronous, and event handling is no exception. It's usually implemented via an infinite loop that "polls" (tries to read data or blocks execution until some data is available) from a set of external I/O sources, and, when at least one of those responds, passes that as an event to the corresponding handler. Once the handler is finished, the control returns to the loop, and it pauses for another poll.
```

jpmedley: >most APIs in system languages like C are synchronous

C is a low-level language, but I've never heard it described as a system language. Do you mean, "most system APIs used in C are synchronous"? Something elese?

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+There are a couple of problems with this approach on the web.
+
+First of all, WebUSB doesn't and cannot expose raw handles of the underlying devices, so polling those directly is not an option. Second, libusb uses `eventfd` and `pipe` APIs for other events as well as for handling transfers on operating systems without raw device handles, but `eventfd` is not currently supported in Emscripten, and `pipe`, while supported, [currently doesn't conform to the spec](https://github.com/emscripten-core/emscripten/issues/13214) and can't wait for events.
```
jpmedley, line 305:
```suggestion
First, WebUSB doesn't and cannot expose raw handles of the underlying devices, so polling those directly is not an option. Second, libusb uses `eventfd` and `pipe` APIs for other events as well as for handling transfers on operating systems without raw device handles, but `eventfd` is not currently supported in Emscripten, and `pipe`, while supported, [currently doesn't conform to the spec](https://github.com/emscripten-core/emscripten/issues/13214) and can't wait for events.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+Finally, the biggest problem is that the web has its own event loop. This global event loop is used for any external I/O operations (including `fetch`, timers, or, in this case, WebUSB), and it invokes event or `Promise` handlers whenever corresponding operation has finished. Executing another, nested, infinite event loop will block the browser's event loop from ever progressing, which means that not only will the UI become unresponsive, but also that the code will never get notifications for the very same I/O events it's waiting for. This usually results in a deadlock, and that's what happened when I tried to use libusb in a demo, too - the page froze.
```
jpmedley, line 307:
```suggestion
Finally, the biggest problem is that the web has its own event loop. This global event loop is used for any external I/O operations (including `fetch`, timers, or, in this case, WebUSB), and it invokes event or `Promise` handlers whenever corresponding operations finish. Executing another, nested, infinite event loop will block the browser's event loop from ever progressing, which means that not only will the UI become unresponsive, but also that the code will never get notifications for the very same I/O events it's waiting for. This usually results in a deadlock, and that's what happened when I tried to use libusb in a demo, too. The page froze.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+Like with other blocking I/O, to port such event loops to the web, developers need to find a way to run those loops without blocking the main thread. One way is to refactor the application to handle I/O events in a separate thread and pass the results back to the main one. The other is to use Asyncify to pause the loop and wait for events in a non-blocking fashion.
```
jpmedley, line 309:
```suggestion
Like with other blocking I/O, to port such event loops to the web, developers need to run those loops without blocking the main thread. One way is to refactor the application to handle I/O events in a separate thread and pass the results back to the main one. The other is to use Asyncify to pause the loop and wait for events in a non-blocking fashion.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+I didn't want to do significant changes to either libusb or gPhoto2, and I've already used Asyncify for `Promise` integration earlier, so that's the path I've chosen. In order to simulate a blocking variant of `poll`, for the initial proof of concept I've used a loop below:
```
jpmedley, line 311:
```suggestion
I didn't want to do significant changes to either libusb or gPhoto2, and I've already used Asyncify for `Promise` integration, so that's the path I've chosen. To simulate a blocking variant of `poll`, for the initial proof of concept I've used a loop as shown below:
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+    if (num_ready != 0) break;
+    // Yield to the browser event loop to handle events.
+    emscripten_sleep(0);
+  } while (emscripten_get_now() < until_time);
+#else
+  num_ready = poll(fds, nfds, timeout_ms);
+#endif
+```
+
+What it does is:
```
jpmedley, line 334: Is this actually a sequence? Many use numbered bullets for things that aren't. If this list isn't a sequence it should be bullets instead of numbers.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+1. Invokes `poll` to check if any events were reported by the backend yet. If there are some, the loop stops, otherwise Emscripten's implementation of `poll` will immediately return with `0`.
```
jpmedley, line 336:
```suggestion
1. Calls `poll()` to check if any events were reported by the backend yet. If there are some, the loop stops. Otherwise Emscripten's implementation of `poll` will immediately return with `0`.
```

jpmedley: Calls: simpler language
(): Make sure functions/methods telegraph themselves as such. Please do this throughout.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+2. Invokes `emscripten_sleep(0)`. This function uses Asyncify and `setTimeout` under the hood and is used here to yield control back to the main browser event loop. This allows the browser to handle any user interactions and I/O events, including WebUSB.
```
jpmedley, line 337:
```suggestion
2. Calls `emscripten_sleep(0)`. This function uses Asyncify and `setTimeout()` under the hood and is used here to yield control back to the main browser event loop. This allows the browser to handle any user interactions and I/O events, including WebUSB.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+3. Check if the specified timeout has expired yet, and, if not, continue the loop.
+
+As the comment mentions, this approach was not optimal, because it kept saving-restoring the entire call stack with Asyncify even when there were no USB events to handle yet (most of the time), and because `setTimeout` itself has a minimal duration of 4ms in modern browsers. Still, it worked well enough to produce 13-14 FPS livestream from DSLR in the initial proof-of-concept.
```
jpmedley, line 340:
```suggestion
As the comment mentions, this approach was not optimal, because it kept saving-restoring the entire call stack with Asyncify even when there were no USB events to handle yet (which is most of the time), and because `setTimeout()` itself has a minimal duration of 4ms in modern browsers. Still, it worked well enough to produce 13-14 FPS livestream from DSLR in the proof-of-concept.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+Later, I've decided to improve it by leveraging the browser event system. There are several ways in which this implementation could be improved further, but for now I've chosen to emit custom events directly on the global object, without associating them with a particular libusb data structure. I've done so via the following wait & notify mechanism:
```
jpmedley, line 342:
```suggestion
Later, I decided to improve it by leveraging the browser event system. There are several ways in which this implementation could be improved further, but for now I've chosen to emit custom events directly on the global object, without associating them with a particular libusb data structure. I've done so via the following wait and notify mechanism:
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+```bash
+emscripten)
+  AC_SUBST(EXEEXT, [.html])
+  # Note: LT_LDFLAGS is not enough here because we need link flags for executable.
+  AM_LDFLAGS="${AM_LDFLAGS} --bind -s ASYNCIFY -s ASSERTIONS -s ALLOW_MEMORY_GROWTH -s INVOKE_RUN=0 -s EXPORTED_RUNTIME_METHODS=['callMain']"
+  ;;
+```
+
+First, executables on Unix platforms normally don't have any extension. Emscripten, however, produces different output depending on which extension you request. I'm using `AC_SUBST(EXEEXT, …)` to change the executable extension to `.html` so that any executable within a package - tests and examples - becomes an HTML with Emscripten's default shell that takes care of loading and instantiating JavaScript and WebAssembly.
```
jpmedley, line 413:
```suggestion
First, executables on Unix platforms normally don't have a file extension. Emscripten, however, produces different output depending on which extension you request. I'm using `AC_SUBST(EXEEXT, …)` to change the executable extension to `.html` so that any executable within a package - tests and examples - becomes an HTML with Emscripten's default shell that takes care of loading and instantiating JavaScript and WebAssembly.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+Second, because I'm using Embind and Asyncify, I need to enable those features (`--bind -s ASYNCIFY`) as well as allow dynamic memory growth (`-s ALLOW_MEMORY_GROWTH`) via linker parameters. Unfortunately, there is no way for a library to report those flags to the linker, so every application that uses this libusb port will have to add the same linker flags into their build configuration as well.
+
+Finally, as mentioned earlier, WebUSB requires device enumeration to be done via a user gesture. libusb examples and tests assume that they can enumerate devices at start-up, and fail with an error without changes. Instead, I had to disable automatic execution (`-s INVOKE_RUN=0`) and expose the manual `callMain` method (`-s EXPORTED_RUNTIME_METHODS=...`).
```
jpmedley, line 417: >fail with an error without changes.

What do you mean by this?

jpmedley, line 417:
```suggestion
Finally, as mentioned earlier, WebUSB requires device enumeration to be done via a user gesture. libusb examples and tests assume that they can enumerate devices at start-up, and fail with an error without changes. Instead, I had to disable automatic execution (`-s INVOKE_RUN=0`) and expose the manual `callMain()` method (`-s EXPORTED_RUNTIME_METHODS=...`).
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+Once all of this was done, I could serve the generated files with a static web server, initialize WebUSB and run those HTML executables manually with the help of DevTools.
```
jpmedley, line 419:
```suggestion
Once all of this was done, I could serve the generated files with a static web server, initialize WebUSB, and run those HTML executables manually with the help of DevTools.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/i6dUiRRDtb0ucKX9gHOH.png", alt="Screenshot showing Chrome window with DevTools open on a locally served `testlibusb` page. DevTools console is evaluating `navigator.usb.requestDevice({ filters: [] })`, which triggered a permission prompt and it's currently asking the user to choose a USB device that should be shared with the page. ILCE-6600 (a Sony camera) is currently selected.", width="800", height="626" %}
```
jpmedley, line 421:
```suggestion
{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/i6dUiRRDtb0ucKX9gHOH.png", alt="Screenshot showing a Chrome window with DevTools open on a locally served `testlibusb` page. DevTools console is evaluating `navigator.usb.requestDevice({ filters: [] })`, which triggered a permission prompt and it's currently asking the user to choose a USB device that should be shared with the page. ILCE-6600 (a Sony camera) is currently selected.", width="800", height="626" %}
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/0FhFjozbwA1sqk1Dhx7f.png", alt="Screenshot of the next step, with DevTools still open. After the device was selected, Console has evaluated a new expression `Module.callMain(['-v'])`, which executed the `testlibusb` app in verbose mode. The output shows various detailed information about previously connected USB camera - manufacturer Sony, product ILCE-6600, serial number, configuration etc.", width="800", height="824" %}
```
jpmedley, line 423:
```suggestion
{% Img src="image/9oK23mr86lhFOwKaoYZ4EySNFp02/0FhFjozbwA1sqk1Dhx7f.png", alt="Screenshot of the next step, with DevTools still open. After the device was selected, Console has evaluated a new expression `Module.callMain(['-v'])`, which executed the `testlibusb` app in verbose mode. The output shows various detailed information about the previously connected USB camera: manufacturer Sony, product ILCE-6600, serial number, configuration etc.", width="800", height="824" %}
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+6. Add the following flags to your application's link arguments: `--bind -s ASYNCIFY -s ALLOW_MEMORY_GROWTH`.
+
+The library currently has a few limitations:
+
+1. No transfer cancellation support. This is a limitation of WebUSB, which, in turn, stems from lack of cross-platform transfer cancellation in libusb itself.
+2. No isochronous transfer support. It shouldn't be hard to add it by following the implementation of existing transfer modes as examples, but it's also a somewhat rare mode and I didn't have any devices to test it on, so for now I left it as unsupported. If you do have such devices, and want to contribute to the library, PRs are welcome!
+3. The [earlier mentioned cross-platform limitations](#heading=h.ik9rifeitesl). Those limitations are imposed by operating systems, so not much we can do here, except ask users to override the driver or permissions. However, if you're porting HID or serial devices, you can follow the libusb example and port some other library to another Fugu API. For example, you could port a C library [hidapi](https://github.com/libusb/hidapi) to [WebHID](/hid/) and side-step those issues, associated with low-level USB access, altogether.
```
jpmedley, line 442:
```suggestion
* No transfer cancellation support. This is a limitation of WebUSB, which, in turn, stems from lack of cross-platform transfer cancellation in libusb itself.
* No isochronous transfer support. It shouldn't be hard to add it by following the implementation of existing transfer modes as examples, but it's also a somewhat rare mode and I didn't have any devices to test it on, so for now I left it as unsupported. If you do have such devices, and want to contribute to the library, PRs are welcome!
* The [earlier mentioned cross-platform limitations](#heading=h.ik9rifeitesl). Those limitations are imposed by operating systems, so not much we can do here, except ask users to override the driver or permissions. However, if you're porting HID or serial devices, you can follow the libusb example and port some other library to another Fugu API. For example, you could port a C library [hidapi](https://github.com/libusb/hidapi) to [WebHID](/hid/) and side-step those issues, associated with low-level USB access, altogether.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+## Conclusion
+
+In this post I've shown how, with the help of Emscripten, Asyncify and Fugu APIs even low-level libraries like libusb can be ported to the web with a few integration tricks.
+
+Porting such essential and widely used low-level libraries is particularly rewarding, because, in turn, it allows to bring higher-level libraries or even whole applications to the web, too. This opens experiences that were previously limited to users of one or two platforms, to all kinds of devices and operating systems, making those experiences available just a link click away.
```
jpmedley, line 448:
```suggestion
Porting such essential and widely used low-level libraries is particularly rewarding, because, in turn, it allows to bringing higher-level libraries or even whole applications to the web, too. This opens experiences that were previously limited to users of one or two platforms, to all kinds of devices and operating systems, making those experiences just a link click away.
```

jpmedley (changes requested):
Thank you for the submission. The writing is pretty tight. My feedback is a mix of corrections and suggestions to improve the language.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
+---
+title: 'Porting USB applications to the web. Part 1: libusb'
```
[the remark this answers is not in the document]
RReverser: I'm on a strict deadline, so I can't really delay this one further. I'm starting on part 2 already though - for now I have a structure, and going to add content over the next few days.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+I'll show how apps that communicate with USB devices can be ported to the web by porting [libusb](https://libusb.info/) - a popular USB library written in C - to WebAssembly (via [Emscripten](https://emscripten.org/)), Asyncify and [WebUSB](/usb/).
+
+{% Aside %}
+Fun fact: WebUSB in Chromium also uses libusb under the hood, so what the port achieves is, in fact, one libusb, compiled to WebAssembly, talking to another libusb, shipped as part of the browser, through an intermediate API layer. Isn't the web fun?
+{% endAside %}
+
+## First things first: a demo
+
+The most important thing to do when porting a library, obviously, is choosing the right demo - something that would showcase the capabilities of the ported library, allow you to test it in a variety of ways, and be visually compelling at the same time.
```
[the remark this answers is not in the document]
RReverser: FWIW it was meant to be tongue-in-cheek joke rather than literal "obviously", so I thought it doesn't fall under that category, but I don't feel strongly about it.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+The idea I've stopped on was DSLR remote control. In particular, an open source project [gPhoto2](http://gphoto.org/) has been in this space long enough to reverse-engineer and implement support for a wide variety of digital cameras. It supports several protocols, but the one I was most interested in was USB support, which it performs via libusb.
+
+I've decided to describe the steps that got me to this working demo in two parts. In this blog post, I'll describe the technical details on how I've ported libusb itself, and what tricks might be necessary to port other popular libraries to Fugu APIs too. In the second post, I'll go into details on porting and integrating gPhoto2 itself.
+
+In the end, I got a working web application that previews live feed from a DSLR and can control its settings over USB. If you're impatient, feel free to jump ahead and check out the demo:
```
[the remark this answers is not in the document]
RReverser: Good catch. Initially I was going to add it in the end and only link to "jump ahead" here, but later decided to add a preview in the beginning of the article. I'll change the wording.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: Hmm "it allows to bringing" doesn't sound right grammatically.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: Since it's "executables" (plural), should this be "file extensions" then?

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: It is.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: "initial" here is to clarify "as opposed to the final version of the demo". Do you think it being a "proof-of-concept" is enough for that distinction?

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: "find a way" seems important in this context, because it's one of those things that are easier said than done.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: Ah interesting, I find `.` distinction between standalone function vs methods helpful, but, of course, general styleguide is more important.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: Should the 2nd part of that sentence be changed to use present tense too perhaps? Something like:

```suggestion
The difficulty is that, unlike on other platforms, there is no way to enumerate all the connected USB devices on the web for security reasons. Instead, the flow is split into two parts. First, the web application requests devices with specific properties via [`navigator.usb.requestDevice()`](https://developer.mozilla.org/docs/Web/API/USB/requestDevice) and the user manually chooses which device they want to expose or rejects the permission prompt. Afterwards, the application can list the already approved and connected devices via [`navigator.usb.getDevices()`](https://developer.mozilla.org/docs/Web/API/USB/getDevices).
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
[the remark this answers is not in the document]
jpmedley: I didn't read it as a joke. Tone in notoriously difficult in writing.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: Yes! Good catch.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: Do you think you need to tell your readers or do you think they already know? If the former, you should make the point explicit. If the latter, you don't need the extra text. How you phrase it is up to you, but here's an example. (Emphasis is so you can see what I added. I'm not recommending bolding it.)

>developers need to run those loops without blocking the main thread, **even if it can sometimes be difficult**.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: Yes. "initial" implies multiple proofs of concept. If you prove it once, why do you need to prove it again?

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: Yes. I fixed the suggestion.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: Yeah, I messed that suggestion up. I removed 'to'.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
jpmedley: I updated the suggestion.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: Why did you remove "available" btw? Without it that sentence doesn't sound right either.

jpmedley: It doesn't sound weird to me. 'Just a click away' means available, making the word redundant.

RReverser: Right but together with the verb before it reads as "making them a click away" which, to me, sounds like it's possible to _make_ them easily rather than _access_ them easily, so "available" or "accessible" feels necessary.

jpmedley: 'Making' has [two definitions in English](https://www.google.com/search?q=making+definition&oq=making+definition&aqs=chrome..69i57j0i512l2j69i60j69i65l3j69i61.2539j1j7&sourceid=chrome&ie=UTF-8). This is entirely idiomatic.

RReverser: Ah. Perhaps it's just one of those things that sound fine to native speakers who are aware of the secondary meaning, but that non-native speakers like me might stumble upon when reading. I can remove the word if you insist though.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
```diff
@@ -0,0 +1,450 @@
...
+
+{% Aside %}
+Fun fact: WebUSB in Chromium also uses libusb under the hood, so what the port achieves is, in fact, one libusb, compiled to WebAssembly, talking to another libusb, shipped as part of the browser, through an intermediate API layer. Isn't the web fun?
+{% endAside %}
+
+## First things first: a demo
+
+The most important thing to do when porting a library, obviously, is choosing the right demo - something that would showcase the capabilities of the ported library, allow you to test it in a variety of ways, and be visually compelling at the same time.
+
+The idea I've stopped on was DSLR remote control. In particular, an open source project [gPhoto2](http://gphoto.org/) has been in this space long enough to reverse-engineer and implement support for a wide variety of digital cameras. It supports several protocols, but the one I was most interested in was USB support, which it performs via libusb.
```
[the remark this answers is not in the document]
RReverser: Can I do at least

```suggestion
Idea I chose was DSLR remote control. In particular, an open source project [gPhoto2](http://gphoto.org/) has been in this space long enough to reverse-engineer and implement support for a wide variety of digital cameras. It supports several protocols, but the one I was most interested in was USB support, which it performs via libusb.
```

I want to clarify that there are various ways to use or showcase the library, I just chose this one for the demo as it seemed most compelling to me.

src/site/content/en/blog/porting-libusb-to-webusb/index.md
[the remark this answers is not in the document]
RReverser: Reworded, added a link to the demo and an image - which is a picture of the demo running on the laptop & connected to a camera. I think I'll replace it with a pre-recorded video tomorrow though, as not all users have DSLRs to try the demo on.

jpmedley (approved):

RReverser:
@jpmedley So what does the approval mean for the leftover comments / discussions?

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser:
```suggestion
Most of the backend code uses `val` and `promise_result` in a similar way as already shown above. There are few more interesting hacks in the data transfer handling code, but those implementation details are less important for the purposes of this article. Make sure to check the code and comments on Github if you're interested.
```

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: "System languages" is a fairly common term. https://en.wikipedia.org/wiki/System_programming_language

src/site/content/en/blog/porting-libusb-to-webusb/index.md
RReverser: I do need to tell the readers since it's one of the gotchas many people switching to the web are not aware of.

PR #7152, opened 2022-01-11
Corrections to HID article

jpmedley:
@beaufortfrancois 

After you knocked off, I noticed that your update forgot to change the year on the `updated` meta element. As long as I was here, I fixed the article on 'HID'.

### DIFF
--- a/src/site/content/en/blog/hid/index.md
+++ b/src/site/content/en/blog/hid/index.md
@@ -6,7 +6,7 @@ subhead: |
 authors:
   - beaufortfrancois
 date: 2020-09-15
-updated: 2021-01-10
+updated: 2022-01-10
 hero: image/admin/05NRg2Lw0w5Rv6TToabY.jpg
 thumbnail: image/admin/AfLwyZZbL7bh4S4RikYi.jpg
 alt: Elgato Stream Deck photo.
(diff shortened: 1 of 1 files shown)


beaufortfrancois (approved):
Thank you @jpmedley for following up!

ajr33:
Saw [this PR](https://github.com/GoogleChrome/web.dev/pull/7141) for an unrelated issue but I was wondering why you switched it from "a HID" to "an HID"? "A HID" is correct. Saying "An HID" is much harder that "A HID".

jpmedley:
The letter 'H' is pronounced with a leading vowel in English. When something begins with a vowel sound, it's proceeded by 'an'.

ajr33:
> The letter 'H' is pronounced with a leading vowel in English. When something begins with a vowel sound, it's proceeded by 'an'.

Yes, correct a vowel sound like 'hour'. You would say "an hour" but you wouldn't say "an home". Unless you pronounce HID like "id", it should not have "an" before it. The original with all "a HID" was correct.

ajr33:
> > The letter 'H' is pronounced with a leading vowel in English. When something begins with a vowel sound, it's proceeded by 'an'.
> 
> Yes, correct a vowel sound like 'hour'. You would say "an hour" but you wouldn't say "an home". Unless you pronounce HID like "id", it should not have "an" before it. The original with all "a HID" was correct. 

I see where the confusion is. You're saying HID as H-I-D whereas I'm saying it as a word. That's a bit confusing because most of the document says HID device which would be saying "Human Interface Device device". HID as a word is still the same but encapsulates it more & makes a bit more sense. I don't know which is used more often but I guess it doesn't matter. I say HID as a word since it's easier & my whole company says it like that but "An H-I-D" is fine as long as you say it like that I guess.

PR #7153, opened 2022-01-12
Update the web-vitals library and metrics

philipwalton:
This PR updates the version of the [web-vitals](https://github.com/GoogleChrome/web-vitals/) library used on web.dev. It also adds two additional metrics (FCP and TTFB) that we can use to help debug performance issues.

I have a few additional performance improvement PRs that I plan to submit shortly, but I'd like to get this PR merged first so we can compare the before and after results.

PR #7155, opened 2022-01-12
Conditionally load sign-in to enable bfcache

philipwalton:
### Summary: 

This PR makes web.dev pages eligible for [bfcache](https://web.dev/bfcache/), which should dramatically improve the load performance for users navigating around the site.

### Problem: 

If you use DevTools to test web.dev pages for [bfcache eligibility](https://web.dev/bfcache/#test-to-ensure-your-pages-are-cacheable), you'll notice that they're ineligible due to an open IndexedDB connection:

Screen Shot 2022-01-11 at 9 42 21 PM

And if you look at the IndexedDB pane in DevTools, you'll see that the only IndexedDB database has "firebase" in the name:

Screen Shot 2022-01-11 at 9 55 03 PM

### Solution: 

After looking through the code for web.dev, as far as I can tell Firebase is only used for sign-in functionality, which (again, as far as I know) was removed after the [recent PSI updates](https://web.dev/whats-new-pagespeed-insights/) rolled out at CDS last fall.

And given that sign-in functionality is no longer being used, it seemed like the easiest solution to fix the bfcache issues would be to just remove all the code that loads firebase (and creates the IndexedDB connection).

This PR removes the following components from the codebase:

- `<profile-switcher-container>`
-`<profile-switcher>`
- `<web-signin-button>`

After removing these components and testing the site locally, it renders and looks perfectly fine (no CSS tweaks needed to update the navigation bar), and I could confirm that bfcache now works as expected:

Screen Shot 2022-01-11 at 9 50 35 PM

This PR also updates the analytics code based on [the recommendations in the bfcache post](https://web.dev/bfcache/#how-bfcache-affects-analytics-and-performance-measurement): It adds a `pageshow` listener to track bfcache restores as pageviews. It also adds a new custom dimension to make it possible to distinguish between pageviews resulting for a bfcache restore vs. other traditional navigation types. Lastly it bumps the tracking version so we can compare performance metrics on the site before and after this change.

**Note:** Before merging this PR, we need to deploy #7153 so we can have data to compare before and after.


src/site/_includes/partials/analytics.njk
```diff
@@ -10,6 +10,11 @@ ga('set', 'transport', 'beacon');
 ga('set', 'anonymizeIp', true);
 ga('set', 'page', window.location.pathname);
 ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version}}');
+try {
+  ga('set', '{{ site.analytics.dimensions.NAVIGATION_TYPE }}', performance.getEntriesByType('navigation')[0].type);
```
philipwalton, line 14: Most modern browsers support `performance.getEntriesByType()`, but to avoid issues with older browsers, it's safer to just wrap this in a try/catch to prevent errors breaking the rest of the analytics initialization code.

philipwalton:
Based on internal conversations with @MichaelSolati, I re-added the sign-in components but updated them to lazily load the Firebase auth library on user interaction with the sign-in button. If the user was previously signed in, then the auth library is loaded right away.

Given that web.dev is not currently doing anything differently for signed-in users, I think it's best to remove the button (we can always re-add later), but either way bfcache now works for all users who are not signed in.

philipwalton:
@devnook @MichaelSolati could one of you take a look at this PR when you get a second? Now that I've merged #7154 we can merge this one at any time.

src/lib/fb.js
```diff
@@ -195,6 +203,7 @@ export async function saveUserUrl(url, auditedOn = null) {
 export async function signIn() {
   let user = null;
   try {
+    initialize();
```
devnook, line 206: Can you add a short comment on why the lazy loading is important here, mentioning indexdb?

src/lib/analytics.js
```diff
@@ -104,14 +103,24 @@ document.addEventListener(
...
+ * https://web.dev/bfcache/#how-bfcache-affects-analytics-and-performance-measurement
+ */
+window.addEventListener(
+  'pageshow',
+  /**
+   * @param {PageTransitionEvent} e
+   */
+  (e) => {
+    if (e.persisted) {
+      ga('set', dimensions.NAVIGATION_TYPE, 'bfcache');
```
devnook, line 119: THis will break the continuity of our analytics for the signed_in dimention, right?

devnook (approved):

src/lib/analytics.js
philipwalton: I re-added the old code since we're planning on keeping the sign-in button on pages that currently have it.

I did want to point out, however, that the existing code isn't actually very useful. It tracks the state change when a user signs in or out, but it doesn't set the stored state at page load time (for the initial pageview), so if a user doesn't sign in or out during their session, we don't actually know whether they were signed in. For example, only 17 out of the 600K pageviews received in the last 7 days have any sign-in information.

If we're planning to use the sign-in functionality more in the future, we should come back to this issue (which would probably require a refactor of the existing analytics code).

src/lib/fb.js
philipwalton: Added (also added a comment above to the `initialize()` function.

PR #7167, opened 2022-01-13
Fix broken link

tcope25:
Fixes #SOME_ISSUE_NUMBER

-
-
-


rachelandrew:
This PR adds the web.dev domain to the front of an internal link, which isn't something we want (and fails the linter because of this).

PR #7174, opened 2022-01-14
Add resource hint that preconnects to imgix

khempenius:
cc @philipwalton

### DIFF
--- a/src/site/_includes/partials/head.njk
+++ b/src/site/_includes/partials/head.njk
@@ -21,6 +21,7 @@
 
 <link rel="preload" as="font" crossorigin href="/fonts/google-sans/regular/latin.woff2">
 <link rel="preload" as="font" crossorigin href="/fonts/google-sans/bold/latin.woff2">
+<link rel="preconnect" href="https://web-dev.imgix.net">
 
 <meta name="theme-color" content="#fff"/>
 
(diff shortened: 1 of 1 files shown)


src/site/_includes/partials/head.njk
```diff
@@ -21,6 +21,7 @@
 
 <link rel="preload" as="font" crossorigin href="/fonts/google-sans/regular/latin.woff2">
 <link rel="preload" as="font" crossorigin href="/fonts/google-sans/bold/latin.woff2">
+<link rel="preconnect" href="https://web-dev.imgix.net">
```
MichaelSolati, line 24: Minor nit, i like the idea. Can we change the href to `https://{{site.imgixDomain}}`. That way we link it to a var, if it we ever change the domain the preconnect will automatically update.

MichaelSolati (changes requested):

src/site/_includes/partials/head.njk
malchata: Another minor nit, and I'm not sure it matters: do we need the trailing slash on the URL?

philipwalton: `preconnect` will turn the URL passed into an absolute URL and then use the origin to preconnect, so as long as there's enough data to get the origin (in this case there is), it should be fine. https://www.w3.org/TR/resource-hints/#dfn-preconnect

philipwalton (approved):
As discussed in our email thread about this, I think in most cases this will not noticeably improve performance because the browser's preload scanner will likely discover the imgix origin via the `preconnect` hint at roughly the same time as it discovers it via  source URLs.

However, it certainly won't hurt performance, and there could be some cases where the source images are far enough down the page that having it here at the top will be helpful.

MichaelSolati (approved):

PR #7177, opened 2022-01-14
feat(select): create `web-select` element, fixes #7176

### LINKED ISSUE #7176: Create new select element
Create select element that looks like courses filter from figma, ie, a material select element.

MichaelSolati:
Fixes #7176

- create web-select component
-
-

### DIFF
--- a/src/lib/components/Select/index.js
+++ b/src/lib/components/Select/index.js
@@ -1,0 +1,153 @@
+import {html} from 'lit-element';
+import {BaseElement} from '../BaseElement';
+
+/**
+ * Element that renders a Material style select element.
+ *
+ * @extends {BaseElement}
+ */
+class Select extends BaseElement {
+  static get properties() {
+    return {
+      active: {type: Boolean, reflect: true},
+      selected: {type: Object},
+    };
+  }
+
+  constructor() {
+    super();
+    this.active = false;
+    this.activeIndex = -1;
+    /** @type {HTMLOptionElement[]} */
+    this.options = [];
+    /** @type {HTMLOptionElement} */
+    this.selected = null;
+    this.value = null;
+
+    this.clickedInside = this.clickedInside.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+    const select = this.querySelector('select');
+    this.options = /** @type {HTMLOptionElement[]} */ (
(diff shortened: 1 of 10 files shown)


src/lib/components/Select/index.js
```diff
@@ -0,0 +1,153 @@
...
+   * Turns option elements into li elements.
+   *
+   * @param {HTMLOptionElement} o
+   * @returns {TemplateResult}
+   */
+  renderOptions(o) {
+    return html`<li
+      ?selected=${o === this.selected}
+      tabindex="0"
+      @click=${() => this.selectOption(o)}
```
philipwalton, line 145: Why is the arrow function needed here? I see you're binding some other callbacks above in the constructor. I think it would make sense to be consistent for all of them.

philipwalton (changes requested):

src/lib/components/Select/index.js
MichaelSolati: In this case I'm not looking for the event, I'm looking for the `option` element that I'm mapping it to. I'll make the other changes though and get back to you next week.

philipwalton: Ahh, sorry, didn't notice you were passing in a variable. SGTM.

src/lib/components/Select/index.js
```diff
@@ -69,21 +69,25 @@ class Select extends BaseElement {
...
-    }
-    e.preventDefault();
-    const increment = e.key === 'ArrowUp' ? -1 : 1;
-    const lis = this.querySelectorAll('li');
+    switch (e.key) {
+      case 'Escape':
+        this.toggleState();
+        break;
+      case 'ArrowUp':
+      case 'ArrowDown':
```
philipwalton, line 77: One other thing I just noticed (difference between this component and a native select) is I can't press the up/down arrows to open a closed select element when it's focused.

devnook:
@philipwalton Can you please take another look?

MichaelSolati:
@philipwalton here the element on a page you can tinker with https://deploy-preview-7177--web-dev-staging.netlify.app/design-system/preview/select/web-select/

matthiasrohmer:
@philipwalton, the accessibility issues you mentioned seem to be fixed for desktop. Can you take another look?

Also some context on this would be helpful. Is this created to have a select that fits web.dev's style? Because that comes with the price of decreased accessibility on Android and iOS devices as they don't open the native select drawer anymore.

philipwalton:
@devnook would have more context on whether or not this is still needed.

PR #7209, opened 2022-01-21
Update web-vitals to 2.1.4

philipwalton:
This PR updates the web-vitals JS library to address an issue fixed in https://github.com/GoogleChrome/web-vitals/issues/201.

This change should have no effect on the functionality of the site, but it'll address an issue with the accuracy of the TTFB data we're reporting to Google Analytics.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "strip-comments": "^2.0.1",
     "terser": "^5.9.0",
     "unistore": "^3.4.1",
-    "web-vitals": "^2.1.3",
+    "web-vitals": "^2.1.4",
     "webdev-infra": "^1.0.28",
     "wicg-inert": "^3.0.1"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -73,7 +73,7 @@
     "strip-comments": "^2.0.1",
     "terser": "^5.9.0",
     "unistore": "^3.4.1",
-    "web-vitals": "^2.1.3",
+    "web-vitals": "^2.1.4",
```
MichaelSolati, line 76: Generally this is fine, but did you just change the number in the JSON or did you `npm i web-vitals@2.1.4`

I would have thought the package-lock.json would have updated as well, which I think we need (i feel like things can get wonky with out automated builds without it).

philipwalton: Ahh, right, I forgot to `npm install` after changing the version. Updated!

malchata (approved):
LGTM

PR #7229, opened 2022-01-24
Updates metric links

malchata:
This pull request updates links across several docs to link to canonical metric pages rather than Lighthouse audit pages:

- Updates links pointing to /first-contentful-paint/ to /fcp/
- Updates links pointing to /interactive/ to /tti/
- Updates links pointing to /largest-contentful-paint/ to /lcp/ (although the former was already redirecting to the latter, but this should reduce redirect latency).


philipwalton (commented):
This looks good so far, but I think it would make sense to also add redirects for these URLs. We don't know what links exist out in the wild, but if any do then we should redirect them to the right place.

malchata:
Agreed, but that would mean we'd have to remove the relevant audit pages before migrating them. I also seeing redirects being one of the last things we do so we have better continuity in our documentation (that is, half of it isn't stuff on d.c.c. and the other half isn't still on web.dev).

philipwalton:
Gotcha, makes sense, SGTM!

philipwalton (approved):

PR #7232, opened 2022-01-24
feat(select): style native select component to look material, fixes #7176

### LINKED ISSUE #7176: Create new select element

MichaelSolati:
Fixes #7176

- Change style to be material-esque.
-
-

### DIFF
--- a/src/scss/blocks/_select.scss
+++ b/src/scss/blocks/_select.scss
@@ -1,0 +1,18 @@
+select {
+  background-color: transparent;
+  border: none;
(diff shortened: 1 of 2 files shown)


devnook:
Can you please add a component for it undes design-system, so it can be tested?

devnook:
I think the idea is to have the whole element styled, not only <select>. Right not the select looks like material design, but the menu it opens does not:

Screenshot 2022-01-26 at 17 52 24

It does not block merging this PR, but please update description ("Fixes" part) so that the issue stays open.

PR #7233, opened 2022-01-25
fix example of position

### LINKED ISSUE #7175: Description of background-position 2-value syntax is wrong.
**Describe the bug**
In the page https://web.dev/learn/css/backgrounds/, the description of background-position 2-value syntax is incorrect.
The answer of Question 3 at the end of the module says `background-position: 25% bottom;` is invalid. But actually it is valid in Chrome and Firefox.

**To Reproduce**
1. Set `background-repeat: no-repeat;`
2. Set `background-position: 25% bottom;`
3. Check that the background image is at the bottom and 25% from left.
4. Set `background-position: bottom 25%;`
5. Both Chrome and Firefox shows the property is invalid.

**Expected behavior**
The document should be changed to match the behavior in browsers.
`background-position: 25% bottom;` is valid, means the background at the bottom, and 25% from left.
`background-position: bottom 25%;` is invalid, even the document says "When CSS values are used alongside keywords, it must be preceded by a keyword."
`background-position: right 25%;` is valid, but it means at the right, and 25% from top.
And the answer of Question 3 at the end of the module should be updated.

rachelandrew:
Fixes #7175 

I think our example conflated two things so I've made it clearly about order of the values.

### DIFF
--- a/src/site/content/en/learn/css/backgrounds/backgrounds.assess.yml
+++ b/src/site/content/en/learn/css/backgrounds/backgrounds.assess.yml
@@ -31,9 +31,9 @@ questions:
     correctAnswers: "1,2,3"
     stem: Which of the following `background-position` declarations are valid?
     options:
-      - content: '`background-position: 25% bottom`'
+      - content: '`background-position: 50% left`'
         rationale: |
-          When CSS values are used alongside keywords, it must be preceded by a keyword.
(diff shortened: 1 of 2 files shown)


rachelandrew:
@una the linked issue was raised, and looking at it I think the article conflates two things. So I've used the simpler case of the order of values in the article and the example.

argyleink (approved):

PR #7237, opened 2022-01-25
Update for Topics

samdutton:


rowan-m (changes requested):
I'm unsure about just _removing_ FLoC. Could we add a sentence to the Topics bullet saying something like the top-level blog post has? e.g. "Topics was informed by community feedback from our earlier FLoC trials, and replaces our FLoC proposal."

samdutton:
> I'm unsure about just _removing_ FLoC. Could we add a sentence to the Topics bullet saying something like the top-level blog post has? e.g. "Topics was informed by community feedback from our earlier FLoC trials, and replaces our FLoC proposal."

Good idea — added.

rowan-m (approved):
LGTM

PR #7254, opened 2022-01-27
fix(i18n): display i18n titles of courses in paths

### LINKED ISSUE #7225: Localized collections should link to localized contents.
**Describe the bug**
Localized collections page is translated properly. However, the articles in that pages are linked to English contents regardless of if they are translated or not. Readers will not be able to find the translated articles.

**To Reproduce**
1. Go to https://web.dev/i18n/ja/secure/
2. Click on "Samesite cookies explained"

**Expected behavior**
Linked to https://web.dev/i18n/ja/samesite-cookies-explained/ so that the article is displayed in Japanese.

In currently implementation, it is linked to Linked to https://web.dev/samesite-cookies-explained/ so it's displayed in English despite it's already translated.

(Note that however if the browser is configured to Japanese as the primary lang for ex, the URLs won't have `/i18n/ja` and always displays Japanese contents where possible. I know this makes it lower priority, but someone how configures English as primary, but prefers Japanese where possible will have hard time finding translated contents.)

MichaelSolati:
Related to #7225

Fixes only titles, links fixed in #7283

- Add `i18n` to urls used to display paths' items' links

### DIFF
--- a/src/site/_includes/partials/item.njk
+++ b/src/site/_includes/partials/item.njk
@@ -1,13 +1,11 @@
 {% for pathItem in section.pathItems %}
-  {% set url = ['', lang, pathItem, ''] | join('/') %}
+  {% set url = ['', 'i18n', lang, pathItem, ''] | join('/') %}
   {% set item = url | findByUrl %}
-
   {# Add a default language fallback in case the article is not translated. #}
   {% if not item %}
     {% set url = ['', 'en', pathItem, ''] | join('/') %}
     {% set item = url | findByUrl %}
   {% endif %}
-
   {% if pathItem.url %}
     <li>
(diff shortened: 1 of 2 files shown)


src/site/_includes/partials/item.njk
```diff
@@ -17,7 +15,7 @@
   {% elif item %}
     {% if not item.data.draft %}
       <li>
-        <a href="{{ item.url }}">
+        <a href="{{ item.url | getDefaultUrl }}">
```
devnook, line 18: So you add i18n to the item in line 2 url just to remove it here. Can you describe what is the expected effect?

devnook (changes requested):
Reviewed

devnook (approved):

PR #7270, opened 2022-01-29
Update index.md for better wordings

nitinmalave:
- Updating docs for better synonyms


jpmedley:
@nitinmalave Thanks for submitting this.

Would you please explain why you think 'unnecessary' is an improvement over 'unneeded'?

PR #7274, opened 2022-01-30
feat(browser-compat): update to use new design system, fixes #6846, fixes #7139

### LINKED ISSUE #6846: Convert browser compat component to use the new design system pattern
Pattern: https://web.dev/design-system/component/browser-compat/
Component: `src/site/_includes/components/BrowserCompat.js`

### LINKED ISSUE #7139: BrowserCompat widget shows "preview" instead of icon.
oh hai!

**Describe the bug**
In [this article](https://web.dev/structured-clone/) the BrowserCompat widget is showing the word "preview" next to Safari, instead of an icon.

image

Also, this is a total nit but, do y'all think the widget would look nicer if the icons were a bit larger?

image

I think the original design from Jiwoong had the icons larger and closer together, and then we made a separate smaller version for Patterns page because it was being used in a smaller area. But it may make sense to have a larger variant for articles, and a smaller one on Patterns.

MichaelSolati:
Fixes #6846
Fixes #7139

- Remove all old SCSS from legacy
- Restructure compat code and attributes to align with new design system
- Use new colors and breakpoints to migrate
- Add compat label/icon/etc for features that are previews


src/scss/_config.scss
```diff
@@ -200,8 +200,11 @@ $gorko-config: (
   ),
   'breakpoints': (
     'xs': '(max-width: 25em)',
+    'max-md': '(max-width: 50em)',
```
devnook, line 203: You should be able to structure your css to use existing breakpoints (they are written this way to favos mobile-first design). Mixing min-width and max-width leads to rather unreadable and unreliable css. If you find the use of it absolutely necessary, please move the reused value (50em etc) to a variable.

devnook (changes requested):
You need to look into spacing of the element on mobile. Some examples:

at 380px:
Screenshot 2022-02-01 at 16 47 42

on Pixel5:
Screenshot 2022-02-01 at 16 47 51

devnook (approved):

devnook (changes requested):
This is still happening:
Screenshot 2022-02-02 at 14 37 24

devnook (approved):

PR #7287, opened 2022-02-01
refactor(author): use `Author` shortcode throughout site, fixes #7285

### LINKED ISSUE #7285: Use `Author` shortcode throughout site
Generally we use the `Author` shortcode to render authors, but in a few places on the site, such as in the courses and component library, we have hardcoded HTML. We should fix/change that so we are using the shortcode instead.

I discovered this by accident when I was exploring and trying to remove the `w-author*` class (this will be removed when #7214, then we'll need to delete `src/styles/pages/_author.scss`)

MichaelSolati:
Fixes #7285

- have author stack use Author shortcode
- add shortcode implementation documentation in design system and handbook


devnook (changes requested):
Discussion needed

src/site/content/en/handbook/web-dev-components/index.md
```diff
@@ -161,6 +162,56 @@ Use the codelab aside to link to an associated codelab.
   [Using Imagemin with Grunt](#)
 {% endAside %}
 
+## Author
```
devnook, line 165: Please remove. This component is not intended to be used by authors inside prose.

src/site/_includes/partials/author-stack.njk
```diff
@@ -1,37 +1,11 @@
 <div class="flow">
```
devnook, line 1: You mentioned this partial is used in only 1 place. It will be better to remove it then and replace with the shortcode directly, since it is not reused anywhere.

devnook (commented):
Reviewed

src/site/_includes/partials/author-stack.njk
MichaelSolati: It's only used in 4 different file (for the same purpose)

- /src/site/content/en/learn/css/conclusion/index.md
- /src/site/content/en/learn/design/conclusion/index.md
- /src/site/content/en/learn/forms/conclusion/index.md
- /src/site/content/en/learn/pwa/conclusion/index.md

Also, /src/site/_includes/partials/post-next.njk has a similar use of "authors". I think we should keep the stacks, but maybe make some changes so it can be used in both.

MichaelSolati: Plz see => https://github.com/GoogleChrome/web.dev/pull/7287/files#diff-136f5a836acb3ac96fd5d99e3248e9a083705c1751508eccc4c558ae10411319

src/site/content/en/learn/css/conclusion/index.md
```diff
@@ -35,4 +35,7 @@ To continue to build your CSS skillset, here are some additional resources to ex
 
 ## Brought to you by
 
-{% include 'partials/author-stack.njk' %}
+<!--lint disable-->
```
devnook, line 38: Why disable all lint rules here?

devnook, line 38: Lint complains here, that you use template tags inside an .md file. This is a hint for you as a developer that you're doing something unusual, and therefore you should have a very strong reason to do so.

Also, you're repeating the same code over 4 files. Again a strong hint that this snippet of code is in the wrong place. 

Maybe it's worth moving this part to the njk template, instead of repeating code in .md files.

devnook (changes requested):
Lint is complaining, because you should not be using a macro in .md file. You can either:
+ go back to the authors-stack template include (there is no reason to move this markup to a macro anyways)
+ add course-conclusion.njk template and use you macro there

devnook (approved):

PR #7296, opened 2022-02-02
Update BCD

rachelandrew:
Update to latest release https://github.com/mdn/browser-compat-data/releases/tag/v4.1.6


MichaelSolati:
As odd of a request as it is, can you run `npm i`

Otherwise the package-lock.json will be out of sync, which isn't the biggest issue, but we'd want in sync.

Otherwise lgtm

MichaelSolati (approved):
Please see comment before merging

PR #7311, opened 2022-02-06
Fix date issues on some Web Vitals content

philipwalton:
This PR makes some minor fixes to a few incorrect dates in some of the Web Vitals posts.


malchata (approved):
LGTM, at least on the face of it. Date formats look correct, etc.

PR #7312, opened 2022-02-06
Add importance=high to likely LCP images

philipwalton:
_**Important!** This PR cannot be merged until https://github.com/GoogleChrome/webdev-infra/pull/21 is merged._

Many of the pages on web.dev currently have an LCP that is higher than it should be. One of the main reasons for this is the images are loaded to late (e.g. the don't start loading until after the fonts and stylesheets finish).

Screen Shot 2022-02-06 at 12 28 47 PM

For example, in the screenshot above you can see, the LCP element's resource ideally should have started loading around 900ms but it doesn't start until almost a second later.

While this issue could be solved with preload, due to the fact that the most pages on web.dev share a common `<head>` template, it would require a major refactor to preload all of these images.

However, the experimental [Priority Hints API](https://web.dev/priority-hints/) directly addresses this issue and is a great solution to the problem web.dev faces. The API is currently in origin trial (so it may change), but I can commit to following the progress of the API and making all future updates necessary.

This PR adds the `importance=high` attribute to likely LCP images. This includes:
- The home page
- Posts with a hero image
- Learning path pages
- The /learn page

It also adds a meta tag for the origin trial token, which I registered. Once Chrome 100 ships, this origin trial will have expired, so it can be safely removed at that time.


devnook:
You need to bump webdev-infra version in package.json to 1.30

package.json
```diff
@@ -74,7 +74,7 @@
     "terser": "^5.9.0",
     "unistore": "^3.4.1",
     "web-vitals": "^2.1.4",
-    "webdev-infra": "^1.0.28",
+    "webdev-infra": "^1.0.30",
```
jeffposnick, line 77: While https://github.com/GoogleChrome/webdev-infra/pull/21 was committed to GitHub, there hasn't been a corresponding publish of `v1.0.30` to `npm`: https://www.npmjs.com/package/webdev-infra

There are some instructions on publishing at https://github.com/GoogleChrome/webdev-infra#-important, but all of the previous versions have been published by folks who are no longer on the team.

@devnook, do you have publish rights? If not, I think we can pull in some folks with admin access to help.

jeffposnick (commented):
The changes seem fine to me, but we need the `npm` deployment of `webdev-infra` to happen before this could be deployed.

package.json
devnook: Published now.

jeffposnick (approved):
It would be great if you could take note of any unexpected knock-on effects that loading the images early has, i.e. whether it leads to more network contention when loading the fonts/styles. But this seems like a good thing to try out for the LCP benefits.

devnook (approved):

philipwalton:
> It would be great if you could take note of any unexpected knock-on effects that loading the images early has, i.e. whether it leads to more network contention when loading the fonts/styles. But this seems like a good thing to try out for the LCP benefits.

@jeffposnick good suggestion. If there are any side-effects due to increased network contention, I think we'd be able to see that manifest as a regression in the FCP metric. So I'll be sure to check what affect this change has on both FCP and LCP.

PR #7313, opened 2022-02-07
Turn browsercompat into table

### LINKED ISSUE #7310: Use a <table> for browser support list in blog posts
**Describe the bug**
The browser support list in articles such as ["New to the web platform in January"](https://web.dev/web-platform-01-2022) has a DOM structure which in absence of styling, for example in an RSS feed reader, makes it difficult to understand.

Using a `<table>` would help group the information better.

**To Reproduce**
1. Subscribe to [web.dev feed](https://web.dev/feed.xml) in an RSS feed reader
2. Find the entry for https://web.dev/web-platform-01-2022
3. See "Browser support" section.

**Screenshots**
Screenshot 2022-02-06 at 12 55 52
*Browser support list as seen in a browser (with styles)*

Screenshot 2022-02-06 at 12 56 21
*Browser support list as seen in an RSS feeed reader (without styles)*

devnook:
Fixes GoogleChrome/webdev-infra#54

### DIFF
--- a/src/scss/blocks/_browser-compat.scss
+++ b/src/scss/blocks/_browser-compat.scss
@@ -20,6 +20,7 @@
   margin-right: 6px;
   display: inline-block;
   background: content-box no-repeat transparent;
+  vertical-align: middle;
 
   &[data-browser='chrome'] {
     background-image: url('/images/icons/chrome.svg');
@@ -43,9 +44,13 @@
   }
 }
 
+.browser-compat__items td {
+  white-space: nowrap;
+  border: 0;
+  padding: 0;
+}
+
 .browser-compat__items {
-  display: flex;
-  align-items: center;
   margin: 1rem 1rem 1rem 0;
 }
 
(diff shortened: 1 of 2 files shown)


rachelandrew:
@devnook I'm wondering if we should make the label "Browser Compat" into a `<caption>`? What would be the best for a11y?

jeffposnick:
FWIW, this is what the `<table>` implementation looks like in Feedly, when I access the RSS feed for the staging server:

Screen Shot 2022-03-10 at 9 28 18 AM

I guess that's better? I'm okay with this PR, but I don't know the answer to Rachel's question about `<caption>` best practices.

(Not to derail the PR, but another approach would be to turn the original `<span>`s into `<li>`s, with a parent `<ul>`. This info strikes me as being "list-like" and the unstyled `<ul>`/`<li>` combo might end up being the most legible in an RSS reader.

matthiasrohmer:
Fixed by https://github.com/GoogleChrome/webdev-infra/pull/60.

PR #7317, opened 2022-02-07
add link to /accessibile from /lighthouse-accessibility

connorjclark:
Lighthouse report will be changing the link here:
image

from https://developers.google.com/web/fundamentals/accessibility to https://web.dev/lighthouse-accessibility/

Adding link to /accessible to give developers a starting point on a11y. Hopefully all the content under that group of pages is equivalent to the old link (I did not verify. I would have liked to linked to it aswell in /lighthouse-accesibility, but I could not work eleventy).

### DIFF
--- a/src/site/_data/i18n/paths/lighthouse_accessibility.yml
+++ b/src/site/_data/i18n/paths/lighthouse_accessibility.yml
@@ -26,6 +26,9 @@ overview:
   zh: 这些检查能够突出可以改进您的 Web 应用无障碍功能的机会。只有一部分无障碍功能问题可以被自动检测到，所以我们也鼓励您进行人工测试。
 
 topics:
+  a11y_overview:
+    en: Accessibility Overview
+
   audit_scoring:
     en: Audit scoring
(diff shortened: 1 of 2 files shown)


src/site/_data/i18n/paths/lighthouse_accessibility.yml
```diff
@@ -26,6 +26,9 @@ overview:
   zh: 这些检查能够突出可以改进您的 Web 应用无障碍功能的机会。只有一部分无障碍功能问题可以被自动检测到，所以我们也鼓励您进行人工测试。
 
 topics:
+  a11y_overview:
+    en: Accessibility Overview
+
```
connorjclark, line 31: How does the i18n pipeline work? Does this get updated later?

devnook: We request new translations in batches. Until the next batch is rganized and contracted, it will fall back to English string.

devnook (approved):

PR #7329, opened 2022-02-10
Type `pointer` cursor for the entire toggle switch

tomayac:

- Make sure the entire toggle switch has a type `pointer` cursor, currently only the `input` has it, which looks bad.

Screen Shot 2022-02-10 at 12 29 35


jeffposnick (approved):
LGTM.

PR #7331, opened 2022-02-10
Promote isConfigSupported()

beaufortfrancois:
This PR is about promoting the pattern to call the static method `VideoEncoder.isConfigSupported()` with the config to check before hand whether the config is supported.

FYI @Djuffin @chcunningham


chcunningham (commented):
LGTM

Djuffin:
LGTM

PR #7333, opened 2022-02-10
Updates the priority hints post re: origin trial info.

### LINKED ISSUE #7326: The Priority Hints post has misleading information about using the origin trial
The [Priority Hints](https://web.dev/priority-hints/) post include a [section about using the origin trial](https://web.dev/priority-hints/#using-the-trial). In this section there's a code example that shows usage via a `<meta>` tag:

```html
<!-- include trial token in the <head> -->
<meta http-equiv="origin-trial" content="{Replace with origin trial token}">
```

However, I found out that this method does not work if you need it to apply to resources loaded right away (one of the main use cases for this API). I confirmed this with Pat Meenan after I noticed that #7312 was not working as expected in production.

I think we should update the post to remove the `<meta>` instruction and clarify that optimal use of the origin trial requires registering it via a header.

cc: @addyosmani

malchata:
Fixes #7326

This PR updates the info regarding the origin trial to recommend specifying the `Origin-Trial` HTTP header rather than in a `<meta>` HTML element using `http-equiv`.

### DIFF
--- a/src/site/content/en/fast/priority-hints/index.md
+++ b/src/site/content/en/fast/priority-hints/index.md
@@ -577,13 +577,13 @@ You can also enable the **Experimental Web Platform Features** flag in Chrome to
 
 The following is a comparison between a set of images and a script file loaded with different priorities. Note that the default priorities for images and scripts are applied on Chrome 94. On the other hand, the priority set through the importance attribute is applied when tested on Chrome 96.
 
-Specifying an origin trial token in an HTML `<meta>` element is often the quickest and most convenient method. For Priority Hints to work during the origin trial, however, you'll need to specify your token in an `Origin-Trial` HTTP response header:
+Specifying an origin trial token in an HTML `<meta>` element is often the quickest and most convenient method. In the case of Priority Hints, however, this may result in hints being overlooked for resources loaded early on during startup. For Priority Hints to work optimally, you should specify the origin trial token in an `Origin-Trial` HTTP response header:
 
 ```http
 Origin-Trial: {Replace with origin trial token}
 ```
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/priority-hints/index.md
```diff
@@ -576,12 +577,15 @@ You can also enable the **Experimental Web Platform Features** flag in Chrome to
 
 The following is a comparison between a set of images and a script file loaded with different priorities. Note that the default priorities for images and scripts are applied on Chrome 94. On the other hand, the priority set through the importance attribute is applied when tested on Chrome 96.
 
-Following is the relevant mark-up included in the HTML file.
+Specifying an origin trial token in an HTML `<meta>` element is often the quickest and most convenient method. For Priority Hints to work during the origin trial, however, you'll need to specify your token in an `Origin-Trial` HTTP response header:
```
philipwalton, line 580: I'd add a qualifier along the lines of: for Priority Hints to work for resources loaded very early in the page, the origin trial token needs to be specified in the HTTP response header.

The complication here is that the original trial _does_ work, it just doesn't apply to the files whose load was initiated by the preload scanner. Files loaded later one _will_ honor the importance attribute.

philipwalton (approved):
LGTM with one suggestion

src/site/content/en/fast/priority-hints/index.md
philipwalton: I think, if we can make it clear that, if you use a `<meta>` tag, the origin trial gets initialized too late to apply to some of the earliest resources loaded, that's fairly clear what's going on.

malchata: I pushed an update to the copy @philipwalton.

PR #7340, opened 2022-02-11
Fix long words in code blocks in titles

gangsthub:

- Avoiding long words from overlapping the sidebar on desktop.
    - Added `word-break: break-all;` in code blocks inside titles.

I've tested it with [this article](https://web.dev/monitor-total-page-memory-usage/). You can see the before and after here:

Before 🤸🏼
image

After ✅
image

- I chose `break-all` instead of `break-word` because this way the long word starts at the same line. Where with `break-word`, it would start on the next line:

🤔
image

### DIFF
--- a/src/scss/next.scss
+++ b/src/scss/next.scss
@@ -395,6 +395,7 @@ h6 code,
   margin: 0;
   padding: 0;
   white-space: normal;
+  word-break: break-all;
 }
 
 /// Sub and sup


gangsthub:
Something cool to see is that in this article there are also `h2`s affected by the long words:

Before
image

After
image

jeffposnick (approved):
Thank you for the contribution!

PR #7393, opened 2022-02-22
Do Not Track short-circuit

toastal:
A quick one-liner can ensure user's consent and privacy is respected. While `sendBeacon` should respect `doNotTrack` setting, `fetch` has no such respect and it would be wasteful to `JSON.stringify` an object to no-op anyhow. Perhaps `Navigator.storeTrackingException` for [granting exceptions](https://www.w3.org/TR/tracking-dnt/#exception-granting) would be more robust, but the keep-it-simple approach should to respect privacy. DNT seems to still be the default for Brave and Firefox, which makes sense given their philosophies.

### DIFF
--- a/src/site/content/en/vitals/index.md
+++ b/src/site/content/en/vitals/index.md
@@ -6,7 +6,7 @@ hero: image/admin/BHaoqqR73jDWe6FL2kfw.png
 authors:
   - philipwalton
 date: 2020-04-30
-updated: 2021-10-26
+updated: 2022-02-22
 tags:
   - metrics
   - performance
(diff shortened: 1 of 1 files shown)


philipwalton:
I don't believe this feature is currently recommended for use or honored by most platforms. See the warning at the top of this page: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DNT

toastal:
Do Not Track still a [part of the standards](https://www.w3.org/TR/tracking-dnt/#dnt-header-field) currently with a mention that it's more or less frozen do to poor uptake. However I'd venture part of the reason it's not well supported is copy-paste from code blocks such as this were not taking DNT into account when sending analytics. As a steward of the web, I feel it's important to include a one-line short-circuit to respect the user agents that _are_ reporting the DNT preference. Due to the structure of the code, _if_ DNT is actually sunset the code will still work but as far as I can tell, many browsers will still be including it.

philipwalton:
I appreciate the PR, but given that it's not widely adopted, I do not think the solution is to try to get all sites on the internet to update their code one by one.

These days, most people use ad blockers or content blockers to prevent tracking, and that seems like a much more effective route to me.

PR #7402, opened 2022-02-24
Graph widget demo - WIP

devnook:
This PR does 2 things:

1. Adds a Graph example demo - code there is based on Rick Viscomi's jsfiddle and serves as an example of an interactive demo with external scripts. It is draft: true, so it does not get included in web.dev/patterns/ page and is not indexed (it's a hidden code pattern).

2. Adds a Widget shortcode to embed any CodePattern in an article. /test-post/ serves as a prevew playground for this feature.

In order to test CSP policy, we need to run in a prod environment, which is enabled here: https://web-dev-production-1--staging-nbjujvug.web.app/test-post/


src/site/content/en/blog/test-post/index.md
```diff
@@ -602,6 +602,16 @@ quis enim blandit, posuere justo dignissim, scelerisque diam. Fusce aliquet urna
 ac blandit ullamcorper. Proin et semper nibh, sit amet imperdiet velit. Morbi at
 quam sem.
 
+{% Widget 'example-set/graph', 520 %}
```
jeffposnick, line 605: It's not a big deal, but `520` is too small for the sample content—the "Highcharts.com" text gets cut off. `540` seems fine.

Screen Shot 2022-03-09 at 11 09 54 AM

src/site/_includes/partials/head.njk
```diff
@@ -52,43 +52,11 @@
...
-{# default.js loads the basic set of custom elements that all pages need. #}
-{% for item in pageScripts %}
-  loadScript('{{ helpers.hashForProd(item) }}', 'module');
-{% else %}
-  loadScript('{{ helpers.hashForProd('/js/default.js') }}', 'module');
-{% endfor %}
-{# {% if process.env.ELEVENTY_ENV === 'prod' %} #}
+{# Script loader for sourced scripts, as they cannot be authorized by a CSP hash. #}
+{% include 'partials/script-loader.njk' %}
+{% set apiScripts %}
```
jeffposnick, line 57: How does `apiScripts` end up getting a value? From what I can see, it's just a template variable that will always be `undefined`.

If it's meant to be used elsewhere from, e.g., child templates, could you add some documentation or explanation for it?

src/site/_includes/components/Widget.js
```diff
@@ -0,0 +1,25 @@
...
+  if (!pattern) {
+    return '';
+  }
+  return `<div class="widget" style="min-height: ${height}px">
+      <iframe
+        src="${pattern.demo}"
+        title="Demo"
+        height="${height}"
+        width="100%"
+        scrolling="no"></iframe>
```
jeffposnick, line 23: What do you think about adding https://web.dev/iframe-lazy-loading/ to this, either by default or via a `loading` parameter, similar to https://github.com/GoogleChrome/webdev-infra/blob/main/shortcodes/Img.js#L113-L115?

src/site/_includes/partials/head.njk
devnook: apiScripts gets value from whatever is between "set" and "endset" tag. The syntax is:

```
{% set variableName %}
  some value here (text)
{% endset %}
```

src/site/_includes/components/Widget.js
devnook: I made it lazy by default, if we get use cases where this is not enough I'll make it a param.

devnook:
PTAL

src/site/_includes/partials/head.njk
jeffposnick: Thanks!

PR #7406, opened 2022-02-24
Fix Algolia multi-byte content indexing

jeffposnick:
This resolves issues our Algolia indexing job has had with mutli-byte content, following up on #6952.

I've also updated the test suite to be a bit more representative of the issues we're seeing, including using both single- and multi-byte characters in the mocked data.

From what I could tell, there were two root causes for the current problems:

## JSON size considerations
Previously, [`byteof`](https://www.npmjs.com/package/byteof) was used to find the size of a given record (and therefore how many bytes needed to be trimmed). But the Algolia API [cares about the size of the serialized JSON object](https://support.algolia.com/hc/en-us/articles/4406981897617-Is-there-a-size-limit-for-my-index-records-#:~:text=The%20record%20size%20limit%20is%20based%20on%20the%20size%20of%20this%20final%20JSON%20file).

In this PR, I've switched off of `byteof` and instead serialize an object to JSON first, and then count those bytes, to match what Algolia checks.

## String truncation

Previously, the trimming logic attempted to find the percentage of the string that needed to be removed by dividing the total bytes in the string by the number of bytes that needed to be dropped, and then called `slice()` to remove that percentage of characters from the start.

This approach would only work if the entire string is made up of characters with a uniform number of bytes (i.e. all one byte, or all two bytes, etc.). When dealing with a string that contains characters with varying byte sizes, there is no guarantee that removing a certain number of characters from the end will reduce the size by a proportional amount. For instance, if the string contains a large number of multi-byte characters at the start, and mostly single-byte characters at the end, then removing the final 25% of characters will not reduce the overall byte size by 25%.

Rather than try to write code that will truncate a string properly while respecting multi-byte character boundaries, I'm turning to the [`truncate-utf8-bytes`](https://www.npmjs.com/package/truncate-utf8-bytes) module, which seems to work as expected.


algolia.js
```diff
@@ -13,36 +13,44 @@
...
+const {sizeOfJSONInBytes} = require('./shared/sizeOfJSONInBytes');
 const {defaultLocale, supportedLocales} = require('./shared/locale');
 
 const maxChunkSizeInBytes = 10000000; // 10,000,000
 const maxChunkSizeInMB = maxChunkSizeInBytes / 1000000;
-const maxItemSizeInBytes = 7000; // 7,000
+// This limit is set by our current Algolia plan (with some headroom).
+// The restriction applies to the size of the JSON serialization of each item.
+// See https://support.algolia.com/hc/en-us/articles/4406981897617-Is-there-a-size-limit-for-my-index-records-#:~:text=The%20record%20size%20limit%20is%20based%20on%20the%20size%20of%20this%20final%20JSON%20file.
+const maxItemSizeInBytes = 9000;
```
jeffposnick, line 32: This was previously set to 7000, even though the API has a max of ~10000, because the old approach did not account for JSON overheard and was not always accurate in its truncation. I'm setting it much higher, so we should end up with more indexed content.

(I still want to leave a little wiggle-room here to account for any extra overhead in the API calls, or edge cases with byte boundaries.)

devnook (approved):

PR #7407, opened 2022-02-25
Fix layout shift by setting a max width on sidebar

### LINKED ISSUE #7345: Courses section has big layout shifts
**Describe the bug**
Visit a page like https://web.dev/learn/css/grid/

Notice that when the page loads, the sidebar appears to initially take up the entire page and then quickly shifts into the correct position. If you run the performance panel you'll see the layout shifts getting flagged.

image

These big shifts are bad for CWV and could cause the course pages to get down-ranked on Search.

hankchizljaw:
Fixes #7345

This layout shift appeared to only be happening intermittently, but setting a max-width on the sidebar should prevent it in the future.

### DIFF
--- a/src/scss/pages/_course.scss
+++ b/src/scss/pages/_course.scss
@@ -150,6 +150,14 @@ $course-side-vert-space: 20px;
 
 /// RESPONSIVE STYLES
 .course {
+  @include media-query('md') {
+    &__sidebar {
+      max-width: var(--sidebar-target-width);
+    }
+  }
+}
+
+.course {
   @include media-query('lg') {
     --web-navigation-drawer-width: 100%;
 


jeffposnick (approved):
I have had a hard time reproducing the original layout shift related to the sidebar on production web.dev, and I notice that even with this PR, there's still a small layout shift related to the `<input>` on the audio player:

Screen Shot 2022-02-25 at 10 49 30 AM

Absent being able to reproduce the original issue, PR does seem fine to me in general, as it theoretically should address what @robdodson originally reported as the larger layout shift.

hankchizljaw:
Yeh I think that one is the audio buffering and subsequently, rendering a progress bar update. No idea how to fix that one tbh.

PR #7408, opened 2022-02-25
Fix captcha alignment

hankchizljaw:
The captcha element on the signup was out of alignment, so this PR fixes that.

### DIFF
--- a/src/scss/web-components/_web-subscribe.scss
+++ b/src/scss/web-components/_web-subscribe.scss
@@ -23,5 +23,7 @@ web-subscribe {
 
   .g-recaptcha {
     height: 78px;
+    width: max-content;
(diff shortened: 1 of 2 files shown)


jeffposnick (commented):
Forgive me if I'm mistaken, as I'm fairly new to this project, but it looks like we're using the `all-center` class elsewhere on the page to center similar elements.

Would this same effect be achievable by just adding the `all-center` class to the `<div>` that currently only has the `g-recaptcha` class?

hankchizljaw:
Ah I forgot about the `all-center` (and I added it 🥴). I've just added that directly to the captcha element which has the same effect as the above.

jeffposnick (approved):

PR #7414, opened 2022-02-28
Update index.md (optimize-lcp)

kanchi240:
The HTML parser will NOT pause if it encounters any external stylesheets (<link rel="stylesheet">)

Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/src/site/content/en/blog/optimize-lcp/index.md
+++ b/src/site/content/en/blog/optimize-lcp/index.md
@@ -199,9 +199,8 @@ Exchanges](/signed-exchanges).
 ## Render blocking JavaScript and CSS {: #render-blocking-resources }
 
 Before a browser can render any content, it needs to parse HTML markup into a
-DOM tree. The HTML parser will pause if it encounters any external stylesheets
-(`<link rel="stylesheet">`) or synchronous JavaScript tags (`<script
-src="main.js">`).
+DOM tree. The HTML parser will pause if it encounters any synchronous JavaScript tags (`<script
+src="main.js">`), which may also be blocked on css processing.
 
 Scripts and stylesheets are both render blocking resources which delay FCP, and
 consequently LCP. Defer any non-critical JavaScript and CSS to speed up loading


jeffposnick:
CC: @jakearchibald @khempenius and @philipwalton, who are all experienced in this area, to confirm that this change is accurate.

src/site/content/en/blog/optimize-lcp/index.md
```diff
@@ -199,9 +199,8 @@ Exchanges](/signed-exchanges).
 ## Render blocking JavaScript and CSS {: #render-blocking-resources }
 
 Before a browser can render any content, it needs to parse HTML markup into a
-DOM tree. The HTML parser will pause if it encounters any external stylesheets
-(`<link rel="stylesheet">`) or synchronous JavaScript tags (`<script
-src="main.js">`).
+DOM tree. The HTML parser will pause if it encounters any synchronous JavaScript tags (`<script
+src="main.js">`), which may also be blocked on css processing.
```
jeffposnick, line 203: "css" should be capitalized as "CSS."

philipwalton: I believe it's correct that CSS is not parser-blocking, but this sentence still doesn't quite read right to me. Including "which may also be blocked on css processing" implies that CSS is parser-blocking, right?

malchata: I think maybe this should be rephrased as:

"Before a browser can render any content, it needs to parse HTML markup into a DOM tree. The renderer will pause if it encounters any external stylesheets (`<link rel="stylesheet">`), and the parser will pause if it encounters synchronous `<script>` tags."

Or is this still technically incorrect? I'm separating "renderer" from "parser" here, which feels like where the lack of clarity might be coming from?

kanchi240: Update:

Before a browser can render any content, it needs to parse HTML markup into a DOM tree and parse CSS into a CSSOM tree.
* By default CSS is render blocking resource. 
* The HTML parser will pause if it encounters any external synchronous JavaScript tags (<script src="main.js">).
* If the browser hasn't finished downloading and building the CSSOM tree, the browser delays script execution and DOM construction until it has finished the CSSOM construction.

So the scripts and stylesheets are both render blocking resources which delay FCP, and consequently LCP. Defer any non-critical JavaScript and CSS to speed up loading of the main content of your web page.

PR #7426, opened 2022-03-02
Fix: changed the color from $PRIMARY_TEXT_COLOR  to var(--color-mid-text) to enhance the contrast in the Dark theme

### LINKED ISSUE #7424: the answer section has a low contrast ratio in "Check your understanding" section in the Dark theme
I think there is a low contrast ratio of `1.06` in the answer that appears after pressing **CHECK** in the **Check your understanding** section when the **Dark theme** is enabled.

1. Go to https://web.dev/labels-and-text-alternatives/#label-buttons-and-links.
2. Scroll down to the end of the article to the **Check you understanding** section.
3. Click on **CHECK**. _(currently existing in SAMPLE 1)_
6. Notice the contrast ratio between the font color and the div background-color.

I think it's expected to have a higher ratio of contrast.

contrast-ratio-issue

gitmibrahim:
Fixes #7424

Changed from `color: $PRIMARY_TEXT_COLOR;` to `color: var(--color-mid-text);`

- enhanced the contrast ratio from `1.06` to `10.14`.
- it's working well in both Dark and Light themes.

### DIFF
--- a/src/lib/components/ResponseThinkAndCheck/_styles.scss
+++ b/src/lib/components/ResponseThinkAndCheck/_styles.scss
@@ -4,7 +4,7 @@
   @include w-body-text();
   background: rgba($SUCCESS_COLOR, .08);
   border-radius: $GLOBAL_RADIUS;
-  color: $PRIMARY_TEXT_COLOR;
+  color: var(--color-mid-text);
   display: block;
   line-height: 1.778 !important; // Override w-body-text mixin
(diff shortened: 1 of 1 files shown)


jeffposnick (approved):
This seems like a reasonable change to me.

I can't say with 100% certainty that that's the "right" color to use, but since it's an improvement over the existing styles, I'm inclined to merge it.

PR #7429, opened 2022-03-02
MigratingAuthors

simgobb:
Added all the authors of the articles labelled as Web.dev in the ContentMigration Sheet

@rachelandrew

### DIFF
--- a/src/site/_data/i18n/authors.yml
+++ b/src/site/_data/i18n/authors.yml
@@ -1,3 +1,81 @@
+paulkinlan:
+  title:
+    en: Paul Kinlan
+  description:
+    en: Paul is a Developer Advocate
+
+dalecurtis:
+  title:
+    en: Dale Curtis
+  description:
+    en: Dale is a Senior Software Engineer
+
+mscales:
+  title:
+    en: Mat Scales
+  description:
+    en: Mat is a contributor to WebFundamentals
+
+ewagasperowicz:
+  title:
+    en: Ewa Gasperowicz
+  description:
+    en: Ewa is a contributor to WebFundamentals
+
+alexdanilo:
+  title:
+    en: Alex Danilo
+  description:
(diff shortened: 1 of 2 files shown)


rachelandrew:
@simgobb this is web.dev, you mentioned DCC. Can you confirm these are authors for web.dev articles not DCC?

simgobb:
@rachelandrew Yes, sorry. I confirm that these are the Web Dev ones I got from the migration sheet. 

I don't think we have any content ready yet as we focused on DCC, but I thought it would be good to sort this side of things as well to avoid the issues we had with DCC.

rachelandrew:
@simgobb it looks like these are going to need images. I think that's why it is not building. You can upload the ones used on WebFundamentals.

src/site/_data/i18n/authors.yml
```diff
@@ -1,3 +1,81 @@
...
+  title:
+    en: Dale Curtis
+  description:
+    en: Dale is a Senior Software Engineer
+
+mscales:
+  title:
+    en: Mat Scales
+  description:
+    en: Mat is a contributor to WebFundamentals
```
rachelandrew, line 17:
```suggestion
    en: Mat is a contributor to web.dev
```

src/site/_data/i18n/authors.yml
```diff
@@ -1,3 +1,81 @@
...
+
+ewagasperowicz:
+  title:
+    en: Ewa Gasperowicz
+  description:
+    en: Ewa is a contributor to WebFundamentals
```
rachelandrew, line 23:
```suggestion
    en: Ewa is a contributor to web.dev
```

rachelandrew (changes requested):
I just realised that a lot of these have a bio of "... is a contributor to WebFundamentals" can we update all of these to web.dev?

rachelandrew (approved):

PR #7455, opened 2022-03-07
Update the BrowserCompat component to filter against stable browsers

### LINKED ISSUE #7447: BrowserCompat widget is displaying pre-release browsers
At some point BCD has started listing pre-release versions of browsers, therefore making it look as if things are available if you don't happen to know the version number that is stable. This makes it look as if things are usable when they are not.

I am wondering if there is anything we can do with our use of the Compat component to filter against stable browsers. The data does include if a browser is current or beta. This is the data for Safari https://github.com/mdn/browser-compat-data/blob/main/browsers/safari.json

You can currently see the problem on the What's New post for Feb releases, it looks like things are available, unless you know that Safari 15.4 is a beta: https://web.dev/web-platform-02-2022/

tannerdolby:
Fixes #7447 

- Introduce a `ReleaseStatement` to lookup `status` info for a given browser release version within compatAria, compatVersion, compatProperty functions.
- Perform a `status` check to determine if the browser release is "current" or not when deciding whether to display the version number or "x" in the compat* functions.

Notes:
The BrowserCompat.js component before my proposed changes was mainly relying on [`SimpleSupportStatement`](https://github.com/mdn/browser-compat-data/blob/312dabab943fe43f7caf7466224d844fa8c4aace/types.d.ts#L134) which provided `version_added`, `version_removed` etc. But this statement type did not include the fields we needed to check the browser version status, as this information was stored in [`ReleaseStatement`](https://github.com/mdn/browser-compat-data/blob/312dabab943fe43f7caf7466224d844fa8c4aace/types.d.ts#L72).

Since we needed to get ahold of a `ReleaseStatement` along with the `SimpleSupportStatement`, I added a bit of logic to do a lookup for a given browser release version provided a `support` object to get the `version_added`, and then check if that release version has a "current" status (not beta). This seems to do the trick for filtering only stable browser release versions, but there is definitely room for improvement for more sturdy filtering based on the possible [`status`](https://github.com/mdn/browser-compat-data/blob/312dabab943fe43f7caf7466224d844fa8c4aace/types.d.ts#L95-L116) values in `ReleaseStatement` object. 

For example, the [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter#browser_compatibility) property displays a browser release version "94" which corresponds to a "[retired](https://github.com/mdn/browser-compat-data/blob/312dabab943fe43f7caf7466224d844fa8c4aace/types.d.ts#L99)" `status` meaning the release is no longer supported, but Chrome 94 does support `scrollbar-gutter` according to [caniuse/scrollbar-gutter](https://caniuse.com/?search=scrollbar-gutter), so this case was something I wanted to mention.

### DIFF
--- a/src/site/_includes/components/BrowserCompat.js
+++ b/src/site/_includes/components/BrowserCompat.js
@@ -30,7 +30,7 @@ const compatVersion = (support, release) => {
   } else if (
     !support.version_removed &&
     typeof support.version_added === 'string' &&
-    release.status === 'current'
+    (release.status === 'current' || release.status === 'retired')
   ) {
     return support.version_added;
   } else {
@@ -49,7 +49,7 @@ const compatProperty = (support, release) => {
   } else if (
     !support.version_removed &&
     support.version_added &&
-    release.status === 'current'
+    (release.status === 'current' || release.status === 'retired')
   ) {
     return 'yes';
   } else {
@@ -69,7 +69,7 @@ const compatAria = (support, release, locale) => {
   } else if (
     !support.version_removed &&
     support.version_added &&
-    release.status === 'current'
+    (release.status === 'current' || release.status === 'retired')
   ) {
     return i18n('i18n.browser_compat.supported', locale);
   } else {
--- a/src/site/_includes/components/BrowserCompat.js
+++ b/src/site/_includes/components/BrowserCompat.js
@@ -15,19 +15,22 @@
  */
 const bcd = require('../../_utils/browserCompat');
 const {i18n, getLocaleFromPath} = require('../../_filters/i18n');
+const defaultBcd = require('@mdn/browser-compat-data');
 
 const browsers = ['chrome', 'firefox', 'edge', 'safari'];
 
 /**
  * @param {import('@mdn/browser-compat-data/types').SimpleSupportStatement} support
+ * @param {import('@mdn/browser-compat-data/types').ReleaseStatement} release
  * @returns {string}
  */
-const compatVersion = (support) => {
+const compatVersion = (support, release) => {
   if (!support.version_removed && support.version_added === 'preview') {
     return '\uD83D\uDC41'; // 👁 eye
   } else if (
     !support.version_removed &&
-    typeof support.version_added === 'string'
+    typeof support.version_added === 'string' &&
+    release.status === 'current'
   ) {
     return support.version_added;
   } else {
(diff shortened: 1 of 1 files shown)


jeffposnick (changes requested):
Thank you for the contribution!

Checking for `release.status === 'current'` does not appear to give the exact behavior we want here, since it excludes features that were added in an earlier browser release.

I.e. Firefox 94 is not "current," but features that were added in Firefox 94 are widely available.

The production rendering for something added in Firefox 94 looks good:

Screen Shot 2022-03-07 at 10 19 28 AM

With this PR, the feature added in Firefox 94 incorrectly ends up with an `X`:

Screen Shot 2022-03-07 at 10 17 28 AM

I think you need to check for `release.status === 'current' || release.status === 'retired'` to get the expected behavior.

tannerdolby:
Hello @jeffposnick! No problem.

> Checking for release.status === 'current' does not appear to give the exact behavior we want here, since it excludes features that were added in an earlier browser release.

Yeah, this is what I was unsure about. At first, I checked "current" || "retired" as I noticed there were more small "x"s than the original page when only filtering against "current" releases, but ultimately even if a browser is retired it still can have widely used features so filtering against both is a better option.

I will add the `release.status === 'current' || release.status === 'retired'` check like you mentioned. Thanks for review!

jeffposnick (approved):
This logic seems good to me.

I'll defer to @rachelandrew for the final review, since she opened the issue to begin with.

tannerdolby:
Ok great. Thanks @jeffposnick

rachelandrew (approved):

tannerdolby:
@rachelandrew Thanks for reviewing/merging this. I see the PR has been reverted here, was there an issue breaking the build? I'm happy to fix any issues to get this feature request landed.

rachelandrew:
Yes I was getting the status error message locally, and it broke the production build. As I couldn't see why I reverted it pending someone from the infra team investigating,

tannerdolby:
Ok, thank you for the heads up. Let me know if I can do anything to help.

devnook:
Hi @tannerdolby thanks for your contribution! 

This PR problem was that it didn't account for non-standard data in MDN BCD, e.g. here: https://github.com/mdn/browser-compat-data/blob/main/html/elements/button.json ("version_added: true"). To fix this, we'd need to add logic for a case where release data (and therefore its status) cannot be found. 
We'd need to add this additional logic to the test file too, to avoid such issues in future (https://github.com/GoogleChrome/web.dev/blob/main/test/unit/src/site/_includes/components/BrowserCompat.js)
Finally, we might consider adding this logic in _data instead (not sure if it makes sense though, just a suggestion).

I cannot commit these changes to your fork, so please let me know if you'd like to fix these issues and submit a new PR, or if you'd prefer that I do it instead (in a new PR too).

@rachelandrew @jeffposnick This PR broke our tests, but it was not caught in time because the $_presubmit label was not added before merging. Please add $_presubmit label to trigger tests before merging a js logic PR.

PR #7460, opened 2022-03-07
add utm_source to security link

topaztee:
Minor fix, adds a utm source to a link


jpmedley:
@devnook This is a bit out of my area. Isn't the information in this change already in the request header?

PR #7470, opened 2022-03-08
Use 'en' codelabs on translated pages

### LINKED ISSUE #7466: CodelabsCallout returns undefined and it was printed
**Describe the bug**

image

> https://web.dev/i18n/ja/http-cache/

It seems that it happens for translated pages.

https://github.com/GoogleChrome/web.dev/blob/abeb34cbb879fbca93b4765989ebde53524f3e5d/src/site/_includes/components/CodelabsCallout.js#L37

If I fix this line to return `''`(empty string, also matches function typings), `undefined` has gone but empty `div` still remains.

**To Reproduce**

1. Go to https://web.dev/i18n/ja/http-cache/
2. Scroll down to bottom
3. See `undefined` 

I found the same issue in 

- https://web.dev/i18n/ja/serve-images-webp/
- https://web.dev/i18n/es/sign-up-form-best-practices/
- etc...

**Expected behavior**

Render Codelabs component properly or render nothing.

jeffposnick:
Fixes #7466

I'm working under the assumption that it's better to show English-language codelabs (and the accompanying English-language description) instead of showing nothing on a localized page where there's no language-specific codelab available.

If you disagree, then instead of falling back to checking for `en`, we can just return `''` if there's no language-specific codelab.

This ends up [looking like](https://deploy-preview-7470--web-dev-staging.netlify.app/i18n/ja/http-cache/):

Screen Shot 2022-03-08 at 12 06 51 PM

### DIFF
--- a/src/site/_includes/components/CodelabsCallout.js
+++ b/src/site/_includes/components/CodelabsCallout.js
@@ -15,10 +15,10 @@
  */
 
 const {html} = require('common-tags');
+
+const {defaultLanguage} = require('../../../lib/utils/language');
 const {findByUrl} = require('../../_filters/find-by-url');
 const md = require('../../_filters/md');
-
-const DEFAULT_LANGUAGE_CODE = 'en';
 
 /**
  * Find matching collection items given a language code.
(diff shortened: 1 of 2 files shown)


devnook (approved):

jeffposnick:
I realized there was a test case to update when running the presubmit, so I just pushed a commit to account for that (and cover the updated logic in a new test).

Issue #7339, opened 2022-02-10
11ty build speed investigation


jeffposnick:
@devnook mentioned (and I noticed myself...) that the `npm run eleventy` step of the build process takes a bit of time, even for subsequent builds where some degree of caching is going on.

I ran `npm --node-options --inspect-brk run eleventy` and then [attached Chrome DevTool's profiler](https://medium.com/@paul_irish/debugging-node-js-nightlies-with-chrome-devtools-7c4a1b95ae27) to take a peak at whether there were any particular hotspots. I'll add more to this thread as I find them.

## lodash/isPlainObject

One area that immediately stood out is calls to [`lodash/isPlainObject`](https://github.com/lodash/lodash/blob/2da024c3b4f9947a48517639de7560457cd4ec6c/isPlainObject.js), inside of 11ty's [`Merge.js`](https://github.com/11ty/eleventy/blob/08fe54e03da025cda24c4b25b5b8a8d0007b4689/src/Util/Merge.js#L1). I used [`module-alias`](https://github.com/ilearnio/module-alias) to swap that out for the implementation in [`typechecker`](https://github.com/bevry/typechecker) (see https://github.com/GoogleChrome/web.dev/commit/d41833e5c711a52e4b85ea0db1911a062ad8b375), and local build times improved _substantially_. The following are comparisons of the build with a warm cache on my M1 Macbook Pro.

Before:

``` sh
$ npm run eleventy 

> web.dev@3.0.0 eleventy /Users/.../git/web.dev
> eleventy --quiet

Eleventy is building, please wait…
Wrote 3372 files in 55.20 seconds (16.4ms each, v0.12.1)
```

After:

``` sh
$ npm --node-options '--require "module-alias/register"' run eleventy

> web.dev@3.0.0 eleventy /Users/.../git/web.dev
> eleventy --quiet

Eleventy is building, please wait…
Wrote 3372 files in 40.55 seconds (12.0ms each, v0.12.1)
```

You can reproduce them by checking out the https://github.com/GoogleChrome/web.dev/tree/isPlainObject-patch branch and running `npm --node-options '--require "module-alias/register"' run eleventy`

I know that we're not on the latest 11ty release, but they seem to still be using `lodash/isPlainObject` in v1.0.0, so I will follow up on the 11ty project to see if they would consider officially switching off of `lodash/isPlainObject`.

## luxon/DateTime

[...]

PR #7473, opened 2022-03-08
Use a Rust-based HTML minifier

jeffposnick:
This is part of the investigation at https://github.com/GoogleChrome/web.dev/issues/7339

Locally, this does improve our production build speed quite substantially, taking it from ~340 seconds to ~245 seconds on my M1 Macbook Pro.

The same set of configuration options does not carry over from `html-minifier` to `@minify-html/js`, but the defaults used by `@minify-html/js` seem reasonable enough, and I was not able to see any issues in a local spot-check of the minified content.

This only changes our HTML minification, and does not change our JS or CSS processing.


package.json
```diff
@@ -54,14 +54,15 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
+    "@11ty/eleventy": "^0.12.1",
```
jeffposnick, line 57: Some dependencies are re-arranged in here because `npm`'s idea of alphabetical order apparently differs from the sorting that I used when I modified `package.json` last.

devnook (approved):

PR #7477, opened 2022-03-09
Sort language list to have Latin script languages appear first and alphabetically sorted, while maintaining relative ordering

### LINKED ISSUE #7430: Listing translations alphabetically by underlying language code may not be best.
The top of web.dev articles lists all all available translations of the article. The order appears random (see the image), but they seem to be ordered alphabetically by language code (except or 'en', which always appears to be last). This is not a problem when the number of translations is small. It could come across as a difficult mess if the list of translations is large.

image

Languages should be grouped by script. Within a script group, languages should be sorted in a way that is appropriate for that script. For example, languages rendered in the Latin script should be sorted alphabetically. 

Off hand, I can't think of any.

tannerdolby:
Fixes #7430 

- Sort the `langhrefs` array with structure ['es', '/i18n/es/some-page', ...] by languages that are within the Latin script group.
- Sort the `langhrefs` array once more to ensure the latin script languages are sorted alphabetically.
- Non-latin script languages will keep their relative ordering
- Add `scriptGroups` object representing language script groups.

Page: https://web.dev/splash-screen/

**Before:**
Available in: [Español](https://web.dev/i18n/es/splash-screen/), [한국어](https://web.dev/i18n/ko/splash-screen/), [Português](https://web.dev/i18n/pt/splash-screen/), [中文](https://web.dev/i18n/zh/splash-screen/), [English](https://web.dev/i18n/en/splash-screen/)

**After:**
Available in: [English](http://localhost:8080/i18n/en/splash-screen/), [Español](http://localhost:8080/i18n/es/splash-screen/), [Português](http://localhost:8080/i18n/pt/splash-screen/), [한국어](http://localhost:8080/i18n/ko/splash-screen/), [中文](http://localhost:8080/i18n/zh/splash-screen/)

cc @jpmedley

### DIFF
--- a/src/lib/utils/language.js
+++ b/src/lib/utils/language.js
@@ -22,6 +22,20 @@ const languageNames = {
 };
 
 /**
+ * A map of script groups to their language codes.
+ * @const
+ */
+const scriptGroups = {
+  latin: ['cs', 'de', 'en', 'es', 'fr', 'it', 'no', 'pl', 'pt', 'sv'],
+  cyrillic: ['bg', 'ru'],
+  nonAlphabetic: ['ta', 'zh'],
+  arabic: 'fa',
+  devanagari: 'hi',
+  kanji: 'ja',
+  hangul: 'ko',
+};
+
+/**
  * A default language for the site.
  * @const
  */
@@ -47,4 +61,5 @@ module.exports = {
   defaultLanguage,
   isValidLanguage,
   supportedLanguages,
+  scriptGroups,
 };
--- a/src/lib/utils/language.js
+++ b/src/lib/utils/language.js
@@ -22,18 +22,23 @@ const languageNames = {
 };
 
 /**
- * A map of script groups to their language codes.
+ * When we display a list of translations available for a given page,
+ * use this ordering for the names of the languages.
+ * See https://github.com/GoogleChrome/web.dev/issues/7430
  * @const
  */
-const scriptGroups = {
-  latin: ['cs', 'de', 'en', 'es', 'fr', 'it', 'no', 'pl', 'pt', 'sv'],
-  cyrillic: ['bg', 'ru'],
-  nonAlphabetic: ['ta', 'zh'],
-  arabic: 'fa',
-  devanagari: 'hi',
-  kanji: 'ja',
-  hangul: 'ko',
-};
+const languageOrdering = [
+  'de',
+  'en',
+  'es',
+  'fr',
+  'pl',
+  'pt',
+  'ru',
+  'zh',
+  'ja',
+  'ko',
+];
 
 /**
  * A default language for the site.
@@ -57,9 +62,9 @@ function isValidLanguage(lang) {
 }
 
 module.exports = {
-  languageNames,
   defaultLanguage,
   isValidLanguage,
+  languageNames,
+  languageOrdering,
   supportedLanguages,
-  scriptGroups,
 };
--- a/src/site/_filters/urls.js
+++ b/src/site/_filters/urls.js
@@ -42,20 +42,34 @@ const getRelativePath = (url, pathPrefix) => path.relative(pathPrefix, url);
 /**
  * Find i18n equivalents of the current url and check if they exist.
  * @param {string} url Url of the original article.
- * @returns {Array} Array of language and url pairs.
+ * @returns {Array<Array<string, string>>} Array of language and url pairs.
  */
 const getTranslatedUrls = (url) => {
   url = getDefaultUrl(url);
   return (
     supportedLocales
-      .map((locale) => [locale, path.join('/', 'i18n', locale, url)])
+      .map((locale) => [locale, getLocaleSpecificUrl(locale, url)])
       // Filter out i18n urls that do not have an existing translated file.
-      .filter((langhref) => !!findByUrl(langhref[1]))
+      .filter((langHref) => !!findByUrl(langHref[1]))
   );
+};
+
+/**
+ * Generate a URL that includes the locale-specific prefix.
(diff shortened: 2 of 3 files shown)


src/site/_includes/components/LanguageList.js
```diff
@@ -16,27 +16,41 @@
...
+        }
+      })
+      .sort((a, b) => {
+        if (
+          scriptGroups.latin.includes(a[0]) &&
+          scriptGroups.latin.includes(b[0])
+        ) {
+          return a[0].localeCompare(b[0]);
+        }
+      })
```
tannerdolby, line 45: I tried to get around using two `.sort()` calls here, but I couldn't think of how to combine the two into a single call. The array needed to be sorted by latin script languages (so they appear at the front) and then if the languages are indeed latin script, I wanted to sort them alphabetically. This would allow the non latin script languages to maintain their relative ordering in the list.

src/site/_includes/components/LanguageList.js
```diff
@@ -16,27 +16,41 @@
...
-    .filter((langhref) => langhref[0] !== lang)
-    .map((langhref) => {
-      const href = langhref[1];
-      return `<a class="w-post-signpost__link"
-        translate="no"
-        lang="${langhref[0]}"
-        href="${href}">
-      ${languageNames[langhref[0]]}</a>`;
-    });
+  let langhrefs = getTranslatedUrls(url).filter(
```
tannerdolby, line 22: I also didn't want to have to change this from `const` to `let`, but since the previous logic was adding the "en" data to `langhrefs` after the initial .filter() and .map() of the getTransledUrls(url) array, it wasn't apart of the sorting when I added the sort logic here initially. So my thought process was to keep the original .filter() call and then do a quick length check before adding the "en" data to `langhrefs` and after this begin the sorting before finally using .map() to format the strings into HTML tags.

jeffposnick (commented):
Thanks a ton for the contribution!

I did have a few ideas for how this approach could be tweaked. Would you be open to making this a collaborative PR, with me pushing a commit to your branch? That might end up being a faster collaboration than if I just requested changes, and additionally, I found a few tweaks to make in utility files that were not modified in your original PR.

tannerdolby:
@jeffposnick You're welcome! Thanks for the review.

Yeah, I'm totally fine with this being a collaborative PR. Go ahead and push any commits to this branch and I will sync with your changes to view locally.

src/lib/utils/language.js
```diff
@@ -21,6 +21,25 @@ const languageNames = {
   fr: 'Français',
 };
 
+/**
+ * When we display a list of translations available for a given page,
+ * use this ordering for the names of the languages.
+ * See https://github.com/GoogleChrome/web.dev/issues/7430
+ * @const
+ */
+const languageOrdering = [
```
jeffposnick, line 30: I think this is a cleaner approach, rather than trying to group the languages into script families and then sort within the script families (as in @tannerdolby's original PR).

The ordering in this array reflects the ordering that we'd like the supported locale translations to appear in.

If we add more supported locales in the future, we'd place the two-letter language code in the appropriate sport in this array.

(If there's a locale that we end up supporting but we forget to add it to this array, the way the `sort()` works, it will just end up being sorted to the front of the list, so it fails-safe.)

jeffposnick:
This builds on @tannerdolby's original set of changes, and also introduces the usage of `Intl`'s `listFormat()`, so that we have a locale-specific alternative to `join(', ')`.

This leads to, e.g., on http://localhost:8080/samesite-cookies-explained/:

Screen Shot 2022-03-09 at 3 34 50 PM

and the Chinese version at http://localhost:8080/i18n/zh/samesite-cookies-explained/:

http://localhost:8080/i18n/zh/samesite-cookies-explained/:

Screen Shot 2022-03-09 at 3 36 29 PM

src/site/_includes/components/LanguageList.js
```diff
@@ -13,35 +13,54 @@
...
   const availableIn = i18n('i18n.post.available_in', lang);
-  return langhrefs.length
-    ? `<span class="w-post-signpost__title">${availableIn}:
-    ${langhrefs.join(', ')}</span>`
-    : '';
+  const listFormat = new Intl.ListFormat(lang);
+
+  return `<span class="w-post-signpost__title">
+    ${availableIn}:
+    ${listFormat.format(languageLinks)}
```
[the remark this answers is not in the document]
tannerdolby: Yeah, this is really good. I forgot how useful [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) would be for ensuring the translation list was localized with the expected separators and conjunction.

src/site/_filters/urls.js
```diff
@@ -42,20 +42,34 @@ const getRelativePath = (url, pathPrefix) => path.relative(pathPrefix, url);
...
  * @param {string} url Url of the original article.
- * @returns {Array} Array of language and url pairs.
+ * @returns {Array<Array<string, string>>} Array of language and url pairs.
  */
 const getTranslatedUrls = (url) => {
   url = getDefaultUrl(url);
   return (
     supportedLocales
-      .map((locale) => [locale, path.join('/', 'i18n', locale, url)])
+      .map((locale) => [locale, getLocaleSpecificUrl(locale, url)])
```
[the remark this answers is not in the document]
tannerdolby: Agreed! This is much safer than path.join(). Nice one.

matthiasrohmer (approved):
Tested locally running a full production build with these changes put on `main`. Works great! Thanks!

tannerdolby:
@matthiasrohmer Awesome! I'm glad it's working as expected. No problem!

PR #7496, opened 2022-03-14
Tweak media queries section for accessible tap targets

patrickhlauke:
Instead of promoting use of `pointer`, which will only catch cases where the browser/OS deems the touchscreen to be the *primary* input (thus leaving out touch-enabled laptops, where mouse/trackpad is still considered the *primary*), use `any-pointer` instead.

This also pads out the throwaway line about testing for width, which came a bit out of nowhere at the end there, and gives it some context.


rachelandrew (approved):
Thanks!

patrickhlauke:
ah, wonder in hindsight if i should have also updated the front matter to change `updated` and possibly squeeze myself in there on the authors list...

PR #7502, opened 2022-03-15
Add location of LH client to /measure/ results

### LINKED ISSUE #4321: Feature Request - Display server location
When we compare web.dev lighthouse results to those from GT Metrix / webpagetest.org we always have certain discrepancies between the scores which we believe can be attributed to the geographically distant web.dev's test server and the target website's server. (i.e. My website server could be located in India and web.dev could be running the test from the US).

Presently the location of the test server is not shown to the end-user in web.dev. The test server location could help us disregard the warnings about high server response time, high LCP (related to loading above the fold banner images), etc if the website is only targetting a specific country.

For the time being, just having visibility into the test server location should help.
A long term goal could be to provide a way to select a data center to run the web.dev test from.

jeffposnick:
Fixes #4321 

This just hardcodes the "central United States" as the location, because the underlying infrastructure isn't multi-region, and I'm doubtful that we'll change it anytime soon.

Happy to tweak the exact wording. It looks like:

Screen Shot 2022-03-15 at 1 59 12 PM

### DIFF
--- a/src/lib/components/LighthouseViewer/index.js
+++ b/src/lib/components/LighthouseViewer/index.js
@@ -38,10 +38,15 @@ class LighthouseViewer extends BaseStateElement {
           />
         </svg>`
       : '';
+    // Hardcode "central United States" based on the current region for
+    // https://lighthouse-dot-webdotdevsite.appspot.com/. If the AppEngine
+    // instance ever moves, we'll need to update this.
+    // See https://github.com/GoogleChrome/web.dev/issues/4321
     return html`
       <div class="text-size-0">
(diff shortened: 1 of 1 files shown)


jeffposnick:
After talking to @devnook, I realize that https://github.com/GoogleChrome/web.dev/blob/709e47a0b5625e1f8932286011334cf786c96239/src/lib/lighthouse-service.js might not be in use anymore, in favor of https://github.com/GoogleChrome/web.dev/blob/709e47a0b5625e1f8932286011334cf786c96239/src/lib/psi-service.js

I'll revisit this.

PR #7503, opened 2022-03-16
Update front matter for accessible tap targets

patrickhlauke:
Follow-up to https://github.com/GoogleChrome/web.dev/pull/7496

PR #7512, opened 2022-03-16
Adds missing authors data

jeffposnick:
#7503 introduced `patrickhlauke` as an author, but there was no corresponding metadata, which broke our build.

I copied info from @patrickhlauke's GitHub profile to bootstrap things and unbreak us; Patrick, feel free to tweak that info in a follow-up PR if you would prefer to be represented differently.

(We will follow-up separately about improving the processes that allowed this breakage to occur.)

### DIFF
--- a/src/site/_data/authorsData.json
+++ b/src/site/_data/authorsData.json
@@ -618,7 +618,7 @@
     "image": "image/admin/KxIs73yjsWwzEPHQaHEA.jpg"
   },
   "nabeelalshamma": {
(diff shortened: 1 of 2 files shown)


patrickhlauke:
once again, apologies for causing so much trouble for what i thought was just a relatively quick edit.

@jeffposnick purely out of interest: how/where are the author images/avatars actually stored? I see some entries reference `images/...`, but the actual images aren't in the repo itself it seems. is there some automation process that grabs them from somewhere else at build time?

PR #7531, opened 2022-03-18
[DO NOT MERGE] Invalid name to trigger failure

jeffposnick:
This is a test PR to confirm that an invalid name in the `authors` YAML field for a post will now trigger a presubmit failure.


jeffposnick:
Looks like the presubmit fails as expected. We're 👍 with the latest 11ty release.

PR #7546, opened 2022-03-22
Separate ScrollSpy into abstract class

devnook:
The intention is to be able to observe arbitrary elements on the page, and be able to build widgets that react to the changes in scroll position of an article.

### DIFF
--- a/src/lib/components/ScrollSpy/index.js
+++ b/src/lib/components/ScrollSpy/index.js
@@ -17,17 +17,23 @@
 import {BaseElement} from '../BaseElement';
 
 /**
- * Element that adds active classes to its children when the user scrolls to
- * a heading this child points to.
+ * Element that fires a 'viewport-entry' event when user scrolls
+ * an observed element into a focus area. The ibserved elements can be
+ * configured by passing a css selector as an element's param.
  * @extends {BaseElement}
- * @final
  */
-class ScrollSpy extends BaseElement {
+export class ScrollSpy extends BaseElement {
+  static get properties() {
+    return {
+      // CSS selector for selecting items to observe.
+      selector: {type: String, reflect: true},
+    };
+  }
+
   constructor() {
     super();
     this.scrollSpy = this.scrollSpy.bind(this);
-    this.tocActiveClass = 'w-scroll-spy__active';
-    this.tocVisibleClass = 'w-scroll-spy__visible';
+    this.selector = '[data-scrollspy]';
   }
 
   connectedCallback() {
@@ -37,15 +43,12 @@ class ScrollSpy extends BaseElement {
     if (!this.articleContent) {
       return;
     }
-
-    this.headings = this.articleContent.querySelectorAll(
-      'h1[id], h2[id], h3[id]',
-    );
-
+    this.items = this.articleContent.querySelectorAll(this.selector);
     this.observer = new IntersectionObserver(this.scrollSpy, {
+      // Observe the upper parth of the viewport (user's focus area)
       rootMargin: '0px 0px -80% 0px',
     });
-    this.headings.forEach((heading) => this.observer.observe(heading));
+    this.items.forEach((item) => this.observer.observe(item));
   }
 
   disconnectedCallback() {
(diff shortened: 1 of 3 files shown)


src/lib/components/TocScrollSpy/index.js
```diff
@@ -0,0 +1,70 @@
...
+ * Element that adds active classes to its children when the user scrolls to
+ * a heading this child points to.
+ * @extends {ScrollSpy}
+ * @final
+ */
+class TocScrollSpy extends ScrollSpy {
+  constructor() {
+    super();
+    this.selector = this.selector = 'h2[id], h3[id]';
+    this.tocActiveClass = 'w-scroll-spy__active';
```
jeffposnick, line 29: This and the following property act as [static class properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static), and would be cleaner if defined as such, instead of being set as instance properties post-construction.

src/lib/components/TocScrollSpy/index.js
```diff
@@ -0,0 +1,70 @@
...
+    this.tocVisibleClass = 'w-scroll-spy__visible';
+  }
+
+  scrollSpy(headings) {
+    const links = new Map(
```
jeffposnick, line 34: This seems like it might be expensive to call each time `scrollSpy()` is invoked—unless there's a need to support HTML that might have been modified post page load.

Could it be created once in the constructor and then references as `this.links` instead?

src/lib/components/TocScrollSpy/index.js
```diff
@@ -0,0 +1,70 @@
...
+      [...this.querySelectorAll('a')].map((l) => [l.getAttribute('href'), l]),
+    );
+    for (const heading of headings) {
+      const href = `#${heading.target.getAttribute('id')}`;
+      const link = links.get(href);
+
+      if (link) {
+        if (heading.intersectionRatio > 0) {
+          link.classList.add(this.tocVisibleClass);
+          this.previouslyActiveHeading = heading.target.getAttribute('id');
```
jeffposnick, line 44: I'm having a hard time figuring out whether this value is being stored as an instance variable (`this.previouslyActiveHeading`) because it needs to persist across invocations of the `scrollSpy()` callback.

If the value does need to persist across calls to `scrollSpy()`, could you please add a comment explaining how it's being used and what the expectations are?

If the value doesn't need to persist, then it seems like this would make more sense if it were declared as a local variable scoped to the function invocation, instead of an instance variable on `this`.

src/lib/components/TocScrollSpy/index.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+class TocScrollSpy extends ScrollSpy {
+  constructor() {
+    super();
+    this.selector = this.selector = 'h2[id], h3[id]';
+    this.tocActiveClass = 'w-scroll-spy__active';
+    this.tocVisibleClass = 'w-scroll-spy__visible';
+  }
+
+  scrollSpy(headings) {
```
jeffposnick, line 33: This function overrides the `scrollSpy()` function in the base `ScrollSpy` class, right? But it feel like it behaves very differently—instead of firing `'viewport-entry'` events, for instance, it performs DOM lookups and CSS class manipulations.

Instead of overriding `scrollSpy()`, what if the `TocScrollSpy` class was responsible for registering the listener for `'viewport-entry'` events, and reacted to those events with logic similar to what's in this `scrollSpy()` implementation? That's kind of what I assumed you wanted to accomplish via the refactor.

jeffposnick (changes requested):

src/lib/components/ScrollSpy/index.js
```diff
@@ -37,57 +43,28 @@ class ScrollSpy extends BaseElement {
...
-
-      if (!firstVisibleLink && this.previouslyActiveHeading) {
-        const last = this.querySelector(
-          `a[href="#${this.previouslyActiveHeading}"]`,
-        );
-        last.classList.add(this.tocActiveClass);
-      }
-    }
+  scrollSpy(items) {
+    const event = new CustomEvent('viewport-entry', {
```
jeffposnick, line 60: What ends up happening if you have, say, three different `ScrollSpy` instances on a given page, and each fires their own `'viewport-entry'` events? The code that listens for the events might only care about the ones fired from a specific `ScrollSpy` instance, but the listener callback will be invoked for all of them. This introduces a burden on the listener callback to filter out and ignore events that it doesn't care about.

src/lib/components/TocScrollSpy/index.js
```diff
@@ -0,0 +1,70 @@
...
+        link.classList.remove(this.tocActiveClass, this.tocVisibleClass);
+      });
+
+      if (firstVisibleLink) {
+        firstVisibleLink.classList.add(this.tocActiveClass);
+      }
+
+      if (!firstVisibleLink && this.previouslyActiveHeading) {
+        const last = this.querySelector(
+          `a[href="#${this.previouslyActiveHeading}"]`,
```
jeffposnick, line 62: Note that if this PR does get merged, this line might need to be addressed to account for the issue @tunetheweb raised in https://github.com/GoogleChrome/web.dev/issues/8259

PR #7601, opened 2022-03-30
typo:(index.md) Timing-Allow-Origin

robberfree:
from Timing-Origin-Allow to Timing-Allow-Origin


jpmedley (approved):

jpmedley:
@jeffposnick @devnook I've been waiting since yesterday for the pullapprove check to finish. That seems a bit long.

jeffposnick:
The PullApprove check is explicitly waiting on @malchata or @sofiayem, who I believe were chosen at random from the list of eligible reviewers. Someone else in the list of reviewers (@jpmedley) giving an approval _should_ obviously be sufficient to merge this, but it does not look like our PullApprove check allows for that.

We can merge this for the time being, and if it's a bigger issue that you've noticed multiple times, we can revisit the PullApprove logic.

PR #7603, opened 2022-03-30
Fix small text threshold explanation

Hatlen:
In the image above the changed paragraph it says:

> Strive to have >60% of page text >= 12px

I think that is the same as what the text ends up saying after the this commit (?) "Lighthouse flags pages on which 40% or more of the text has a font size smaller
than 12&nbsp;px."

Maybe it would be better to change the explanation to explain this the other way around like "Unless your page have 60% or more text with text size bigger than 12px"?

I think I'm leaning towards it being more intuitive the way that the text ended up in this commit though :)

/ Trond froding
PS: Thanks a lot for all amazing content on web.dev!

### DIFF
--- a/src/site/content/en/lighthouse-seo/font-size/index.md
+++ b/src/site/content/en/lighthouse-seo/font-size/index.md
@@ -22,7 +22,7 @@ with font sizes that are too small to read easily on mobile:
   {% Img src="image/tcFciHGuF3MxnTr1y5ue01OGLBn2/ky2VDt8ZtedleWFLn1Gt.png", alt="Lighthouse audit showing page has illegible font sizes", width="800", height="225" %}
 </figure>
 
-Lighthouse flags pages on which 60% or more of the text has a font size smaller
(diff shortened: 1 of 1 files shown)


malchata:
Sorry for the staleness on this, but I'm not sure if I'm the right person to review this. @brendankenny, would you have bandwidth to verify this for accuracy (and also to make sure the change makes it to the migrated docs on d.c.c.)?

brendankenny (approved):
Wow, thank you for catching this, @Hatlen!

(sorry I missed the ping, @malchata)

brendankenny:
> I think I'm leaning towards it being more intuitive the way that the text ended up in this commit though :)

Yeah, agreed. It reads well, as long as the number is actually correct :)

sofiayem (approved):

sofiayem:
Making sure this makes it to d.c.c.: https://github.com/GoogleChrome/developer.chrome.com/pull/3122

PR #7615, opened 2022-03-31
Removes redirects for new Workbox docs and replaces them with destination links.

malchata:
Since we've been in the swing of moving d.g.c. content and redirecting to web.dev, I wanted to see how many of those redirects were for the Workbox documentation. Turns out there were about 200 links. While the redirects work, they add latency to navigation requests, and we should eventually get to a point where we can replace all redirected links to d.g.c. to their destination URL. This PR does that, but just for Workbox documentation links.


jeffposnick (approved):
I didn't clickthrough to every link, but I eyeballed them, and the new URLs all seem to make sense. Thank you for all the effort! It's also a nice confirmation that we don't have significant gaps in the docs that we've migrated over.

PR #7634, opened 2022-04-01
Fix table markup in Defining CWV metrics article

thinkverse:
Minor fix to the [Defining the Core Web Vitals metrics thresholds](https://web.dev/defining-core-web-vitals-thresholds) article. The first `.table-wrapper` div is missing its closing tag causing the remaining article to not display properly - missing margins, making it difficult to read the article.

Checked the translations and it seems this is only an issue in the English version.

<details><summary>Before fix</summary>

Screen Shot 2022-04-02 at 12 49 54 AM
</details>

<details><summary>After fix</summary>

Screen Shot 2022-04-02 at 12 50 19 AM
</details>


malchata (approved):
Good catch. LGTM!

PR #7639, opened 2022-04-04
Treat missing type attribute as type=text in CSS

### LINKED ISSUE #7638: Broken styling on Developer Newsletter form
**Describe the bug**

On web.dev website, under Developer Newsletter form, first two inputs are missing type on input element, for that reason styling is not applied.

**To Reproduce**
1. Go to https://web.dev/ and scroll down to the form
2. Inspect first two input elements (first name, last name)

**Expected behavior**
- Form should be styled.

**Screenshots**
Screenshot 2022-04-04 at 13 47 52

jeffposnick:
Fixes #7638

Our current HTML minifier removes the `type="text"` attribute from `<input>` elements, since that's the implicit value. (I don't see any [option](https://docs.rs/minify-html/latest/minify_html/struct.Cfg.html) to disable this removal.)

Our current CSS uses the presence of `type="text"` to style certain form fields, including the newsletter signup field. When those fields are unstyled, they can look jarring, especially in Dark Mode.

This PR attempts to address this by finding all the places we have explicit styling for `input[type='text']`, and adds `input:not([type])` as an additional selector to apply the same styling.

I could have also removed the `type="text"` attributes from the original HTML (since they're going to be removed during minification anyway), to keep the development and production HTML similar, but I didn't in this PR. Let me know if you think I should.

It would be good to hear from @hankchizljaw to see if I'm missing something when it comes to these styles, since I believe they worked them originally.

### DIFF
--- a/src/scss/next.scss
+++ b/src/scss/next.scss
@@ -503,6 +503,8 @@ details summary + * {
 
 /// Form fields
 
+// See https://github.com/GoogleChrome/web.dev/issues/7638#issuecomment-1087607518
+input:not([type]),
 input[type='text'],
 input[type='email'],
 input[type='password'],
@@ -526,6 +528,8 @@ textarea {
   @include apply-utility('leading', 'short');
 }
 
+// See https://github.com/GoogleChrome/web.dev/issues/7638#issuecomment-1087607518
+input:not([type]),
 input[type='text'],
 input[type='email'],
 input[type='password'],
--- a/src/scss/next.scss
+++ b/src/scss/next.scss
@@ -503,6 +503,8 @@ details summary + * {
 
 /// Form fields
 
+// Style inputs without a type attribute the same as type="text", since they
+// are semantically equivalent.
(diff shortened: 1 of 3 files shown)


src/scss/next.scss
```diff
@@ -503,6 +503,8 @@ details summary + * {
 
 /// Form fields
 
+// See https://github.com/GoogleChrome/web.dev/issues/7638#issuecomment-1087607518
```
devnook, line 506: I would add an explicit comment that this targets text inputs by default, and then follow by the link

devnook:
I'd consider adding explicit comment.

devnook (approved):

wilburlikesmith:
image

I was a little late in the game, so I turned my screenshot of Inspoect Element into a doodle art for the sake of my time spent already and then took the day off...

src/scss/next.scss
jeffposnick: Done.

PR #7646, opened 2022-04-05
Move some components to webdev-infra

devnook:
+ Use some components from webdev-infra instead of local
+ small css fixes
+ remove instructions from test-post (as deprecated)


src/scss/blocks/_prose.scss
```diff
@@ -50,6 +50,10 @@
     --flow-space: #{get-space('size-2')};
   }
 
+  details + details {
```
jeffposnick, line 53: FWIW, nesting `{% Details %}` seems to result in invalid HTML (at least as per https://github.com/GoogleChrome/web.dev/pull/7648#issuecomment-1088927674). Maybe we shouldn't encourage it until we can find a fix for that?

devnook: This is not for nesting, but for "next sibling", so that a list of details does not have margin between items:

Screenshot 2022-04-05 at 18 30 27

jeffposnick: Sorry! My lack of CSS knowledge is showing.

PR #7648, opened 2022-04-05
Fix for the Aside component

### LINKED ISSUE #7640: Using `Aside` directive in Markdown list context leads to invalid HTML
**Describe the bug**
The articles that are using `Details` directive are broken.

**To Reproduce**
1. Go to https://web.dev/security-headers/ or https://web.dev/strict-csp/
2. Scroll down to the bottom

**Expected behavior**
You can see the bottom of the page, but it's actually terminated at `Details` directive.

jeffposnick:
This PR does three things:

1. Fixes #7640, by ensuring that the string returned by the `{% Aside %}` component won't interact with the parent's Markdown parsing, e.g.  by including multiple newlines, which could cause a `</li>` tag to be potentially added to the middle of the returned string. The approach used here [mimics what's done](https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/components/Label.js) in the `{% Label %}` component.
2. Adds a test case to `/test-post/` to trigger this issue. I can confirm that the test results in incorrect HTML prior to this PR, and correct HTML with this PR.
3. Removes unused code that was leftover from before our design system migration.

### DIFF
--- a/src/site/_includes/components/Aside.js
+++ b/src/site/_includes/components/Aside.js
@@ -15,18 +15,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const {i18n, getLocaleFromPath} = require('../../_filters/i18n');
-const isDesignSystemContext = require('../../../lib/utils/is-design-system-context');
+
+const {html} = require('common-tags');
 const fs = require('fs');
+const md = require('markdown-it')();
 const path = require('path');
 
-/* NOTE: This component is in a transition period to support both new design system contexts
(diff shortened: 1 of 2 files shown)


jeffposnick:
I may have to perform similar cleanup in the `Details` component, because I see that https://deploy-preview-7648--web-dev-staging.netlify.app/strict-csp/ still has some rendering issues due to invalid HTML.

devnook:
@jeffposnick please keep in mind that Details are being moved out of this repo: https://github.com/GoogleChrome/web.dev/pull/7646

devnook (approved):

Issue #7134, opened 2022-01-06
Drag and Drop demo broken in Incognito


pwnall:
**Describe the bug**
The demo on https://web.dev/drag-and-drop/ is broken in Incognito.

**To Reproduce**
Please see the description and steps in https://crbug.com/1273552

**Expected behavior**
Please see the Chromium bug for a description. The Glitch demo should stop using Storage (localStorage), or we should find a way to serve our demos from a 1P context.

**Screenshots**
Please see the Chromium bug for a video demo.

dstmarthe:
Thanks for raising this issue. We are migrating to a new platform and issue tracker. I have moved this to our new tracker at: https://issuetracker.google.com/issues/298799199

PR #7649, opened 2022-04-05
Update drag and drop article.

jpmedley:
@rachelandrew 

Gov'ner,

I reviewed this along with everything else in the the new IA I'm working on. There are a few things you should note.

* I removed the abbreviation 'DnD' and spelled it out. I'm generally opposed to excessive use of abbreviations, and I'm not sure what purpose it serves in this article. And 'DnD' doesn't have good SEO.
* I could not find a reason for testing whether `.preventDefault()` exists in `dragover`. If you know of one, please let me know. I thought maybe it was there since the API isn't supported on Android. I tested the page with remove debugging. The events to don't even fire on Android.
* The ghosting when dragging the Google logo doesn't seem to work any more. I've added a new example, but I'd prefer one that more closely parallels the original. If you have an idea, please let me know.
* After consulting the spec and doing some experiments, it seems that the reason for using `preventDefault()` is broader than what the article originally stated. Again, if you know something that I don't...

### DIFF
--- a/src/site/content/en/blog/drag-and-drop/index.md
+++ b/src/site/content/en/blog/drag-and-drop/index.md
@@ -4,10 +4,10 @@ authors:
   - ericbidelman
   - rachelandrew
 date: 2010-09-30
-updated: 2022-01-28
+updated: 2022-04-05
 description: >
-  The HTML5 Drag and Drop (DnD) API means that we can make almost any element on our page draggable.
-  In this post we’ll explain the basics of Drag and Drop.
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):
Hi Joe - thanks for this, I think these updates are all good. This article was a move from one of the other sites back when I was contracting for Kayce so probably includes some legacy stuff.

rachelandrew:
@jpmedley while we are looking at this, not sure if you had thoughts on this related issue: https://github.com/GoogleChrome/web.dev/issues/7134

PR #7654, opened 2022-04-05
Update DLC article with new launch handler API manifest changes

joycetoh8:
The Launch Handler API has changed slightly, so the Declarative Link Capturing page which has a section on migration to Launch Handler needs to be updated.


tomayac (approved):
LGTM 

(I had made these changes In the [main article](https://web.dev/launch-handler/#the-launch_handler-manifest-member) already, thanks for reflecting the changes here, too.)

jpmedley (approved):

PR #7673, opened 2022-04-09
use extension

jimmywarting:
you talk about esm but you aren't truly using esm, b/c esm is strict about the path
this only works cuz you are using a bundler...

also extensionless file lookups makes a lot of unnecessary stats lookup on the HDD
made a benchmark on a module that included 5000 empty dummy files with and without `.js` extension and the result was that it took like like 0.5 - 1.5s longer to find all those small modules.


malchata (approved):
LGTM. Sorry for the delay.

PR #7719, opened 2022-04-14
Change order of Learn PWA modules.

jpmedley:
This moves the list of capabilities to the end so that we can later replace it with something more robust at a higher level without disrupting the numbering of the parts.

@rachelandrew I marked this p0 because it blocks PRs or remaining parts in Learn PWA


jpmedley:
@rachelandrew I'll be this needs an LGTM from Jeff or Ewa, doesn't it?

jeffposnick (approved):

rachelandrew (approved):

PR #7739, opened 2022-04-18
Info about image CDN and dimensions in handbook

jeffposnick:
This updated recommendation for image dimensions in the handbook matches what I've observed the best practice to be when uploading images to the CDN.

The current wording was a source of confusions for @malchata in https://github.com/GoogleChrome/developer.chrome.com/pull/2611#discussion_r852238823

@devnook, it would be great if you could confirm that this update describes what is in fact the best practice.


malchata (approved):
This guidance looks good to me!

heyawhite (approved):

PR #7761, opened 2022-04-21
Fix broken metric threshold diagrams

philipwalton:
This PR is a slight reformatting of the existing threshold diagrams in the metrics docs, that I think were a regression in redesign (just a guess, not sure when).

The main thing this changes is the `width` and `height` attributes so they're not too small on desktop views.

Ideally our CSS would default to full width (or have an option to do so), but I didn't see anything in the existing CSS, so this was a quick fix.


malchata (approved):
I manually examined these new figures and LGTM. However, we should be considering adding a media query inside the SVG itself for dark mode colors at some point so that we can lay them on a dark background without a light fill color behind it. But we definitely don't need to do it now

Otherwise, LGTM!

PR #7788, opened 2022-04-25
minor improvements to stale-while-revalidate blog

ZYinMD:
Before the change, there is one way to misunderstand the sentence, for example:
```
Cache-Control: max-age=100, stale-while-revalidate=200
```
can be understood as "if the cache age is **covered by** 200, then it can be served from cache. If the cache age is **not covered** by 200, e.g. if it's 250-second-old, then it cannot be served from cache". That's a wrong understanding because the cache is good for 300 seconds.

(Sorry I saw instruction of "don't forget to add the `$-presubmit` label" but I don't know what that means.)


rachelandrew (approved):
Thank you!

PR #7790, opened 2022-04-25
Explicitly set crossorigin="anonymous"

### LINKED ISSUE #7770: Fonts preload without CORS
**Describe the bug**
See console errors in screenshot below.
Also, notice that something seems off about the minified HTML - missing quotes, missing spaces, etc.

**To Reproduce**
1. Go to https://web.dev/optimize-webfont-loading/ (or any other page, really)
2. Open devtools
4. See error in console

**Expected behavior**
No errors in console.

**Screenshots**
Screen Shot 2022-04-21 at 1 28 49 PM

Source looks correct:

https://github.com/GoogleChrome/web.dev/blob/e5e87746c542aa87d7314681f1bb0037035c4e28/src/site/_includes/partials/head.njk#L19-L23

Perhaps something went wrong during the build?

jeffposnick:
Fixes #7770 

This shouldn't be necessary, but it works around https://github.com/wilsonzlin/minify-html/issues/81 in the HTML minifier we switched to.


malchata (approved):
LGTM.

PR #7808, opened 2022-04-28
Add analytics for user color scheme preference

philipwalton:
This PR updates the analytics implementation to capture the user's color scheme preference (i.e. light vs. dark mode) as a custom dimension.

I originally wanted to include other user preferences here as well, but given that the site's analytics initialization code runs in the HTML templates, I didn't want to add too much bloat if we didn't have specific plans to use that information (please let me know if we do).


devnook:
Thanks @philipwalton ! Could you please add a note about our use of Analytics dimensions in the Wiki?

devnook (approved):

PR #7837, opened 2022-05-04
Fix virtualKeyboard geometrychange example

HolgerJeromin:
`const { x, y, width, height } = event.target;`  is incorrect. 
moving to `const { x, y, width, height } = event.target.boundingRect;` 

And switched order of paragraphs so allow referencing to the right one.

### DIFF
--- a/src/site/content/en/blog/virtualkeyboard/index.md
+++ b/src/site/content/en/blog/virtualkeyboard/index.md
@@ -110,7 +110,7 @@ console.log('Virtual keyboard geometry:', x, y, width, height);
 ### Being informed of geometry changes
 
 Whenever the virtual keyboard appears or disappears, the `geometrychange` event is dispatched. The
-event's `target` property contains the virtualKeyboard which (as discussed above) contains the new geometry of the virtual keyboard inset as a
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/virtualkeyboard/index.md
```diff
@@ -95,31 +95,31 @@ a `geometrychange` event if the virtual keyboard previously was shown.
...
 
 ```js
 const { x, y, width, height } = navigator.virtualKeyboard.boundingRect;
 console.log('Virtual keyboard geometry:', x, y, width, height);
 ```
 
+### Being informed of geometry changes
+
+Whenever the virtual keyboard appears or disappears, the `geometrychange` event is dispatched. The
+event's `target` property contains the virtualKeyboard which (as discussed above) contains the new geometry of the virtual keyboard inset as a
```
tomayac, line 113:
```suggestion
event's `target` property contains the `virtualKeyboard` object which (as discussed above) contains the new geometry of the virtual keyboard inset as a
```

tomayac (approved):
LGTM with the one nit. Thank you!

Just verified that the erroneous destructuring code is correct in the [demo](https://glitch.com/edit/#!/virtualkeyboard).

tomayac (approved):

PR #7848, opened 2022-05-06
Don't imply getDevices after notification click

beaufortfrancois:
As raised in https://bugs.chromium.org/p/chromium/issues/detail?id=1322413, the article implied developers could jump straight from the notification to calling getDevices(). This is not the case.

FYI @reillyeon

### DIFF
--- a/src/site/content/en/blog/usb/index.md
+++ b/src/site/content/en/blog/usb/index.md
@@ -5,7 +5,7 @@ subhead: |
 authors:
   - beaufortfrancois
 date: 2016-03-30
-updated: 2022-02-25
+updated: 2022-05-06
 hero: image/admin/hhnhxiNuRWMfGqy4NSaH.jpg
 thumbnail: image/admin/RyaGPB8fHCuuXUc9Wj9Z.jpg
 alt: A photo of an Arduino Micro board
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/usb/index.md
```diff
@@ -119,7 +119,7 @@ are simply ECMAScript 2015 [Arrow functions].
 
 You can either prompt the user to select a single connected USB device using
 `navigator.usb.requestDevice()` or call `navigator.usb.getDevices()` to get a
-list of all connected USB devices the origin has access to.
+list of all connected USB devices the website has been granted access to.
```
jpmedley, line 122: 'Website' is less specific than 'origin'. Is this really what you want it to say?

jpmedley (changes requested):

src/site/content/en/blog/usb/index.md
beaufortfrancois: In that context, I think it's okay to say that. That's a pattern I've used in other Device APIs. See https://web.dev/serial/#:~:text=returns%20a%20list%20of%20serial%20ports%20the%20website%20has%20been%20granted%20access%20to.

jpmedley (approved):

PR #7882, opened 2022-05-10
Let readers see the demo sources.

jpmedley:
@tomayac I thought we should make these a little easier to play with. Do you agree?


tomayac (approved):
LGTM, thanks

PR #7897, opened 2022-05-11
Fix `@media min-width` example

buondevid:

- Fixes an example in [new media query width syntax](https://web.dev/state-of-css-2022/#was-in-min-width-or-max-width).

I think the example is wrong, can you check if I'm right?

### DIFF
--- a/src/site/content/en/blog/state-of-css-2022/index.md
+++ b/src/site/content/en/blog/state-of-css-2022/index.md
@@ -1438,7 +1438,7 @@ articulate over and under conditions. It may look like this:
 After media query ranges, the same media query could look like this:
 
 ```css
-@media (320px >= width) {
+@media (width >= 320px) {
   …
 }
 ```


src/site/content/en/blog/state-of-css-2022/index.md
```diff
@@ -1438,7 +1438,7 @@ articulate over and under conditions. It may look like this:
 After media query ranges, the same media query could look like this:
 
 ```css
-@media (320px >= width) {
+@media (width >= 320px) {
```
argyleink, line 1441: nice catch, thank you! this update makes the query match the min-width version 👍🏻

argyleink (approved):

src/site/content/en/blog/state-of-css-2022/index.md
buondevid: thanks for the amazing article!

PR #7921, opened 2022-05-13
ContentMigrationBatch#SH5R9

simgobb:
https://www.html5rocks.com/en/tutorials/casestudies/racer/
https://www.html5rocks.com/en/tutorials/casestudies/roll-it/
https://www.html5rocks.com/en/tutorials/casestudies/subsonic/
https://www.html5rocks.com/en/tutorials/casestudies/sunlight_streamcongress/

@rachelandrew

### DIFF
--- a/src/site/content/en/blog/racer/index.md
+++ b/src/site/content/en/blog/racer/index.md
@@ -1,0 +1,289 @@
+---
+layout: post
+title: Case Study - Building Racer
+authors:
+  - activetheory
+date: 2013-06-11
+tags:
+  - blog
+  - case-study
+---
+## Introduction
+
+[Racer](http://g.co/racer) is a web-based mobile Chrome Experiment developed by [Active Theory](http://activetheory.net). Up to 5 friends can connect their phones or tablets to race across every screen. Armed with the concept, design and prototype from Google Creative Lab and sound from Plan8 we iterated on builds for 8 weeks leading up to the launch at I/O ‘13. Now that the game has been live for a few weeks we’ve had the chance to field some questions from the developer community on how it works. Below is a breakdown of the key features and answers to the questions we’re most often asked.
+
+## The Track
+
+A fairly obvious challenge we faced was how to make a web based mobile game that works well across a wide variety of devices. Players needed to be able to construct a race with different phones and tablets. One player could have a Nexus 4 and want to race against his friend who had an iPad. We needed to come up with a way to determine a common track size for each race. The solution had to involve using different size tracks depending on the specs for each device included in the race.
+
+### Calculating Track Dimensions
+
+As each player joins, information about their device is sent to the server and shared with other players. When the track is being built, this data is used to calculate the height and width of the track. We calculate the height by finding the height of the smallest screen, and the width is the total width of all screens. So in the example below the track would have a width of 1152 pixels and a height of 519 pixels.
+
+<figure>
+{% Img src="image/T4FyVKpzu4WKF1kBNvXepbi08t52/pGzIg4YvKiY4kK4kypOG.jpg", alt="The red area shows the total width and height of the track for this example.", width="731", height="617" %}
+<figcaption>The red area shows the total width and height of the track for this example.</figcaption>
+</figure>
+
+```js
+this.getDimensions = function() {
+    var response = {};
+    response.width = 0;
+    response.height = _gamePlayers[0].scrn.h; //first screen height
+    response.screens = [];
+    
+    for (var i = 0; i < _gamePlayers.length; i++) {
+        var player = _gamePlayers[i];
+        response.width += player.scrn.w;
+        if (player.scrn.h < response.height) {
+            //find the smallest screen height
+            response.height = player.scrn.h;
+        }
+        
+        response.screens.push(player.scrn);
+    }
+    
+    return response;
+}
+```
+
+### Drawing the Track
+
+[Paper.js](http://paperjs.org) is an open source vector graphics scripting framework that runs on top of HTML5 Canvas. We found Paper.js was the perfect tool to create vector shapes for the tracks, so we used its capabilities to render the SVG tracks that were built in Adobe Illustrator on a Canvas element. To create the track, the TrackModel class appends the SVG code to the DOM and gathers information about the original dimensions and positioning to be passed to the TrackPathView which will draw the track to a canvas.
+
+```js
+paper.install(window);
+_paper = new paper.PaperScope();
+_paper.setup('track_canvas');
+                    
+var svg = document.getElementById('track');
+var layer = new _paper.Layer();
+
+_path = layer.importSvg(svg).firstChild.firstChild;
+_path.strokeColor = '#14a8df';
+_path.strokeWidth = 2;
+```
+
+Once the track is drawn, each device finds its x offset based on it’s position in the device line-up order, and positions the track accordingly.
+
+```js
+var x = 0;
+for (var i = 0; i < screens.length; i++) {
+        if (i < PLAYER_INDEX) {
+                x += screens[i].w;
+        }
+}
+```
+
+<figure>
+{% Img src="image/T4FyVKpzu4WKF1kBNvXepbi08t52/rhkZNlB2FiqNkNNcfTWU.jpg", alt="The x offset can then be used to show the appropriate portion of the track.", width="681", height="942" %}
+<figcaption>The x offset can then be used to show the appropriate portion of the track</figcaption>
+</figure>
+
+
+## CSS Animations
+
+Paper.js uses a lot of CPU processing to draw the track lanes and this process will take more or less time on different devices. To handle this, we needed a loader to loop until all devices finish processing the track. The problem was that any JavaScript-based animation would skip frames due to the CPU requirements of Paper.js. Enter CSS animations, which run on a separate UI thread, allowing us to smoothly animate the sheen across the "BUILDING TRACK” text.
+
+```css
+.glow {
+    width: 290px;
+    height: 290px;
+    background: url(img/track-glow.png) 0 0 no-repeat;
+    background-size: 100%;
+    top: 0px;
+    left: -290px;
+    z-index: 1;
+    -webkit-animation: wipe 1.3s linear 0s infinite;
+}
+
+@-webkit-keyframes wipe {
+    0%   { -webkit-transform: translate(-300px,0); }
+    25%  { -webkit-transform: translate(-300px,0); }
+    75%  { -webkit-transform: translate(920px,0); }
+    100%  { -webkit-transform: translate(920px,0); }
+}
+```
+
+## CSS Sprites
+
+CSS also came in handy for in-game effects. Mobile devices, with their limited power, are kept busy animating the cars running across the tracks. So for additional excitement we used sprites as a way of implementing pre-rendered animations into the game. In a CSS sprite, transitions apply a step-based animation which changes the background-position property, creating the car explosion
+
+```css
+#sprite {
+    height: 100px; 
+    width: 100px;
+    background: url(sprite.jpg) 0 0 no-repeat;
+    -webkit-animation: play-sprite 0.33s linear 0s steps(9) infinite;
+}
+
+@-webkit-keyframes play-sprite {
+    0%     { background-position: 0px 0px; }
+    100%   { background-position: -900px 0px; }
+}
+```
+
+The problem with this technique is you can only use sprite sheets laid out on a single row. In order to loop through multiple rows, the animation must be chained through multiple keyframe declarations.
+
+```css
+#sprite {
+    height: 100px; 
+    width: 100px;
+    background: url(sprite.jpg) 0 0 no-repeat;
+    -webkit-animation-name: row1, row2, row3;
+    -webkit-animation-duration: 0.2s;
+    -webkit-animation-delay: 0s, 0.2s, 0.4s;
+    -webkit-animation-timing-function: steps(5), steps(5), steps(5);
+    -webkit-animation-fill-mode: forwards;
+}
+
+@-webkit-keyframes row1 {
+    0%     { background-position: 0px 0px; }
+    100%   { background-position: -500px 0px; }
+}
+
+@-webkit-keyframes row2 {
+    0%     { background-position: 0px -100px; }
+    100%   { background-position: -500px -100px; }
+}
+
+@-webkit-keyframes row3 {
+    0%     { background-position: 0px -200px; }
+    100%   { background-position: -500px -200px; }
+}
+```
+
+## Rendering the Cars
+
+As with any car racing game we knew it was important to give the user a feeling of acceleration and handling. Applying a different amount of traction was important for game balancing and the fun factor, so that once a player got a feel for the physics, they’d get a sense of accomplishment and become a better racer. 
+
+Once again we called on Paper.js which comes with an extensive set of math utilities. We used some of its methods to move the car along the path, while adjusting the car position and rotation smoothly each frame.
+
+```js
+var trackOffset = _path.length - (_elapsed % _path.length);
+var trackPoint = _path.getPointAt(trackOffset);
+var trackAngle = _path.getTangentAt(trackOffset).angle;
+
+_velocity.length += _throttle; //apply the throttle
+
+if (!_throttle) {
+    //slow down since the throttle is off
+    _velocity.length *= FRICTION;
+}
+
+if (_velocity.length > MAXVELOCITY) {
+    _velocity.length = MAXVELOCITY;
+}
+
+_velocity.angle = trackAngle;
+
+trackOffset -= _velocity.length;
+_elapsed += _velocity.length;
+
+//find if a lap has been completed
+if (trackOffset < 0) {
+    while (trackOffset < 0) trackOffset += _path.length;
+    trackPoint = _path.getPointAt(trackOffset);
+    console.log('LAP COMPLETE!');
+}
+
+if (_velocity.length > 0.1) {
+    //render the car if there is actually velocity
+    renderCar(trackPoint);
+}
+```
+
+While we were optimizing car rendering, we found an interesting point. On iOS, the best performance was achieved by applying a translate3d transform to the car:
+
+```js
+_car.style.webkitTransform = 'translate3d('+_position.x+'px, '+_position.y+'px, 0px)rotate('+_rotation+'deg)';
+```
+
+On Chrome for Android, the best performance was achieved by calculating the matrix values and applying a matrix transform:
+
+```js
+var rad = _rotation.rotation * (Math.PI * 2 / 360);
+var cos = Math.cos(rad);
+var sin = Math.sin(rad);
+var a = parseFloat(cos).toFixed(8);
+var b = parseFloat(sin).toFixed(8);
+var c = parseFloat(-sin).toFixed(8);
+var d = a;
+_car.style.webkitTransform = 'matrix('+a+', '+b+', '+c+', '+d+', '+_position.x+', '+_position.y+')';
+```
+
+## Keeping the Devices Synced
+
+The most important (and difficult) part of development was to make sure the game synced across devices. We thought users could be forgiving if a car occasionally skipped a few frames because of a slow connection but it wouldn’t be much fun if your car is jumping around, appearing on multiple screens at once.  Resolving this required a ton of trial and error, but we eventually settled on a few tricks which made it work.
+
+### Calculating Latency
+
+The starting point for syncing devices is knowing how long it takes for messages to be received from the Compute Engine relay. The tricky part is that the clocks on each device will never be completely in sync. To get around this, we needed to find the difference in time between the device and the server.
+
+To find the time offset between the device and the main server, we send a message with the current device timestamp. The server will then reply with the original timestamp along with the timestamp of the server. We use the response to calculate the actual difference in time.
+
+```js
+var currentTime = Date.now();
+var latency = Math.round((currentTime - e.time) * .5);
+var serverTime = e.serverTime;
+currentTime -= latency;
+var difference = currentTime - serverTime;
+```
+
+Doing this once is not enough, as the round trip to the server is not always symmetrical, meaning it may take longer for the response to get to the server than it does for the server to return it. To get around this, we poll the server multiple times, taking the median result. This gets us within 10ms of the actual difference between device and server.
+
+### Acceleration/Deceleration
+
+When Player 1 presses or releases the screen, the acceleration event is sent to the server. Once received, the server adds its current timestamp and then passes that data along to every other player.
+
+When an “accelerate on” or “accelerate off” event is received by a device, we are able to use the server offset (calculated above) to find out how long it took for that message to be received. This is useful, because Player 1 may receive the message in 20ms, but Player 2 might take 50ms to receive it. This would result in the car being in two different places because device 1 would start the acceleration sooner.
+
+We can take the time it took to receive the event and convert that into frames. At 60fps, each frame is 16.67ms -- so we can add more velocity (acceleration) or friction (deceleration) on of the car to account for the frames it missed.
+
+```js
+var frames = time / 16.67;
+var onScreen = this.isOnScreen() && time < 75;
+for (var i = 0; i < frames; i++) {
+    if (onScreen) {
+        _velocity.length += _throttle * Math.round(frames*.215);
+    } else {
+        _this.render();
+    }
+}
+```
+
+In the example above, if Player 1 has the car on its screen and the time it took to receive the message is less than 75ms, it will adjust the car’s velocity, speeding it up to make up the difference. If the device is not on screen or the message took too long, it will run the render function and actually make the car jump to where it needs to be.
+
+### Keeping the Cars Synced
+
+Even after accounting for latency in acceleration, the car could still come out of sync and appear on multiple screens at once; specifically when transitioning from one device to the next. In order to prevent this, update events are sent frequently to keep the cars in the same position on the track across all screens.
+
+The logic is simple: every 4 frames, if the car is visible on screen, that device sends it’s values to each of the other devices. If the car is not visible, the app updates the values with those received and then moves the car forward based on the time it took to get the update event.
+
+```js
+this.getValues = function() {
+    _values.p = _position.clone();
+    _values.r = _rotation;
+    _values.e = _elapsed;
+    _values.v = _velocity.length;
+    _values.pos = _this.position;
+    return _values;
+}
+
+this.setValues = function(val, time) {
+    _position.x = val.p.x;
+    _position.y = val.p.y;
+    _rotation = val.r;
+    _elapsed = val.e;
+    _velocity.length = val.v;
+
+    var frames = time / 16.67;
+    for (var i = 0; i < frames; i++) {
+            _this.render();
+    }
+}
+```
+
+## Conclusion
+
+As soon as we heard the concept for Racer, we knew it had the potential to be a very special project. We quickly built a prototype which gave us a rough idea of how to overcome latency and network performance. It was a challenging project that kept us busy during late nights and long weekends, but it was a great feeling when the game started to take shape. Ultimately, we’re very happy with the end result. Google Creative Lab’s concept pushed the limits of browser technology in a fun way, and as developers we couldn’t ask for more.
--- a/src/site/content/en/blog/racer/index.md
+++ b/src/site/content/en/blog/racer/index.md
@@ -10,7 +10,7 @@ tags:
 ---
 ## Introduction
 
-[Racer](http://g.co/racer) is a web-based mobile Chrome Experiment developed by [Active Theory](http://activetheory.net). Up to 5 friends can connect their phones or tablets to race across every screen. Armed with the concept, design and prototype from Google Creative Lab and sound from Plan8 we iterated on builds for 8 weeks leading up to the launch at I/O ‘13. Now that the game has been live for a few weeks we’ve had the chance to field some questions from the developer community on how it works. Below is a breakdown of the key features and answers to the questions we’re most often asked.
+[Racer](http://g.co/racer) is a web-based mobile Chrome Experiment developed by [Active Theory](http://activetheory.net). Up to 5 friends can connect their phones or tablets to race across every screen. Armed with the concept, design and prototype from Google Creative Lab and sound from Plan8 we iterated on builds for 8 weeks leading up to the launch at I/O 2013. Now that the game has been live for a few weeks we've had the chance to field some questions from the developer community on how it works. Below is a breakdown of the key features and answers to the questions we're most often asked.
 
 ## The Track
 
@@ -26,30 +26,31 @@ As each player joins, information about their device is sent to the server and s
 </figure>
 
 ```js
-this.getDimensions = function() {
-    var response = {};
-    response.width = 0;
-    response.height = _gamePlayers[0].scrn.h; //first screen height
-    response.screens = [];
-    
-    for (var i = 0; i < _gamePlayers.length; i++) {
-        var player = _gamePlayers[i];
-        response.width += player.scrn.w;
-        if (player.scrn.h < response.height) {
-            //find the smallest screen height
-            response.height = player.scrn.h;
-        }
-        
-        response.screens.push(player.scrn);
+this.getDimensions = function () {
+  var response = {};
+  response.width = 0;
+  response.height = _gamePlayers[0].scrn.h; // First screen height
+  response.screens = [];
+  
+  for (var i = 0; i < _gamePlayers.length; i++) {
+    var player = _gamePlayers[i];
+    response.width += player.scrn.w;
+
+    if (player.scrn.h < response.height) {
+      // Find the smallest screen height
+      response.height = player.scrn.h;
     }
-    
-    return response;
+      
+    response.screens.push(player.scrn);
+  }
+  
+  return response;
 }
 ```
 
 ### Drawing the Track
 
-[Paper.js](http://paperjs.org) is an open source vector graphics scripting framework that runs on top of HTML5 Canvas. We found Paper.js was the perfect tool to create vector shapes for the tracks, so we used its capabilities to render the SVG tracks that were built in Adobe Illustrator on a Canvas element. To create the track, the TrackModel class appends the SVG code to the DOM and gathers information about the original dimensions and positioning to be passed to the TrackPathView which will draw the track to a canvas.
+[Paper.js](http://paperjs.org) is an open source vector graphics scripting framework that runs on top of HTML5 Canvas. We found Paper.js was the perfect tool to create vector shapes for the tracks, so we used its capabilities to render the SVG tracks that were built in Adobe Illustrator on a `<canvas>` element. To create the track, the `TrackModel` class appends the SVG code to the DOM and gathers information about the original dimensions and positioning to be passed to the `TrackPathView` which will draw the track to a canvas.
 
 ```js
 paper.install(window);
@@ -68,10 +69,11 @@ Once the track is drawn, each device finds its x offset based on it’s position
 
 ```js
 var x = 0;
+
 for (var i = 0; i < screens.length; i++) {
-        if (i < PLAYER_INDEX) {
-                x += screens[i].w;
-        }
+  if (i < PLAYER_INDEX) {
+    x += screens[i].w;
+  }
 }
 ```
 
@@ -80,46 +82,62 @@ for (var i = 0; i < screens.length; i++) {
 <figcaption>The x offset can then be used to show the appropriate portion of the track</figcaption>
 </figure>
 
-
 ## CSS Animations
 
-Paper.js uses a lot of CPU processing to draw the track lanes and this process will take more or less time on different devices. To handle this, we needed a loader to loop until all devices finish processing the track. The problem was that any JavaScript-based animation would skip frames due to the CPU requirements of Paper.js. Enter CSS animations, which run on a separate UI thread, allowing us to smoothly animate the sheen across the "BUILDING TRACK” text.
+Paper.js uses a lot of CPU processing to draw the track lanes and this process will take more or less time on different devices. To handle this, we needed a loader to loop until all devices finish processing the track. The problem was that any JavaScript-based animation would skip frames due to the CPU requirements of Paper.js. Enter CSS animations, which run on a separate UI thread, allowing us to smoothly animate the sheen across the "BUILDING TRACK" text.
 
 ```css
 .glow {
-    width: 290px;
-    height: 290px;
-    background: url(img/track-glow.png) 0 0 no-repeat;
-    background-size: 100%;
-    top: 0px;
-    left: -290px;
-    z-index: 1;
-    -webkit-animation: wipe 1.3s linear 0s infinite;
+  width: 290px;
+  height: 290px;
+  background: url('img/track-glow.png') 0 0 no-repeat;
+  background-size: 100%;
+  top: 0;
+  left: -290px;
+  z-index: 1;
+  -webkit-animation: wipe 1.3s linear 0s infinite;
 }
 
 @-webkit-keyframes wipe {
-    0%   { -webkit-transform: translate(-300px,0); }
-    25%  { -webkit-transform: translate(-300px,0); }
-    75%  { -webkit-transform: translate(920px,0); }
-    100%  { -webkit-transform: translate(920px,0); }
+  0% {
+    -webkit-transform: translate(-300px, 0);
+  }
+
+  25% {
+    -webkit-transform: translate(-300px, 0);
+  }
+
+  75% {
+    -webkit-transform: translate(920px, 0);
+  }
+
+  100% {
+    -webkit-transform: translate(920px, 0);
+  }
+}
 }
 ```
 
 ## CSS Sprites
 
-CSS also came in handy for in-game effects. Mobile devices, with their limited power, are kept busy animating the cars running across the tracks. So for additional excitement we used sprites as a way of implementing pre-rendered animations into the game. In a CSS sprite, transitions apply a step-based animation which changes the background-position property, creating the car explosion
+CSS also came in handy for in-game effects. Mobile devices, with their limited power, are kept busy animating the cars running across the tracks. So for additional excitement we used sprites as a way of implementing pre-rendered animations into the game. In a CSS sprite, transitions apply a step-based animation which changes the `background-position` property, creating the car explosion.
 
 ```css
 #sprite {
-    height: 100px; 
-    width: 100px;
-    background: url(sprite.jpg) 0 0 no-repeat;
-    -webkit-animation: play-sprite 0.33s linear 0s steps(9) infinite;
+  height: 100px; 
+  width: 100px;
+  background: url('sprite.jpg') 0 0 no-repeat;
+  -webkit-animation: play-sprite 0.33s linear 0s steps(9) infinite;
 }
 
 @-webkit-keyframes play-sprite {
-    0%     { background-position: 0px 0px; }
-    100%   { background-position: -900px 0px; }
+  0% {
+    background-position: 0 0;
+  }
+
+  100% {
+    background-position: -900px 0;
+  }
 }
 ```
 
@@ -127,29 +145,44 @@ The problem with this technique is you can only use sprite sheets laid out on a 
 
 ```css
 #sprite {
-    height: 100px; 
-    width: 100px;
-    background: url(sprite.jpg) 0 0 no-repeat;
-    -webkit-animation-name: row1, row2, row3;
-    -webkit-animation-duration: 0.2s;
-    -webkit-animation-delay: 0s, 0.2s, 0.4s;
-    -webkit-animation-timing-function: steps(5), steps(5), steps(5);
-    -webkit-animation-fill-mode: forwards;
+  height: 100px; 
+  width: 100px;
+  background: url('sprite.jpg') 0 0 no-repeat;
+  -webkit-animation-name: row1, row2, row3;
+  -webkit-animation-duration: 0.2s;
+  -webkit-animation-delay: 0s, 0.2s, 0.4s;
+  -webkit-animation-timing-function: steps(5), steps(5), steps(5);
+  -webkit-animation-fill-mode: forwards;
 }
 
 @-webkit-keyframes row1 {
-    0%     { background-position: 0px 0px; }
-    100%   { background-position: -500px 0px; }
+  0% {
+    background-position: 0 0;
+  }
+
+  100% {
+    background-position: -500px 0;
+  }
 }
 
 @-webkit-keyframes row2 {
-    0%     { background-position: 0px -100px; }
-    100%   { background-position: -500px -100px; }
+  0% {
+    background-position: 0 -100px;
+  }
+
+  100% {
+    background-position: -500px -100px;
+  }
 }
 
 @-webkit-keyframes row3 {
-    0%     { background-position: 0px -200px; }
-    100%   { background-position: -500px -200px; }
+  0% {
+    background-position: 0 -200px;
+  }
+
+  100% {
+    background-position: -500px -200px;
+  }
 }
 ```
 
@@ -164,36 +197,37 @@ var trackOffset = _path.length - (_elapsed % _path.length);
 var trackPoint = _path.getPointAt(trackOffset);
 var trackAngle = _path.getTangentAt(trackOffset).angle;
 
-_velocity.length += _throttle; //apply the throttle
+// Apply the throttle
+_velocity.length += _throttle;
 
 if (!_throttle) {
-    //slow down since the throttle is off
-    _velocity.length *= FRICTION;
+  // Slow down since the throttle is off
+  _velocity.length *= FRICTION;
 }
 
 if (_velocity.length > MAXVELOCITY) {
-    _velocity.length = MAXVELOCITY;
+  _velocity.length = MAXVELOCITY;
 }
 
 _velocity.angle = trackAngle;
-
 trackOffset -= _velocity.length;
 _elapsed += _velocity.length;
 
-//find if a lap has been completed
+// Find if a lap has been completed
 if (trackOffset < 0) {
-    while (trackOffset < 0) trackOffset += _path.length;
-    trackPoint = _path.getPointAt(trackOffset);
-    console.log('LAP COMPLETE!');
+  while (trackOffset < 0) trackOffset += _path.length;
+
+  trackPoint = _path.getPointAt(trackOffset);
+  console.log('LAP COMPLETE!');
 }
 
 if (_velocity.length > 0.1) {
-    //render the car if there is actually velocity
-    renderCar(trackPoint);
+  // Render the car if there is actually velocity
+  renderCar(trackPoint);
 }
 ```
 
-While we were optimizing car rendering, we found an interesting point. On iOS, the best performance was achieved by applying a translate3d transform to the car:
+While we were optimizing car rendering, we found an interesting point. On iOS, the best performance was achieved by applying a `translate3d` transform to the car:
 
 ```js
 _car.style.webkitTransform = 'translate3d('+_position.x+'px, '+_position.y+'px, 0px)rotate('+_rotation+'deg)';
@@ -209,12 +243,12 @@ var a = parseFloat(cos).toFixed(8);
 var b = parseFloat(sin).toFixed(8);
 var c = parseFloat(-sin).toFixed(8);
 var d = a;
-_car.style.webkitTransform = 'matrix('+a+', '+b+', '+c+', '+d+', '+_position.x+', '+_position.y+')';
+_car.style.webkitTransform = 'matrix(' + a + ', ' + b + ', ' + c + ', ' + d + ', ' + _position.x + ', ' + _position.y + ')';
 ```
 
 ## Keeping the Devices Synced
 
-The most important (and difficult) part of development was to make sure the game synced across devices. We thought users could be forgiving if a car occasionally skipped a few frames because of a slow connection but it wouldn’t be much fun if your car is jumping around, appearing on multiple screens at once.  Resolving this required a ton of trial and error, but we eventually settled on a few tricks which made it work.
+The most important (and difficult) part of development was to make sure the game synced across devices. We thought users could be forgiving if a car occasionally skipped a few frames because of a slow connection but it wouldn't be much fun if your car is jumping around, appearing on multiple screens at once. Resolving this required a ton of trial and error, but we eventually settled on a few tricks which made it work.
 
 ### Calculating Latency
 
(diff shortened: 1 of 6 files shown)


src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+authors:
+  - activetheory
+date: 2013-06-11
+tags:
+  - blog
+  - case-study
+---
+## Introduction
+
+[Racer](http://g.co/racer) is a web-based mobile Chrome Experiment developed by [Active Theory](http://activetheory.net). Up to 5 friends can connect their phones or tablets to race across every screen. Armed with the concept, design and prototype from Google Creative Lab and sound from Plan8 we iterated on builds for 8 weeks leading up to the launch at I/O ‘13. Now that the game has been live for a few weeks we’ve had the chance to field some questions from the developer community on how it works. Below is a breakdown of the key features and answers to the questions we’re most often asked.
```
malchata, line 13:
```suggestion
[Racer](http://g.co/racer) is a web-based mobile Chrome Experiment developed by [Active Theory](http://activetheory.net). Up to 5 friends can connect their phones or tablets to race across every screen. Armed with the concept, design and prototype from Google Creative Lab and sound from Plan8 we iterated on builds for 8 weeks leading up to the launch at I/O 2013. Now that the game has been live for a few weeks we've had the chance to field some questions from the developer community on how it works. Below is a breakdown of the key features and answers to the questions we're most often asked.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+        if (player.scrn.h < response.height) {
+            //find the smallest screen height
+            response.height = player.scrn.h;
+        }
+        
+        response.screens.push(player.scrn);
+    }
+    
+    return response;
+}
```
malchata, line 47:
```suggestion
this.getDimensions = function () {
  var response = {};
  response.width = 0;
  response.height = _gamePlayers[0].scrn.h; // First screen height
  response.screens = [];
  
  for (var i = 0; i < _gamePlayers.length; i++) {
    var player = _gamePlayers[i];
    response.width += player.scrn.w;

    if (player.scrn.h < response.height) {
      // Find the smallest screen height
      response.height = player.scrn.h;
    }
      
    response.screens.push(player.scrn);
  }
  
  return response;
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+### Drawing the Track
+
+[Paper.js](http://paperjs.org) is an open source vector graphics scripting framework that runs on top of HTML5 Canvas. We found Paper.js was the perfect tool to create vector shapes for the tracks, so we used its capabilities to render the SVG tracks that were built in Adobe Illustrator on a Canvas element. To create the track, the TrackModel class appends the SVG code to the DOM and gathers information about the original dimensions and positioning to be passed to the TrackPathView which will draw the track to a canvas.
```
malchata, line 52:
```suggestion
[Paper.js](http://paperjs.org) is an open source vector graphics scripting framework that runs on top of HTML5 Canvas. We found Paper.js was the perfect tool to create vector shapes for the tracks, so we used its capabilities to render the SVG tracks that were built in Adobe Illustrator on a `<canvas>` element. To create the track, the `TrackModel` class appends the SVG code to the DOM and gathers information about the original dimensions and positioning to be passed to the `TrackPathView` which will draw the track to a canvas.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+Once the track is drawn, each device finds its x offset based on it’s position in the device line-up order, and positions the track accordingly.
+
+```js
+var x = 0;
+for (var i = 0; i < screens.length; i++) {
+        if (i < PLAYER_INDEX) {
+                x += screens[i].w;
+        }
+}
```
malchata, line 75:
```suggestion
var x = 0;

for (var i = 0; i < screens.length; i++) {
  if (i < PLAYER_INDEX) {
    x += screens[i].w;
  }
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+<figure>
+{% Img src="image/T4FyVKpzu4WKF1kBNvXepbi08t52/rhkZNlB2FiqNkNNcfTWU.jpg", alt="The x offset can then be used to show the appropriate portion of the track.", width="681", height="942" %}
+<figcaption>The x offset can then be used to show the appropriate portion of the track</figcaption>
+</figure>
+
+
+## CSS Animations
```
malchata, line 84:
```suggestion
</figure>

## CSS Animations
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+Paper.js uses a lot of CPU processing to draw the track lanes and this process will take more or less time on different devices. To handle this, we needed a loader to loop until all devices finish processing the track. The problem was that any JavaScript-based animation would skip frames due to the CPU requirements of Paper.js. Enter CSS animations, which run on a separate UI thread, allowing us to smoothly animate the sheen across the "BUILDING TRACK” text.
```
malchata, line 86:
```suggestion
Paper.js uses a lot of CPU processing to draw the track lanes and this process will take more or less time on different devices. To handle this, we needed a loader to loop until all devices finish processing the track. The problem was that any JavaScript-based animation would skip frames due to the CPU requirements of Paper.js. Enter CSS animations, which run on a separate UI thread, allowing us to smoothly animate the sheen across the "BUILDING TRACK" text.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+    left: -290px;
+    z-index: 1;
+    -webkit-animation: wipe 1.3s linear 0s infinite;
+}
+
+@-webkit-keyframes wipe {
+    0%   { -webkit-transform: translate(-300px,0); }
+    25%  { -webkit-transform: translate(-300px,0); }
+    75%  { -webkit-transform: translate(920px,0); }
+    100%  { -webkit-transform: translate(920px,0); }
```
malchata, line 104:
```suggestion
.glow {
  width: 290px;
  height: 290px;
  background: url('img/track-glow.png') 0 0 no-repeat;
  background-size: 100%;
  top: 0;
  left: -290px;
  z-index: 1;
  -webkit-animation: wipe 1.3s linear 0s infinite;
}

@-webkit-keyframes wipe {
  0% {
    -webkit-transform: translate(-300px,0);
  }

  25% {
    -webkit-transform: translate(-300px,0);
  }

  75% {
    -webkit-transform: translate(920px,0);
  }

  100% {
    -webkit-transform: translate(920px,0);
  }
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+## CSS Sprites
+
+CSS also came in handy for in-game effects. Mobile devices, with their limited power, are kept busy animating the cars running across the tracks. So for additional excitement we used sprites as a way of implementing pre-rendered animations into the game. In a CSS sprite, transitions apply a step-based animation which changes the background-position property, creating the car explosion
```
malchata, line 110:
```suggestion
CSS also came in handy for in-game effects. Mobile devices, with their limited power, are kept busy animating the cars running across the tracks. So for additional excitement we used sprites as a way of implementing pre-rendered animations into the game. In a CSS sprite, transitions apply a step-based animation which changes the `background-position` property, creating the car explosion.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+    height: 100px; 
+    width: 100px;
+    background: url(sprite.jpg) 0 0 no-repeat;
+    -webkit-animation: play-sprite 0.33s linear 0s steps(9) infinite;
+}
+
+@-webkit-keyframes play-sprite {
+    0%     { background-position: 0px 0px; }
+    100%   { background-position: -900px 0px; }
+}
```
malchata, line 123:
```suggestion
#sprite {
  height: 100px; 
  width: 100px;
  background: url('sprite.jpg') 0 0 no-repeat;
  -webkit-animation: play-sprite 0.33s linear 0s steps(9) infinite;
}

@-webkit-keyframes play-sprite {
  0% {
    background-position: 0 0;
  }

  100% {
    background-position: -900px 0;
  }
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+@-webkit-keyframes row2 {
+    0%     { background-position: 0px -100px; }
+    100%   { background-position: -500px -100px; }
+}
+
+@-webkit-keyframes row3 {
+    0%     { background-position: 0px -200px; }
+    100%   { background-position: -500px -200px; }
+}
```
malchata, line 153:
```suggestion
#sprite {
  height: 100px; 
  width: 100px;
  background: url('sprite.jpg') 0 0 no-repeat;
  -webkit-animation-name: row1, row2, row3;
  -webkit-animation-duration: 0.2s;
  -webkit-animation-delay: 0s, 0.2s, 0.4s;
  -webkit-animation-timing-function: steps(5), steps(5), steps(5);
  -webkit-animation-fill-mode: forwards;
}

@-webkit-keyframes row1 {
  0% {
    background-position: 0 0;
  }

  100% {
    background-position: -500px 0;
  }
}

@-webkit-keyframes row2 {
  0% {
    background-position: 0 -100px;
  }

  100% {
    background-position: -500px -100px;
  }
}

@-webkit-keyframes row3 {
  0% {
    background-position: 0 -200px;
  }

  100% {
    background-position: -500px -200px;
  }
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+if (trackOffset < 0) {
+    while (trackOffset < 0) trackOffset += _path.length;
+    trackPoint = _path.getPointAt(trackOffset);
+    console.log('LAP COMPLETE!');
+}
+
+if (_velocity.length > 0.1) {
+    //render the car if there is actually velocity
+    renderCar(trackPoint);
+}
```
malchata, line 193:
```suggestion
var trackOffset = _path.length - (_elapsed % _path.length);
var trackPoint = _path.getPointAt(trackOffset);
var trackAngle = _path.getTangentAt(trackOffset).angle;

// Apply the throttle
_velocity.length += _throttle;

if (!_throttle) {
  // Slow down since the throttle is off
  _velocity.length *= FRICTION;
}

if (_velocity.length > MAXVELOCITY) {
  _velocity.length = MAXVELOCITY;
}

_velocity.angle = trackAngle;
trackOffset -= _velocity.length;
_elapsed += _velocity.length;

// Find if a lap has been completed
if (trackOffset < 0) {
  while (trackOffset < 0) trackOffset += _path.length;

  trackPoint = _path.getPointAt(trackOffset);
  console.log('LAP COMPLETE!');
}

if (_velocity.length > 0.1) {
  // Render the car if there is actually velocity
  renderCar(trackPoint);
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+While we were optimizing car rendering, we found an interesting point. On iOS, the best performance was achieved by applying a translate3d transform to the car:
```
malchata, line 196:
```suggestion
While we were optimizing car rendering, we found an interesting point. On iOS, the best performance was achieved by applying a `translate3d` transform to the car:
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+```js
+var rad = _rotation.rotation * (Math.PI * 2 / 360);
+var cos = Math.cos(rad);
+var sin = Math.sin(rad);
+var a = parseFloat(cos).toFixed(8);
+var b = parseFloat(sin).toFixed(8);
+var c = parseFloat(-sin).toFixed(8);
+var d = a;
+_car.style.webkitTransform = 'matrix('+a+', '+b+', '+c+', '+d+', '+_position.x+', '+_position.y+')';
```
malchata, line 212:
```suggestion
_car.style.webkitTransform = 'matrix(' + a + ', ' + b + ', ' + c + ', ' + d + ', ' + _position.x + ', ' + _position.y + ')';
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+## Keeping the Devices Synced
+
+The most important (and difficult) part of development was to make sure the game synced across devices. We thought users could be forgiving if a car occasionally skipped a few frames because of a slow connection but it wouldn’t be much fun if your car is jumping around, appearing on multiple screens at once.  Resolving this required a ton of trial and error, but we eventually settled on a few tricks which made it work.
```
malchata, line 217:
```suggestion
The most important (and difficult) part of development was to make sure the game synced across devices. We thought users could be forgiving if a car occasionally skipped a few frames because of a slow connection but it wouldn't be much fun if your car is jumping around, appearing on multiple screens at once. Resolving this required a ton of trial and error, but we eventually settled on a few tricks which made it work.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+var difference = currentTime - serverTime;
+```
+
+Doing this once is not enough, as the round trip to the server is not always symmetrical, meaning it may take longer for the response to get to the server than it does for the server to return it. To get around this, we poll the server multiple times, taking the median result. This gets us within 10ms of the actual difference between device and server.
+
+### Acceleration/Deceleration
+
+When Player 1 presses or releases the screen, the acceleration event is sent to the server. Once received, the server adds its current timestamp and then passes that data along to every other player.
+
+When an “accelerate on” or “accelerate off” event is received by a device, we are able to use the server offset (calculated above) to find out how long it took for that message to be received. This is useful, because Player 1 may receive the message in 20ms, but Player 2 might take 50ms to receive it. This would result in the car being in two different places because device 1 would start the acceleration sooner.
```
malchata, line 239:
```suggestion
When an "accelerate on" or "accelerate off" event is received by a device, we are able to use the server offset (calculated above) to find out how long it took for that message to be received. This is useful, because Player 1 may receive the message in 20ms, but Player 2 might take 50ms to receive it. This would result in the car being in two different places because device 1 would start the acceleration sooner.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+We can take the time it took to receive the event and convert that into frames. At 60fps, each frame is 16.67ms -- so we can add more velocity (acceleration) or friction (deceleration) on of the car to account for the frames it missed.
```
malchata, line 241:
```suggestion
We can take the time it took to receive the event and convert that into frames. At 60fps, each frame is 16.67ms&mdash;so we can add more velocity (acceleration) or friction (deceleration) on of the car to account for the frames it missed.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```js
+var frames = time / 16.67;
+var onScreen = this.isOnScreen() && time < 75;
+for (var i = 0; i < frames; i++) {
+    if (onScreen) {
+        _velocity.length += _throttle * Math.round(frames*.215);
+    } else {
+        _this.render();
+    }
+}
```
malchata, line 252:
```suggestion
var frames = time / 16.67;
var onScreen = this.isOnScreen() && time < 75;

for (var i = 0; i < frames; i++) {
  if (onScreen) {
    _velocity.length += _throttle * Math.round(frames * .215);
  } else {
    _this.render();
  }
}}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+In the example above, if Player 1 has the car on its screen and the time it took to receive the message is less than 75ms, it will adjust the car’s velocity, speeding it up to make up the difference. If the device is not on screen or the message took too long, it will run the render function and actually make the car jump to where it needs to be.
```
malchata, line 255:
```suggestion
In the example above, if Player 1 has the car on its screen and the time it took to receive the message is less than 75ms, it will adjust the car's velocity, speeding it up to make up the difference. If the device is not on screen or the message took too long, it will run the render function and actually make the car jump to where it needs to be.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Keeping the Cars Synced
+
+Even after accounting for latency in acceleration, the car could still come out of sync and appear on multiple screens at once; specifically when transitioning from one device to the next. In order to prevent this, update events are sent frequently to keep the cars in the same position on the track across all screens.
+
+The logic is simple: every 4 frames, if the car is visible on screen, that device sends it’s values to each of the other devices. If the car is not visible, the app updates the values with those received and then moves the car forward based on the time it took to get the update event.
```
malchata, line 261:
```suggestion
The logic is that, every 4 frames, if the car is visible on screen, that device sends its values to each of the other devices. If the car is not visible, the app updates the values with those received and then moves the car forward based on the time it took to get the update event.
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+    _position.y = val.p.y;
+    _rotation = val.r;
+    _elapsed = val.e;
+    _velocity.length = val.v;
+
+    var frames = time / 16.67;
+    for (var i = 0; i < frames; i++) {
+            _this.render();
+    }
+}
```
malchata, line 284:
```suggestion
this.getValues = function () {
  _values.p = _position.clone();
  _values.r = _rotation;
  _values.e = _elapsed;
  _values.v = _velocity.length;
  _values.pos = _this.position;

  return _values;
}

this.setValues = function (val, time) {
  _position.x = val.p.x;
  _position.y = val.p.y;
  _rotation = val.r;
  _elapsed = val.e;
  _velocity.length = val.v;

  var frames = time / 16.67;

  for (var i = 0; i < frames; i++) {
    _this.render();
  }
}
```

src/site/content/en/blog/racer/index.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+## Conclusion
+
+As soon as we heard the concept for Racer, we knew it had the potential to be a very special project. We quickly built a prototype which gave us a rough idea of how to overcome latency and network performance. It was a challenging project that kept us busy during late nights and long weekends, but it was a great feeling when the game started to take shape. Ultimately, we’re very happy with the end result. Google Creative Lab’s concept pushed the limits of browser technology in a fun way, and as developers we couldn’t ask for more.
```
malchata, line 289:
```suggestion
As soon as we heard the concept for Racer, we knew it had the potential to be a very special project. We quickly built a prototype which gave us a rough idea of how to overcome latency and network performance. It was a challenging project that kept us busy during late nights and long weekends, but it was a great feeling when the game started to take shape. Ultimately, we're very happy with the end result. Google Creative Lab's concept pushed the limits of browser technology in a fun way, and as developers we couldn't ask for more.
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+## 3D Workflow
+
+One of the struggles in the beginning was figuring out the best way to bring 3D models from our software to a web ready file format. After creating the assets inside [Cinema 4D](http://en.wikipedia.org/wiki/Cinema_4D), the models were simplified and converted to low-polygon meshes. Each mesh was given certain polygon selection tags to differentiate between parts of the object for coloring and texturing. We were then able to export as a Collada 1.5 (.dae) file and import into [Blender](http://www.blender.org/), an open source 3D program, in order to make compatible files for three.js. Once we ensured our models were imported correctly, we exported the mesh as a .json file and the lighting was applied using code. Here's a more detailed look at the steps we took:
```
malchata, line 18:
```suggestion
One of the struggles in the beginning was figuring out the best way to bring 3D models from our software to a web ready file format. After creating the assets inside [Cinema 4D](http://en.wikipedia.org/wiki/Cinema_4D), the models were simplified and converted to low-polygon meshes. Each mesh was given certain polygon selection tags to differentiate between parts of the object for coloring and texturing. We were then able to export as a Collada 1.5 (.dae) file and import into [Blender](http://www.blender.org/), an open source 3D program, in order to make compatible files for three.js. Once we ensured our models were imported correctly, we exported the mesh as a JSON file and the lighting was applied using code. Here's a more detailed look at the steps we took:
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+### The development environment
+
+Most of Roll It's original code was written with [CoffeeScript](http://coffeescript.org/) - a clean and concise language that transcompiles to well-formed and [linted](http://www.javascriptlint.com/) JavaScript. CoffeeScript shines for OOP development with its great inheritance model and cleaner scope handling. The CSS was written with the [SASS](http://sass-lang.com/) framework, which gives the developer a number of great tools to enhance and manage a project's stylesheets. Adding these systems to the build process takes a little bit of time to set up, but the payoff is definitely worth it, especially for a larger project like Roll It. We set up a simple [Ruby on Rails](http://rubyonrails.org/) server to auto-compile our assets during development, so all of these compilation steps became transparent.
```
malchata, line 66:
```suggestion
Most of Roll It's original code was written with [CoffeeScript](http://coffeescript.org/)&mdash;a clean and concise language that transcompiles to well-formed and [linted](http://www.javascriptlint.com/) JavaScript. CoffeeScript shines for OOP development with its great inheritance model and cleaner scope handling. The CSS was written with the [SASS](http://sass-lang.com/) framework, which gives the developer a number of great tools to enhance and manage a project's stylesheets. Adding these systems to the build process takes a little bit of time to set up, but the payoff is definitely worth it, especially for a larger project like Roll It. We set up a [Ruby on Rails](http://rubyonrails.org/) server to auto-compile our assets during development, so all of these compilation steps became transparent.
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+Beyond creating a streamlined and comfortable coding environment, we manually optimized assets to minimize requests in order to load the site faster. We ran every image through a couple of compression programs - [ImageOptim](http://imageoptim.com/) and [ImageAlpha](http://pngmini.com/). Each program optimizes images in their own way - lossless and lossy, respectively. With the right combination of settings, they can significantly cut down an image's file size. This not only saves bandwidth when loading external images, but once optimized, your images will translate into a much smaller base64 encoded strings for inline embedding in HTML, CSS and JavaScript. While on the subject of base64 encoding, we also embedded our [Open Sans](http://www.google.com/fonts/specimen/Open+Sans) .woff and .svg font files directly into the CSS using this technique, which resulted in even fewer total requests.
```
malchata, line 68:
```suggestion
Beyond creating a streamlined and comfortable coding environment, we manually optimized assets to minimize requests in order to load the site faster. We ran every image through a couple of compression programs&mdash;[ImageOptim](http://imageoptim.com/) and [ImageAlpha](http://pngmini.com/). Each program optimizes images in their own way&mdash;lossless and lossy, respectively. With the right combination of settings, they can significantly cut down an image's file size. This not only saves bandwidth when loading external images, but once optimized, your images will translate into a much smaller base64 encoded strings for inline embedding in HTML, CSS and JavaScript. While on the subject of base64 encoding, we also embedded our [Open Sans](http://www.google.com/fonts/specimen/Open+Sans) WOFF and SVG font files directly into the CSS using this technique, which resulted in even fewer total requests.
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+### The physics-enabled 3D scene
+
+[THREE.js](http://threejs.org/) is the ubiquitous 3D JavaScript library for the web. It wraps up low-level 3D math and hardware-based WebGL optimizations that enable mere mortals to easily create well-lit and beautiful interactive 3D scenes without having to write custom shaders or perform manual matrix transformations. [Physijs](https://github.com/chandlerprall/Physijs) is a THREE.js-specific wrapper for a popular C++ physics library that has been translated to JavaScript. We took advantage of this library to simulate the ball rolling, jumping, and bouncing towards its destination in 3D.
+
+From the start, we set out to not only make the physical experience of rolling the ball feel realistic, but also to make sure that the objects in the game felt real. This required many iterations of adjusting the overall gravity of the Physijs scene, the speed of the ball as it rolls from the player's throw, the slope of the lane's jump, and the friction and restitution (bounciness) properties of the ball and lane materials. The combination of more gravity and more speed resulted in a more realistic gaming experience.
+
+
+### Smoothing it out
```
malchata, line 77:
```suggestion

### Smoothing it out
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+Most modern browser and video card combinations should take advantage of native hardware-based anti-aliasing in the WebGL environment, but some won't play nice. In the case that anti-aliasing doesn't work natively, any hard and contrasted edges in the THREE.js scene will will be jagged and ugly (to our discerning eyes, at least). 
+
+Luckily there's a fix: through a snippet of code, we can detect whether the platform will natively support antialiasing. If it does, then we're good to go. If it doesn't, there are a series of post-processing shaders that come with THREE.js that can help us. Namely, the FXAA anti-aliasing filter. By redrawing the rendered scene every frame with this shader, we're generally left with a much smoother look to our lines and edges. See the demo below:
+
+```js
+// check for native platform antialias support via the THREE renderer
```
malchata, line 84:
```suggestion
// Check for native platform antialias support via the THREE renderer
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+Detecting Roll It's main "roll" gesture starts with tracking the 10 most recent accelerometer updates that come from the window's ‘[deviceorientation](http://www.html5rocks.com/en/tutorials/device/orientation/)' event. By subtracting the previous tilt value from the current tilt value, we store the angle delta between events. Then, by constantly summing up the last ten angle deltas, we can detect continuous rotation as the phone moves through space. When the phone passes a threshold of sweeping angle change, we trigger a roll. Then, by finding the largest single tilt delta in that sweep, we can estimate a speed for the ball. In Roll It, this speed is normalized using timestamps that we attach to each accelerometer update. This helps smooth out the variable speed in which accelerometer updates stream into the browser on different devices.
```
malchata, line 93:
```suggestion
Detecting Roll It's main "roll" gesture starts with tracking the 10 most recent accelerometer updates that come from the window's [`deviceorientation`](http://www.html5rocks.com/en/tutorials/device/orientation/) event. By subtracting the previous tilt value from the current tilt value, we store the angle delta between events. Then, by constantly summing up the last ten angle deltas, we can detect continuous rotation as the phone moves through space. When the phone passes a threshold of sweeping angle change, we trigger a roll. Then, by finding the largest single tilt delta in that sweep, we can estimate a speed for the ball. In Roll It, this speed is normalized using timestamps that we attach to each accelerometer update. This helps smooth out the variable speed in which accelerometer updates stream into the browser on different devices.
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+### WebSockets communication
+
+Once the player rolls the ball with their phone, a message is sent from the phone to the laptop telling it to launch the ball. This "roll" message is sent via a [JSON](http://en.wikipedia.org/wiki/JSON) data object through a [WebSocket](http://en.wikipedia.org/wiki/WebSocket) connection between the two machines. The JSON data is small, mainly consisting of a message type, throw speed, and aim direction. 
+
+```json
+{type: "device:ball-thrown", speed: 0.5, aim: 0.1}
```
malchata, line 100:
```suggestion
{
  "type": "device:ball-thrown",
  "speed": 0.5,
  "aim": 0.1
}
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+All of the communication between the laptop and phone happens via small JSON messages like this. Every time the game updates its state on the desktop, or the user tilts or taps a button on the phone, a WebSocket message is transmitted between the machines. In order to keep this communication simple and easy to manage, the WebSockets messages are broadcast using a single exit point from either browser. Conversely, there's a single entry point on the receiving browser, with one WebSocket object handling all incoming and outgoing messages on both ends. When a WebSocket message is received, the JSON data is rebroadcast within the JavaScript app using jQuery's trigger() method. At this point, the incoming data behaves just like any other custom DOM event, and can be picked up and processed by any other object in the application.
```
malchata, line 103:
```suggestion
All of the communication between the laptop and phone happens via small JSON messages like this. Every time the game updates its state on the desktop, or the user tilts or taps a button on the phone, a WebSocket message is transmitted between the machines. In order to keep this communication simple and easy to manage, the WebSockets messages are broadcast using a single exit point from either browser. Conversely, there's a single entry point on the receiving browser, with one WebSocket object handling all incoming and outgoing messages on both ends. When a WebSocket message is received, the JSON data is rebroadcast within the JavaScript app using jQuery's `trigger()` method. At this point, the incoming data behaves just like any other custom DOM event, and can be picked up and processed by any other object in the application. 
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Wrap up
+
+Roll It is a sign of the times. Between the open source projects that powered its development, the processing power of the devices on our desks and in our pockets, and the state of the web as a platform, it's a truly exciting and transformative time to be connected on the open web. Just a few years ago, much of this technology only existed in proprietary systems, unavailable to freely use and distribute. Today, complex experiences can be realized with less work and more imagination as we create and share new pieces of the puzzle every day. So, what are you waiting for? Build something great and share it with the world!
+
+
+<figure>
```
malchata, line 135:
```suggestion

<figure>
```

src/site/content/en/blog/roll-it/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+Once the tilt values are picked up on the computer, they're interpolated over time using [requestAnimationFrame](http://updates.html5rocks.com/2012/05/requestAnimationFrame-API-now-with-sub-millisecond-precision) to keep a smooth feel. The end result is a rotating menu and a ball that rolls to help indicate the user's selection. As the phone sends tilt data, these DOM elements are updated in real time by recalculating a CSS transform inside of the requestAnimationFrame loop. The menu's container simply rotates, but the ball seems to roll along the floor. To achieve this effect, we implement some basic trigonometry to relate the balls x-coordinate to its rotation. The simple equation is: rotations = x / (diameter * π)
```
malchata, line 128:
```suggestion
Once the tilt values are picked up on the computer, they're interpolated over time using [`requestAnimationFrame`](http://updates.html5rocks.com/2012/05/requestAnimationFrame-API-now-with-sub-millisecond-precision) to keep a smooth feel. The end result is a rotating menu and a ball that rolls to help indicate the user's selection. As the phone sends tilt data, these DOM elements are updated in real time by recalculating a CSS transform inside of the `requestAnimationFrame` loop. The menu's container simply rotates, but the ball seems to roll along the floor. To achieve this effect, we implement some basic trigonometry to relate the balls x-coordinate to its rotation. The simple equation is: rotations = x / (diameter * π)
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+  - seanmiddleditch
+date: 2012-07-25
+tags:
+  - blog
+  - case-study
+---
+
+## Introduction
+
+Last summer I worked as the technical lead on a commercial WebGL game called [SONAR](https://chrome.google.com/webstore/detail/bnnkhdiodblfknjkjcifnpcmbkckmiaf).  The project took around three months to complete, and was done completely from scratch in JavaScript.  During the development of SONAR, we had to find innovative solutions to a number of problems in the new and untested HTML5 waters.  In particular, we needed a solution to a seemingly simple problem: how do we download and cache 70+ MB of game data when the player starts the game?
```
malchata, line 14:
```suggestion
Last summer I worked as the technical lead on a commercial WebGL game called [SONAR](https://chrome.google.com/webstore/detail/bnnkhdiodblfknjkjcifnpcmbkckmiaf). The project took around three months to complete, and was done completely from scratch in JavaScript. During the development of SONAR, we had to find innovative solutions to a number of problems in the new and untested HTML5 waters. In particular, we needed a solution to a seemingly simple problem: how do we download and cache 70+ MB of game data when the player starts the game?
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+Other platforms have ready-made solutions for this problem.  Most consoles and PC games load resources off of a local CD/DVD or from a hard-drive.  Flash can package all resources as part of the SWF file that contains the game, and Java can do the same with JAR files.  Digital distribution platforms like Steam or the App Store ensure that all resources are downloaded and installed before the player can even start the game.
```
malchata, line 16:
```suggestion
Other platforms have ready-made solutions for this problem. Most consoles and PC games load resources off of a local CD/DVD or from a hard-drive. Flash can package all resources as part of the SWF file that contains the game, and Java can do the same with JAR files. Digital distribution platforms like Steam or the App Store ensure that all resources are downloaded and installed before the player can even start the game.
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+HTML5 doesn’t give us these mechanisms, but it does give us all of the tools we need to build our own game resource download system.  The upside to building our own system is that we get all the control and flexibility we need, and can build a system that exactly matches our needs.
```
malchata, line 18:
```suggestion
HTML5 doesn'’'t give us these mechanisms, but it does give us all of the tools we need to build our own game resource download system. The upside to building our own system is that we get all the control and flexibility we need, and can build a system that exactly matches our needs.
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+if (xhr.readyState == 4 &amp;&amp; xhr.status == 200) {
+    callback(path, xhr.response, self);
+
+    if (--self.pending == 0) {
+    self.oncomplete();
+    }
+}
+};
+xhr.send();
+};
```
malchata, line 49:
```suggestion
function ResourceLoader() {
  this.pending = 0;
  this.baseurl = './';
  this.oncomplete = function() {};
}

ResourceLoader.prototype.request = function(path, callback) {
  var xhr = new XmlHttpRequest();
  xhr.open('GET', this.baseurl + path);
  var self = this;

  xhr.onreadystatechange = function() {
    if (xhr.readyState == 4 && xhr.status == 200) {
      callback(path, xhr.response, self);

      if (--self.pending == 0) {
        self.oncomplete();
      }
    }
  };

  xhr.send();
};
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+The `oncomplete` callback attached to the main `ResourceLoader` instance will only be called after all resources are loaded.  The game loading screen can just wait for that callback to be invoked before transitioning to the next screen.
+
+Quite a bit more can be done with this interface, of course.  As exercises for the reader, a few additional features that are worth investigating are adding progress/percentage support, adding image loading (using the Image type), adding automatic parsing of JSON files, and of course error handling.
+
+The most important feature for this article is the baseurl field, which allows us to easily switch the source of files we request.  It's easy to set up the core engine to allow a ?uselocal type of query parameter in the URL to request resources from a URL served by the same local Web server (like `python -m SimpleHTTPServer`) that served the main HTML document for the game, while using the cache system if the parameter is not set.
```
malchata, line 58:
```suggestion
The usage of this interface is pretty simple but also quite flexible. The initial game code can request some data files that describe the initial game level and game objects. These might be simple JSON files, for instance. The callback used for these files then inspect that data and can make additional requests (chained requests) for dependencies. The game objects definition file might list models and materials, and the callback for materials might then request texture images.

The `oncomplete` callback attached to the main `ResourceLoader` instance will only be called after all resources are loaded. The game loading screen can just wait for that callback to be invoked before transitioning to the next screen.

Quite a bit more can be done with this interface, of course. As exercises for the reader, a few additional features that are worth investigating are adding progress/percentage support, adding image loading (using the Image type), adding automatic parsing of JSON files, and of course, error handling.

The most important feature for this article is the baseurl field, which allows us to easily switch the source of files we request. It's easy to set up the core engine to allow a `?uselocal` type of query parameter in the URL to request resources from a URL served by the same local Web server (like `python -m SimpleHTTPServer`) that served the main HTML document for the game, while using the cache system if the parameter is not set.
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+One of the most important fields is the type field.  This is a single digit octal number which tells us what type of file the record contains.  The only two interesting record types for our purposes are regular files ('0') and directories ('5').  If we were dealing with arbitrary TAR files we might also care about symbolic links ('2') and possibly hard links ('1').
+
+Each header is followed immediately by the contents of the file described by the header (excepting file types that have no contents of their own, like directories).  The file contents are then followed by padding to ensure that every header begins on a 512-byte boundary.  Thus, to calculate the total length of a file record in a TAR file, we first must read the header for the the file.  We then add the length of the header (512 bytes) with the length of the file contents extracted from the header.  Finally, we add any padding bytes necessary to make the offset align to 512 bytes, which can be done easily by dividing the file length by 512, taking the ceiling of the number, and then multiplying by 512.
```
malchata, line 74:
```suggestion
One problem with chained loading of resources is that there's no way to get a complete byte count of all the data. The consequence of this is that there's no way to make a simple, reliable progress dialog for downloads. Since we're going to be downloading all the content and caching it, and this can take a rather long time for larger games, giving the player a nice progress dialog is pretty important.

The easiest fix for this problem (which also gives us a few other nice advantages) is to package all the resource files into a single bundle, which we'll download with a single XHR call, which gives us the progress events we need to display a nice progress bar.

Building a custom bundle file format is not terribly hard, and would even solve a few problems, but would require creating a tool for creating the bundle format. An alternative solution is to use an existing archive format for which tools already exist, and then need to write a decoder to run in the browser. We don't need a compressed archive format because HTTP can already compress data using gzip or deflate algorithms just fine. For these reasons, we settled on the TAR file format.

TAR is a relatively simple format. Every record (file) has a 512 byte header, followed by the file contents padded out to 512 bytes. The header has only a few relevant or interesting fields for our purposes, mainly the file type and name, which are stored at fixed positions within the header.

Header fields in the TAR format are stored at fixed locations with fixed sizes in the header block. For example, the last modification timestamp of the file is stored at 136 bytes from the start of the header, and is 12 bytes long. All numeric fields are encoded as octal numbers stored in ASCII format. To parse the fields, then, we extract the fields from our array buffer, and for numeric fields we call `parseInt()` being sure to pass in the second parameter to indicate the desired octal base.

One of the most important fields is the type field. This is a single digit octal number which tells us what type of file the record contains. The only two interesting record types for our purposes are regular files (`'0'`) and directories (`'5'`). If we were dealing with arbitrary TAR files we might also care about symbolic links (`'2'`) and possibly hard links (`'1'`).

Each header is followed immediately by the contents of the file described by the header (excepting file types that have no contents of their own, like directories). The file contents are then followed by padding to ensure that every header begins on a 512-byte boundary. Thus, to calculate the total length of a file record in a TAR file, we first must read the header for the the file. We then add the length of the header (512 bytes) with the length of the file contents extracted from the header. Finally, we add any padding bytes necessary to make the offset align to 512 bytes, which can be done easily by dividing the file length by 512, taking the ceiling of the number, and then multiplying by 512.
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+state.index = offset + Math.ceil(size / 512) * 512;
+
+// return the descriptor with the relevant fields we care about
+return {
+name : name,
+size : size,
+type : type,
+offset : offset
+};
+};
```
malchata, line 133:
```suggestion
// Read a string out of an array buffer with a maximum string length of 'len'.
// state is an object containing two fields: the array buffer in 'buffer' and
// the current input index in 'index'.
function readString (state, len) {
  var str = '';

  // We read out the characters one by one from the array buffer view.
  // this actually is a lot faster than it looks, at least on Chrome.
  for (var i = state.index, e = state.index + len; i != e; ++i) {
    var c = state.buffer[i];

    if (c == 0) { // at NUL byte, there's no more string
      break;
    }

    str += String.fromCharCode(c);
  }

  state.index += len;

  return str;
}

// Read the next file header out of a tar file stored in an array buffer.
// state is an object containing two fields: the array buffer in 'buffer' and
// the current input index in 'index'.
function readTarHeader (state) {
  // The offset of the file this header describes is always 512 bytes from
  // the start of the header
  var offset = state.index + 512;

  // The header is made up of several fields at fixed offsets within the
  // 512 byte block allocated for the header.  fields have a fixed length.
  // all numeric fields are stored as octal numbers encoded as ASCII
  // strings.
  var name = readString(state, 100);
  var mode = parseInt(readString(state, 8), 8);
  var uid = parseInt(readString(state, 8), 8);
  var gid = parseInt(readString(state, 8), 8);
  var size = parseInt(readString(state, 12), 8);
  var modified = parseInt(readString(state, 12), 8);
  var crc = parseInt(readString(state, 8), 8);
  var type = parseInt(readString(state, 1), 8);
  var link = readString(state, 100);

  // The header is followed by the file contents, then followed
  // by padding to ensure that the next header is on a 512-byte
  // boundary.  advanced the input state index to the next
  // header.
  state.index = offset + Math.ceil(size / 512) * 512;

  // Return the descriptor with the relevant fields we care about
  return {
    name : name,
    size : size,
    type : type,
    offset : offset
  };
};
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+I settled on loading the TAR files as an `ArrayBuffer` directly from the XHR request and adding a small convenience function for converting chunks from the `ArrayBuffer` to a string.  Currently my code only handles basic ANSI/8-bit characters, but this can be fixed once a more convenient conversion API is available in browsers.
+
+The code simply scans over the `ArrayBuffer` parsing out record headers, which includes all of the relevant TAR header fields (and a few not-so-relevant ones) as well as the location and size of the file data within the `ArrayBuffer`.  The code can also optionally extract the data as an `ArrayBuffer` view and store that in the returned record headers list.
+
+The code is freely available under a friendly, permissive Open Source license at [https://github.com/subsonicllc/TarReader.js](https://github.com/subsonicllc/TarReader.js).
```
malchata, line 144:
```suggestion
I looked around for existing TAR readers, and found a few, but none that did not have other dependencies or which would easily fit into our existing codebase. For this reason, I chose to write my own. I also took the time to optimize the loading as best as possible, and ensure that the decoder easily handles both binary and string data within the archive.

One of the first problems I had to solve was how to actually get the data loaded from an XHR request. I originally started with a "binary string" approach. Unfortunately, converting from binary strings to more readily usable binary forms like an `ArrayBuffer` is not straightforward, nor are such conversions particularly quick. Converting to `Image` objects is equally painful.

I settled on loading the TAR files as an `ArrayBuffer` directly from the XHR request and adding a small convenience function for converting chunks from the `ArrayBuffer` to a string. Currently my code only handles basic ANSI/8-bit characters, but this can be fixed once a more convenient conversion API is available in browsers.

The code simply scans over the `ArrayBuffer` parsing out record headers, which includes all of the relevant TAR header fields (and a few not-so-relevant ones) as well as the location and size of the file data within the `ArrayBuffer`. The code can also optionally extract the data as an `ArrayBuffer` view and store that in the returned record headers list.

The code is freely available under a friendly, permissive Open Source license at [https://github.com/subsonicllc/TarReader.js](https://github.com/subsonicllc/TarReader.js).
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+
+The code is a little obtuse, but it's not particularly tricky nor difficult to write.
+
+A second important thing to take into account is the file size limit of the FileSystem API's `PERSISTENT` storage.  We wanted persistent storage because the temporary storage can be cleared at any time, including while the user is in the middle of playing our game right before it tries to load the evicted file.
+
+For apps targeting the Chrome Web Store, there are no storage limits when using the `unlimitedStorage` permission in the application's manifest file.  However, regular web apps can still request space with the experimental quota request interface.
```
malchata, line 158:
```suggestion
For actually storing file contents and accessing them later, we used the FileSystem API. The API is quite new but already has some great documentation, including the excellent [HTML5 Rocks FileSystem article](https://www.html5rocks.com/tutorials/file/filesystem/).

The FileSystem API is not without its caveats. For one thing, its an event-driven interface; this both makes the API non-blocking which is great for UI but also makes it a pain to use. [Using the FileSystem API from a WebWorker](https://www.html5rocks.com/tutorials/file/filesystem-sync/) can alleviate this problem, but that would require splitting the entire downloading and unpacking system into a WebWorker. That might even be the best approach, but it's not the one I went with due to time constraints (I was not familiar with WorkWorkers yet), so I had to deal with the asynchronous event-driven nature of the API.

Our needs are mostly focused on writing out files into a directory structure. This requires a series of steps for each file. First, we need to take the file path and turn it into a list, which is easily done by splitting the path string on the path separator character (which is always the forward-slash, like URLs). Then we need to iterate over each element in the resulting list save for the last, recursively creating a directory (if necessary) in the local filesystem. Then we can create the file, and then create a `FileWriter`, and finally write out the file contents.

A second important thing to take into account is the file size limit of the FileSystem API's `PERSISTENT` storage. We wanted persistent storage because the temporary storage can be cleared at any time, including while the user is in the middle of playing our game right before it tries to load the evicted file.

For apps targeting the Chrome Web Store, there are no storage limits when using the `unlimitedStorage` permission in the application's manifest file. However, regular web apps can still request space with the experimental quota request interface.
```

src/site/content/en/blog/subsonic/index.md
```diff
@@ -0,0 +1,171 @@
...
+function allocateStorage(space_in_bytes, success, error) {
+webkitStorageInfo.requestQuota(
+webkitStorageInfo.PERSISTENT,
+space_in_bytes,
+function() {
+    webkitRequestFileSystem(PERSISTENT, space_in_bytes, success, error);      
+},
+error
+);
+}
```
malchata, line 170:
```suggestion
function allocateStorage(space_in_bytes, success, error) {
  webkitStorageInfo.requestQuota(
    webkitStorageInfo.PERSISTENT,
    space_in_bytes,
    function() {
      webkitRequestFileSystem(PERSISTENT, space_in_bytes, success, error);      
    },
    error
  );
}
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
+---
+layout: post
+title: Case Study - Real-time Updates in Stream Congress
```
malchata, line 3:
```suggestion
title: Case Study: Real-time Updates in Stream Congress
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Starting with WebSockets
+
+The WebSockets spec has gotten quite a bit of attention for what it enables: a stable, bi-directional [TCP socket](http://www.csc.villanova.edu/~mdamian/Sockets/TcpSockets.htm) between the browser and server. There is no data format imposed on the TCP socket; the developer is free to define  a messaging protocol. In practice, passing JSON objects around as strings is most convenient. The client-side JavaScript code to listen for live updates is clean and simple:
```
malchata, line 18:
```suggestion
The WebSockets spec has gotten quite a bit of attention for what it enables: a stable, bi-directional [TCP socket](http://www.csc.villanova.edu/~mdamian/Sockets/TcpSockets.htm) between the browser and server. There is no data format imposed on the TCP socket; the developer is free to define a messaging protocol. In practice, passing JSON objects around as strings is most convenient. The client-side JavaScript code to listen for live updates is clean and simple:
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+```js
+var liveSocket = new WebSocket("ws://streamcongress.com:8080/live");
+liveSocket.onmessage = function(payload) {
+addToStream(JSON.parse(payload.data).reverse());
+};
```
malchata, line 24:
```suggestion
var liveSocket = new WebSocket("ws://streamcongress.com:8080/live");

liveSocket.onmessage = function (payload) {
  addToStream(JSON.parse(payload.data).reverse());
};
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+While browser support for WebSockets is straightforward, server-side support is still in the formative stage. [Socket.IO](http://socket.io/) on Node.js provides one of the most mature and robust server-side solutions. An event-driven server like Node.js is the right fit for WebSockets. For alternative implementations, Python developers can use [Twisted](http://twistedmatrix.com/trac/) and [Tornado](http://www.tornadoweb.org/), while Ruby developers have [EventMachine](http://rubyeventmachine.com/).
+
+### Introducing Cramp
+
+[Cramp](https://github.com/lifo/cramp) is an asynchronous Ruby web framework that runs on top of EventMachine. It's written by [Pratik Naik](http://m.onkey.org/), a member of the Ruby on Rails core team. Providing a simple domain specific language (DSL) for real-time web apps, Cramp is an ideal choice for Ruby web developers. Those familiar with writing controllers in Ruby on Rails will recognize Cramp's style:
```
malchata, line 31:
```suggestion
[Cramp](https://github.com/lifo/cramp) is an asynchronous Ruby web framework that runs on top of EventMachine. It's written by [Pratik Naik](http://m.onkey.org/), a member of the Ruby on Rails core team. Providing a domain specific language (DSL) for real-time web apps, Cramp is an ideal choice for Ruby web developers. Those familiar with writing controllers in Ruby on Rails will recognize Cramp's style:
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+```js
```
malchata, line 33:
```suggestion
```ruby
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+Because Cramp sits on top of the non-blocking EventMachine, there are several considerations to keep in mind:
+
+- Non-blocking database drivers must be used, like [MySQLPlus](https://github.com/oldmoe/mysqlplus) and [em-mongo](https://github.com/bcg/em-mongo). 
+- Event-driven web servers must be used. Support is built in for [Thin](http://code.macournoyer.com/thin/) and [Rainbows](http://rainbows.rubyforge.org/).
+- The Cramp app must be run separately from the main Rails app that powers Stream Congress, restarted and monitored independently.
+
+### Current Limitations
+
+WebSockets suffered a setback on December 8, 2010 when a security vulnerability was publicized. Both [Firefox and Opera](http://hacks.mozilla.org/2010/12/websockets-disabled-in-firefox-4/) removed browser support for WebSockets. While no pure JavaScript polyfill exists, there is a [Flash fallback](https://github.com/gimite/web-socket-js/) that has  been widely adopted. However, relying on Flash is far from ideal. Even though Chrome and Safari continue to support WebSockets, it became clear that to support all modern browsers without relying on Flash, WebSockets would need to be replaced.
```
malchata, line 69:
```suggestion
WebSockets suffered a setback on December 8, 2010 when a security vulnerability was publicized. Both [Firefox and Opera](http://hacks.mozilla.org/2010/12/websockets-disabled-in-firefox-4/) removed browser support for WebSockets. While no pure JavaScript polyfill exists, there is a [Flash fallback](https://github.com/gimite/web-socket-js/) that has been widely adopted. However, relying on Flash is far from ideal. Even though Chrome and Safari continue to support WebSockets, it became clear that to support all modern browsers without relying on Flash, WebSockets would need to be replaced.
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+The decision was made to move away from WebSockets and back to "old-school" AJAX polling. While much less efficient from a disk and network I/O perspective, AJAX polling simplified the technical implementation of Stream Congress. Most significantly, the need for a separate Cramp app was eliminated. The AJAX endpoint was instead provided by the Rails app. The client-side code was modified to support jQuery AJAX polling:
+
+```js
+var fillStream = function(mostRecentActivity) {
+$.getJSON(requestURL, function(data) {
+addToStream(data.reverse());
+setTimeout(function() {fillStream(recentActivities.last());}, 15000);
+});
+};
+```
```
malchata, line 82:
```suggestion
```js
var fillStream = function(mostRecentActivity) {
  $.getJSON(requestURL, function(data) {
    addToStream(data.reverse());

    setTimeout(function() {
      fillStream(recentActivities.last());
    }, 15000);
  });
};
```
```

File: src/site/content/en/blog/sunlight-streamcongress/index.md [hunk 49]
```diff
@@ -0,0 +1,158 @@
...
+
+AJAX polling, though, is not without its downsides. Relying on the HTTP request/response cycle means that the server sees constant load even when there aren't any new updates. And of course, AJAX polling doesn't take advantage of what HTML5 has to offer.
+
+## EventSource: The right tool for the job
+
+Up to this point, a key factor was  ignored about the nature of Stream Congress: the app only needs to stream updates one way, from server to client - downstream. It didn't need to be real-time, upstream client-to-server communication.
```
malchata, line 88:
```suggestion
Up to this point, a key factor was ignored about the nature of Stream Congress: the app only needs to stream updates one way, from server to client - downstream. It didn't need to be real-time, upstream client-to-server communication. 
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+- A similar, simple JavaScript API on the browser side.
+- The open connection is HTTP-based, not dropping to the low level of TCP.
+- Automatic reconnection when the the connection is closed.
+
+### Going Back to Cramp
+
+In recent months, Cramp has added support for EventSource. The code is very similar to the WebSockets implementation:
+
+```js
```
malchata, line 102:
```suggestion
```ruby
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+routes = HttpRouter.new do
+add('/').to(LiveEvents)
+end
+run routes
+```
+
+A significant issue to keep in mind with EventSource is that cross-domain connections are not allowed. This means that the Cramp app must be served from the same streamcongress.com domain as the main Rails app. This can be accomplished with proxying at the web server. Assuming the Cramp app is powered by Thin and running on port 8000, the Apache configuration looks like so:
+
+```js
```
malchata, line 124:
```suggestion
```apache
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+Options -MultiViews
+</Directory>
+
+<Proxy balancer://thin>
+BalancerMember http://localhost:8000
+</Proxy>
+ProxyPass /live balancer://thin/
+ProxyPassReverse /live balancer://thin/
+ProxyPreserveHost on
+
```
malchata, line 146:
```suggestion
LoadModule  proxy_module             /usr/lib/apache2/modules/mod_proxy.so
LoadModule  proxy_http_module        /usr/lib/apache2/modules/mod_proxy_http.so
LoadModule  proxy_balancer_module    /usr/lib/apache2/modules/mod_proxy_balancer.so

<VirtualHost *:80>
  ServerName streamcongress.com
  DocumentRoot /projects/streamcongress/www/current/public
  RailsEnv production
  RackEnv production

  <Directory /projects/streamcongress/www/current/public>
    Order allow,deny
    Allow from all
    Options -MultiViews
  </Directory>

  <Proxy balancer://thin>
    BalancerMember http://localhost:8000
  </Proxy>

  ProxyPass /live balancer://thin/
  ProxyPassReverse /live balancer://thin/
  ProxyPreserveHost on
```

src/site/content/en/blog/sunlight-streamcongress/index.md
```diff
@@ -0,0 +1,158 @@
...
+</VirtualHost>
+```
+
+This configuration sets an EventSource endpoint  at `streamcongress.com/live`.
```
malchata, line 150:
```suggestion
This configuration sets an EventSource endpoint at `streamcongress.com/live`.
```

malchata (approved):
Nits, but LGTM!

rachelandrew:
@simgobb this one has conflicts.

PR #7937, opened 2022-05-17
[Local Font article] Add desktop-only

tomayac:

- Add desktop-only.


src/site/content/en/blog/local-fonts/index.md
```diff
@@ -179,6 +179,11 @@ if ('queryLocalFonts' in self) {
 }
 ```
 
+{% Aside %}
+The Local Font Access API is currently only available on desktop versions of Chromium, but not on
+Android.
```
beaufortfrancois, line 184: I would not mention Android. As in iOS should be mentioned as well and others.
Or maybe say "mobile" instead of Android. Your choice.

beaufortfrancois (approved):
LGTM with nit

jpmedley (approved):

PR #7958, opened 2022-05-20
Support url=... query parameter in /measure/

### LINKED ISSUE #7957: [Feature Request] AutoFill & AutoStart benchmark from URL parameters like on "PageSpeed"
I use link on websites prefilled URLs to benchmarks so when someone clicks on the link a new fresh benchmark is performed.

With [PageSpeed](https://pagespeed.web.dev) I do it like this:
https://pagespeed.web.dev/report?url=https://www.heldmayer.com

But with [web.dev/measure/](https://web.dev/measure/) I still do have some trouble, as the same thing here does not work:
https://web.dev/measure/?url=https://www.heldmayer.com

Are there any other workarounds, to create a link that automatically and prefilled performs a benchmark at [web.dev/measure/](https://web.dev/measure/)?

Thanks in advance!

jeffposnick:
Fixes #7957

This adds support for navigating to URLs like http://web.dev/measure/?url=https%3A%2F%2Fexample.com which will prefill the `value` of the `<input>` field used to determine the URL that's being measured.

It also updates the current page's URL to add in the corresponding `url=...` query parameter when a report is started.

### DIFF
--- a/src/lib/components/UrlChooser/index.js
+++ b/src/lib/components/UrlChooser/index.js
@@ -159,7 +159,7 @@ class UrlChooser extends BaseElement {
 
     const currentURL = new URL(location.href);
     currentURL.searchParams.set('url', this._urlInput.value);
-    history.pushState({}, '', currentURL.href);
+    history.replaceState({}, '', currentURL.href);
   }
 
   onSwitchUrl() {
--- a/src/lib/components/UrlChooser/index.js
+++ b/src/lib/components/UrlChooser/index.js
@@ -30,6 +30,9 @@ class UrlChooser extends BaseElement {
   }
 
   render() {
+    const currentURL = new URL(location.href);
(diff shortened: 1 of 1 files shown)


src/lib/components/UrlChooser/index.js
```diff
@@ -50,6 +53,7 @@ class UrlChooser extends BaseElement {
             pattern="https?://.+"
             minlength="7"
             @keyup="${this.onUrlKeyup}"
+            value="${urlSearchParam}"
```
jeffposnick, line 56: I'm pretty sure that the `html` template tag handles all the necessary escaping here, and that it's safe to pass in user-defined input. If that's not the case, please let me know.

devnook: I think it would be better to pass it as a param into the component. But since this component is not reused atm, let's keep it as is.

devnook (approved):

PR #7980, opened 2022-05-23
Fix d.c.c. redirects

### LINKED ISSUE #7979: Redirects to other domains only work with trailing `/`
If the user enters a URL without the trailing slash `/`, they are redirected to the version with `/` (for example, https://web.dev/svgcode → https://web.dev/svgcode/). This feature is lost with redirects to other domains:

❌ Doesn't work: https://web.dev/fugu-status → N/A
✅ Works: https://web.dev/fugu-status/ → https://developer.chrome.com/blog/fugu-status/

jeffposnick:
Fixes #7979

In general, the `source` field should not have a trailing `/`, as per https://github.com/GoogleChrome/web.dev/blob/5475eaf29bc1f8fae3d75018f5a6da591f5f90e4/redirects.yaml#L1-L6

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -291,25 +291,23 @@ redirects:
     destination: /media-eme
 
   # Redirects to https://developer.chrome.com/.
-  - source: /browser-flags/
+  - source: /browser-flags
     destination: https://developer.chrome.com/blog/browser-flags/
-  - source: /origin-trials/
-    destination: https://developer.chrome.com/blog/origin-trials/
-  - source: /third-party-origin-trials/
+  - source: /third-party-origin-trials
     destination: https://developer.chrome.com/blog/third-party-origin-trials/
-  - source: /fugu-status/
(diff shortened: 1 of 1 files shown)


redirects.yaml
```diff
@@ -291,25 +291,23 @@ redirects:
     destination: /media-eme
 
   # Redirects to https://developer.chrome.com/.
-  - source: /browser-flags/
+  - source: /browser-flags
     destination: https://developer.chrome.com/blog/browser-flags/
-  - source: /origin-trials/
```
jeffposnick, line 296: There are two `/origin-trials/` entries in this file. Only one of them will "win."

https://developer.chrome.com/docs/web-platform/origin-trials/ and https://developer.chrome.com/blog/origin-trials/ are the two destinations. The non-`/blog/` one seems more "authoritative," so I'm removing the `/blog/` entry.

sofiayem (approved):

PR #7998, opened 2022-05-26
Add SPC to payments collection

heyawhite:
This PR is most certainly not going to work. It's a placeholder for updates, per https://github.com/GoogleChrome/developer.chrome.com/pull/2906

AI: Get documentation for handbook on how to create and add new items to a collection.

### DIFF
--- a/src/site/_data/paths/payments.json
+++ b/src/site/_data/paths/payments.json
@@ -36,13 +36,21 @@
           ]
         },
         {
-          "title": "i18n.paths.payments.topics.build_a_web_based_payment_app",
-          "pathItems": [
-            "web-based-payment-apps-overview",
-            "registering-a-web-based-payment-app",
-            "orchestrating-payment-transactions",
-            "handling-optional-payment-information"
-          ]
+           "title": "i18n.paths.payments.topics.build_a_web_based_payment_app",
+           "pathItems": [
+             "web-based-payment-apps-overview",
+             "registering-a-web-based-payment-app",
+             "orchestrating-payment-transactions",
+             "handling-optional-payment-information"
+           ]
+        },
+        {
+          "title": "i18n.paths.payments.topics.secure_payment_confirmation",
+	    "urls": [
+		   "https://developer.chrome.com/articles/secure-payment-confirmation",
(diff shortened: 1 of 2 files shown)


src/site/_data/paths/payments.json
```diff
@@ -36,13 +36,21 @@
...
+           "pathItems": [
+             "web-based-payment-apps-overview",
+             "registering-a-web-based-payment-app",
+             "orchestrating-payment-transactions",
+             "handling-optional-payment-information"
+           ]
+        },
+        {
+          "title": "i18n.paths.payments.topics.secure_payment_confirmation",
+	    "urls": [
```
heyawhite: This should be pathItems *but* these live on dcc. What's the protocol?

rachelandrew: I think we need another way to do this, because we want to make it clear that these are elsewhere and have a different status that things on this site. I think Ewa is out today but we likely need her input.

heyawhite: Definitely yes -- it will need i18n paths. My assumption is that there isn't an existing way. Once we've sorted it out, I'll make handbook updates

agektmr: Not sure if I understand the issue here precisely but /identity could have external links without any problems. https://web.dev/identity/
Only thing I wish we had is a clear indication that a link is for an external site.

rachelandrew: yeah we seem to be able to do it, my suggestion for now is we put the external links in one group with a clear heading.

rachelandrew: @heyawhite I got this to work https://deploy-preview-7998--web-dev-staging.netlify.app/payments/

We should bikeshed the wording, hopefully we can come up with a better method to indicate the experimental things (maybe we need those icons of Jake and Una in lab coats?)

heyawhite: That makes sense to me. I'm going to open a new issue around that, but publish for now. We can continue to address the question of collections / topic organization, after publication.

rachelandrew: I'll approve this then and you can merge after merging the articles on d.c.c.

rachelandrew (approved):

PR #8002, opened 2022-05-26
Fixes patterns landing page table of contents copy to match blog posts

argyleink:

- Patterns page now says "On this page" instead of "Layout patterns"

It's confusing going to a patterns page that isnt about layout and seeing the ToC titled "Layout patterns" when really it's the same component as a blog post has which reads "On this page". This normalizes the text and disambiguates pattern detail pages table of contents.

side note / feature request:
if the pattern page has no sub categories, this ToC component shouldn't be present on the page because it's pointless to show "Table of contents > All". thoughts?


jeffposnick (approved):
Thanks, Adam!

PR #8011, opened 2022-05-27
Clean up fenced code blocks in Web Push article

jeffposnick:
There are some issues with https://web.dev/push-notifications-web-push-protocol/ following the migration to web.dev, due to the way the Markdown fenced code blocks were translated over.

Fenced blocks marked as `text` don't get their `>` and `<` characters automatically escaped, for one thing. There is also some extra indentation in some JavaScript blocks.


rachelandrew (approved):
thank you!

PR #8013, opened 2022-05-28
Add vietnamese language

hackerpro536:
Add vi_VN language to website, make more people in our country can understand and learn on website easy more.


rachelandrew:
@hackerpro536 thanks for your PR however we don't currently have plans to translate content to Vietnamese. We support only a small number of languages currently as we have limited capacity to translate content.

PR #8035, opened 2022-05-31
Use correct attribute for image source

lejtzen:
The `<source>` tag needs the `srcset` attribute when used as a child of the `<picture>` tag. Only when used in `<audio>` and `<video>` the `src` attribute is valid. With this change the correct sources will be used on the different breakpoints.


lejtzen:
I'm not able to add the `$-presubmit` label. How should I proceed with this PR?

malchata (approved):
LGTM.

PR #8051, opened 2022-06-01
Add polyfill

tomayac:
Thanks, @rauschma.

- Add polyfill

### DIFF
--- a/src/site/content/en/blog/streams/index.md
+++ b/src/site/content/en/blog/streams/index.md
@@ -7,7 +7,7 @@ description: |
 authors:
   - thomassteiner
 date: 2021-02-19
-updated: 2022-01-11
+updated: 2022-06-01
 hero: image/8WbTDNrhLsU0El80frMBGE4eMCD3/TuciUuOQOd3u7uMgDZBi.jpg
 alt: A forest stream with colored fallen leaves.
 tags:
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/streams/index.md
```diff
@@ -341,37 +341,24 @@ for await (const chunk of stream) {
...
-
-  return {
-    next() {
-      // Stream reads already resolve with {done, value}, so
-      // we can just call read:
-      return reader.read();
-    },
-    return() {
-      // Release the lock if the iterator terminates.
+if (ReadableStream.prototype[Symbol.asyncIterator]) {
```
jeffposnick, line 347: Shouldn't there be a leading `!` here, so that you only do the assignment if it's not already defined?

jeffposnick (changes requested):

src/site/content/en/blog/streams/index.md
tomayac: Duh, of course. Forgot to negate after testing in the Console. Good catch.

tomayac, line 347:
```suggestion
if (!ReadableStream.prototype[Symbol.asyncIterator]) {
```

jeffposnick (approved):

PR #8094, opened 2022-06-06
update sanitizer api article align with current release plan

### LINKED ISSUE #8093: Update Sanitizer api article align with new release plan

https://web.dev/sanitizer/

Sanitizer API will be shipped only with MVP API's described below.

https://groups.google.com/a/chromium.org/g/blink-dev/c/KOpwkS-bgl0/m/hmJ6gfU7AQAJ

Article should only rely on `setHTML()` not `sanitize()` and `sanitizeFor()`

### What's the deadline?

M105

#### Is it a hard deadline?

Yes. To avoid confusing among developers.

JackJey:
Fixes #8093 

- remove `sanitize()` from article
- remove `sanitizeFor()` too
- update status table

CC: @otherdaniel

### DIFF
--- a/src/site/content/en/blog/sanitizer/index.md
+++ b/src/site/content/en/blog/sanitizer/index.md
@@ -76,8 +76,13 @@ The [Sanitizer API](https://wicg.github.io/sanitizer-api/) is used in the follow
 ```js
 const $div = document.querySelector('div')
 const user_input = `<em>hello world</em><img src="" onerror=alert(0)>`
-const sanitizer = new Sanitizer()
-$div.setHTML(user_input, sanitizer) // <div><em>hello world</em><img src=""></div>
+$div.setHTML(user_input, { sanitizer: new Sanitizer() }) // <div><em>hello world</em><img src=""></div>
+```
+
+However, the `{ sanitizer: new Sanitizer() }` is the default argument. So it can be just like below.
+
+```js
+$div.setHTML(user_input) // <div><em>hello world</em><img src=""></div>
 ```
 
 It is worth noting that `setHTML()` is defined on [`Element`](https://developer.mozilla.org/docs/Web/API/Element). Being a method of `Element`, the context to parse is self-explanatory (`<div>` in this case), the parsing is done once internally, and the result is directly expanded into the DOM.
@@ -85,9 +90,10 @@ It is worth noting that `setHTML()` is defined on [`Element`](https://developer.
 To get the result of sanitization as a string, you can use `.innerHTML` from the `setHTML()` results.
 
 ```js
-$div.setHTML(user_input, sanitizer).innerHTML // <em>hello world</em><img src="">
+const $div = document.createElement('div')
+$div.setHTML(user_input)
+$div.innerHTML // <em>hello world</em><img src="">
 ```
-
 
 ### Customize via configuration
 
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):
LGTM.

src/site/content/en/blog/sanitizer/index.md
```diff
@@ -78,32 +78,14 @@ const $div = document.querySelector('div')
...
 const sanitizer = new Sanitizer()
 $div.setHTML(user_input, sanitizer) // <div><em>hello world</em><img src=""></div>
-
 ```
 
 It is worth noting that `setHTML()` is defined on [`Element`](https://developer.mozilla.org/docs/Web/API/Element). Being a method of `Element`, the context to parse is self-explanatory (`<div>` in this case), the parsing is done once internally, and the result is directly expanded into the DOM.
 
-If you don't want to expand directly into the DOM, you can also get the result as an `HTMLElement`.
-
+If you want to get the result of sanitization as a string, you can use `.innerHTML` from the `setHTML()` results.
```
mihajlija, line 85:
```suggestion
To get the result of sanitization as a string, you can use `.innerHTML` from the `setHTML()` results.
```

mihajlija (approved):

JackJey:
Please block for @otherdaniel 's approval.

rachelandrew:
@JackJey I've added the do not merge label until we get that.

src/site/content/en/blog/sanitizer/index.md
```diff
@@ -78,32 +78,14 @@ const $div = document.querySelector('div')
...
-```js
-sanitizer.sanitizeFor("div", user_input).innerHTML // <em>hello world</em><img src="">
-```
-
-If you already have a user-controlled `DocumentFragment` and you want to remove only the harmful stuff from it, use `sanitize()`.
-
-
-```js
-$div.replaceChildren(s.sanitize($userDiv));
+$div.setHTML(user_input, sanitizer).innerHTML // <em>hello world</em><img src="">
```
otherdaniel, line 88: This shouldn't quite work, because .setHTML returns void. There's also one more thing (that I think is important): If $div is connected, that is, if it's part of the main document, then inserting it would trigger some actions. For example, it would start loading images.

I think the concept is important, though. I'd describe this as follows:

If you don't want to expand directly into the DOM, you can create a new element and sanitize into that.
const div = document.createElement("div");
div.setHTML(user_input, sanitizer);
return div.innerHTML;

This has the advantage the $div is definitely not connected to the DOM, and thus neither loads images (or other content), and doesn't have active event handlers.

-----------------

There is some ongoing discussion whether we should improve the API to allow this directly. But as it presently stands, this is probably the best way to do it.

One interesting thing - which I'm just dropping here, even though it doesn't directly apply to this issue - is a paper that was recently published on sanitizer use in HTML (https://swag.cispa.saarland/papers/klein2022hand.pdf). I think it's interesting because it gives us more information about the audience of the Sanitizer API, and hence this article: They found that many sites "sanitize" their input using string operations (.replace, or regexp matching), and often do it wrong. IMHO, these people are our primary audience, and I hope it pays off to be a bit more clean on how we think this should be used.

src/site/content/en/blog/sanitizer/index.md
```diff
@@ -138,16 +120,16 @@ The following options specify how the sanitization result should treat the speci
 ```js
 const str = `hello <b><i>world</i></b>`
 
-new Sanitizer().sanitizeFor("div", str)
+$div.setHTML(new Sanitizer(), str)
```
otherdaniel, line 123: [Here and below:]

The syntax is:
.setHTML(string, {sanitizer: s});

For the specific example:
$div.setHTML(str, {sanitizer: new Sanitizer()});

However, the "new Sanitizer()" is the default. 

Maybe we can teach it this way:
- default case .setHTML(str);   // note: no sanitizer. Still sanitizes, but with default settings.
- other cases: .setHTML(str, {sanitizer: new Sanitizer({....})});  // advanced usage.

src/site/content/en/blog/sanitizer/index.md
otherdaniel, line 88: Hah. You still have to tell us what user_input is. I think we need to keep that line for this example to make sense. :-)

otherdaniel (commented):
The general structure looks good!

Two minor issues:
- Order of arguments + options dictionary (in several examples)
- setHTML returns void, so one example needs to be modified.

JackJey:
Thanks @otherdaniel 

I fixed sample codes.

JackJey:
Thanks @otherdaniel !

PR #8099, opened 2022-06-06
Adds lazy loading guidance to LCP metric page.

malchata:

- Adds lazy loading guidance to LCP metric page.

### DIFF
--- a/src/site/content/en/metrics/lcp/index.md
+++ b/src/site/content/en/metrics/lcp/index.md
@@ -387,7 +387,7 @@ Lazy loading is a common performance optimization in which images and other elem
 
 However, lazy loading can affect LCP negatively when used improperly. As a general rule, these is one rule of lazy loading to follow if your LCP candidate is an image: If an LCP candidate is above the fold (or reasonably close), do _not_ lazy load it. The reason for this depends on how you're lazy loading images:
 
-- **If you're using JavaScript to lazy load images**, lazy loading your LCP candidate will delay LCP because the image will be dependent on a script in order to display. This not only causes a delay in that the necessary JavaScript requires time to parse, compile, and execute, but [such a pattern defeats the browser preload scanner](https://web.dev/preload-scanner/#lazy-loading-with-javascript).
+- **If you're using JavaScript to lazy load images**, lazy loading your LCP candidate will delay LCP because the image will be dependent on a script in order to display. This not only causes a delay in that the necessary JavaScript requires time to parse, compile, and execute, but [such a pattern defeats the browser preload scanner](/preload-scanner/#lazy-loading-with-javascript).
 - **If you're using [browser-level lazy loading]()**, specifying `loading=lazy` on your LCP candidate image will cause the browser to deprioritize the request for the image, allocating it _less_ bandwidth during startup.
 
 Both of these will result in a slower LCP, so avoid these patterns. Regardless of your lazy loading method, it's worth pointing out that lazy loading below-the-fold imagery will result in less bandwidth contention, meaning that your LCP candidate image could load faster.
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/lcp/index.md
```diff
@@ -381,6 +381,17 @@ performance techniques that can also improve LCP, see:
 * [Optimize your JavaScript](/fast#optimize-your-javascript) (for
   client-rendered sites)
 
+### When and when not to lazy load images
+
+Lazy loading is a common performance optimization in which images and other elements are deferred from loading until they approach the viewport as the user scrolls. It's a time-tested technique that saves bandwidth and improves load performance.
+
+However, lazy loading can affect LCP negatively when used improperly. As a general rule, these is one rule of lazy loading to follow if your LCP candidate is an image: If an LCP candidate is above the fold (or reasonably close), do _not_ lazy load it. The reason for this depends on how you're lazy loading images:
```
philipwalton, line 388: s/these is/there is/

philipwalton, line 388: Saying "rule" twice here reads a bit awkwardly. How about "In general, there is one rule..."?

src/site/content/en/metrics/lcp/index.md
```diff
@@ -381,6 +381,17 @@ performance techniques that can also improve LCP, see:
...
+
+- **If you're using JavaScript to lazy load images**, lazy loading your LCP candidate will delay LCP because the image will be dependent on a script in order to display. This not only causes a deglay in that the necessary JavaScript requires time to parse, compile, and execute, but [such a pattern defeats the browser preload scanner](/preload-scanner/#lazy-loading-with-javascript).
```
philipwalton, line 390: s/deglay/delay/

src/site/content/en/metrics/lcp/index.md
philipwalton, line 388: Also, I wonder if we want to make this advice stronger and more concise. E.g. "Never lazy load any images that are likely to be your LCP element."

Since an LCP candidate needs to be in the viewport in order to be an LCP candidate, I'm not sure we want to talk about "above the fold". My sense is the guidance is two steps:

1. Use devtools/lighthouse to figure out what your LCP element is across a number of common screen sizes
2. Make sure you're not lazy loading any images identified in No. 1

philipwalton (commented):
Overall I think this is good content that we should publish, but my concern is that the LCP metric doc might not be the right place for it.

WDYT? There are lots of anti-patterns that are important for us to tell people not to do, but we don't include any of those in the LCP metric doc currently. Should we?

Aside: I've been meaning to update the "How to improve LCP" section of this doc now that I've updated the Optimize LCP doc, so I'm open to exploring new ways to organize this content across these two docs.

src/site/content/en/metrics/lcp/index.md
malchata:
```suggestion
However, lazy loading can affect LCP negatively when used improperly. In general, there is one rule of lazy loading to follow if your LCP candidate is an image: If an LCP candidate is above the fold (or reasonably close), do _not_ lazy load it. The reason for this depends on how you're lazy loading images:
```

src/site/content/en/metrics/lcp/index.md
```diff
@@ -4,7 +4,7 @@ title: Largest Contentful Paint (LCP)
 authors:
   - philipwalton
```
malchata, line 5: @philipwalton, should we keep this to one author? Not sure what the threshold is for adding a byline when you add content.

malchata:
> Overall I think this is good content that we should publish, but my concern is that the LCP metric doc might not be the right place for it.
> 
> WDYT? There are lots of anti-patterns that are important for us to tell people not to do, but we don't include any of those in the LCP metric doc currently. Should we?
> 
> Aside: I've been meaning to update the "How to improve LCP" section of this doc now that I've updated the Optimize LCP doc, so I'm open to exploring new ways to organize this content across these two docs.

IMO, we should include the anti-patterns in that optimize guide, but link out to them in the docs they reside in.

PR #8105, opened 2022-06-07
Don't set inert on web-header when nav drawer opens

### LINKED ISSUE #7749: Burger Menu can not navigate to page or dismiss
**Describe the bug**
You can not navigate on mobile via the burger menu 😅

**To Reproduce**
1. Go to https://web.dev/ on a small screen/mobile
2. Click on burger menu
3. Can not navigate to a page
4. Can not close overlay

**Expected behavior**
- Burger Menu allows me to navigate to different pages
- I can close the Burger menu by clicking outside or on the "x"

**Version Info**
- OS: macOS Monterey (v12.3)
- Browser: Chrome: Version 100.0.4896.127 (Official Build) (x86_64)

**Video**

https://user-images.githubusercontent.com/24378/164080292-05c44c3a-f378-489e-9168-9549acabd268.mp4

jeffposnick:
Fixes #7749

This seems to behave reasonably, as far as I can tell, in Safari and Chrome (which now natively support `inert`) as well as in Firefox (where the `wicg-inert` polyfill will still get used).

My guess is that there's a difference between native `inert` and the polyfill with regards to the children of a parent element that has `inert` set on it—but that's just a guess. In any case, removing `inert` from the parent `web-header` when the navigation drawer is open looks like it makes the navigation drawer interactive again.

### DIFF
--- a/src/lib/actions.js
+++ b/src/lib/actions.js
@@ -115,13 +115,10 @@ const disablePage = () => {
   /** @type {HTMLElement|object} */
   const main = document.querySelector('main') || {};
   /** @type {HTMLElement|object} */
-  const header = document.querySelector('web-header') || {};
-  /** @type {HTMLElement|object} */
   const footer = document.querySelector('.w-footer') || {};
 
   document.body.classList.add('overflow-hidden');
   main.inert = true;
-  header.inert = true;
   footer.inert = true;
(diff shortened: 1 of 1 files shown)


agektmr (approved):
LGTM

PR #8134, opened 2022-06-13
Make web-navigation-drawer inert when search UI is open

### LINKED ISSUE #8131: Menu Links Still Clickable on Website When Search is Open
**Describe the bug**

When the search bar is open, the main nav links are still right there in the bar and are still clickable if you are clickout out of the search box, removing focus. You end up on another page completely. Short screen recording of the bug happening on Chrome 102 below.

**Screenshots**
Screen Recording of Bug - https://app.claap.io/threed-software/search-bug-on-web-dev-c-HGm4RhKene-Tq448OJK57E3

jeffposnick:
Fixes #8131

It would be great to get your take on this approach, @devnook, as I feel like I'm addressing these in an ad hoc fashion, and don't really understand how, e.g., the whole actions/state model for controlling how elements on the page behave when specific actions are being performed.

That being said, I'm basically re-adding the same logic around setting `inert` that was removed #8105, except:

- only when the search UI interface is open **and**
- only on the `<web-navigation-drawer>` element, not `<web-header>` (since the search element is a child of `<web-header>`)

### DIFF
--- a/src/lib/components/Search/index.js
+++ b/src/lib/components/Search/index.js
@@ -305,11 +305,6 @@ class Search extends BaseStateElement {
     this.updateComplete.then(() => {
       this.inputEl.focus();
     });
-
-    // See https://github.com/GoogleChrome/web.dev/issues/8131
-    /** @type {HTMLElement|object} */
-    const navDrawer = document.querySelector('web-navigation-drawer') || {};
-    navDrawer.inert = true;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/lib/components/Search/index.js
```diff
@@ -306,14 +305,23 @@ class Search extends BaseStateElement {
     this.updateComplete.then(() => {
       this.inputEl.focus();
     });
+
+    // See https://github.com/GoogleChrome/web.dev/issues/8131
+    /** @type {HTMLElement|object} */
+    const navDrawer = document.querySelector('web-navigation-drawer') || {};
+    navDrawer.inert = true;
```
devnook, line 312: I think the drawer should react to isSearchExpanded: true change of state instead. Here's the state listener for drawer: https://github.com/GoogleChrome/web.dev/blob/f5ccf1921c92cc60fc21146441b6b733863d8d3e/src/lib/components/NavigationDrawer/index.js#L89

Since search already updates that field in the state store, Drawer can use it (instead of having inert state set from outside.)

jeffposnick:
Thanks for the feedback! Please take another look at this alternative implementation.

devnook (approved):

PR #8198, opened 2022-06-17
BrowserCompat UI for deprecated and flagged features

### LINKED ISSUE #7798: The `BrowserCompat` macro should indicate when features are deprecated
The new How-to section that @tomayac is working on lists both an old and a new way to accomplish a task. Some of the old methods of doing things are only provided as a fallback while the new gains market share. 

Many of them are deprecated. The BrowserCompat macro should show them as such.

The records in the browser compatibility data contain a boolean member at *`category.feature`*`.status.deprecated` to indicate whether the *specification* is deprecated. For example, the deprecation status of `navigator.canShare()` is located at `api.Navigator.canShare.status.deprecated`. 

When `deprecated` is false, the BrowserCompat macro should show it in some way.

We could add a note to the how-to articles. Since this is being tracked for us anyway, we should take advantage of the available data.

**Additional context**
If we find that the deprecation status of a feature is out of date, we can submit a PR to the compat database and all users of that data including MDN will benefit.

### LINKED ISSUE #7936: BrowserCompat widget should not mark features behind flags as supported
While reading https://web.dev/state-of-css-2022/, I noticed that the widget does not look at Experimental Flags.

Just to highlight one issue, [color-mix()](https://web.dev/state-of-css-2022/#color-mix()) is [only available behind Flags](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix#browser_compatibility) in Firefox 88 and Safari 15. But the widget displays a green background.

jeffposnick:
Fixes #7936 and fixes #7798

This adds logic to the existing `BrowserCompat` shortcode to do two things:

- If a feature is currently marked as being available behind a flag, show a flag icon, not a version number for the browser:
Screen Shot 2022-06-17 at 2 38 22 PM

- If a feature is deprecated, show the trash icon, not a version number for the browser:
Screen Shot 2022-06-17 at 2 42 01 PM

### DIFF
--- a/src/scss/blocks/_browser-compat.scss
+++ b/src/scss/blocks/_browser-compat.scss
@@ -69,6 +69,16 @@
     color: get-utility-value('color', 'state-warn-text');
   }
 
+  &[data-compat='flag'] {
+    background: get-utility-value('bg', 'state-warn-bg');
+    color: get-utility-value('color', 'state-warn-text');
+  }
+
+  &[data-compat='deprecated'] {
+    background: get-utility-value('bg', 'state-bad-bg');
+    color: get-utility-value('color', 'state-bad-text');
+  }
+
   &[data-compat='no'] {
     background: get-utility-value('bg', 'state-bad-bg');
     color: get-utility-value('color', 'state-bad-text');
--- a/src/site/_data/i18n/browser_compat.yml
+++ b/src/site/_data/i18n/browser_compat.yml
@@ -42,3 +42,9 @@ preview:
   pt: Visualizar
   ru: Предварительный просмотр
   zh: 预览
+
+flag:
+  en: Behind a flag
+
+deprecated:
+  en: Deprecated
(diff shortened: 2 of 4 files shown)


src/scss/blocks/_browser-compat.scss
```diff
@@ -69,6 +69,16 @@
     color: get-utility-value('color', 'state-warn-text');
   }
 
+  &[data-compat='flag'] {
```
jeffposnick, line 72: We could theoretically collapse these into the existing `preview` and `no` styles, but I figured we might want to style them separately at some point?

src/site/_data/i18n/browser_compat.yml
```diff
@@ -42,3 +42,9 @@ preview:
   pt: Visualizar
   ru: Предварительный просмотр
   zh: 预览
+
+flag:
```
jeffposnick, line 46: I don't know what the process is for requesting a translation, so for now, localizations will have to fall back to the English.

rachelandrew (approved):
This looks great - thank you!

jeffposnick:
https://deploy-preview-8198--web-dev-staging.netlify.app/state-of-css-2022/#color-mix() has a live example of the Flag, and https://deploy-preview-8198--web-dev-staging.netlify.app/handbook/web-dev-components/#browsercompat is an example of the Trash.

jpmedley (approved):
Awesome! Thank you.

PR #8245, opened 2022-06-28
Adds Farfetch Core Web Vitals case study

malchata:

- Adds the Farfetch Core Web Vitals case study.


rachelandrew (approved):
A bunch of suggestions and it does need an image, but I've approved so as not to block it.

PR #8251, opened 2022-06-30
New on the web June

rachelandrew:
Adds the June web platform post.

PR #8252, opened 2022-06-30
Update @mdn/browser-compat-data to v5.x

jeffposnick:
As discussed with @rachelandrew, this is a major semver bump but from what I can tell, the breaking changes are related to TypeScript definitions that we are not dependent on. Things look okay when tested locally.

Still, we're breaking this up into a separate PR to accompany #8251 to keep it simpler if we need to roll things back.


jeffposnick:
https://deploy-preview-8252--web-dev-staging.netlify.app/web-platform-04-2022/ and https://deploy-preview-8252--web-dev-staging.netlify.app/web-platform-05-2022/ both seem to render fine with this change.

rachelandrew (approved):

PR #8260, opened 2022-07-01
Update bfcache article with section on measuring cache ratio hit

tunetheweb:

- Adds a section on measuring bfcache hit ratio

Also includes some other misc changes
- Removed unnecessary `=== true` code section as follow on from #4272
- Adds case study from Smashing Magazine (full disclosure - I wrote this one).

### DIFF
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -533,10 +533,10 @@ the bfcache, including:
 - when the user duplicates a tab
 - when the user closes a tab and uncloses it
 
-So, website owners are not aiming for 100% bfcache hit ratio on all
-`back_forward` navigations, but measuring their ratio can be useful to identify
-pages where the page itself is preventing bfcache usage for a high proportion
-of navigations.
+So, website owners should not be expecting a 100% bfcache hit ratio for all
+`back_forward` navigations. However, measuring their ratio can be useful to
+identify pages where the page itself is preventing bfcache usage for a high
+proportion of back and forward navigations.
 
 The Chrome team is working on a
 (`NotRestoredReason API`)[https://github.com/rubberyuzu/bfcache-not-retored-reason/blob/main/NotRestoredReason.md]
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -87,7 +87,7 @@ Chrome usage data shows that 1 in 10 navigations on desktop and 1 in 5 on mobile
 are either back or forward. With bfcache enabled, browsers could eliminate the
 data transfer and time spent loading for billions of web pages every single day!
 
-### How the "cache" works {: #how-the-cache-works }
+### How the "cache" works
 
 The "cache" used by bfcache is different from the [HTTP cache](/http-cache/)
 (which is also useful in speeding up repeat navigations). The bfcache is a
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -501,7 +501,7 @@ window.addEventListener('pageshow', (event) => {
 ### Measuring your bfcache hit ratio
 
 As well as measuring pageview measurements for bfcache page navigations, you
-may wish to also track whether the bfcached was used to help identify pages
+may wish to also track whether the bfcache was used to help identify pages
 that are not utilising the bfcache. For example:
 
 ```js
@@ -520,7 +520,8 @@ window.addEventListener('pageshow', (event) => {
     gtag('event', 'bfcache', {
       'event_category': 'page_load_details',
       'event_label': window.location.pathname,
-      'value': event.persisted });
+      'value': event.persisted
+    });
   }
 });
 ```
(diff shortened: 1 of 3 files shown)


malchata (approved):
LGTM, small nits aside.

src/site/content/en/blog/bfcache/index.md
```diff
@@ -491,13 +491,59 @@ should be similar for other analytics tools:
...
+
+window.addEventListener('pageshow', (event) => {
+
+  // Send another pageview if the page is restored from bfcache.
+  if (event.persisted) {
     gtag('event', 'page_view');
   }
+
+  // For Back/Forward navigations, send track bfcache usage.
+  if (performance.getEntriesByType('navigation')[0].type == "back_forward") {
```
philipwalton, line 519: There needs to be an `else` here because a page that was previously of type `back_forward` could later be restored from bfcache, and you wouldn't want to send two events.

src/site/content/en/blog/bfcache/index.md
```diff
@@ -491,13 +491,59 @@ should be similar for other analytics tools:
...
+    gtag('event', 'bfcache', {
+      'event_category': 'page_load_details',
+      'event_label': window.location.pathname,
+      'value': event.persisted
+    });
+  }
 });
 ```
 
+It is important to realise that there are a number of scenarios, outside
```
philipwalton, line 529: s/realise/realize/

philipwalton (changes requested):

src/site/content/en/blog/bfcache/index.md
```diff
@@ -491,13 +491,59 @@ should be similar for other analytics tools:
...
+    gtag('event', 'page_view');
+  }
+});
+```
+
+### Measuring your bfcache hit ratio
+
+As well as measuring pageview measurements for bfcache page navigations, you
+may wish to also track whether the bfcache was used to help identify pages
+that are not utilising the bfcache. For example:
```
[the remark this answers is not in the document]
tunetheweb: This is gonna catch me a lot :-)

I've actually set my machine and Google Docs language as US English, but sadly won't help with little edits here directly done in VS Code.

src/site/content/en/blog/bfcache/index.md
tunetheweb: I'm sending a `page_view` event in the first instance (only for restores for tracking the page load), and a generic event in the second instance (for tracking bfcache usage).

I agree that is two events for bfcache restores, but I did look into whether you could add `event_label`, `event_category` and `value` on a `page_view` event to combine them into one event, but didn't seem to the case - even though it's still an "event".

Can give it another go if you think this should be possible?

philipwalton: Sure, tracking this as a dimension is [how the web-vitals JS library does it] and probably makes sense here as well.

In that case, you'd want to make sure that every `page_view` event has a `navigation_type` (or whatever name you want) dimension set, and in cases of a bfcache restore you would manually set `bfcache` or `back_forward_cache`.

tunetheweb: Yeah I thought about custom dimensions, but the relatively low limit on those (at least in old school GA - not looked at GA4) means I try to avoid over using them. Plus the explanation of setting those up versus the simplicity here for this simple code snippet puts me off.

philipwalton: FYI, GA4 doesn't necessarily require any custom dimension set (see: [web.dev/vitals-ga4](https://web.dev/vitals-ga4/)), and given that GA3 is deprecated, I think it makes sense to assume GA4 going forward.

Also, we don't necessarily have to make these GA examples, for simplicity we could update them to be generic functions, e.g.:

```js
sendToAnalytics({
  eventName: 'page_view',
  navigationType: 'back_forward_cache',
});

Comment (tunetheweb) [reply src/site/content/en/blog/bfcache/index.md:519]: So that becomes something like this?

```js
// Send a pageview when the page is first loaded, including the navigation type
gtag('event', 'page_view',
  { dimensionX: performance.getEntriesByType('navigation')[0].type }
);

window.addEventListener('pageshow', (event) => {

  // Send another pageview if the page is restored from bfcache.
  if (event.persisted) {
    gtag('event', 'page_view',
      { dimensionX: "back_forward_cache" }
    );
  }

});
```

Then we get the following events in GA:
- `navigation`
- `reload`
- `back_forward`
- `back_forward_cache`
- ...etc.

It works, but then we need to explain to the user to measure it they need to total `back_forward` and `back_forward_cache` and divide by `back_forward_cache` to get the bfcache hit ratio, ignoring any non-baack/force navigations.

It does save one event I admit, but I think the current code and reporting of it is just a heck of a lot easy for readers to understand.

Plus with the current code it's easy to see this ratio in the standard event report in GA:

image

> Sure, tracking this as a dimension is [how the web-vitals JS library does it] and probably makes sense here as well.

Well that's not really true. The web vitals JS library still [sends a "Web Vitals" event](https://github.com/GoogleChrome/web-vitals#using-gtagjs-universal-analytics) (in fact one per CWV so multiple events for each page load) - it just adds the optional debugging information in the dimensions. The actual values are sent in the event `value` field, exactly like I'm doing with the current code.

philipwalton (approved):

src/site/content/en/blog/bfcache/index.md
tunetheweb: Discussed offline and updated.

PR #8275, opened 2022-07-04
Add cross-browser tag

tomayac:

- Add cross-browser tag.

### DIFF
--- a/src/site/_data/i18n/tags.yml
+++ b/src/site/_data/i18n/tags.yml
@@ -298,24 +298,6 @@ cors:
     ru: Последние новости и статьи о CORS.
     zh: 我们关于 CORS 的最新资讯、动态和故事。
 
-cross-browser:
-  title:
-    en: Cross-browser
-    es: Multi-navegador
-    ja: クロスブラウザ
-    ko: 브라우저 간
-    pt: Vários navegadores
-    ru: кроссбраузерный
-    zh: 跨浏览器
-  description:
-    en: Our latest news, updates, and stories about features that are supported cross-browser.
-    es: Nuestras últimas noticias, actualizaciones e historias sobre funciones compatibles con todos los navegadores.
-    ja: クロスブラウザでサポートされている機能に関する最新のニュース、アップデート、ストーリー。
-    ko: 브라우저 간 지원되는 기능에 대한 최신 뉴스, 업데이트 및 스토리.
-    pt: Nossas notícias, atualizações e histórias mais recentes sobre recursos compatíveis com vários navegadores.
-    ru: Наши последние новости, обновления и истории о функциях, поддерживаемых в разных браузерах.
-    zh: 我们关于跨浏览器支持的功能的最新新闻、更新和故事。
-
 css:
   title:
     en: CSS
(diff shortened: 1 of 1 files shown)


src/site/_data/i18n/tags.yml
```diff
@@ -298,6 +298,24 @@ cors:
...
+  title:
+    en: Cross-browser
+    es: Multi-navegador
+    ja: クロスブラウザ
+    ko: 브라우저 간
+    pt: Vários navegadores
+    ru: кроссбраузерный
+    zh: 跨浏览器
+  description:
+    en: Our latest news, updates, and stories about features that are supported cross-browser.
```
rachelandrew, line 311:
```suggestion
    en: Our latest news, updates, and stories about features that have become supported cross-browser.
```

rachelandrew: Ideally everything on this site is supported cross browser, the aim of this tag is to highlight the recent things (I'll likely remove it as things get to be a year old or something).

src/site/_data/i18n/tags.yml
```diff
@@ -298,6 +298,24 @@ cors:
     ru: Последние новости и статьи о CORS.
     zh: 我们关于 CORS 的最新资讯、动态和故事。
 
+cross-browser:
+  title:
+    en: Cross-browser
```
rachelandrew, line 303: I don't want to call it cross browser as per my comment below. It's more "newly interoperable" which is not a very snappy tag to be fair. I'll have a think.

rachelandrew (changes requested):
See comments inline re naming.

src/site/_data/i18n/tags.yml
tomayac: Your suggestion `newly-interoperable` has the small problem that it ages poorly. What's new today is old tomorrow. Other ideas:

`universally-usable`?
`universally-supported`?

rachelandrew: It aging badly is the entire point. As mentioned, I'd intend to remove the tags because everything on the site is intended to be interoperable so there would be no point to the tag if we added it to everything ;) The aim is to show case the stuff that's landed within a recent period.

tomayac: In this case it makes sense. I just brought it up because whenever we would write "new" in a post, Joe would make us remove it. This is different, though, so `newly-interoperable` works for me then. Changing to that.

src/site/_data/i18n/tags.yml
```diff
@@ -886,6 +868,24 @@ new-to-the-web:
   description:
     en: A series of posts to keep you updated with what's landing in web browsers each month.
 
+newly-interopaerable:
```
rachelandrew, line 871:
```suggestion
newly-interoperable:
```

src/site/_data/i18n/tags.yml
```diff
@@ -886,6 +868,24 @@ new-to-the-web:
...
+  title:
+    en: Newly interoperable
+    es: Interoperable recientemente
+    ja: 新しく相互運用可能
+    ko: 새롭게 상호 운용 가능
+    pt: Recentemente interoperável
+    ru: Новый интероперабельный
+    zh: 新的互操作性
+  description:
+    en: Our latest news, updates, and stories about features that are supported cross-browser.
```
rachelandrew, line 881:
```suggestion
    en: Features that have recently become supported in all three engines.
```

rachelandrew (changes requested):
spelling and a better suggestion for the description (as it explains what we mean by interoperable).

src/site/_data/i18n/tags.yml
tomayac: Duh

malchata (approved):
LGTM!

rachelandrew (approved):

PR #8279, opened 2022-07-05
[TransformStream article] Use new tag and aside

tomayac:
Use the new tag and aside.

### DIFF
--- a/src/site/content/en/blog/transformstream/index.md
+++ b/src/site/content/en/blog/transformstream/index.md
@@ -13,7 +13,7 @@ tags:
 ---
 
 {% Aside 'celebration' %}
-Woohoo, this web feature is now usable on all browsers.
+This web feature is now available in all three browser engines!
 {% endAside %}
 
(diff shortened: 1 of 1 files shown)


tomayac:
Let’s see how it looks.

tomayac:
- Tag: https://deploy-preview-8279--web-dev-staging.netlify.app/tags/newly-interoperable/
- Aside: https://deploy-preview-8279--web-dev-staging.netlify.app/transformstream/

I like it 👍.

src/site/content/en/blog/transformstream/index.md
```diff
@@ -9,8 +9,13 @@ alt: Transformer toy figurine.
 author: thomassteiner
 tags:
   - blog
+  - newly-interoperable
 ---
 
+{% Aside 'celebration' %}
+Woohoo, this web feature is now usable on all browsers.
```
rachelandrew, line 16:
```suggestion
This web feature is now available in all three engines!
```

rachelandrew: Try to avoid "all browsers" as frequently that is not the case, even when it is available in all engines.

rachelandrew (approved):

PR #8285, opened 2022-07-05
Migrates navigation preload post from d.c.c.

malchata:

- Now that service worker navigation preload is supported in all major browser engines, I think we should migrate @jakearchibald's article on the API from d.c.c. to web.dev, as it is not a Chrome-specific feature, but one that anyone can use cross-browser now that Safari has landed it.
- Small tweaks in language to focus on this being a generally available feature, rather than just in Chrome.
- A related PR (GoogleChrome/developer.chrome.com#3136) removes this post from the d.c.c. repo and puts in place a redirect that will point users to the web.dev version.


src/site/content/en/blog/navigation-preload/index.md
```diff
@@ -0,0 +1,273 @@
...
+  Navigation preload lets you overcome service worker startup time by making requests in parallel.
+tags:
+  - blog
+  - performance
+  - progressive-web-apps
+---
+
+{% Aside 'celebration' %}
+Service Worker Navigation Preload is [now supported by all major browser engines](https://caniuse.com/mdn-api_navigationpreloadmanager).
+{% endAside %}
```
rachelandrew, line 19: it would be good to add the compat widget below this.

jeffposnick (approved):

PR #8317, opened 2022-07-12
Fetch streaming requests: Update article for Chrome 105

### LINKED ISSUE #8288: content: Streaming requests with the fetch API seemly not work

[fetch-upload-streaming](https://web.dev/fetch-upload-streaming)

The article describes that Request can use Stream as a body in Chrome 95, but I opened the demo in subsequent versions  and was told that the current browser does not support this feature

Also, [caniuse](https://caniuse.com/mdn-api_request_request_readablestream_request_body) says that any version does not support this feature. However, I don't see the article revealing that this feature requires a specific flag to be enabled.

jakearchibald:
Fixes #8288

Preview: https://deploy-preview-8317--web-dev-staging.netlify.app/fetch-upload-streaming/

- Clarifies that feature is launching in Chrome 105
- Removes origin trial details, and other references to experimentation
- Updates API references and feature detect
- Reordering article a bit, as some of the restrictions became more important
- Updated details on duplexing, since it's a new option
- Removed references to the HTTP/1 opt-in, as it's no longer a thing
- Changed the hero image because it was low contrast and it annoyed me

### DIFF
--- a/src/site/content/en/blog/fetch-upload-streaming/index.md
+++ b/src/site/content/en/blog/fetch-upload-streaming/index.md
@@ -290,7 +290,7 @@ Safari _does_ support streams in request objects, but _doesn't_ allow them to be
 used with `fetch`, so the `duplex` option is tested, which Safari doesn't
 currently support.
 
-### Writable streams
+## Using with writable streams
 
 Sometimes it's easier to work with streams when you have a `WritableStream`. You
 can do this using an 'identity' stream, which is a readable/writable pair that
--- a/src/site/content/en/blog/fetch-upload-streaming/index.md
+++ b/src/site/content/en/blog/fetch-upload-streaming/index.md
@@ -29,7 +29,7 @@ You could use this to:
   wait until the user presses 'send' before sending the data they entered.
 - Gradually send data generated on the client, such as audio, video, or input
   data.
-- Recreate web sockets over HTTP/2.
+- Recreate web sockets over HTTP/2 or HTTP/3.
 
 But since this is a low-level web platform feature, don't be limited by _my_
 ideas. Maybe you can think of a much more exciting use-case for request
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/fetch-upload-streaming/index.md
```diff
@@ -19,18 +19,21 @@ feedback:
   - api
 ---
 
-From Chrome 95, you can start a request before you have the whole body available by using the streams API.
+From Chrome 105, you can start a request before you have the whole body
```
tomayac, line 22:
```suggestion
From Chromium&nbsp;105, you can start a request before you have the whole body
```

jakearchibald:
@rachelandrew fwiw, this should probably be on d.c.c

tomayac (approved):
LGTM, just a nit for Chrome → Chromium (on web.dev).

tomayac:
(If it goes out on d.c.c., we can purely talk about Chrome of course.)

rachelandrew:
@jakearchibald @tomayac yeah this should get moved. It's not a good fit here.

jakearchibald:
Should I open a PR over on d.c.c and you can set up the redirect? (I don't know how to do that bit)

rachelandrew:
@jakearchibald sounds great - I can sort the redirect

nwtgck:
Hi, thank you so much for the update!

> The fetch will be rejected if the connection isn't HTTP/2.

I'm not sure this is right. In my experiment HTTP/3 connection was allowed in Version 105.0.5148.2 (Official Build) dev (x86_64) as follows.

image

nwtgck:
@jakearchibald Thank you for the update!

jakearchibald:
Moved to https://github.com/GoogleChrome/developer.chrome.com/pull/3214

PR #8321, opened 2022-07-12
Some small Accessibility tweaks to score 100 in Lighthouse

### LINKED ISSUE #8320: Articles get 88 score in Lighthouse accessibility audit
**Describe the bug**

This appears to be due to two items:
1. Our Hamburger menu button on mobile doesn't have an accessible name - easily fixed with an aria-label.
2. Our Article tags are in a `role="list"` but there are no list items within that list (as it's a `div` instead of `ul`, and there are no `li` items). Probably easiest fix is to just remove the `role="list"` for now as changing the `div` to `ul` causes an indentation, it would require wrapping the links in `<li>`s (possibly requiring further CSS changes), and other "lists" like language links do are not in `role="lists"` anyway.

**To Reproduce**
1. Go to https://web.dev/measure/?url=https%3A%2F%2Fweb.dev%2Fbrowser-level-image-lazy-loading%2F
3. Run Lighthouse Audit
4. Notice we score 88 for a couple of simple markup reasons

**Expected behavior**
As a best practice website, we should be looking to score 100 in a11y audits.

**Screenshots**
N/A

Happy to throw in a PR for this as seems fairly simple fixes.

tunetheweb:
Fixes #8320 

- Our Hamburger menu button on mobile doesn't have an accessible name - added an aria-labelledby to point to tool tip (which is hidden by default), similar to other code.
- Our Article tags are in a `role="list"` but there are no list items within that list (as it's a `div` instead of `ul`, and there are no `li` items). I changed it to a `<nav>` instead (which has an implicit `navigation` role as changing the `div` to `ul` causes an indentation (requiring CSS changes), it would require wrapping the links in `<li>`s (possibly requiring further CSS changes), and other "lists" like language links do are not in `role="lists"` anyway. Plus `<nav>` is actually more accurate.

### DIFF
--- a/src/site/_includes/partials/post-next.njk
+++ b/src/site/_includes/partials/post-next.njk
@@ -118,14 +118,14 @@
         {% StackOverflow stack_overflow_tag %}
       {% endif %}
 
-      <div class="cluster gutter-base flow-space-size-2" aria-label="tags">
+      <nav class="cluster gutter-base flow-space-size-2" aria-label="tags">
         {% for tagKey in tags or [] %}
           {% if tagKey in collections.tags %}
             {% set tag = collections.tags[tagKey] %}
             <a class="pill" href="{{tag.url}}">{{tag.overrideTitle or tag.title | i18n(locale)}}</a>
           {% endif %}
         {% endfor %}
-      </div>
+      </nav>
 
       <div class="text-size-0 color-mid-text">
         <span>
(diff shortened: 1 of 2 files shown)


malchata (approved):
These accessibility tweaks LGTM!

malchata:
@tunetheweb it appears there's a Percy failure, so I'll leave this up to you to investigate and merge.

tunetheweb:
> @tunetheweb it appears there's a Percy failure, so I'll leave this up to you to investigate and merge.

Wondered about those myself. Seemed completely unrelated so wasn't sure if this was a known flakey test. Anyway, I reran it and passed this time so guess was a false positive.

So good to merge now.

PR #8336, opened 2022-07-15
Use checkmark icon in BrowserCompat when version_added is true

### LINKED ISSUE #8333: Browser Compatibility component shows incorrect data
**Describe the bug**

the browser compat macro in [Building a button component](https://web.dev/building-a-button-component/) seems to show that `<button>` elements are only available in Edge. [According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#browser_compatibility), the source this macro pulls from, they have been available to all browsers for a very long time.

image

**To Reproduce**
Follow the links above to compare data that should be identical.

**Expected behavior**
The component on web.dev should show the data that's on MDN.

jeffposnick:
Fixes #8333 by using a ✅ icon when there's a feature whose `version_added` field is set to `true` (as opposed to being set to an actual version number).

CC: @jpmedley

### DIFF
--- a/src/site/_includes/components/BrowserCompat.js
+++ b/src/site/_includes/components/BrowserCompat.js
@@ -56,6 +56,15 @@ function getInfoFromSupportStatement(support, status, locale) {
         icon: support.version_added,
       };
     }
+
+    // See https://github.com/GoogleChrome/web.dev/issues/8333
+    if (support.version_added === true) {
+      return {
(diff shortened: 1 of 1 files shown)


jeffposnick:
As per https://deploy-preview-8336--web-dev-staging.netlify.app/building-a-button-component/, this renders as

Screen Shot 2022-07-15 at 10 58 56 AM

rachelandrew (approved):

PR #8347, opened 2022-07-17
Adds guidance on serving WebP images using the HTTP `Accept` header

### LINKED ISSUE #3314: content: update WebP image guide to mention the `Accept` response header?
This is both a question _and_ a proposal to update existing content. I've noticed that the section for using WebP images in the image optimization guide doesn't mention scanning the content of the `Accept` request header on the server side. Is this something that should be covered?

If folks think this would be worth adding, I'd be happy to contribute and add a small section plus code snippets to demonstrate how this might be done on the back end.

https://web.dev/serve-images-webp/

Using `Accept` on the server side to detect WebP support in the browser is beneficial because we can rewrite image requests for legacy formats to WebP. This means we can send less `<picture>` markup, which can save some bytes—especially for complex `<picture>` use cases with lots of image candidates.

malchata:
Fixes #3314

- Outlines how to use the HTTP `Accept` header to automatically rewrite requests for WebP images based on its contents.

### DIFF
--- a/src/site/content/en/fast/serve-images-webp/index.md
+++ b/src/site/content/en/fast/serve-images-webp/index.md
@@ -8,11 +8,12 @@ description: |
   magnitude of a 25–35% reduction in filesize. This decreases page sizes and
   improves performance.
 date: 2018-11-05
-updated: 2020-04-06
+updated: 2022-07-18
 codelabs:
   - codelab-serve-images-webp
 tags:
   - performance
+  - web-vitals
 feedback:
   - api
 ---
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/serve-images-webp/index.md
```diff
@@ -156,6 +156,30 @@ ignore the tags it doesn't support. Thus, it only "sees" the
...
+RewriteCond %{HTTP:Content-Disposition} !attachment [NC]
+RewriteCond %{DOCUMENT_ROOT}/$1.webp -f [NC]
+RewriteRule (.+)\.(png|jpe?g)$ $1.webp [T=image/webp,L]
+```
+
+The rewrite rule above will look for a WebP version of any requested JPEG or PNG image. If a WebP alternate is found, it will be served with the proper `Content-Type`  header. This will allow you to use image markup similar to the following with automatic WebP support:
+
+```html
+<img src="flower-320w.jpg" srcset="flower-320w.jpg 320w, flower-640w.jpg 640w, flower-960w.jpg 960w">
+```
```
tunetheweb, line 181: Personally I don't like serving webp images with a .jpg extension but that's just me :-)
But I do agree it can help reduce the markup considerably.

At the very least though, we should add a `vary` header to these responses, so they are cached correctly at any CDN or other intermediary cache.

malchata: Agreed on `Vary`. Just need to look up the syntax for setting the `Vary` header in `mod_rewrite`.

tunetheweb (approved):
LGTM!

PR #8348, opened 2022-07-17
Addresses LCP concerns for autoplaying video as animated GIF replacement.

malchata:

- Adds a section at the end of the article discussing the effects of autoplaying videos on LCP.

### DIFF
--- a/src/site/content/en/fast/replace-gifs-with-videos/index.md
+++ b/src/site/content/en/fast/replace-gifs-with-videos/index.md
@@ -3,6 +3,7 @@ layout: post
 title: Replace animated GIFs with video for faster page loads
 authors:
   - houssein
+  - jlwagner
 description: |
   Have you ever seen an animated GIF on a service like Imgur or Gfycat,
   inspected it in your dev tools, only to find out that GIF was really a video?
--- a/src/site/content/en/fast/replace-gifs-with-videos/index.md
+++ b/src/site/content/en/fast/replace-gifs-with-videos/index.md
@@ -9,7 +9,7 @@ description: |
   There's a good reason for that. Animated GIFs can be downright huge! By
   converting large GIFs to videos, you can save big on users' bandwidth.
 date: 2018-11-05
-updated: 2019-08-29
+updated: 2022-07-18
 codelabs:
   - codelab-replace-gifs-with-video
 tags:
@@ -122,3 +122,9 @@ Browsers don't speculate about which `<source>` is optimal, so the order of
 browser supports WebM, browsers will skip the WebM `<source>` and use the MPEG-4
 instead. If you prefer a WebM `<source>` be used first, specify it first!
 {% endAside %}
+
+## Effect on Largest Contentful Paint (LCP)
+
+It should be noted that while `<img>` elements are candidates for LCP, `<video>` elements are not [LCP candidates](/lcp/#what-elements-are-considered) in all cases. When using videos as a replacement for animated GIFs provides better content efficiency, `<video>` elements without a [`poster` image](https://developer.mozilla.org/docs/Web/HTML/Element/video#attr-poster) are not considered for LCP. This will be the case if you're using videos to replace animated GIFs.
+
+What does this mean for your website? For one, if your largest element on screen is an autoplaying video that acts as an animated GIF replacement, the next largest node that qualifies as an LCP candidate will be selected instead. This may change in the future as LCP evolves.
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/replace-gifs-with-videos/index.md
```diff
@@ -122,3 +123,9 @@ Browsers don't speculate about which `<source>` is optimal, so the order of
 browser supports WebM, browsers will skip the WebM `<source>` and use the MPEG-4
 instead. If you prefer a WebM `<source>` be used first, specify it first!
 {% endAside %}
+
+## Effect on Largest Contentful Paint (LCP)
+
+It should be noted that while `<img>` elements are candidates for LCP, `<video>` elements are not [LCP candidates](/lcp/#what-elements-are-considered) in all cases. When using videos as a replacement for animated GIFs provides better content efficiency, `<video>` elements without a [`poster` image](https://developer.mozilla.org/docs/Web/HTML/Element/video#attr-poster) are not considered for LCP. This will be the case if you're using videos to replace animated GIFs.
```
tunetheweb, line 129: Is this the only case where `video` is not considered? If so could we merge the first and second sentences to make this clear?

src/site/content/en/fast/replace-gifs-with-videos/index.md
```diff
@@ -122,3 +123,9 @@ Browsers don't speculate about which `<source>` is optimal, so the order of
...
+
+What does this mean for your website? For one, if your largest element on screen is an autoplaying video that acts as an animated GIF replacement, the next largest node that qualifies as an LCP candidate will be selected instead. This may change in the future as LCP evolves. Consult the [LCP documentation](/lcp/) for the most up-to-date information.
```
tunetheweb, line 131: Should we also explicitly recommend a poster image when using video? If so why? And if not why?

I just worry this current wording suggests sticking with GIFs - which we know are notoriously inefficient and so likely larger and likely impacting LCP times anyway because of that.

malchata: I just think it depends on the use case. A `poster` will not faithfully emulate the animated GIF-like behavior of autoplaying and looping, so I think this is more a caveat that such images won't be treated as an LCP candidate. Maybe I could be more clear...

tunetheweb: That info is good to know in and of itself!

Does a poster always prevent autoplaying? Or is it just what's there until the video downloads and then it's ignore for autoplaying videos? To be honest I don't know enough about videos hence why I'm asking...

Maybe this is not necessary but thought I'd ask the question. Have a think, and if happy to keep as is, then that works for me.

malchata: IIRC, the effect is a wasted request, where the poster image is downloaded, but immediately discarded. There could be differences in how browser engines handle this, but using `poster` with `autoplay` doesn't make a whole lot of sense to begin with.

tunetheweb: That makes sense to me.

Maybe some wording along the lines of "...and given that autoplaying videos that are likely to replace GIFs do not really benefit from an poster image (since the intent is to auto place)..."

I wonder also if there is a benefit of this? GIFs, as I say, are typically terrible when it comes to compression and so performance so is this _yet more reason_ to switch to video elements instead? Or is that a bit opinionated?

I may be over thinking this but just trying to avoid the "web.dev told me video was worse than GIF for LCP reasons, hence why I stuck with GIF" type thinking.

src/site/content/en/fast/replace-gifs-with-videos/index.md
```diff
@@ -122,3 +123,9 @@ Browsers don't speculate about which `<source>` is optimal, so the order of
 browser supports WebM, browsers will skip the WebM `<source>` and use the MPEG-4
 instead. If you prefer a WebM `<source>` be used first, specify it first!
 {% endAside %}
+
+## Effect on Largest Contentful Paint (LCP)
+
+It should be noted that while `<img>` elements are candidates for LCP, `<video>` elements without a [`poster` image](https://developer.mozilla.org/docs/Web/HTML/Element/video#attr-poster) are not [LCP candidates](/lcp/#what-elements-are-considered). The solution in the case of emulating animated GIFs is _not_ to add `poster` attribute to your `<video>` elements, because that image will go unused.
+
+The solution here is to stick with using a `<video>` instead of an animated GIF, but with the understanding that such media will not be a candidate for LCP, and the next largest candidate will be used instead.
```
tunetheweb, line 131: Double use of "the solution".

WDYT about this?

```suggestion
What does this mean for your website? The recommendation is to stick with using a `<video>` instead of an animated GIF, but with the understanding that such media will not be a candidate for LCP, and the next largest candidate will be used instead. As GIFs and `<video>`s are typically larger and so slower to download, moving to a different LCP candidate will likely even improve the site's LCP.
```

tunetheweb (approved):
LGTM!

Thanks for dealing with my concerns. Think it's stronger for it though!

PR #8357, opened 2022-07-18
Twitter Spaces Series

una:
Adds initial Spaces schedule

### DIFF
--- a/src/site/_includes/spaces.njk
+++ b/src/site/_includes/spaces.njk
@@ -147,9 +147,11 @@ layout: default-next
                 <span class="ellipsis__more">{{ 'i18n.spaces.see_more' | i18n(locale) }}</span>
               </summary>
             </details>
+            {% if space.data.audio %}
             <p>
               <audio src="{{ space.data.audio }}" class="audio-player__element" controls=""></audio>
             </p>
+            {% endif %}
           </p>
         </td>
       </tr>
--- a/src/site/_includes/spaces.njk
+++ b/src/site/_includes/spaces.njk
@@ -34,6 +34,7 @@ layout: default-next
       {% endfor %}
     </ul>
     <time class="eyebrow">{{ space.data.event_date | prettyDate('DATE_FULL') }}</time>
+    <p class="eyebrow"> {{ space.data.event_time }} </p>
     <p class="gap-top-size-1 gap-bottom-size-2">{{ space.data.description }}</p>
   </div>
   {% if space.data.twitter_link %}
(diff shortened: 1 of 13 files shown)


src/site/_includes/spaces.njk
```diff
@@ -34,13 +34,19 @@ layout: default-next
...
     <time class="eyebrow">{{ space.data.event_date | prettyDate('DATE_FULL') }}</time>
+    <p class="eyebrow"> {{ space.data.event_time }} </p>
     <p class="gap-top-size-1 gap-bottom-size-2">{{ space.data.description }}</p>
   </div>
   {% if space.data.twitter_link %}
     <a class="button" href="{{ space.data.twitter_link }}" target="_blank">
       {% include "icons/twitter.svg" %}
       <span>{{ 'i18n.spaces.see_on_twitter' | i18n(locale) }}</span>
     </a>
+  {% elif space.data.cal_link %}
```
jeffposnick, line 45: Ideally, the `calendarLink()` filter would be flexible enough to take in a time of day as well as the date, and it could replace this hardcoded `space.data.cal_link`.

src/site/content/en/spaces/2022-08-04-ui-ama/index.md
```diff
@@ -0,0 +1,16 @@
...
+hosts:
+  - bramus
+  - jhey
+  - una
+primary_host:
+  - una
+event_date: 2022-08-04
+event_time: 9AM PT
+cal_link: "https://calendar.google.com/calendar/u/0/r/eventedit?&text=Ask+me+anything+with+the+UI+DevRel+Team&details=Join+the+Chrome+UI+DevRel+Team+to+ask+your+questions+about+building+modern+web+interfaces.&location=Twitter+Spaces&dates=20220804T160000Z%2F20220804T170000Z"
+audio: tbd
```
jeffposnick, line 13: Hardcoding `"tbd"` for the audio URL isn't _great_.

Maybe leave it blank, and have the template only conditionally output an `<audio>` element if the property is set?

jeffposnick (approved):
I had a couple of comments, but nothing that would necessarily block this from going live, given the time constraints.

jeffposnick (commented):
Now that `space.data.audio` might not be set, it would be preferable to change https://github.com/GoogleChrome/web.dev/blob/2d621fc887457b224f85ecfd5316607fdc128faa/src/site/_includes/spaces.njk#L157 so that outputting the `<audio>` tag was conditional, and only happened if it was set. Otherwise you'll end up with a `<audio src="">`

jeffposnick (approved):
I believe that @rachelandrew needs to give a LGTM in order to make the review bot fully happy, but I'm fine with admin-merging given the time constraints.

PR #8374, opened 2022-07-20
Avoid calling minify with the empty string twice

jakearchibald:


jakearchibald:
@jeffposnick discovered the issue, I'm just stealing the glory so I feel like I achieved something today

PR #8391, opened 2022-07-22
Clarifies that scrolling with the space bar factors into INP.

malchata:

- Updates INP metric documentation to clarify that scrolling with the space bar factors into INP.

### DIFF
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -116,6 +116,10 @@ As far as INP goes, **only the following interaction types are observed:**
 - Tapping on a device with a touchscreen.
 - Pressing a key on either a physical or onscreen keyboard.
 
+{% Aside 'important' %}
+Hovering and scrolling does not factor into INP. However, scrolling with the space bar counts as a keystroke and _is_ factored into INP, as the main thread may block key inputs in the presence of long tasks.
+{% endAside %}
+
 Interactions may consist of two parts, each with multiple events. For example, a keystroke consists of the `keydown`, `keypress`, and `keyup` events. Tap interactions contain `pointerup` and `pointerdown` events. The event with the longest duration within the interaction is chosen as the interaction's latency.
 
 <figure>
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/inp/index.md
```diff
@@ -116,6 +116,10 @@ As far as INP goes, **only the following interaction types are observed:**
 - Tapping on a device with a touchscreen.
 - Pressing a key on either a physical or onscreen keyboard.
 
+{% Aside 'important' %}
+Hovering and scrolling does not factor into INP. However, scrolling with the space bar counts as a keystroke and _is_ factored into INP, as the main thread may block key inputs in the presence of long tasks.
```
mmocny, line 120: Continuous events like hovering and scrolling are not measured by INP.

However, sometimes keyboard everts (space bar, arrow buttons, PgUp PgDown) can start scrolling.  The scroll itself is not measured, but INP will measure the keyboard events themselves.

That is because long tasks on main thread may block the keyboard event which needs to be processed before you can even start the scroll.  The scroll itself is never measured.

Note, JavaScript like the following can prevent scrolling via keyboard, so its clear that the INP is just for the key itself not the scroll

```
document.addEventListener('keydown', (evt) => {
    evt.preventDefault();
});
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -140,7 +144,7 @@ INP is more than about first impressions. By sampling all interactions, responsi
 It's possible that a page can return no INP value. This can happen for a number of reasons:
 
 - The page was loaded, but the user never clicked, tapped, or pressed a key on their keyboard.
-- The page was loaded, but the user interacted with the page using gestures that did not involve clicking, tapping, or using the keyboard. For example, scrolling or hovering over elements does not factor into how INP is calculated.
+- The page was loaded, but the user interacted with the page using gestures that did not involve clicking, tapping, or using the keyboard. For example, scrolling or hovering over elements does not factor into how INP is calculated (unless the space bar was used to scroll the page).
```
mmocny, line 147: Given the above clarity, I think we can remove this.  Scroll is never measured, its always just keyboard.  The fact that some keyboard actions scroll by default I think is addressed in the detail above.

src/site/content/en/metrics/inp/index.md
malchata: urgh—I already merged this one, but I can open a second PR to address these. Sorry I got all merge-happy!

mmocny:
I added some comments inline but I didn't sneak in before merge :P

I think "scrolling is not measured unless is scrolling via space button" is the wrong message imho, and doesn't cover the myriad of other ways to scroll with keyboard.

I would suggest the message the scrolling itself is never measured-- keyboard events are always measured.  Some keyboard events can kick off a scroll (or not), and in those cases we still measure the keyboard event.  So its not "dont measure scroll unless" its "always measure keyboard even when"

PR #8394, opened 2022-07-22
Adds BBC HSTS short post

malchata:

- Adds a short blog post that discusses how BBC rolled out a new HSTS policy, as well as some info on HSTS itself.

### DIFF
--- a/src/site/content/en/blog/bbc-hsts/index.md
+++ b/src/site/content/en/blog/bbc-hsts/index.md
@@ -37,10 +37,10 @@ While redirects work fine, they can be misconfigured in ways that still allow ac
 
 {% BrowserCompat 'http.headers.Strict-Transport-Security' %}
 
-[HSTS](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic, rather than the server. This prevents the request being intercepted, since it never leaves the browser, so is more secure. As and added bonus, as there is no network request needed to get the redirect, these types of redirect are extremely fast, and so prevent noticeable latency during an HTTP-to-HTTPS hop.
+[HSTS](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic, rather than the server. This prevents the request being intercepted, since it never leaves the browser, so is more secure. As an added bonus, these types of redirect are extremely fast, so any noticeable latency during an HTTP-to-HTTPS hop is eliminated.
 
 <figure>
-  {% Img src="image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/J9vQLJCLmnHQPb6spsRe.png", alt="A 307 internal redirect from HTTP to HTTPS, triggered by an HSTS header. The 307 redirect only takes 2 milliseconds.", width="800", height="97" %}
+  {% Img src="image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/RCCZMoqNd4aLb91Uj1hx.png", alt="A 307 internal redirect from HTTP to HTTPS, triggered by an HSTS header. The 307 redirect only takes 2 milliseconds.", width="800", height="84" %}
 </figure>
 
 Similar in syntax to `Cache-Control`'s `max-age` directive, an HSTS header specifies a `max-age` directive. This directive takes a value in seconds that specifies how long the policy is effective for:
--- a/src/site/content/en/blog/bbc-hsts/index.md
+++ b/src/site/content/en/blog/bbc-hsts/index.md
@@ -18,7 +18,7 @@ tags:
 
 HTTPS adoption has been steadily increasing in recent years. Per the HTTP Archive's 2021 Web Almanac, [around 91% of all requests for both desktop and mobile were served over HTTPS](https://almanac.httparchive.org/en/2021/security#transport-security). HTTPS isn't just here to stay, it's a necessary prerequisite to use features such as [Service Worker](https://developer.mozilla.org/docs/Web/API/Service_Worker_API) and modern protocols such as [HTTP/2](/performance-http2/) and [HTTP/3](https://www.cloudflare.com/learning/performance/what-is-http3/).
 
-Recently Neil Craig&mdash;a lead technical Architect at the BBC&mdash;tweeted that [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) is being slowly rolled out for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
+Recently Neil Craig&mdash;a lead technical Architect at the BBC&mdash;tweeted that [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is being slowly rolled out for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
 
 ## The problem
 
--- a/src/site/content/en/blog/bbc-hsts/index.md
+++ b/src/site/content/en/blog/bbc-hsts/index.md
@@ -1,5 +1,5 @@
 ---
-title: The BBC rolls out HTTP Strict Transport Security (HSTS)
+title: How the BBC is rolling out HSTS for better security and performance.
 subhead: |
   The BBC is rolling out HSTS for their website to improve security and performance. Find out what it means, and how HSTS can help you.
 authors:
@@ -12,17 +12,17 @@ thumbnail: image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/ew6tI1HdAi2g3vxQwdfM.png
 alt: The BBC logo set on a black background.
 tags:
   - blog
-  - performance
   - security
+  - performance
 ---
 
 HTTPS adoption has been steadily increasing in recent years. Per the HTTP Archive's 2021 Web Almanac, [around 91% of all requests for both desktop and mobile were served over HTTPS](https://almanac.httparchive.org/en/2021/security#transport-security). HTTPS isn't just here to stay, it's a necessary prerequisite to use features such as [Service Worker](https://developer.mozilla.org/docs/Web/API/Service_Worker_API) and modern protocols such as [HTTP/2](/performance-http2/) and [HTTP/3](https://www.cloudflare.com/learning/performance/what-is-http3/).
 
-Recently Neil Craig&mdash;a lead technical Architect at the BBC&mdash;tweeted that [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is being slowly rolled out for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
+Recently [Neil Craig](https://twitter.com/tdp_org)&mdash;a lead technical Architect at the BBC&mdash;[tweeted that HTTP Strict Transport Security (HSTS) is being slowly rolled out](https://twitter.com/tdp_org/status/1549028675328573440?s=20&t=S5mm1eEqGUbxZ4jfi-SvtA) for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
 
 ## The problem
 
-Web servers these days tend to listen for requests on ports 80 and 443. Port 80 is for insecure HTTP requests, whereas 443 is for secure HTTPS. This can create a problem, because when you enter an address into your address bar without the https:// protocol prefix&mdash;like most users tend to do&mdash;some browsers will direct traffic to the insecure HTTP version of a site.
+Web servers often listen for requests on both ports 80 and 443. Port 80 is for insecure HTTP requests, whereas 443 is for secure HTTPS. This can create a problem, because when you enter an address into your address bar without the `https://` protocol prefix&mdash;like most users tend to do&mdash;some browsers will direct traffic to the insecure HTTP version of a site, for legacy reasons ([though this isn't always the case](https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html)).
 
 A common way to ensure users don't access an unsecured version of a website is to place an HTTP-to-HTTPS redirect for all requests. This certainly works, but it kicks off the following chain of events:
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+alt: The BBC logo set on a black background.
+tags:
+  - blog
+  - performance
+  - security
+---
+
+HTTPS adoption has been steadily increasing in recent years. Per the HTTP Archive's 2021 Web Almanac, [around 91% of all requests for both desktop and mobile were served over HTTPS](https://almanac.httparchive.org/en/2021/security#transport-security). HTTPS isn't just here to stay, it's a necessary prerequisite to use features such as [Service Worker](https://developer.mozilla.org/docs/Web/API/Service_Worker_API) and modern protocols such as [HTTP/2](/performance-http2/) and [HTTP/3](https://www.cloudflare.com/learning/performance/what-is-http3/).
+
+Recently Neil Craig&mdash;a lead technical Architect at the BBC&mdash;tweeted that [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is being slowly rolled out for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
```
tunetheweb, line 21: I follow Neil so should we know his Twitter handle - should we use that?

```suggestion
Recently [Neil Craig](https://twitter.com/tdp_org)&mdash;a lead technical Architect at the BBC&mdash;tweeted that [HTTP Strict Transport Security (HSTS)](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is being slowly rolled out for [bbc.com](https://www.bbc.com/). Let's find out what that means for the BBC, and what it could mean for you.
```

Also wonder if we should link his tweet since you've referenced it? https://twitter.com/tdp_org/status/1549028675328573440?s=20&t=S5mm1eEqGUbxZ4jfi-SvtA

Maybe link that rather than HSTS and link that later?

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+## The problem
+
+Web servers these days tend to listen for requests on ports 80 and 443. Port 80 is for insecure HTTP requests, whereas 443 is for secure HTTPS. This can create a problem, because when you enter an address into your address bar without the https:// protocol prefix&mdash;like most users tend to do&mdash;some browsers will direct traffic to the insecure HTTP version of a site.
```
tunetheweb, line 25:
```suggestion
Web servers these days tend to listen for requests on ports 80 and 443. Port 80 is for insecure HTTP requests, whereas 443 is for secure HTTPS. This can create a problem, because when you enter an address into your address bar without the `https://` protocol prefix&mdash;like most users tend to do&mdash;some browsers will direct traffic to the insecure HTTP version of a site, for legacy reasons.
```

Maybe link to this post: https://blog.chromium.org/2021/03/a-safer-default-for-navigation-https.html

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+A common way to ensure users don't access an unsecured version of a website is to place an HTTP-to-HTTPS redirect for all requests. This certainly works, but it kicks off the following chain of events:
+
+1. The server receives a request via HTTP.
+2. The server issues a redirect to go to the HTTPS equivalent of the requested resource.
+3. The server via HTTPS must negotiate a secure connection with the browser.
+4. The content loads as usual.
+
+The problem here is that the 301 redirect delays rendering of content, forcing the user to wait longer, and unless redirects are properly set up, you can still end up sending users to an insecure version of a web site.
```
tunetheweb, line 34: Much as I love performance, the main reason for HSTS is for security reasons - web perf is a nice added benefit.

"and unless redirects are properly set up, you can still end up sending users to an insecure version of a web site."

This doesn't make sense. You HAVE sent users to an insecure version of your site (well the browser has). While I agree you can mess up the redirects after that, the security impact is already there.

The primary reason for HSTS is to stop a middle man intercepting the request to http://www.bbc.com (which they can do!) and redirecting to https://wwww.evil-bbc.com (or some less obviously named bad site!).

I think you need to either reframe this article with a. better title ("How the BBC improved performance by rolling out HSTS"), or concentrate on the primary benefit of HSTS, and only then mention the performance benefits of it. Right now, as someone who used to be equally interested in web security and web performance, the framing seems off. Maybe it would be fine if web.dev was all about performance, but it's not. So downplaying the main, security, reasons for this technology seems off.

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Enter HSTS
+
+{% BrowserCompat 'HTTP.Headers.Strict-Transport-Security' %}
+
+HSTS is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic. These types of redirect are extremely fast, and add no noticeable latency during an HTTP-to-HTTPS hop.
```
tunetheweb, line 40:
```suggestion
HSTS is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic, rather than second it to the server. This prevents the request being intercepted, since it never leaves the browser, so is more secure. As and added bonus, as there is no network request needed to get the redirect, these types of redirect are extremely fast, and so prevent noticeable latency during an HTTP-to-HTTPS hop.
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+The main drawback of deploying HSTS is if you're not ready to treat your origin as strictly secure. Let's say you have a number of subdomains you're serving resources from, but maybe not all of them are secure. In this scenario, an HSTS header could break your website.
+
+The BBC took the right approach to deploying HSTS. [As Neil Craig said in his tweet](https://twitter.com/tdp_org/status/1549022691772059649), the initial value that was set for bbc.com was `max-age=10`.
+
+This approach means that the policy was only initially effective for ten seconds. This doesn't provide much of a benefit, but the idea is to feel out whether there may be issues with applying HSTS at all. As time goes on, you can increase the policy incrementally and see if issues occur. At present, bbc.com is specifying an HSTS policy of `max-age=3600`, and that will almost certainly increase.
+
+You certainly don't want to come out of the gate with a long `max-age` value when deploying HSTS. You could find yourself suddenly scrambling to fix issues while users experience issues. Start small, and increment over time! When you're confident, you can set your `max-age` directive to a much longer period of time.
```
tunetheweb, line 62:
```suggestion
You certainly don't want to come out of the gate with a long `max-age` value when deploying HSTS. You could find yourself suddenly scrambling to fix issues while users experience issues. Start small, and increment over time! When you're confident, you can set your `max-age` directive to a much longer period of time. It is [recommended to set this value to one or two years](https://hstspreload.org/#deployment-recommendations) when it is fully rolled out.
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Get faster initial navigations with the HSTS preload list
```
tunetheweb, line 64:
```suggestion
## Get safer and faster initial navigations with the HSTS preload list
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+An HSTS policy only takes effect after the first visit to a website, so the benefit isn't noticeable until a user comes back. However, you can _preload_ your HSTS policy by submitting your website to the [HSTS preload list](https://hstspreload.org/), which is a hardcoded list of websites that the browser knows are strictly HTTPS. When your site is on the preload list, HTTP-to-HTTPS redirect latency via HSTS will be instantaneous&mdash;**even on the first visit!**
```
tunetheweb, line 66: Again pivot to security first, and then performance. Plus add a warning about this (there is a loooong history about sites messing this up).

```suggestion
An HSTS policy only takes effect after the first visit to a website, so the benefits are not present for the first visit to the site. This will still require the insecure redirect. However, you can _preload_ your HSTS policy by submitting your website to the [HSTS preload list](https://hstspreload.org/), which is a hardcoded list of websites that the browser knows are strictly HTTPS. When your site is on the preload list, the first visit is also protected and HTTP-to-HTTPS redirect latency via HSTS will be instantaneous.

{% Aside 'warning' %}
Once your site is on the preload list, the HSTS will be embedded in browser code. This means if you need to revert this, you need to wait for the HSTS preload list to be updated and rolled out to users&mdash;a process that can take several months. Ensure you have rolled out HSTS with the header only and with a large `max-age` and given that time to bed in&mdash;like the BBC is doing&mdash;before taking the next step to add your domain to the preload list.
{% endAside %}
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+
+{% Aside %}
+If you have a domain with a .dev TLD, your website is required to have a valid SSL certificate to function at all. This means you don't need to add your .dev website to the HSTS preload list&mdash;it will already be assumed as strictly secure by the browser!
```
tunetheweb, line 69: Don't like the implication that SSL implies HSTS (it's the other way around).

```suggestion
If you have a domain with a `.dev` top level domain (TLD), your website is already protected by HSTS as the whole `.dev` TLD has been included in preload list. It will already be assumed as strictly secure by the browser and is required to have a valid SSL certificate to function at all.
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+{% endAside %}
+
+## Try it out for yourself
+
+If the BBC feels comfortable testing out HSTS, there's a good chance that you can do the same for your website. Give it a shot for your website, and&mdash;if you're really looking to rev things up&mdash;add it to the HSTS preload list when you're confident there are no bugs to give your users a secure _and_ fast experience.
```
tunetheweb, line 74:
```suggestion
If the BBC feels comfortable testing out HSTS, there's a good chance that you can do the same for your website. Give it a shot for your website, and&mdash;once you're confident there are no bugs&mdash;to really rev things up, add it to the HSTS preload list to give your users a secure _and_ fast experience.
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,74 @@
...
+date: 2022-07-25
+description: |
+  The BBC is rolling out HSTS for their website to improve security and performance. Find out what it means, and how HSTS can help you.
+hero: image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/BbUZ7XxZ991yp4lpD5Tr.png
+thumbnail: image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/ew6tI1HdAi2g3vxQwdfM.png
+alt: The BBC logo set on a black background.
+tags:
+  - blog
+  - performance
+  - security
```
tunetheweb, line 16:
```suggestion
  - security
  - performance
```

tunetheweb (changes requested):
A bit of feedback as I have some history on HSTS.

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,78 @@
...
+4. The content loads as usual.
+
+While redirects work fine, they can be misconfigured in ways that still allow access to the insecure version of a site. Even if everything _is_ configured properly, there is still a security issue in that the user will still connect over insecure HTTP during the redirect phase, which exposes users to the possibility of dangerous [man-in-the-middle attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack).
+
+## Enter HSTS
+
+{% BrowserCompat 'http.headers.Strict-Transport-Security' %}
+
+[HSTS](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic, rather than second it to the server. This prevents the request being intercepted, since it never leaves the browser, so is more secure. As and added bonus, as there is no network request needed to get the redirect, these types of redirect are extremely fast, and so prevent noticeable latency during an HTTP-to-HTTPS hop.
```
tunetheweb, line 40: Any better word that "second"? Always reads to me like time.

```suggestion
[HSTS](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security) is dictated by the `Strict-Transport-Security` HTTP response header for HTTPS requests. When set, return visits to a website will trigger a special redirect known as a "307 Internal Redirect", which is when the browser handles the redirect logic, rather than the server. This prevents the request being intercepted, since it never leaves the browser, so is more secure. As and added bonus, as there is no network request needed to get the redirect, these types of redirect are extremely fast, and so prevent noticeable latency during an HTTP-to-HTTPS hop.
```

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+<figure>
+  {% Img src="image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/J9vQLJCLmnHQPb6spsRe.png", alt="A 307 internal redirect from HTTP to HTTPS, triggered by an HSTS header. The 307 redirect only takes 2 milliseconds.", width="800", height="97" %}
```
tunetheweb, line 43: Think it would be less confusing to have a screenshot with a 200, rather than 304 response?

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,78 @@
...
+</figure>
+
+Similar in syntax to `Cache-Control`'s `max-age` directive, an HSTS header specifies a `max-age` directive. This directive takes a value in seconds that specifies how long the policy should take effect:
```
tunetheweb, line 46: Your wording would be read as "when the policy will start to take effect", rather than "how long it is effective for".

```suggestion
Similar in syntax to `Cache-Control`'s `max-age` directive, an HSTS header specifies a `max-age` directive. This directive takes a value in seconds that specifies how long the policy is effective for:
```

tunetheweb (changes requested):
Looks much better, but still got a couple more suggestions.

src/site/content/en/blog/bbc-hsts/index.md
malchata: I suppose so, I'd just need to recapture it, which might take me a bit.

malchata:
I still need to get a better screenshot, but if you can approve this, I can merge it when I have the new screenshot that you requested, @tunetheweb.

tunetheweb (approved):
Btw I presume BBC and/or Neil has OKed this? Or do they not need to?

malchata:
I hadn't asked, but if you wanted to tag Neil on this, I'd be open either way to it. I was of the mind that we didn't necessarily need to ask, since it was a public tweet in the first place, but WDYT?

tunetheweb:
@neilstuartcraig any thoughts on this?

neilstuartcraig:
> @neilstuartcraig any thoughts on this?

I'm on my phone on holiday at the moment but if it's just including info on our hsts rollout then that's absolutely fine. If any more info would be helpful then please let me know. More than happy to provide it if it's going to help people

rachelandrew (approved):

src/site/content/en/blog/bbc-hsts/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+{% Aside 'caution' %}
+Take care to read the information in this section **very carefully** before you decide to deploy HSTS! Botching an HSTS deployment can result in serious problems for users!
+{% endAside %}
+
+The main drawback of deploying HSTS is if you're not ready to treat your origin as strictly secure. Let's say you have a number of subdomains you're serving resources from, but maybe not all of them are secure. In this scenario, **an HSTS header could break your website.**
+
+The BBC took the right approach to deploying HSTS. As [Neil Craig](https://twitter.com/tdp_org) mentioned in his tweet, the initial value that was set for bbc.com was `max-age=10`.
+
+This approach means that the policy was only initially effective for ten seconds. This doesn't provide much of a benefit, but the idea is to feel out whether there may be issues with applying HSTS at all. As time goes on, you can increase the policy incrementally and see if issues occur. At the time of this writing, bbc.com is specifying an HSTS policy of `max-age=3600`, and that will almost certainly increase over time.
```
tunetheweb, line 64: They've upped it: https://twitter.com/tdp_org/status/1551513913082216448?s=20&t=Z04sbUm1193SCTvau7jRsw

```suggestion
This approach means that the policy was only initially effective for ten seconds. This doesn't provide much of a benefit, but the idea is to feel out whether there may be issues with applying HSTS at all. As time goes on, you can increase the policy incrementally and see if issues occur. At the time of this writing, bbc.com is specifying an HSTS policy of `max-age=86400`, and that will almost certainly increase over time.
```

neilstuartcraig:
JFYI we bumped max-age to 86400 earlier today. 
Also, we did similar in BBC.co.uk last year. I have a paper doc which explains more. Happy to share if it's of use.

tunetheweb:
Spotted that just in time - see my last review above 4 hours ago 😁

malchata:
We updated the note in the doc to reflect that. I assume it will increase, and I think when you eventually get to a long-term HSTS policy and/or apply for the preload list, we'd like to update the doc to note that accomplishment!

Also happy to link to the BBC.co.uk paper in the post if you think it'd be helpful.

neilstuartcraig:
Awesome, thanks @malchata !
We'll bump max age 2x more, 1 month then 1 year. 
After that, I'm hoping we can add to our special case for preloading and include BBC.co.uk as well since that's on 1 year already.

malchata:
Cool! Feel free to update us here when you've gotten to the 1 year mark on your `max-age` directive so I know when to update the post.

neilstuartcraig:
Yep, will do. Should be in a few weeks if I remember my schedule correctly. Cheers and thanks for including us. This is one of those things where I think we should use our public service position to encourage others and give some guidance on how folks might go about it.

PR #8400, opened 2022-07-25
Adjust avatar offset for bigger number of hosts

devnook:
Adjust the spacing of the avatars:
1. Shrink the avatars overlap tp 30% so that the hosts faces are better visible.
2. Expand the overlap to 50% for 5 or 6 hosts to prevent current overflow:

Screenshot 2022-07-25 at 12 22 41


jeffposnick (approved):
👍

PR #8419, opened 2022-07-27
Unconditionally run presubmit workflow for all PRs

### LINKED ISSUE #8385: Consider unconditionally running the presubmit checks for all PRs
Many folks don't realize that they need to add the `$-presubmit` label, and our builds are a bit faster now, to the point where it's probably a good idea.

Here's a recent example of a PR that went through even though it eventually broke the build, but there was no indication of that in the PR: https://github.com/GoogleChrome/web.dev/pull/8384

jeffposnick:
Fixes #8385

As per recent discussion, the build time is short enough and the risk of breakage is high enough that we do want to go ahead and run presubmits for all PRs, without having to remember to add in the `$-presubmit` label.


devnook (commented):
Shouldn't this check run prod build instead of staging? Or would that be too long / expensive?

jeffposnick:
Historically we've been running staging in our presubmit, so I just left that as-is. Staging takes 6 minutes.

I'll add a commit to flip it to production and we can see how long it takes.

jeffposnick:
So, it's 18 minutes for the prod build, vs. 6 minutes for the staging build.

A staging build will be sufficient to catch any of the 11ty-related errors we've run into lately. A prod build would be most useful if you were actually examining the output of, e.g., the HTML minifier, but you can't do that for the GitHub Actions build anyway.

I would say we should just stick with a staging build for the presubmit check. What do you think, @devnook?

jeffposnick:
> A staging build will be sufficient to catch any of the 11ty-related errors we've run into lately.

I did want to acknowledge that a staging build would _not_ have caught the HTML minifier error addressed in https://github.com/GoogleChrome/web.dev/pull/8374, though. So it's a tradeoff.

rachelandrew (approved):
I'm keen we switch this on, will this run if someone makes a new commit on an already opened PR? That's the situation where we have had recent breakage, people run presubmit initially but then forget to do so later.

jeffposnick:
Yes, this should re-run each time there's an additional commit.

It's configured for `on: pull_request`, and as per the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

> Runs your workflow when activity on a pull request in the workflow's repository occurs. For example, if no activity types are specified, the workflow runs when a pull request is opened or reopened or when the head branch of the pull request is updated.

devnook (approved):

PR #8430, opened 2022-07-29
docs: add webtransport.day

luzhenqian:
docs: add webtransport.day


jecfish (approved):

jecfish:
@heyawhite The translation is accurate but the presubmit seems like not running. Do you know how can I proceed to merge it?

rachelandrew:
@jecfish presubmit now runs on each commit to a PR but that means older ones aren't running, we can admin merge them so let me know if there are others.

PR #8437, opened 2022-07-31
Adds a note in the INP documentation about asynchronous work.

malchata:

- [Per this tweet thread](https://twitter.com/isAdrisal/status/1552871710277263360),  the original INP doc does not specify whether asynchronous work is considered for INP. This update adds clarifying language to address this issue.

### DIFF
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -110,7 +110,7 @@ Since INP is experimental, the guidance around thresholds may change over time a
 
 The primary driver of interactivity is often JavaScript, though browsers do provide interactivity through controls _not_ powered by JavaScript, such as checkboxes, radio buttons, and controls powered by CSS.
 
-As far as INP goes, **only the following _synchronous_ interaction types are observed:**
+As far as INP goes, **only the following interaction types are observed:**
 
 - Clicking with a mouse.
 - Tapping on a device with a touchscreen.
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -110,14 +110,19 @@ Since INP is experimental, the guidance around thresholds may change over time a
 
 The primary driver of interactivity is often JavaScript, though browsers do provide interactivity through controls _not_ powered by JavaScript, such as checkboxes, radio buttons, and controls powered by CSS.
 
-As far as INP goes, **only the following interaction types are observed:**
+As far as INP goes, **only the following _synchronous_ interaction types are observed:**
 
 - Clicking with a mouse.
 - Tapping on a device with a touchscreen.
 - Pressing a key on either a physical or onscreen keyboard.
 
 {% Aside 'important' %}
+<p>
 Hovering and scrolling does not factor into INP. However, scrolling with the keyboard (space bar, page up, page down, and so forth) involves a keystroke, which may trigger other events that INP _does_ measure. Any resulting scrolling is _not_ factored into how INP is calculated.
+</p>
+<p>
+Additionally, only synchronous work is considered by INP. Asynchronous work from a previous interaction _may_ factor into the latency observed in a later interaction if the result of that asynchronous work kicks off a task that blocks the main thread.
+</p>
 {% endAside %}
 
 Interactions may consist of two parts, each with multiple events. For example, a keystroke consists of the `keydown`, `keypress`, and `keyup` events. Tap interactions contain `pointerup` and `pointerdown` events. The event with the longest duration within the interaction is chosen as the interaction's latency.


tunetheweb (approved):

malchata:
Thanks, @tunetheweb! Before merging, I'll have either @anniesullie or @mmocny take a look at the language to make sure it's good.

src/site/content/en/metrics/inp/index.md
```diff
@@ -110,14 +110,19 @@ Since INP is experimental, the guidance around thresholds may change over time a
 
 The primary driver of interactivity is often JavaScript, though browsers do provide interactivity through controls _not_ powered by JavaScript, such as checkboxes, radio buttons, and controls powered by CSS.
 
-As far as INP goes, **only the following interaction types are observed:**
+As far as INP goes, **only the following _synchronous_ interaction types are observed:**
```
anniesullie, line 113: I think adding the word synchronous to describe interaction types is confusing. Interactions themselves aren't really defined as synchronous or asynchronous; instead some of the work they kick off is synchronous and some of it asynchronous.

For example, let's go back to the tweet thread with the two codepens.

In [Adriano's example](https://codepen.io/isadrisal/pen/abYVVGd), there *is* some synchronous work, just not very much:

```
  const target = evt.target;
  if (target.tagName !== 'BUTTON') return;
  
  const originalText = target.textContent;
  const delay = target.dataset.slow ? 2000 : 0;
  
  setTimeout(()
```

That will be measured for INP, and we'll generally get a score of 0 because there is so little work. But it's still measured. The rest of the work is asynchronous, because it doesn't need to happen in the event handler or in order to display a frame:

```
    target.textContent = 'Clicked';
    target.classList.add('clicked');
    setTimeout(() => {
      target.textContent = originalText;
      target.classList.remove('clicked');
    }, 400);
```

If we waited for _all_ the asynchronous work to complete, we'd wait 2000ms for the first setTimeout, then ~0ms for the code to run, then 400ms for the second setTimeout, and then ~16ms for another frame. And that would be really unfair, because we'd be counting not just the time to update the UI, but also the animation to un-highlight the button. And that's one reason we don't currently count asynchronous work; it's very hard to understand what the user is or isn't waiting for.

In [my example](https://codepen.io/anniesullie/pen/eYMeQBo) there is also synchronous work:

```
  const target = evt.target;
  if (target.tagName !== 'BUTTON') return;
  
  const originalText = target.textContent;
  const delay = target.dataset.slow ? 2000 : 0;
  
  const startTime = performance.now();
  let currentTime = performance.now();
  while (currentTime - startTime < delay) {
    currentTime = performance.now();
  }
  
  target.textContent = 'Clicked';
  target.classList.add('clicked');
  setTimeout(() 
```

This work takes much longer and blocks the main thread because of the busy loop. Since I put it in the event handler, it needs to run before the main thread for the page can do other work, so the UI is locked for 2 seconds. And I get a resulting bad INP score of 2016ms (the 2 seconds plus the time to update the UI)

But my version _also_ does some asynchronous work, this code inside the setTimeout:
```
    target.textContent = originalText;
    target.classList.remove('clicked');
```

This additional wait for 400ms to run this code and the time to run the code and produce a frame won't be counted towards INP.

anniesullie: To clarify, my feedback would be to not change line 113.

src/site/content/en/metrics/inp/index.md
```diff
@@ -110,14 +110,19 @@ Since INP is experimental, the guidance around thresholds may change over time a
...
 - Clicking with a mouse.
 - Tapping on a device with a touchscreen.
 - Pressing a key on either a physical or onscreen keyboard.
 
 {% Aside 'important' %}
+<p>
 Hovering and scrolling does not factor into INP. However, scrolling with the keyboard (space bar, page up, page down, and so forth) involves a keystroke, which may trigger other events that INP _does_ measure. Any resulting scrolling is _not_ factored into how INP is calculated.
+</p>
+<p>
+Additionally, only synchronous work is considered by INP. Asynchronous work from a previous interaction _may_ factor into the latency observed in a later interaction if the result of that asynchronous work kicks off a task that blocks the main thread.
```
anniesullie, line 124: This looks good to me, but would love to hear @mmocny's thoughts as well. It's a holiday in Canada so may need to wait till tomorrow.

src/site/content/en/metrics/inp/index.md
philipwalton: I agree with not changing this line.

I think the place to mention that it's only sync/blocking work that INP considers is either in the "What is INP?" or "What's in an interaction?" sections. Both of those sections break down INP and define the terms "interaction latency" and "event duration", and I think that's where we want to clarify that any work that gets scheduled/postpone to run in a future frame will not contribute to INP.

Also (meta point), I'm not 100% sure we want to refer to this a sync vs. async work, since microtasks are async but they can and do block the main thread.

src/site/content/en/metrics/inp/index.md
malchata: @mmocny WDYT?

src/site/content/en/metrics/inp/index.md
malchata: I think the aside I placed later on in this section clarifies this sufficiently, @philipwalton, but let me know what you think. The changes are on lines 120-124.

src/site/content/en/metrics/inp/index.md
mmocny: (thanks for ping, I forgot to review this patch)

I... worry that we are trying to add clarifications in ways that actually makes it less clear.  Sorry :(

How should I interpret that "**only synchronous work** is considered for INP" but also "Asynchronous work [...] may affect INP"?

Also it's not just "async work _from a previous interaction_"... it's **any long tasks** on main regardless where they come from.

There is a nuanced point to make that if you move code that used to run sync inside the handlers of an Event and was blocking next frame, by scheduling it to run in a future task after frame... it will still be bad for performance in the long run -- but we can just make that point directly: **INP is affected by long tasks**.  Shuffling long tasks around can help sometimes but is still problematic.

Scott H recently put together a very good document explaining how scheduling affects INP and I'll share that in case it helps.

---

Here's what I think I would suggest: we may need to **double down** on explaining what it means to measure **Interaction** to **Next Paint** so folks understand it so intuitively, they don't need clarifications.  If folks ("really") understood that concept I think these nuanced questions would become more obvious.  My 2cents.

Perhaps that means reminding folks about [rendering architecture](https://developer.chrome.com/articles/renderingng/), or [the event loop](https://www.youtube.com/watch?v=cCOL7MC4Pl0), or explaining the life of a frame in intuitive terms by using an interactive demo like Annie's cookie clicker.

Philip's Optimizing LCP presentation from Google I/O, which breaks down the time that blocks LCP into intuitive segments, I think is a good aspirational goal.  But it took years to get to the point where we had that level of intuition, so it may just take time?

---

Besides just understanding what INP measures at a more intuitive level, I think some things folks are specifically confused about INP definition:

- When there is no need for a new paint (i.e. no visual update) folks assume we will wait and measure until some eventual visual update: 
  - Maybe that's a misunderstanding of the term "Paint" which we need to explain?

- When the Next Paint is "unrelated" to the interaction itself, folks don't consider the Interaction "complete", and so also assume we will keep measuring:
  - Maybe we explain that INP measures the effects of the interaction on the page performance overall, not on the quality/semantic-value of the feedback to the interaction itself.

malchata: I think changing the focus to long tasks is a good first step, but do you think this would benefit from a short explanation about the event loop? Nothing hairy like we did in the watch party today, but a simple explanation that explains _why_ long tasks block? Or is this too much background information for what could be a more direct explanation?

malchata:
Going to take another shot at this with an event loop-centric approach and stronger emphasis on the role of long tasks.

PR #8444, opened 2022-08-01
Improve regex in markdown hack

jakearchibald:
The previous regex didn't correctly handle more than two newlines in a row. This now matches each blank line separately.


jakearchibald:
Should I be worried about that Percy test?

rachelandrew:
no, it's generally spurious

rachelandrew (approved):

PR #8452, opened 2022-08-02
Fixes pattern embed path

argyleink:
A pattern isn't embedding, it was missed by me when I updated the others to their newer `patternId`'s. This updates that pattern as well.


malchata (approved):
LGTM.

PR #8454, opened 2022-08-03
Adds guidance around LCP for preconnects.

malchata:

- Adds guidance around LCP for `preconnect`.

### DIFF
--- a/src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
+++ b/src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
@@ -150,9 +150,13 @@ Implementing `dns-prefetch` fallback in the same `<link>` tag causes a bug in Sa
 
 ## Effect on Largest Contentful Paint (LCP)
 
-The goal of `dns-prefetch` and `preconnect` is to lessen the amount of time it takes to connect to another origin. The ultimate result is that the time to discover a resource loaded from another origin should be reduced, even if only somewhat.
+Using `dns-prefetch` and `preconnect` allows sites to reduce the amount of time it takes to connect to another origin. The ultimate aim is that the time to load a resource from another origin should be minimized as much as possible.
 
-Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, the least you can do to load a cross-origin resource faster is to use `preconnect`&mdash;but a well-placed [`preload`](/preload-critical-assets/) with a [`fetchpriority` value of `"high"`](/priority-hints/#:~:text=Note%20that%20preload%20is%20still%20required%20for%20the%20early%20discovery%20of%20LCP%20images%20included%20as%20CSS%20backgrounds%20and%20can%20be%20combined%20with%20priority%20hints%20by%20including%20the%20fetchpriority%3D%27high%27%20on%20the%20preload%2C%20otherwise%20it%20will%20still%20start%20with%20the%20default%20"Low"%20priority%20for%20images) would most likely be better for resources that aren't immediately discoverable, since [LCP candidates](/lcp/#what-elements-are-considered) are crucial parts of the user experience. A `preconnect` is sensible enough, but only if a `preload` isn't a fit.
+Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, it is better that resources are immediately discoverable, since [LCP candidates](/lcp/#what-elements-are-considered) are crucial parts of the user experience. A [`fetchpriority` value of `"high"`](/priority-hints/when-would-you-need-priority-hints) on LCP resources can further improve this by signalling the importance of this asset to the browser so it can fetch it early.
+
+Where it is not possible to make LCP assets immediately discoverable, a [`preload`](/preload-critical-assets/) link&mdash;also with the `fetchpriority` value of `"high"`&mdash;still allows the browser to load the resource as soon as possible.
+
+If neither of these options are available&mdash;because the exact resource will not be known until later in the page load&mdash;you can use `preconnect` on cross-origin resources to reduce the impact of the late discovery of the resource as much as possible.
 
 Additionally, `preconnect` is less expensive than `preload` in terms of bandwidth usage, but still not without its risks. As is the case with excessive `preload` hints, excessive `preconnect` hints still consume bandwidth where TLS certificates are concerned. Be careful not to preconnect to too many origins, as this may cause bandwidth contention.
 
--- a/src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
+++ b/src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
@@ -3,7 +3,7 @@ title: Establish network connections early to improve perceived page speed
 subhead: |
     Learn about rel=preconnect and rel=dns-prefetch resource hints and how to use them.
 date: 2019-07-30
-updated: 2022-08-03
+updated: 2022-08-07
 hero: image/admin/Dyccd1RLN0fzhjPXswmL.jpg
 alt: Adam's Creation by Michelangelo on Sistine Chapel ceiling
 authors:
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
```diff
@@ -146,5 +148,14 @@ Implementing `dns-prefetch` fallback in the same `<link>` tag causes a bug in Sa
 
 {% endCompare %}
 
+## Effect on Largest Contentful Paint (LCP)
+
+The goal of `dns-prefetch` and `preconnect` is to lessen the amount of time it takes to connect to another origin. The ultimate result is that the time to discover a resource loaded from another origin should be reduced, even if only somewhat.
+
+Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, the least you can do to load a cross-origin resource faster is to use `preconnect`&mdash;but a well-placed [`preload`](/preload-critical-assets/) would most likely be better, as an [LCP candidate](/lcp/#what-elements-are-considered) is a crucial part of the user experience. That said, a `preconnect` can only help, even if not as much as a `preload`.
```
tunetheweb, line 155: A preload image without `fetchpriority="high"` would still not be processed until later. So actually a `preconnect` may be faster. 🤔

Best would be `preload` with `fetchpriority="high"` rather than just `preload`. It's a common misconception that it will cause it to be fetched with high priority just be using the `preload` but that's not the case (and don't think it ever was?). See [here for more info](https://web.dev/priority-hints/#:~:text=Note%20that%20preload%20is%20still%20required%20for%20the%20early%20discovery%20of%20LCP%20images%20included%20as%20CSS%20backgrounds%20and%20can%20be%20combined%20with%20priority%20hints%20by%20including%20the%20fetchpriority%3D%27high%27%20on%20the%20preload%2C%20otherwise%20it%20will%20still%20start%20with%20the%20default%20%22Low%22%20priority%20for%20images), and maybe consider linking to it.

However, I think I would add a line about "for best results serve critical results like LCP resources from the same origin, but where that is not possible, you can at least reduce the impact with `preconnect` or `preload`.

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
```diff
@@ -146,5 +148,14 @@ Implementing `dns-prefetch` fallback in the same `<link>` tag causes a bug in Sa
...
+
+All that said, `preconnect` is less expensive than `preload` in terms of bandwidth usage. While multiple `preload` hints can cause bandwidth contention, the same is less true for `preconnect`. Use it to reduce resource discovery time, which can have a positive effect on LCP.
```
tunetheweb, line 157: "That said" in last paragraph. "All that said" in this paragraph. Can we reword to avoid the repetition.

And it's true that `preconnect` is less expensive, but still can use up a bit of bandwith with TLS certificates - especially at this critical stage of the initial load of the page. We've seen some sites use a quite-frankly [ridiculous 1286 preconnects](https://docs.google.com/spreadsheets/d/1Mw6TjkIClRtlZPHbij5corOZbaSUp-vgTVq3Ig18IwQ/edit#gid=547570509) which surely has an impact on bandwidth.

So maybe tone down this sentence or add a warning?

tunetheweb (changes requested):

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
tunetheweb: And actually, another point is that, with the LCP image discoverable in the initial HTML and a `fetchpriority="high"` you might not even need, nor benefit, from the preload. It’s better for hidden resources.

philipwalton: +1 to Barry's last comment. I don't think we want to be recommending preconnect in situations where the LCP resource is discoverable from within the main document. It's really only useful for resources that are not discovered right away, and then the connection can be established in parallel in that discoveraby phrase.

Overall I think the main thing this document is missing as a clear explanation of _when_ preconnect is useful and when it's not (which is true for a lot of these older docs).

As a meta point: In its current state, this document makes it sounds a bit like `preconnect` is one weird trick that will always improve your load performance by 100-500ms.

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
```diff
@@ -146,5 +148,14 @@ Implementing `dns-prefetch` fallback in the same `<link>` tag causes a bug in Sa
 
 {% endCompare %}
 
+## Effect on Largest Contentful Paint (LCP)
+
+The goal of `dns-prefetch` and `preconnect` is to lessen the amount of time it takes to connect to another origin. The ultimate result is that the time to discover a resource loaded from another origin should be reduced, even if only somewhat.
```
tunetheweb, line 153: Not sure it's a "goal" as such. A "benefit" maybe.

```suggestion
Using `dns-prefetch` and `preconnect` allows sites to reduce the amount of time it takes to connect to another origin. The ultimate aim is that the time to load a resource from another origin should be minimized as much as possible.
```

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
```diff
@@ -146,5 +148,14 @@ Implementing `dns-prefetch` fallback in the same `<link>` tag causes a bug in Sa
...
+
+Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, the least you can do to load a cross-origin resource faster is to use `preconnect`&mdash;but a well-placed [`preload`](/preload-critical-assets/) with a [`fetchpriority` value of `"high"`](/priority-hints/#:~:text=Note%20that%20preload%20is%20still%20required%20for%20the%20early%20discovery%20of%20LCP%20images%20included%20as%20CSS%20backgrounds%20and%20can%20be%20combined%20with%20priority%20hints%20by%20including%20the%20fetchpriority%3D%27high%27%20on%20the%20preload%2C%20otherwise%20it%20will%20still%20start%20with%20the%20default%20"Low"%20priority%20for%20images) would most likely be better for resources that aren't immediately discoverable, since [LCP candidates](/lcp/#what-elements-are-considered) are crucial parts of the user experience. A `preconnect` is sensible enough, but only if a `preload` isn't a fit.
```
tunetheweb, line 155: WDYT about chaning the ordering of this to make it more readable:

```suggestion
Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, it is better that resources are immediately discoverable, since [LCP candidates](/lcp/#what-elements-are-considered) are crucial parts of the user experience. A [`fetchpriority` value of `"high"`](/priority-hints/when-would-you-need-priority-hints) on LCP resources can further improve this by signalling the importance of this asset to the browser so it can fetch it early.

Where it is not possible to make LCP assets immediately discoverable, a [`preload`](/preload-critical-assets/) link&mdash;also with the `fetchpriority` value of `"high"`&mdash;still allows the browser to load the resource as soon as possible.

If neither of these options are available&mdash;because the exact resource will not be known until later in the page load&mdash;you can use `preconnect` on cross-origin resources to reduce the impact of the late discovery of the resource as much as possible.
```

tunetheweb (changes requested):

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
malchata: Agreed. Added to the batch,

src/site/content/en/blog/preconnect-and-dns-prefetch/index.md
```diff
@@ -3,10 +3,12 @@ title: Establish network connections early to improve perceived page speed
 subhead: |
     Learn about rel=preconnect and rel=dns-prefetch resource hints and how to use them.
 date: 2019-07-30
+updated: 2022-08-03
```
malchata, line 6:
```suggestion
updated: 2022-08-07
```

tunetheweb (approved):

PR #8463, opened 2022-08-05
Do not make web-navigation-drawer inert. Fixes #8462

### LINKED ISSUE #8462: When search input has focus, you can't type anything into the field
**Describe the bug**
When you attempt to search for content on the site, the input field loses focus, and you are unable to type anything into the field. You can still browse the page, however.

Chrome, Edge Chromium and Safari are impacted, but not Firefox. Occurs on both Windows and Mac OS operating systems. Issue also occurs in Chrome on Android.

Although I can use Firefox for now, it's clearly quite a nasty little bug for newcomers and regular visitors to your otherwise great site, if using other browsers.

**To Reproduce**
1. Go to [https://web.dev/learn/css/](https://web.dev/learn/css/)
2. Click on search input field.
3. Input field loses focus, you are unable to type anything.

**Expected behavior**
I should be able to search for content. This occurs in Edge Chromium on both Windows and Mac OS, and Safari. However, the input field works in Firefox.

**Screenshots**
I had thought that the console error screenshotted below was the cause of the problem. But no further output is generated, whether you click on the search field or not. This same error is output in Firefox's console log, but you can still search.

console-error

bramus:
This PR takes a different approach to solving #8134, thereby fixing the unpleasant side-effect as described in #8462. 

Because the whole `web-navigation-drawer` was made `inert` in #8134, the search box –which is a child of `web-navigation-drawer` on some pages– would no longer work.

Instead of making the entire `web-navigation-drawer` `inert`, this PR sets its `pointer-events` to `none`. This solves #8134 and fixes #8462 as well.

### DIFF
--- a/src/scss/blocks/_site-header.scss
+++ b/src/scss/blocks/_site-header.scss
@@ -138,6 +138,7 @@
 
   .site-header[class*='expanded'] web-navigation-drawer {
     opacity: 0;
+    pointer-events: none !important;
   }
 
   .site-header__nav {
--- a/src/lib/components/NavigationDrawer/index.js
+++ b/src/lib/components/NavigationDrawer/index.js
@@ -86,11 +86,8 @@ export class NavigationDrawer extends BaseStateElement {
     this.addEventListener('click', closeNavigationDrawer);
   }
 
-  onStateChanged({isSearchExpanded, isNavigationDrawerOpen, currentUrl}) {
+  onStateChanged({isNavigationDrawerOpen, currentUrl}) {
     this.open = isNavigationDrawerOpen;
-
-    // See https://github.com/GoogleChrome/web.dev/issues/8131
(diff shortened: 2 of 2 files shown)


src/scss/blocks/_site-header.scss
```diff
@@ -138,6 +138,7 @@
 
   .site-header[class*='expanded'] web-navigation-drawer {
     opacity: 0;
+    pointer-events: none !important;
```
bramus, line 141: FWIW: The reason I used `!important` here is because a few lines down it has `pointer-events` explicitly set. I don’t want to remove the line below, as I can’t gauge which side-effects that would cause.

rachelandrew (approved):
This fixes the original issue and the new issue so let's go with it. I'll flag it up to Ewa next week. Thank you!

PR #8464, opened 2022-08-05
fix stale-while-revalidate example in Learn/PWA

lionralfs:

- Fixes the example code for the stale-while-revalidate caching pattern [example code](https://web.dev/learn/pwa/serving/#stale-while-revalidate) in the Learn PWA section. Makes sure the network response is used (otherwise, `networkFetch` would be a promise resolving to `undefined`).


rachelandrew (approved):

matthiasrohmer:
Thanks for your contribution, this has already been fixed in another PR!

PR #8465, opened 2022-08-05
Add flags to the ffmpeg command for creating GIFs

blowery:
This adds flags to the ffmpeg command for creating an MP4 from a GIF that tag the output in the sYCC color space, which fixes the "washed out" colors and color shifts often reported when doing this conversion. Without the flags, browsers tend to interpret the colorspace incorrectly.

Note this currently does not affect Firefox, but does improve display in Chrome and Safari.

### DIFF
--- a/src/site/content/en/fast/replace-gifs-with-videos/index.md
+++ b/src/site/content/en/fast/replace-gifs-with-videos/index.md
@@ -47,7 +47,7 @@ To use FFmpeg to convert the GIF, `my-animation.gif` to an MP4 video, run the
 following command in your console:
 
 ```bash
-ffmpeg -i my-animation.gif -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420p my-animation.mp4
+ffmpeg -i my-animation.gif -b:v 0 -crf 25 -f mp4 -vcodec libx264 -pix_fmt yuv420p -color_primaries bt709 -color_trc iec61966_2_1 -colorspace smpte170m my-animation.mp4
 ```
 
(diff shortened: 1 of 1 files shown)


matthiasrohmer (approved):
Thanks for your contribution! Just verified this, the right one is a little more saturated in Chrome.  The difference might get lost by screenshot and GitHub compression though...

image

Can you please [sign our CLA](https://github.com/GoogleChrome/web.dev/pull/8465/checks?check_run_id=7699100626)? Only then we are able to merge this 🙂

blowery:
@matthiasrohmer After further testing, we (Tumblr) did find a problem with these flags: they don't appear to work for any browser on Windows, or Firefox on any platform. So, while I still think they're correct, I'm not sure if you want to take them for docs just yet? 

Amazes me that it's 2022 and we're still dealing with colorspace issues, as an industry.

blowery:
> Just verified this, the right one is a little more saturated in Chrome.

There's also a Digital Color Meter utility on macOS you can use to compare point values

matthiasrohmer:
Oh, awesome! Thanks a lot for getting back on this, @blowery. Yes, then we better hold off from merging this and keep the universally working command for now. If you find a solution that works across browsers and platforms and has consistent colour representation we would really appreciate having this on web.dev. Until then, closing.

blowery:
> If you find a solution that works across browsers and platforms and has consistent colour representation...

Sadly that's ... on the browsers, the OS and the video card folks. Maybe y'all can push on https://bugs.chromium.org/p/chromium/issues/detail?id=1229547&q=component%3AInternals%3EMedia%3EVideo%20color&can=2 a bit? :)

PR #8472, opened 2022-08-07
Adds CLS and browser preload scanner guidance to the image dimensions guide.

malchata:

- The image dimensions guide deals largely with image file size, but there is a good place in the document to call out information on CLS-related issues, as well as evaluating whether an `` element is a better choice over CSS background images due to browser preload scanner optimizations.

### DIFF
--- a/src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
+++ b/src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
@@ -124,26 +124,23 @@ Responsive Images guide.
 
 ## Avoid layout shifts by specifying dimensions
 
-While this guide discusses image dimensions in the context of performance, it's important to note that reserving space for images in the layout is a crucial component of minimizing a page's [Cumulative Layout Shift](/cls/) metric. When serving images in HTML, be sure to use proper `width` and `height` attributes so that the browser knows how much space to allocate in the layout for the image:
+While this guide discusses image dimensions in the context of reducing the amount of unnecessary bytes downloaded, it's important to note that reserving the correct space for images in the layout is another crucial component of minimizing a page's [Cumulative Layout Shift](/cls/) metric. When serving images in HTML, be sure to use proper `width` and `height` attributes so that the browser knows how much space to allocate in the layout for the image:
 
 ```html
 <img src="flower.jpg" width="640" height="480" alt="A picture of a siberian iris.">
 ```
 
-Without these attributes, the browser has no idea how much space the image will take up until it has loaded. This will cause layout shifts in the document, which can be frustrating to users when they "miss" their intended hit target and end up clicking on something else they didn't intend to during page load.re-run
+Without these attributes, of the equivalent CSS sizing, the browser has no idea how much space the image will take up until it has loaded. This will cause layout shifts in the document, which can be frustrating to users when they "miss" their intended hit target and end up clicking on something else they didn't intend to during page load.re-run
 
-If you're using images in CSS, be sure to use the [`aspect-ratio` property](https://developer.mozilla.org/docs/Web/CSS/aspect-ratio) on the image's container. This has the same effect on an element's size that `width` and `height` attributes do in the sense that the container will maintain a consistent aspect ratio.
+An alternative to providing the width and height explicitly, is to use the  CSS [`aspect-ratio` property](https://developer.mozilla.org/docs/Web/CSS/aspect-ratio) on the image. This has a similar effect on an element's size that `width` and `height` attributes do in the sense that the container will maintain a consistent aspect ratio. However, the difference is that this may result in a different aspect-ratio being used than the image is provided in, so you will likely want to use an `object-fit` setting to ensure the image is not distorted in this explicit 16/9 view:
 
 ```css
-.masthead {
-  background: url('/flower.jpg');
+img {
   aspect-ratio: 16 / 9;
+  width: 100%;
+  object-fit: cover;
 }
 ```
-
-{% Aside 'important' %}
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
```diff
@@ -121,6 +122,29 @@ If you'll be resizing many images, you may find it more convenient to use a
...
+
+Without these attributes, the browser has no idea how much space the image will take up until it has loaded. This will cause layout shifts in the document, which can be frustrating to users when they "miss" their intended hit target and end up clicking on something else they didn't intend to during page load.re-run
+
+If you're using images in CSS, be sure to use the [`aspect-ratio` property](https://developer.mozilla.org/docs/Web/CSS/aspect-ratio) on the image's container. This has the same effect on an element's size that `width` and `height` attributes do in the sense that the container will maintain a consistent aspect ratio.
+
+```css
+.masthead {
+  background: url('/flower.jpg');
+  aspect-ratio: 16 / 9;
+}
```
tunetheweb, line 141: I find this section weird in this context and think we should remove it.

A CSS background-image will not resize the element after the image downloads to match the image's dimensions. Therefore the `aspect-ratio: 16 / 9;` is needed to size the element anyway. There will be no CLS without it (though the image also won't be shown!).

You can use `aspect-ratio` on an `img` element (Jake talked about it recently in this post: https://jakearchibald.com/2022/img-aspect-ratio/), but that's not as a background image and doesn't have quite "the same effect on an element's size that `width` and `height` attributes do".

So maybe rephrase it as:

```suggestion
An alternative to providing the width and height explicitly, is to use the  CSS [`aspect-ratio` property](https://developer.mozilla.org/docs/Web/CSS/aspect-ratio) on the image. This has a similar effect on an element's size that `width` and `height` attributes do in the sense that the container will maintain a consistent aspect ratio. However, the difference is that this may result in a different aspect-ratio being used than the image is provided in, so you will likely want to use an `object-fit` setting to ensure the image is not distorted in this explicit 16/9 view:

```css
img {
  aspect-ratio: 16 / 9;
  width: 100%;
  object-fit: cover;
}
```

src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
```diff
@@ -121,6 +122,29 @@ If you'll be resizing many images, you may find it more convenient to use a
...
+```
+
+{% Aside 'important' %}
+Always evaluate whether you really need CSS background images. Sometimes an HTML `<img>` element is a better choice. Using `<img~` will lead to better loading performance and a faster [Largest Contentful Paint (LCP)](/lcp/) (_if_ the image is an [LCP candidate](/lcp/#what-elements-are-considered)) as the browser's preload scanner will be able to start loading the image while rendering is blocked. The browser preload scanner [can't discover images ahead of time in CSS](/preload-scanner/#css-background-images), and so loading for them will be delayed until [the CSSOM is constructed and applied to the DOM](/critical-rendering-path-render-tree-construction/).
+{% endAside %}
+
```
tunetheweb, line 147: I'd remove this if taking above advice. It's no longer relevant now we're not talking about background-images anymore.

```suggestion
```

tunetheweb (changes requested):

src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
```diff
@@ -8,6 +8,7 @@ description: |
   page. The image looks fine, but it is wasting users' data and hurting page
   performance.
 date: 2018-11-05
+updated: 2022-08-09
```
malchata, line 11:
```suggestion
updated: 2022-08-07
```

src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
```diff
@@ -121,6 +122,26 @@ If you'll be resizing many images, you may find it more convenient to use a
...
 
+## Avoid layout shifts by specifying dimensions
+
+While this guide discusses image dimensions in the context of reducing the amount of unnecessary bytes downloaded, it's important to note that reserving the correct space for images in the layout is another crucial component of minimizing a page's [Cumulative Layout Shift](/cls/) metric. When serving images in HTML, be sure to use proper `width` and `height` attributes so that the browser knows how much space to allocate in the layout for the image:
+
+```html
+<img src="flower.jpg" width="640" height="480" alt="A picture of a siberian iris.">
+```
+
+Without these attributes, of the equivalent CSS sizing, the browser has no idea how much space the image will take up until it has loaded. This will cause layout shifts in the document, which can be frustrating to users when they "miss" their intended hit target and end up clicking on something else they didn't intend to during page load.re-run
```
tunetheweb, line 133: Typo of mine `of` - > `or`.

Also I think mis-clicks is only one impact of CLS. The "losing your place" is more frequent IMHO, more annoying, though at least is easier resolved. WDYT about this?:

```suggestion
Without these attributes, or the equivalent CSS sizing, the browser has no idea how much space the image will take up until it has loaded. This will cause layout shifts in the document, which can be frustrating to users when content moves after they have started consuming it. This can result in losing your place when reading, or to "miss" their intended hit target and end up clicking on something else they didn't intend to during page load.
```

tunetheweb (changes requested):

src/site/content/en/fast/serve-images-with-correct-dimensions/index.md
malchata: The subject shifts in the edited paragraph. I recommend:

"This can result in losing your place when reading, or to "miss" **your** intended hit target..."

tunetheweb: Good spot. Edited the suggestion the other way as more consistent to what you originally suggested.

Feel free to change to "your" if you prefer.

tunetheweb (approved):

PR #8487, opened 2022-08-09
Update Lighthouse audit source code links

tunetheweb:
In https://github.com/GoogleChrome/lighthouse/pull/14242 Lighthouse moved the audit source codes that we reference in the docs on web.dev and now the links are broken.

This PR updates them to the new location.


malchata (commented):
Obviously didn't check every link, but I think I get what you're trying to do here—and if I do, LGTM.

tunetheweb:
> Obviously didn't check every link, but I think I get what you're trying to do here—and if I do, LGTM.

Yeah it was a simple bulk update as just the folder name changed.

Can you approve and then I'll merge the change?

malchata (approved):
My bad, @tunetheweb, I thought I approved this change already.

PR #8489, opened 2022-08-09
Adds full viewport image guidance for LCP.

malchata:

- Updates the LCP candidate information by specifying that full-viewport images are no longer considered for LCP [since Chrome 88](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/speed/metrics_changelog/2020_11_lcp.md).

### DIFF
--- a/src/site/content/en/metrics/lcp/index.md
+++ b/src/site/content/en/metrics/lcp/index.md
@@ -100,7 +100,7 @@ considered for Largest Contentful Paint are:
   elements containing text nodes or other inline-level text elements children.
 
 {% Aside 'important' %}
-Starting in Chrome 88, [images that occupy the entire viewport are no longer considered LCP candidates](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/speed/metrics_changelog/2020_11_lcp.md).
+Images that occupy the entire viewport are not considered LCP candidates.
 {% endAside %}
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/lcp/index.md
```diff
@@ -99,6 +99,10 @@ considered for Largest Contentful Paint are:
 * [Block-level](https://developer.mozilla.org/docs/Web/HTML/Block-level_elements)
   elements containing text nodes or other inline-level text elements children.
 
+{% Aside 'important' %}
+Starting in Chrome 88, [images that occupy the entire viewport are no longer considered LCP candidates](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/speed/metrics_changelog/2020_11_lcp.md).
+{% endAside %}
```
tunetheweb, line 104: Chrome 88 was a long time ago. Do we need to be this specific? And should this still be an important aside - which I'd use for a breaking change, like the current wording is, but less so for what is now part of the regular definition.

```suggestion

{% Aside 'important' %}
Images that occupy the entire viewport are not considered LCP candidates.
{% endAside %}
```

malchata: Yeah, maybe the specificity is a bit much, but it is a historical artifact. I'm fine with the change, though.

PR #8498, opened 2022-08-10
Remove experimental items from Manifest article

jpmedley:
The article contained information that is still experimental and is behind a flag in Chrome. I'm working on an update to an article that covers the removed items.

### DIFF
--- a/src/site/content/en/progressive-web-apps/add-manifest/index.md
+++ b/src/site/content/en/progressive-web-apps/add-manifest/index.md
@@ -331,34 +331,6 @@ truncates longer descriptions
 right in your browser."
 ```
 
-#### `screenshots` {: #screenshots }
-
-The `screenshots` property is used with the description to show an installation
-prompt that contains more than just an application name and an installation
-button. It is an array of image objects representing your app in common usage
-scenarios. Each object must include the `src`, a `sizes` property, and the
-`type` of image. The `screenshots` property may contain no more than eight image
-objects.
-
-<figure>
-  {% Img src="image/xizoeLGxYNf3VLUHc5BsIoiE1Af1/SpStAtUk8Zp5iwi9yqKP.jpg", 
-alt="Richer Android installation prompt", width="342", height="722" %}
-  <figcaption>Richer Android installation prompt.</figcaption>
-</figure>
-
-In Chrome, the image must respond to certain criteria:
-
-* Width and height must be at least 320px and at most 3840px.
-* The maximum dimension can't be more than 2.3 times as long as the minimum
-  dimension.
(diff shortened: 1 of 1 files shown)


tomayac:
This is already [documented on MDN](https://developer.mozilla.org/en-US/docs/Web/Manifest/screenshots) and it's [completely non-controversial](https://github.com/w3c/manifest-app-info/issues/25). I suggest to leave the information in the article.

tomayac (commented):
This is already [documented on MDN](https://developer.mozilla.org/en-US/docs/Web/Manifest/screenshots) and it's [completely non-controversial](https://github.com/w3c/manifest-app-info/issues/25). I suggest to leave the information in the article.

jpmedley:
They should not be on MDN because they're still behind a flag. (I hope it wasn't me.) I'll check about when these might be shipping by default, but I may need to take them down from MDN as well.

These are in the new article draft I wrote for richer installs.

beaufortfrancois:
Those items are neither behind a flag, nor experimental. They are supported in stable channel on Chrome for Android and soon™️  on Desktop.

Regarding a new article, we already have https://developer.chrome.com/blog/richer-pwa-installation/. Why is another required?

I wished @jpmedley  would have been pinged about this as I was the one authoring this part of the article.

jpmedley:
It's technically an update because new members are being added to `screenshots`. Rewriting seemed like the better approach, but I realize that I misunderstood the current state.

beaufortfrancois:
@jpmedley I'd suggest we add `platform` to this article and explain it is supported only in Chrome 106 for now.

jpmedley:
I agree, except that we can't have experimental features on web.dev any more.

Closing this PR because there's nothing I can do to it right now.

beaufortfrancois:
As @tomayac said, the `platform` member is not experimental. It is standardised: https://www.w3.org/TR/manifest-app-info/#platform-member

I believe we should document it in web.dev.

beaufortfrancois:
@jpmedley Do you have plans to update this web.dev article once dcc article is live? If not, I'll do it.

jpmedley:
I do. Thank you for the offer.

beaufortfrancois:
@jpmedley What is the current status?

jpmedley:
I closed this PR because I was misinformed when I opened it.

beaufortfrancois:
Sorry. I meant about https://github.com/GoogleChrome/web.dev/pull/8498#issuecomment-1229901381

tropicadri:
@beaufortfrancois I'm taking over the article from @jpmedley I'll sync with you tomorrow

PR #8510, opened 2022-08-12
There are two values for attribute “loading”, not three

IsaacLK:
The document said there was three possible values for the attribute “loading”, when only two `exist`


tunetheweb (approved):
Thanks! My fault as I recently update this doc to remove the non-standard `auto` which is no longer used. Obviously missed the number above this 🤦‍♂️

malchata (approved):
LGTM!

PR #8512, opened 2022-08-12
Support BYOB reader for SerialPort's ReadableStreams

beaufortfrancois:
Following https://bugs.chromium.org/p/chromium/issues/detail?id=1182905#c11, this PR documents BYOB readers in the web.dev/serial article.

@reillyeon Please have a look

### DIFF
--- a/src/site/content/en/blog/serial/index.md
+++ b/src/site/content/en/blog/serial/index.md
@@ -258,8 +258,7 @@ To specify the buffer into which data is read instead of the stream allocating a
 ```js
 try {
   const reader = port.readable.getReader({ mode: "byob" });
-  let buffer = new ArrayBuffer(512);
-  // Call reader.read() to read data into the buffer...
+  // Call reader.read() to read data into a buffer...
 } catch (error) {
   if (error instanceof TypeError) {
     // BYOB readers are not supported.
@@ -272,46 +271,16 @@ Here's an example of how to reuse the buffer out of `value.buffer`:
 
 ```js
 const bufferSize = 8 * 1024; // 8kB
-let reader;
-while (port.readable) {
-  try {
-    try {
-      reader = port.readable.getReader({ mode: "byob" });
-    } catch {
-      reader = port.readable.getReader();
-    }
+let buffer = new ArrayBuffer(bufferSize);
 
-    let buffer = null;
-    while (true) {
-      const { value, done } = await (async () => {
-        if (reader instanceof ReadableStreamBYOBReader) {
-          if (!buffer) {
-            buffer = new ArrayBuffer(bufferSize);
-          }
-          const { value, done } = await reader.read(
-            new Uint8Array(buffer, 0, bufferSize)
-          );
-          buffer = value?.buffer;
-          return { value, done };
-        } else {
-          return await reader.read();
-        }
-      })();
-      if (done) {
-        // reader.cancel() has been called.
-        break;
-      }
-      // Handle value.
-    }
-  } catch (e) {
-    // Handle error...
-  } finally {
-    if (reader) {
-      // Allow the serial port to be closed later.
-      reader.releaseLock();
-      reader = undefined;
-    }
+const reader = port.readable.getReader({ mode: "byob" });
+while (true) {
+  const { value, done } = await reader.read(new Uint8Array(buffer));
+  if (done) {
+    break;
   }
+  buffer = value.buffer;
+  // Handle `value`.
 }
 ```
 
@@ -332,6 +301,9 @@ async function readInto(reader, buffer) {
   }
   return buffer;
 }
+
+// Set `bufferSize` on open() to at least the size of the buffer.
+await port.open({ baudRate: 9600, bufferSize: 1024 });
 
 const reader = port.readable.getReader({ mode: "byob" });
 let buffer = new ArrayBuffer(512);
--- a/src/site/content/en/blog/serial/index.md
+++ b/src/site/content/en/blog/serial/index.md
@@ -4,7 +4,7 @@ subhead: The Web Serial API allows websites to communicate with serial devices.
 authors:
   - beaufortfrancois
 date: 2020-08-12
-updated: 2022-08-12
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/serial/index.md
```diff
@@ -253,6 +253,21 @@ while (true) {
 }
 ```
 
+To specify the buffer into which data is read instead of the stream allocating a buffer for each chunk, get a [ReadableStreamBYOBReader] by calling `port.readable.getReader({ mode: "byob" })` and read data into a new `ArrayBuffer`. Note that the Web Serial API supports "Bring Your Own Buffer" readers in Chrome 106 or later.
+
+```js
+try {
+  const reader = port.readable.getReader({ mode: "byob" });
+  let buffer = new ArrayBuffer(512);
+  // Call reader.read() to read data into the buffer...
```
reillyeon, line 262: Can you include some examples of how to call `read()` with a BYOB reader? When writing the tests I encountered a few gotchas:

1. You have to pass an `ArrayBufferView` (like `Uint8Array`) rather than a raw ArrayBuffer.
2. The buffer is transferred so if you want to reuse it (which is half the point) you need to extract it out of `value.buffer`.

See the `readInto()` function from https://github.com/web-platform-tests/wpt/blob/master/serial/serialPort_readable_byob.https.any.js#L5 for an example of how to use a BYOB reader to read a specific amount of data from the port.

See https://github.com/GoogleChromeLabs/serial-terminal/blob/main/src/index.ts#L271 for how I implemented fallback and reuse of the buffer.

beaufortfrancois: Thanks for sharing those examples. I've added some.

src/site/content/en/blog/serial/index.md
```diff
@@ -253,6 +253,94 @@ while (true) {
...
+  if (error instanceof TypeError) {
+    // BYOB readers are not supported.
+    // Fallback to port.readable.getReader()...
+  }
+}
+```
+
+Here's an example of how to reuse the buffer out of `value.buffer`:
+
+```js
```
reillyeon, line 273: This example ends up being kinda complicated because it's combining a bunch of things together. It handles transient errors with the `while (port.readable)` and also has a fallback if BYOB buffers aren't supported. Do you think we should pull it all together into a single example or keep it short and just demonstrate one concept. Like,

```js
const bufferSize = 8 * 1024; // 8kB
let buffer = new ArrayBuffer(bufferSize);

const reader = port.readable.getReader({ mode: 'byob' });
while (true) {
  const { value, done } = await reader.read(new Uint8Array(buffer));
  if (done) {
    break;
  }
  buffer = value.buffer;
  // Handle `value`.
}
reader.releaseLock();
```

src/site/content/en/blog/serial/index.md
```diff
@@ -253,6 +253,94 @@ while (true) {
...
+const bufferSize = 8 * 1024; // 8kB
```
reillyeon, line 274: Another potential footgun, if the developer is providing a large buffer to `read()` they should make sure to set `bufferSize` on `open()` to at least as much.

beaufortfrancois: I've updated one example to include that hint.

src/site/content/en/blog/serial/index.md
beaufortfrancois: I've kept it short. WDYT?

src/site/content/en/blog/serial/index.md
reillyeon: Can you put that in this example? It's not important for the "read exactly this much" example because in that case we retry until the whole buffer is filled. In the case where we are explicitly trying to reuse the buffer from offset 0 however there's no point specifying a large buffer if the buffer size specified on open is smaller.

beaufortfrancois: Done. Sorry for that.

reillyeon (approved):

src/site/content/en/blog/serial/index.md
```diff
@@ -253,6 +253,66 @@ while (true) {
 }
 ```
 
+To specify the buffer into which data is read instead of the stream allocating a buffer for each chunk, get a [ReadableStreamBYOBReader] by calling `port.readable.getReader({ mode: "byob" })` and read data into a new `ArrayBuffer`. Note that the Web Serial API supports "Bring Your Own Buffer" readers in Chrome 106 or later.
```
jpmedley, line 256:
```suggestion
To specify the buffer into which data is read don't allocate a buffer for each chunk. Instead, get a [ReadableStreamBYOBReader] by calling `port.readable.getReader({ mode: "byob" })` and read data into a new `ArrayBuffer`. Note that the Web Serial API supports "Bring Your Own Buffer" readers in Chrome 106 or later.
```

jpmedley:
Sorry for the delay. I'm a bit behind with my email.

src/site/content/en/blog/serial/index.md
reillyeon: It is the stream allocating the buffer, not the developer. We're telling them that with a BYOB reader they can take over allocating buffers and reuse them as well. The message needs to be that it puts them in control.

jpmedley: Reworded to clarify.

reillyeon: How about,

```suggestion
You can take control of how memory is allocated when you read from the stream using a "Bring Your Own Buffer" reader. Call `port.readable.getReader({ mode: "byob" })` to get the [ReadableStreamBYOBReader] interface and provide your own `ArrayBuffer` when calling `read(). Note that the Web Serial API supports this feature in Chrome 106 or later.
```

jpmedley: Sounds good.

beaufortfrancois:
@jpmedley I've commited @reillyeon's suggestion. Please review.

jpmedley (approved):

PR #8513, opened 2022-08-12
Update Optimize LCP post with more fetchpriority guidance

tunetheweb:

- Include `fetchpriority="high"` when preloading images.
- Change lazy-loading text "deprioritized" to "delay" to more accurately reflect what this does.
- Move lazy loading warning to be beside lazy loading text.
- Explain why images are not fetched with a high priority even if discovered early.


malchata (approved):
LGTM.

PR #8520, opened 2022-08-12
Increase Memory for Percy to prevent CI failures

tunetheweb:
See if this fixes Percy since it started failing 

- Increase memory size for node for Percy CI.

### DIFF
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -15,6 +15,7 @@ on:
       - '**.scss'
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


tunetheweb:
@rachelandrew @malchata any ideas who should review this? I'm guessing we've just got too big for the default memory size.

Increasing it seems to allow the Percy CI check to complete (though now complaining that there are changes, but that's more because of what's been merged since this stopped working).

malchata:
Not certain, though it seems like it might be a task for web.dev eng. @devnook, do you know who would be best to look this over?

devnook (approved):

PR #8312, opened 2022-07-11
Various updates to image lazy-loading articles

tunetheweb:
Fixes #7588 and did quite a few other updates after doing a full review of the article.

Changes proposed in this pull request:

- Updates with latest browser support
- Removes mention of `auto` (since it's effectively gone now) and clarifies `eager`
- Adds section and links for [Priority Hints](https://web.dev/priority-hints/).
- Removes all but one Lite Mode references.
- Add reference to [The performance effects of too much lazy-loading](https://web.dev/lcp-lazy-loading/) article, and remove note that the impact is small.
- Rewrite section on third-party libraries and remove note when using both (I don't see how that worked when they must have used `data-src` rather than `src`?).
- Adds reference to [It's time to lazy-load offscreen iframes!](https://web.dev/iframe-lazy-loading/) article.

Quite a few changes here, so let me know your thoughts if you disagree with any of them, or have better wording in mind.

I've also updated the [iframe lazy loading article](https://web.dev/iframe-lazy-loading/) to remove references to `auto` (including translations for this, as smaller change).

PR #8529, opened 2022-08-15
Fix counts in iframes lazy loading article

tunetheweb:
Follow on from #8510 

- Changes two to three in iframes article after `auto` was removed in #8312 
- This PR also manages translations and updates dates. Can merge #8510 first then this. Or just this.


malchata (approved):
LGTM.

Issue #8262, opened 2022-07-01
Cookies banner overflows screen max width on mobile


charisTheo:
**Describe the bug**
The cookies banner is overflowing the screen max width on mobile, on [this page](https://web.dev/text-fragments/) but not on [this page](https://web.dev/farewell-html5rocks/).

**To Reproduce**
1. Navigate to https://web.dev/text-fragments/ on mobile
2. Scroll further down the page until the cookie banner appears
3. Meantime the screen is scrollable horizontally
4. Once the cookie banner appears it looks like in the screenshots below.

**Expected behavior**
Cookie banner should be contained within the available screen width like in [this page](https://web.dev/farewell-html5rocks/).

**Screenshots**

<img width="320" alt="screenshot on mobile showing cookie banner overflowing screen width" src="https://user-images.githubusercontent.com/9759549/176904695-cb2507f6-d9dd-4329-bd57-6df82e42bd8e.png">

<img width="320" alt="screenshot on mobile showing the page having a bigger width than the available screen width." src="https://user-images.githubusercontent.com/9759549/176904703-e539bcf0-cec9-4f70-b7e7-bed58bdf2e0e.png">

PR #8534, opened 2022-08-15
Gracefully handle anchor creation for non-latin headlines

### LINKED ISSUE #8528: Table of Content ignores first non-roman H2 title
For non-roman translations, especially when the first title doesn't include any non-roman characters and therefore renders the anchor as `/#`, the table of content ignores the header and doesn't render.

In the below screenshot, "미니앱이 아닌 것들" is an H2 header, but it doesn't contain any roman characters and therefore has an anchor of `/#` and doesn't appear on the table of contents.

Screen Shot 2022-08-15 at 5 46 04 PM

matthiasrohmer:
Fixes #8528. Adds a fallback slug method for headlines with non-latin characters.


devnook:
LGTM, it looks like it might be worthwhile to add a test with some non-latin cases for this plugin though?

src/site/_plugins/markdown.js
```diff
@@ -36,7 +36,20 @@ const markdown = md({
     permalinkClass: 'w-headline-link',
     permalinkSymbol: '#',
     // @ts-ignore
-    slugify: (s) => slugify(s, {lower: true, remove: /[$*_+~.()'"!\-:@]+/g}),
+    slugify: (s) => {
+      let slug = slugify(s, {lower: true, remove: /[$*_+~.()'"!\-:@]+/g});
```
tunetheweb, line 40: @matthiasrohmer is this still needed now #8262 is merged?

If so, and you want to URL encode non-English chars, which I think [is technically the right thing](https://stackoverflow.com/questions/40636281/what-are-the-eligible-characters-in-a-urls-fragment-location-hash), but leads to ugly URLs when they seem to work fine without encoding them, then I think you need to change this as this code no longer works?

tunetheweb (changes requested):

src/site/_plugins/markdown.js
matthiasrohmer: Thanks for flagging here! I'll check and then update accordingly, but probably this is fixed with your change 🙂

tunetheweb: Good stuff. FYI I’ve a enhancement to that in #8549 as missed question marks in my first PR.

matthiasrohmer:
Superseded by #8263 and #8549.

PR #8535, opened 2022-08-15
Migrates intersection observer post from d.c.c.

malchata:

- The [`IntersectionObserver` post](https://developer.chrome.com/blog/intersectionobserver/#how-much-work-can-i-do-in-the-callback) by Surma from d.c.c. should be migrated to web.dev, as the feature now has full support in modern browsers. Edits to this post's copy to reflect the API's support in modern browsers has also been made, including adding a compat component.
- A dependent pull request (to be created) will be added to redirect requests for the d.c.c. version to the web.dev version once this has landed.

### DIFF
--- a/src/site/content/en/blog/intersectionobserver/index.md
+++ b/src/site/content/en/blog/intersectionobserver/index.md
@@ -1,0 +1,143 @@
+---
+layout: post
+title: IntersectionObserver's coming into view
+subhead: IntersectionObservers let you know when an observed element enters or exits the browser's viewport.
+authors:
+  - surma
+description: IntersectionObservers let you know when an observed element enters or exits the browser's viewport.
+date: 2016-04-20 
+updated: 2022-08-15
+tags:
+  - blog
+  - javascript
+  - performance
+---
+
+{% BrowserCompat 'api.IntersectionObserver' %}
+
+Let's say you want to track when an element in your DOM enters the visible [viewport](https://en.wikipedia.org/wiki/Viewport). You might want to do this so you can lazy-load images just in time or because you need to know if the user is actually looking at a certain ad banner. You can do that by hooking up the scroll event or by using a periodic timer and calling [`getBoundingClientRect()`](https://developer.mozilla.org/docs/Web/API/Element/getBoundingClientRect)
+on that element.
+
+This approach, however, is painfully slow as each call to `getBoundingClientRect()` forces the browser to [re-layout the entire page](https://gist.github.com/paulirish/5d52fb081b3570c81e3a) and will introduce considerable jank to your website. Matters get close to impossible when you know your site is being loaded inside an iframe and you want to know when the user can see an element. The Single Origin Model and the browser won't let you access any data from the web page that contains the iframe. This is a common problem for ads for example, that are frequently loaded using iframes.
+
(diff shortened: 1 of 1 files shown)


tunetheweb:
We also have https://web.dev/intersectionobserver-v2/ (which also refers to this post by the old URL so should probably be updated). My understanding is that this is only available in Chromium browsers so probably makes sense to keep both posts, but probably should add a note to the v1 post directing users to this one?

src/site/content/en/blog/intersectionobserver/index.md
```diff
@@ -0,0 +1,143 @@
...
+});
+```
+
+### `<iframe>` magic
+
+`IntersectionObserver`s were designed specifically with ads services and social network widgets in mind, which frequently use `<iframe>` elements and could benefit from knowing whether they are in view. If an `<iframe>` observes one of its elements, both scrolling the `<iframe>` as well as scrolling the window *containing the `<iframe>`* will trigger the callback at the appropriate times. For the latter case, however, `rootBounds` will be set to `null` to avoid leaking data across origins.
+
+## What is `IntersectionObserver` *Not* about?
+
+Something to keep in mind is that `IntersectionObserver` is intentionally neither pixel perfect nor low latency. Using them to implement endeavours like scroll-dependent animations are bound to fail, as the data will be&mdash strictly speaking&mdash;out of date by the time you'll get to use it. The [explainer](https://github.com/WICG/IntersectionObserver/blob/gh-pages/explainer.md) has more details about the original use cases for `IntersectionObserver`.
```
tunetheweb, line 133: Missed a semi-colon. Also when do we have to use `&mdash;` as opposed to `—` out of interest?

```suggestion
Something to keep in mind is that `IntersectionObserver` is intentionally neither pixel perfect nor low latency. Using them to implement endeavours like scroll-dependent animations are bound to fail, as the data will be&mdash;strictly speaking&mdash;out of date by the time you'll get to use it. The [explainer](https://github.com/WICG/IntersectionObserver/blob/gh-pages/explainer.md) has more details about the original use cases for `IntersectionObserver`.
```

malchata:
I can add an aside to the post at the top.

tunetheweb (changes requested):

src/site/content/en/blog/intersectionobserver/index.md
```diff
@@ -0,0 +1,147 @@
...
+updated: 2022-08-15
+tags:
+  - blog
+  - javascript
+  - performance
+---
+
+{% Aside %}
+`IntersectionObserver` V2 has landed in Chromium-based browsers. [Read up on the updates](/intersectionobserver-v2/) for how that API improves on the first version of `IntersectionObserver`.
+{% endAside %}
```
malchata, line 18: This look good, @tunetheweb?

tunetheweb: Can't say I know enough about it, but seems it's not a complete replacement](https://web.dev/intersectionobserver-v2/#:~:text=For%20that%20reason%2C%20Intersection%20Observer%C2%A0v2%20is%20not%20intended%20to%20be%20used%20broadly%20in%20the%20way%20that%20Intersection%20Observer%C2%A0v1%20is.) so yeah I think that's good and generic enough.

tunetheweb (approved):

PR #8536, opened 2022-08-15
Consolidate duplicate capturing images article

matthiasrohmer:
This PR redirects https://web.dev/capturing-images/ to https://web.dev/media-capturing-images/ and removes the duplicate source file. Later one has the better performing URL with 3815 vs 6 views last month.


rachelandrew:
Thanks! I imagine the better performing one is the one we are actually redirecting to from the old site, the other just being a dupe.

devnook (approved):

PR #8545, opened 2022-08-16
Removes scrolling performance post and redirects to the intersection …

malchata:

- Removes the [scrolling performance post](https://web.dev/speed-scrolling/) and redirects it to the migrated [intersection observer post](https://web.dev/intersectionobserver/).


tunetheweb:
These don't seem to be _that_ similar.

The Speed Scrolling one talks about the work the browser has to do on scrolling.
The Intersection Observer ones walks about how to see if you're in the viewport.

Or is it just we feel the Speed Scrolling one is no longer relevant as too old, and this is the nearest thing we have on the topic, even if it's not really an exact match?

malchata:
The issue I take with the original post (from 2012) is that it covers scroll event handlers at all, which are expensive and generally unnecessary now that observers give us a simpler and faster way to detect things like element size and visibility. I could be moved to keep if I could be convinced there are any use cases in which these expensive operations—debounced or not—which can't be solved with a modern approach.

tunetheweb:
To be honest I don't think the articles relevant any more. Not 100% sure it's right to redirect to intersectionobserver article but guess we don't have a better one.

tunetheweb (approved):

PR #8548, opened 2022-08-16
CWV iframes info

### LINKED ISSUE #8547: Make iframe CWV eligibility clearer

Provide a list of URLs.
- https://web.dev/vitals/
- https://web.dev/lcp/
- https://web.dev/cls/
- https://web.dev/fid/
- https://web.dev/inp/

After reading my [CrUX RUM Doc](https://web.dev/crux-and-rum-differences/) Someone raised to me that [they didn't realise](https://webperformance.slack.com/archives/C04BK7K1X/p1660663925699789?thread_ts=1660572444.598339&cid=C04BK7K1X) that content within an iframe would affect the top-level page CWV, rather than the iframe CWV. They "got it" when it was explained to them (and someone else said: "CrUX is trying to capture the experience for end users. End users have no idea if a piece of content in a tab comes from the main frame or an iframe. That's why the metrics are counted against the main frame.")

I had a hunt around our docs and we never made this clear.

We have pieces [like this](https://web.dev/lcp/#differences-between-the-metric-and-the-api):

> The API does not consider elements within iframes, but to properly measure LCP you should consider them. Sub-frames can use the API to report their `largest-contentful-paint` entries to the parent frame for aggregation.

Which kind of implies this, but never explicitly says it.

And we've often said ads, and third-party content can contribute to LCP and CLS issues, but people could have thought this was due to the whole iframe loading late, rather than content within it.

Do you think we should add something to the main web vitals page (https://web.dev/vitals/) to make this clear? And/or each metric' "What elements are considered?" section to make this clear?

### What's the deadline?

No deadline

#### Is it a hard deadline?

Nope

tunetheweb:
Fixes #8547

- Add note that iframes count to CWV docs
- Remove "CLS has changed" warning as over a year old now.
- Update `master` to `main` as web-vitals.js has moved.

### DIFF
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -120,7 +120,7 @@ As far as INP goes, **only the following interaction types are observed:**
 Hovering and scrolling does not factor into INP. However, scrolling with the keyboard (space bar, page up, page down, and so forth) involves a keystroke, which may trigger other events that INP _does_ measure. Any resulting scrolling is _not_ factored into how INP is calculated.
 {% endAside %}
 
-Interactions happen in the main document or in iframes embedded in the document—for example clicking play on an embedded video. End users will not be aware what is in an iframe or not. Therefore, INP within iframes are needed to measure the user experience for the top level page. Note JavaScript Web APIs will not have access to the iframe contents so may not be able to measure FID within an iframe and this will [show as a difference between CrUX and RUM](https://web.dev/crux-and-rum-differences/#iframes).
+Interactions happen in the main document or in iframes embedded in the document—for example clicking play on an embedded video. End users will not be aware what is in an iframe or not. Therefore, INP within iframes are needed to measure the user experience for the top level page. Note JavaScript Web APIs will not have access to the iframe contents so may not be able to measure FID within an iframe and this will [show as a difference between CrUX and RUM](/crux-and-rum-differences/#iframes).
 
 
 Interactions may consist of two parts, each with multiple events. For example, a keystroke consists of the `keydown`, `keypress`, and `keyup` events. Tap interactions contain `pointerup` and `pointerdown` events. The event with the longest duration within the interaction is chosen as the interaction's latency.
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -120,7 +120,7 @@ As far as INP goes, **only the following interaction types are observed:**
 Hovering and scrolling does not factor into INP. However, scrolling with the keyboard (space bar, page up, page down, and so forth) involves a keystroke, which may trigger other events that INP _does_ measure. Any resulting scrolling is _not_ factored into how INP is calculated.
 {% endAside %}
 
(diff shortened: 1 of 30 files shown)


rviscomi:
Looking more closely at the docs, I do see now that each metric's [Differences between the metric and the API](https://web.dev/fid/#differences-between-the-metric-and-the-api) sections include a point about iframes. Not sure we need to explicitly call that out again in the main definition. WDYT?

tunetheweb:
Yeah I did say that in #8547 

> We have pieces [like this](https://web.dev/lcp/#differences-between-the-metric-and-the-api):
> 
>> The API does not consider elements within iframes, but to properly measure LCP you should consider them. Sub-frames can use the API to report their largest-contentful-paint entries to the parent frame for aggregation.
>
> Which kind of implies this, but never explicitly says it.

I don’t think it’s really obvious from that. But think my wording is a bit long winded so happy to take suggestions on how to cut it down a bit.

rviscomi:
Oh ok thanks for pointing that out.

> Which kind of implies this, but never explicitly says it.

So could we rephrase it to be more explicit?

malchata (approved):
This seems okay to me. From an INP perspective, the guidance may get relocated/reworded depending on what additional guidance/restructuring is added to that doc in the future.

tunetheweb:
I've moved the changes down to those sections as requested.

@malchata I did still leave the paragraph in INP as only just saw your message. Let me know if you'd prefer it to be removed from this PR.

malchata:
It's all good. I think this is reasonable guidance to have.

src/site/content/en/metrics/inp/index.md
```diff
@@ -120,6 +120,9 @@ As far as INP goes, **only the following interaction types are observed:**
 Hovering and scrolling does not factor into INP. However, scrolling with the keyboard (space bar, page up, page down, and so forth) involves a keystroke, which may trigger other events that INP _does_ measure. Any resulting scrolling is _not_ factored into how INP is calculated.
 {% endAside %}
 
+Interactions happen in the main document or in iframes embedded in the document—for example clicking play on an embedded video. End users will not be aware what is in an iframe or not. Therefore, INP within iframes are needed to measure the user experience for the top level page. Note JavaScript Web APIs will not have access to the iframe contents so may not be able to measure FID within an iframe and this will [show as a difference between CrUX and RUM](/crux-and-rum-differences/#iframes).
```
malchata, line 123:
```suggestion
Interactions happen in the main document or in iframes embedded in the document—for example clicking play on an embedded video. End users will not be aware what is in an iframe or not. Therefore, INP within iframes are needed to measure the user experience for the top level page. Note JavaScript Web APIs will not have access to the iframe contents so may not be able to measure INP within an iframe and this will [show as a difference between CrUX and RUM](/crux-and-rum-differences/#iframes).
```

malchata: Missed this bit here.

tunetheweb: Yup - reused same text for both 😁

rviscomi (approved):

PR #8560, opened 2022-08-18
Remove HTTP/2 Server Push recommendations

tunetheweb:

- Minor tweaks where we push HTTP/2 push (pardon the pun), but shouldn't since it's going away.


malchata (approved):
LGTM.

PR #8567, opened 2022-08-19
fix performance.measureUserAgentSpecificMemory demo and update article

brendankenny:

- fixes/updates the demo (broken since the API was renamed in Jan 2021)
- small content updates to remove the majority of the references to the original Origin Trial, flags, etc, since the API has been shipping since m89
- demo code changes to hopefully make failure cases clearer. `performance.measureUserAgentSpecificMemory` doesn't exist when not cross-origin isolated, making it easy to confuse with the API being unsupported altogether (the report I got was that all the demo code was broken when copy/pasted into the DevTools console on some non-isolated test page)

### DIFF
--- a/src/site/content/en/blog/monitor-total-page-memory-usage/index.md
+++ b/src/site/content/en/blog/monitor-total-page-memory-usage/index.md
@@ -5,10 +5,11 @@ subhead: >
   Learn how to measure memory usage of your web page in production to detect regressions.
 description: >
   Learn how to measure memory usage of your web page in production to detect regressions.
-updated: 2020-10-19
+updated: 2022-08-19
 date: 2020-04-13
 authors:
   - ulan
+  - bckenny
 hero: image/admin/Ne2U4cUtHG6bJ0YeIkt5.jpg
 alt: >
   Green RAM stick. Photo by Harrison Broadbent on Unsplash.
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/monitor-total-page-memory-usage/index.md
```diff
@@ -97,68 +88,34 @@ are less useful. Example use cases:
...
+
+Support can be detected at runtime:
 
 ```javascript
-if (performance.measureUserAgentSpecificMemory) {
+if (!window.crossOriginIsolated) {
+  console.log('performance.measureUserAgentSpecificMemory() is only available in cross-origin-isolated pages');
+} else if (!performance.measureUserAgentSpecificMemory) {
+  console.log('performance.measureUserAgentSpecificMemory() is not available in this browser');
+} else {
```
brendankenny, line 118: This is admittedly a little awkward, but it's important to be aware of the three possible states, not just if the API is available or not. Happy to adjust further.

malchata (commented):
It scans for me, but I'm not a subject matter expert here. Do you know of anyone else who might be able to review this for technical accuracy?

src/site/content/en/blog/monitor-total-page-memory-usage/index.md
```diff
@@ -255,59 +210,60 @@ async function performMeasurement() {
...
+scheduleMeasurement();
+```
+
 The result may look as follows:
 
 ```javascript
 // Console output:
 {
-  bytes: 60_000_000,
+  bytes: 60_100_000,
```
philipwalton, line 225: Only nit comment is it's not clear why this changed.

philipwalton (approved):
LGTM. I just talked this through with Brendan, and the main SME who could have reviewed is no longer at Google, so I think it's fine to merge as is.

src/site/content/en/blog/monitor-total-page-memory-usage/index.md
brendankenny: it's because there's the new `types: ['DOM']` entry below with 100,000 bytes. I could just make it 10,000,000 bytes of DOM and the total could be an even 70,000,000 if the value is distracting?

philipwalton: Ahh, that's right (you explained this to me before).

I think it's only distracting in the diff; I think it won't be distraction to someone viewing the post as a whole.

PR #8517, opened 2022-08-12
Update importance="high" to fetchpriority="high"

tunetheweb:
Changes proposed in this pull request:

- Changes the sites use of `importance="high"` to `fetchpriority="high"` for banner images and the like, since this was renamed and Chrome no longer supports `importance`
- Update one doc (and it's translation) for similar reasons.

PR #8576, opened 2022-08-23
Update webdev-infra to enable fetchpriority use

philipwalton:
This PR continues the work done in https://github.com/GoogleChrome/web.dev/pull/8517 and updates the [webdev-infra](https://github.com/GoogleChrome/webdev-infra) package, so `fetchpriority` works with the image shortcode.

### DIFF
--- a/src/site/_data/site.js
+++ b/src/site/_data/site.js
@@ -53,7 +53,7 @@ module.exports = {
       NAVIGATION_TYPE: 'dimension6',
       COLOR_SCHEME_PREFERENCE: 'dimension7',
     },
-    version: 10,
+    version: 11,
   },
   firebase: {
     prod: {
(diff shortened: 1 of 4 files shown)


src/site/_data/site.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
       NAVIGATION_TYPE: 'dimension6',
       COLOR_SCHEME_PREFERENCE: 'dimension7',
     },
-    version: 10,
+    version: 11,
```
tunetheweb, line 56: What's this for, out of interest?

tunetheweb (changes requested):
Need to update the lock file took.

src/site/_data/site.js
philipwalton: Bumping the analytics version is useful in case you want to compare performance before and after. This corresponds to a custom dimension (`dimenson5`) sent on all analytics beacons.

philipwalton:
> Need to update the lock file took.

Thanks, done.

tunetheweb (approved):
LGTM - assuming tests pass

PR #8588, opened 2022-08-24
Do not add an id to h1 in posts

### LINKED ISSUE #8574: Headers have the same id
Hello,
I was wanting to share a link [Choose the Right Image Format](https://web.dev/choose-the-right-image-format/#choose-the-right-image-format) with a new colleague and when I clicked the link it brought me to the top of the page instead of the 1st sub header within the article.

**To Reproduce**
1. Go to [Choose the Right Image Format](https://web.dev/choose-the-right-image-format/#choose-the-right-image-format)
2. Scroll down to the header text right above the picture, and click the hashtag tag that appears to link to that section.

**Expected behavior**
Expected the page to display right at that header, but it went to the top of the page. The article and this header have the same text, which made them have the same ID on the elements.

**Screenshots**
web dev-header-id

https://github.com/GoogleChrome/web.dev/blob/1c7e48e7e9ea20e8e41f4123b6e34473a8480333/src/site/content/en/fast/choose-the-right-image-format/index.md?plain=1#L3

https://github.com/GoogleChrome/web.dev/blob/1c7e48e7e9ea20e8e41f4123b6e34473a8480333/src/site/content/en/fast/choose-the-right-image-format/index.md?plain=1#L43
-------
I wasn't sure what it should be changed to, but figured I would let ya'll know.

Great article nonetheless! Thanks for all your hard work!

matthiasrohmer:
Resolves #8574. This PR stops adding the slugified title as an id to the `h1` of our posts. As the headline is not linkable from the UI (other headlines show a # you can use to copy a deep link) this should not affect anyone.

### DIFF
--- a/src/site/_includes/partials/post-next.njk
+++ b/src/site/_includes/partials/post-next.njk
@@ -76,7 +76,7 @@
           </nav>
         {% endif %}
 
-        <h1 id="{{ safeTitle | slugify }}">{{ safeTitle }}</h1>
+        <h1>{{ safeTitle }}</h1>
         {% if subhead | length %}
           <p class="color-mid-text flow-space-base">
             {{ subhead | md | safe }}


Yohn:
I like it, that's a nice and clean way to fix it too.

devnook (approved):

PR #8595, opened 2022-08-25
Add note about CSP and unsafe-inline

larowlan:

- Clarify that the `onload` technique is not compatible with CSP


malchata (approved):
LGTM. @larowlan, please sign the CLA so this can be merged.

PR #8596, opened 2022-08-25
fix: fix typo what are pwas in japanese

takashimamorino:


matthiasrohmer (approved):
Verified with Google Translate.

PR #8597, opened 2022-08-25
Add CompressionStream article

tomayac:
Fixes `N/A` (submitted content proposal, but no Issue was created.)

- Add the article ([staging preview](https://deploy-preview-8597--web-dev-staging.netlify.app/compressionstream/)).

### DIFF
--- a/src/site/content/en/blog/compressionstream/index.md
+++ b/src/site/content/en/blog/compressionstream/index.md
@@ -8,7 +8,6 @@ alt: Clamp.
 author: thomassteiner
 tags:
   - blog
-  - newly-interoperable
 ---
 
 {% Aside 'celebration' %} This web feature is now available in two major browser engines!
--- a/src/site/content/en/blog/compressionstream/index.md
+++ b/src/site/content/en/blog/compressionstream/index.md
@@ -1,17 +1,14 @@
 ---
-title: CompressionStream is now supported in Safari and Chrome
+title: Compression and decompression in the browser with the Compression Streams API
 subhead: >
-  Now that compression streams are supported in Chrome and Safari, they're ready for prime time!
-date: 2022-08-25
+  Write smaller web apps that don't need to ship their owm compression or decompression library
+date: 2022-08-26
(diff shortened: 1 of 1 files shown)


rachelandrew:
Not sure why you didn't get an issue, however this one isn't supported in all three engines so can't be included as an interop moment. This is one we looked at in the meeting and realised there was no Firefox support. Do you know if Firefox are currently implementing?

src/site/content/en/blog/compressionstream/index.md
```diff
@@ -0,0 +1,41 @@
...
+title: CompressionStream is now supported in Safari and Chrome
+subhead: >
+  Now that compression streams are supported in Chrome and Safari, they're ready for prime time!
+date: 2022-08-25
+hero: image/8WbTDNrhLsU0El80frMBGE4eMCD3/koW7cNlWa6BUlJsQR4wI.jpg
+alt: Clamp.
+author: thomassteiner
+tags:
+  - blog
+  - newly-interoperable
```
rachelandrew, line 11: This is not interoperable

rachelandrew (changes requested):
To be included this needs to have landed support in all three engines. If that is coming then add a comment as to when you think Firefox is landing support and we'll hold it till then.

rachelandrew:
and it looks as if the script failed as I have the info in the table but it didn't get as far as creating an issue. If that happens again let us know, as the reason to have a content proposal form is to discuss proposals before making a PR! At which point I'd have asked about the actual interop status of this.

src/site/content/en/blog/compressionstream/index.md
tomayac, line 11:
```suggestion
```

tomayac:
Can it be moved to dcc instead? The only place we mention `CompressionStream` is https://developer.chrome.com/articles/fetch-streaming-requests/, so it might make sense to at least have a small article like this. What do you think?

tomayac:
Closing in favor of https://github.com/GoogleChrome/developer.chrome.com/pull/3557.

PR #8599, opened 2022-08-25
Upgrade web-vitals to 3.0.0 and add INP monitoring

tunetheweb:

- Upgrade to the new 3.0.0 web-vitals library
- Switch to the preferred `onXXX` syntax
- Add onINP monitoring

@philipwalton should we add attribution tracking? Would be good to have a reference implementation to points to. Just need a Custom Dimension set up in GA.

### DIFF
--- a/src/lib/analytics.js
+++ b/src/lib/analytics.js
@@ -1,4 +1,4 @@
-import {onCLS, onFCP, onFID, onLCP, onTTFB, onINP} from 'web-vitals';
+import {onCLS, onFCP, onFID, onLCP, onTTFB, onINP} from 'web-vitals/attribution';
 import {dimensions} from 'webdev_analytics';
 import {store} from './store';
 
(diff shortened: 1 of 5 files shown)


philipwalton:
> @philipwalton should we add attribution tracking? Would be good to have a reference implementation to points to. Just need a Custom Dimension set up in GA.

Yeah, I think we should. We can do more comprehensive attribution tracking when we switch to GA4, but for now I think something basic like this would work:

```js
function sendToGoogleAnalytics(metric) {
  // ...
  
  let dimension1;

  switch (metric.name) {
    case 'CLS':
      dimension1 = metric.attribution.largestShiftTarget;
      break;
    case 'LCP':
      dimension1 = metric.attribution.element;
      break;
    case 'FID':
    case 'INP':
      dimension1 = metric.attribution.eventTarget;
      break;
  }

  ga('send', 'event', {
    dimension1,
    // ...
  });
}
```

WDYT?

tunetheweb:
Yup exactly. That’s what we do on HTTPArchive.org (just upgraded to 3.0.0 today?) and almanac.httparchive.org (upgraded and will be in next release).

tunetheweb:
Done! PTAL when you get a chance.

src/lib/analytics.js
```diff
@@ -58,7 +65,27 @@ export function trackError(error, message = '', fatal = false) {
...
+      break;
+    case 'FID':
+    case 'INP':
+      webVitalInfo = attribution.eventTarget;
+      break;
+    case 'LCP':
+      webVitalInfo = attribution.element;
+      break;
+    case 'TTFB':
+      webVitalInfo = attribution.connectionTime;
```
philipwalton, line 83: I'm not sure we need to track debug info for TTFB and FCP, but if we are going to—and since this is being set on a custom dimension param—we should pick a dimension (string value) rather than a metric (numeric value).

We *could* create custom metrics for `connectionTime`, `firstByteToFCP`, etc., but unless we have a plan to build reports with that data, I think we should hold off to keep things simple (we can always add it later).

(BTW, when we update to GA4, I definitely think we should include all of this info.)

tunetheweb: And they're gone!

philipwalton (approved):
LGTM with one last comment (unrelated to your changes):

Can you change [this line](https://github.com/GoogleChrome/web.dev/blob/12cd082eaee91e397607a79199257c8825d3bb49/src/lib/analytics.js#L154) from `bfcache` to `back-forward-cache` for consistency with what the web-vitals library will report?

PR #8607, opened 2022-08-27
Normalizes core web vitals terminology on the learn web vitals landing page.

malchata:

- There has been a desire to correct the "web vitals" term to "core web vitals" to improve consistency in how the term is used. This pull request changes all the language on the existing [learn web vitals](https://web.dev/learn-web-vitals/) page to "Core Web Vitals" from "Web Vitals".
- Redirects /learn-web-vitals/ to /learn-core-web-vitals/ and corrects all existing links to point to that page.


tunetheweb (commented):
The change itself looks fine.

However, we do still have this https://web.dev/vitals/#other-web-vitals suggesting we, at least in the past, considered the non-CWV metrics as "web vitals". So is the plan to clean that up?

Also, I'm a bit worried that the more commonly referenced `/vitals/` URL assuming you do similar in a later PR? Though I guess in theory with a redirect, it should all be fine.

WDYT @philipwalton ?

malchata:
Yeah, the plan is to do a broader pass to normalize terminology and start calling anything outside of the three Core Web Vitals "metrics". Phil and I have been discussing this for a while, and I think the approach will be incremental, but I would welcome what Phil thinks of this change.

philipwalton (approved):

Inikowu:
> 

image/j2RDdG43oidUy6AL6LovThjeX9c2/NcqmhRbQCFP5twoNcOHg.svg
---

PR #8608, opened 2022-08-27
Removes and redirects old appcache post to service worker lifecycle guide.

malchata:

- With AppCache deprecated and removed in most browsers, any content on AppCache should redirect to service worker documentation. This removes the [/workingoffthegrid/](https://web.dev/workingoffthegrid/) post migrated from HTML5 Rocks and redirects it to the [service worker overview](https://developer.chrome.com/docs/workbox/service-worker-overview/) doc on d.c.c.


rachelandrew (approved):

redirects.yaml
```diff
@@ -326,6 +326,8 @@ redirects:
     destination: https://developer.chrome.com/docs/web-platform/handwriting-recognition/
   - source: /virtualkeyboard
     destination: https://developer.chrome.com/docs/web-platform/virtual-keyboard/
+  - source: /workingoffthegrid
+  - destination: https://developer.chrome.com/docs/workbox/service-worker-overview/
```
tunetheweb, line 330: I think this is a bug, and dash shouldn't be there @malchata . Will open a follow up PR.

PR #8610, opened 2022-08-28
Adds core web vitals guidance to the remove unused code guide.

malchata:

- Updates the [remove unused code post](https://web.dev/remove-unused-code/) intro with a note about how unused code can affect certain Core Web Vitals and other metrics.

### DIFF
--- a/src/site/content/en/fast/remove-unused-code/index.md
+++ b/src/site/content/en/fast/remove-unused-code/index.md
@@ -25,9 +25,11 @@ download and use over _half a million_ public packages. But we often include
 libraries we're not fully utilizing. To fix this issue, **analyze your bundle**
 to detect unused code. Then remove **unused** and **unnecessary** libraries.
 
-By removing unused code, you can improve your website's [Core Web Vitals](/vitals/). [Largest Contentful Paint](/lcp/), for example, can be affected by unused code by increased bandwidth contention caused by large assets for [LCP candidates](/lcp/#what-elements-are-considered). LCP can also be affected if large JavaScript assets that render markup solely on the client [contain references to LCP candidates](/preload-scanner/#rendering-markup-with-client-side-javascript) by [delaying when these resources can load](/optimize-lcp/#1-eliminate-resource-load-delay).
+## Impact on Core Web Vitals
 
-Other metrics such as [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/) can also be affected by unused code during startup, because even unused JavaScript must be downloaded, parsed, compiled, and then executed. The unused code can introduce unnecessary delays in resource load time and main thread activity that contribute to poor page responsiveness.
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/remove-unused-code/index.md
```diff
@@ -24,6 +25,12 @@ download and use over _half a million_ public packages. But we often include
 libraries we're not fully utilizing. To fix this issue, **analyze your bundle**
 to detect unused code. Then remove **unused** and **unnecessary** libraries.
 
+By removing unused code, you can improve your website's [Core Web Vitals](/vitals/). [Largest Contentful Paint](/lcp/), for example, can be affected by unused code by increased bandwidth contention caused by large assets for [LCP candidates](/lcp/#what-elements-are-considered). LCP can also be affected if large JavaScript assets that render markup solely on the client [contain references to LCP candidates](/preload-scanner/#rendering-markup-with-client-side-javascript) by [delaying when these resources can load](/optimize-lcp/#1-eliminate-resource-load-delay).
```
tunetheweb, line 28: Suggest adding a heading.

Also even if the LCP isn't in the large bundle, it can still be slowed by contention so tweaked that slightly.

```suggestion
## Impact on Core Web Vitals

By removing unused code, you can improve your website's [Core Web Vitals](/vitals/). [Largest Contentful Paint](/lcp/), for example, can be affected by unused code by increased bandwidth contention caused by larger than necessary assets. LCP can also be affected if large JavaScript assets that render markup solely on the client [contain references to LCP candidates](/preload-scanner/#rendering-markup-with-client-side-javascript) by [delaying when these resources can load](/optimize-lcp/#1-eliminate-resource-load-delay).
```

src/site/content/en/fast/remove-unused-code/index.md
```diff
@@ -24,6 +25,12 @@ download and use over _half a million_ public packages. But we often include
...
+
+Other metrics such as [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/) can also be affected by unused code during startup, because even unused JavaScript must be downloaded, parsed, compiled, and then executed. The unused code can introduce unnecessary delays in resource load time and main thread activity that contribute to poor page responsiveness.
```
tunetheweb, line 30:
```suggestion
Other metrics such as [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/) can also be affected by unused code, because even unused JavaScript must be downloaded, parsed, compiled, and then executed. The unused code can introduce unnecessary delays in resource load time, memory usage, and main thread activity that contribute to poor page responsiveness.
```

Out of interest how would FID and INP be impacted by load time? E.g. if an element is rendered, but no click handler is registered, and the user clicks, and THEN the JavaScript adds event handlers is that earlier event eligible for FID and/or INP? I would have thought not as no Event Handler at time of click - a gap in the metrics currently. But not 100% sure on that to be honest. An either way it's affecting the user experience, even if CWV doesn't capture it currently.

tunetheweb (changes requested):

src/site/content/en/fast/remove-unused-code/index.md
malchata: FID is a technically a load responsiveness metric, and INP can be affected if interactions are made while the page is starting up, as is my understanding. INP is calculated even for some elements without event handlers (form controls, for example), so this is a possibility.

tunetheweb (approved):

PR #8612, opened 2022-08-28
Fix description of checks for "bypass" Lighthouse audit based on Axe

thibaudcolas:
This audit is based on Axe’s [bypass](https://dequeuniversity.com/rules/axe/3.3/bypass) rule, which checks headings, not header elements. The error might stem from Axe’s check for this being called [header-present](https://github.com/dequelabs/axe-core/blob/develop/lib/checks/navigation/header-present.json), but it does check h1…h6 elements, not `<header>`.

I looked for similar copy elsewhere and this seems to be the only place with the issue.

### DIFF
--- a/src/site/content/en/lighthouse-accessibility/bypass/index.md
+++ b/src/site/content/en/lighthouse-accessibility/bypass/index.md
@@ -27,7 +27,7 @@ Lighthouse flags pages that don't provide a way to skip repetitive content:
 </figure>
 
 Lighthouse checks that the page contains at least one of the following:
-- A `<header>` element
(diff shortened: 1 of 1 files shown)


heyawhite (approved):

AaronForinton:
I'm closing this because this content was migrated to [https://developer.chrome.com](https://developer.chrome.com) and then substantially refactored.

thibaudcolas:
Ah, that’s good to know! @AaronForinton would you happen to have a link to the new content? I couldn’t find it in the site search, I’d be keen to see how this is now described.

PR #8634, opened 2022-09-03
Make text more exact

simevidas:
I think it’s a good idea to always say “the three *major* browser engines” because there are other, smaller browser engines.


malchata (approved):
LGTM.

PR #8648, opened 2022-09-07
Fixed: loop dropping out at first non file entry

abitofcode:
Fixes an issue in the file system access blog post that results in the exiting the loop of entries early.

https://web.dev/file-system-access/#opening-a-directory-and-enumerating-its-contents


tomayac (approved):
LGTM, approval

Thank you, the other `break` of course… 🤦‍♂️

PR #7404, opened 2022-02-24
Removes and redirects Lighthouse accessibility docs to axe docs.

malchata:
This PR is part of the effort around resolving this issue: GoogleChrome/developer.chrome.com#1997

Changes proposed in this pull request:

- Deletes Lighthouse accessibility audit docs in all languages (minus the accessibility scoring doc in English)
- Redirects deleted articles to the appropriate axe documentation.

Note: `DO NOT MERGE` label will be dropped once it's safe to deploy these changes.

PR #8654, opened 2022-09-08
redirect Lighthouse accessibility docs to axe docs

brendankenny:
#7404 had diverged enough that rebasing on `main` was a lot of work, so this takes over from there.

There are still a few files in `src/site/content/en/lighthouse-accessibility/` that turn out to need to be migrated to d.c.c, which I'll do in a follow up PR.


malchata (approved):
LGTM. I didn't click through every link, obviously, but I assume these destinations are appropriate.

PR #8658, opened 2022-09-10
Fix Chinese translation

xfq:
邮件列表 is the common translation of the term "mailing list".

Ref:

* https://zh.wikipedia.org/wiki/%E9%82%AE%E4%BB%B6%E5%88%97%E8%A1%A8
* https://www.zhihu.com/question/23757640


matthiasrohmer (approved):
Verified via Google Translate and Deepl.

PR #8669, opened 2022-09-13
Fix invalid/broken container query syntax

ChezFre:
The shorthand `container` property requires the `container-name` to be specified. As it's missing in the container query it uses `inline-size` as the name of the container and the default value of `normal` for the `container-type`.

The article could be slightly expanded to point out that the `@container` rule can also take the name of a container.

The same fix is required in the linked codepen at the bottom https://codepen.io/una/pen/xxLPwBX


una:
Updated the Codepen, cheers

matthiasrohmer (approved):
Thanks for pointing this out and fixing it!

PR #8675, opened 2022-09-13
Clarify code for checking if the navigation menu is currently open

AleksandrHovhannisyan:
I was reading the latest article on [building the main navigation for a website](https://web.dev/website-navigation/) and found this bit of code confusing the way it was written since it suggests that `isOpen` is true when `aria-expanded="false"` (but `aria-expanded="false"` means the menu `isClosed`). Imo, it makes more sense to check `isOpen` directly but negate the condition when updating the `aria-expanded` attribute.

(Note: I'm not sure how to update the associated CodePens, or if that's even necessary since the code is functionally the same.)

### DIFF
--- a/src/site/content/en/blog/website-navigation/index.md
+++ b/src/site/content/en/blog/website-navigation/index.md
@@ -370,12 +370,9 @@ const button = burgerClone.querySelector('button');
 
 // Toggle aria-expanded attribute
 button.addEventListener('click', e => {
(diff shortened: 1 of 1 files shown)


AleksandrHovhannisyan:
Not sure if there's anything for me to do on this PR. I know checks are failing, but I just want to clarify that's not because of changes I made.

rachelandrew (approved):

PR #8689, opened 2022-09-19
Fix missing link and word

xiaohai-huang:
1. The blog [Constructing the Object Model](https://web.dev/critical-rendering-path-constructing-the-object-model/) contains a broken link __render tree__. I think that link should point to this blog [Render-tree Construction, Layout, and Paint](https://web.dev/critical-rendering-path-render-tree-construction/). image
2. The first sentence should be "The final output is a render **tree** that contains..."
   > The final output is a render that contains both the content and style information of all the visible content on the screen. With the render tree in place, we can proceed to the "layout" stage.


matthiasrohmer (approved):
Oh wow, thanks for fixing both this and the missing word!

PR #8695, opened 2022-09-20
[lighthouse] complete accessibility content redirects to d.c.c

brendankenny:
Closes #7404

follow up to #8654
requires https://github.com/GoogleChrome/developer.chrome.com/pull/3768 to deploy before redirects will work

- adds remaining redirects to accessibility content (soon to be) migrated to d.c.c
- deletes web.dev lighthouse-accessibility infrastructure

Everything appears to be fine with the deleted files, but the lighthouse content has existed for so long and pokes into enough places that something may end up breaking. Let's find out!


redirects.yaml
```diff
@@ -395,84 +395,116 @@ redirects:
...
+  # Redirects for Lighthouse accessibility content.
+  - source: /lighthouse-accessibility
+    destination: https://developer.chrome.com/docs/lighthouse/accessibility/
+  - source: /accessibility-scoring
+    destination: https://developer.chrome.com/docs/lighthouse/accessibility/scoring/
+  # Redirects accessibility audits, now deferring to axe docs.
+  - source: /accesskeys
     destination: https://dequeuniversity.com/rules/axe/4.4/accesskeys
-  - source: /bypass/
+  - source: /bypass
```
brendankenny, line 406: I realized we are 404ing for anyone visiting without a trailing slash (e.g. https://web.dev/bypass currently 404s, though https://web.dev/bypass/ correctly redirects). Dropping the trailing slash handles both cases.

redirects.yaml
```diff
@@ -395,84 +395,116 @@ redirects:
...
     destination: https://dequeuniversity.com/rules/axe/4.4/html-lang-valid
-  - source: /valid-lang/
+  - source: /valid-lang
     destination: https://dequeuniversity.com/rules/axe/4.4/valid-lang
+  - source: /aria-name
+    destination: https://dequeuniversity.com/rules/axe/4.4
+  - source: /duplicate-id
+    destination: https://dequeuniversity.com/rules/axe/4.4/duplicate-id-active
+  - source: /heading-levels
+    destination: https://dequeuniversity.com/rules/axe/4.4/heading-order
```
brendankenny, line 489: technically these three don't have exact axe equivalents:
- `aria-name` was a catchall for a bunch of `aria-*-name` audits (e.g. [`aria-meter-name`](https://dequeuniversity.com/rules/axe/4.4/aria-meter-name)). A link to the overall rules is the best we can do, I think
- `duplicate-id` was removed by axe and replaced with `duplicate-id-active` and (and somewhat less directly, `duplicate-id-aria`). The `duplicate-id` page was kept around to keep old links alive, but we've had 37 visitors to it in the last year, so seems ok to redirect it to the more modern version of the audit
- `heading-levels` was an old manual Lighthouse audit that eventually got a real `heading-order` audit from axe. It similarly had 11 visitors in the last year, so seems ok to redirect to the modern version

src/site/content/en/learn/learn.11tydata.js
```diff
@@ -37,7 +37,6 @@ module.exports = function () {
     allPaths['lighthouse-performance'],
     allPaths['lighthouse-pwa'],
     allPaths['lighthouse-best-practices'],
-    allPaths['lighthouse-accessibility'],
```
brendankenny, line 40: this removes `lighthouse-accessibility` from the https://web.dev/learn/ Lighthouse collection. I think this is the right thing to do? I assume when I complete the migration we just won't have a Lighthouse collection under learn/ at all

cc @rachelandrew if that's something you've thought about

src/site/content/es/accessible/headings-and-landmarks/index.md
```diff
@@ -7,7 +7,7 @@ date: 2018-11-18
 description: Al utilizar los elementos correctos para los títulos y los landmark, puedes mejorar la experiencia de navegación para los usuarios de tecnología de asistencia
 ---
 
-{% include 'content/lighthouse-accessibility/why-headings.njk' %}
+{% include 'content/why-headings.njk' %}
```
brendankenny, line 10: this was the only `lighthouse-accessibility` _include left, and it's not used in lighthouse content anymore, so moved it up a directory

malchata (approved):
At a high level, this seems fine to me. Good catch on the trailing slashes!

Issue #8697, opened 2022-09-20
Corrections to Creative List Styling article


mbarker84:
* Use this template to request updates to existing content.
* Delete any fields that aren't relevant to your issue. 
* Update the title of this issue with a brief summary of the work to be done.

### What page(s) need to be updated?

https://web.dev/creative-list-styling/

### Why is this update needed?

Two of the list demos do not work as described in Chrome and Safari.

- Replace demo https://codepen.io/web-dot-dev/pen/QWmBrGp with https://codepen.io/michellebarker/pen/KKRqWNb/48a9ddf770277cf0cf7e1a7debffac6f.
- Replace demo https://codepen.io/web-dot-dev/pen/YzajLpm with https://codepen.io/michellebarker/pen/ExLXWZj/b6bb07c800ce71fb37c2b2904a5be1a8

heyawhite:
@mbarker84 I copied your updates into our pens.

PR #8698, opened 2022-09-20
Adding note about `start` attribute

mbarker84:
Addresses #8697. (Codepen demos will also need to be updated.)

- Add line detailing use of `start` attribute in order for list markers to count down correctly across all browsers when using `::before`.


rachelandrew (approved):
thank you! I'll update the demos.

PR #8736, opened 2022-09-27
Remove opinionated take w.r.t. optional end tags

mathiasbynens:
The main point of this patch is to avoid saying "it isn't good practice" about something that in fact is a good practice per Google's HTML style guide.

### DIFF
--- a/src/site/content/en/learn/html/overview/index.md
+++ b/src/site/content/en/learn/html/overview/index.md
@@ -46,7 +46,7 @@ HTML is very, very forgiving. For example, if we omit the closing `</li>` tags, 
 </ul>  
 ```
 
-It is valid to not close an `<li>`, and some style guides (such as [Google’s](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags)) even recommend doing so. The closing `</ul>` is mandatory. If the unordered list's closing tag is omitted, the browser will try to determine where your list and list items end, but you might not agree with the decision.
+It is valid to not close an `<li>`, but the closing `</ul>` is mandatory. If the unordered list's closing tag is omitted, the browser will try to determine where your list and list items end, but you might not agree with the decision.
 
 The specification for each element lists whether the closing tag is mandatory or not. "Neither tag is omissible" in the specification means both an opening tag and a closing tag are required. The [specification provides a list of all the required closing tags](https://html.spec.whatwg.org/multipage/syntax.html#syntax-tag-omission). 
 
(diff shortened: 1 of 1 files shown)


rachelandrew:
I think for most developers leaving optional tags unclosed causes more problems than it solves. We're not writing this course for people working on Google products, so the style guide isn't relevant here, and would be confusing to add as it is in itself an opinionated take.

mathiasbynens:
The current statement (“it isn't good practice”) is an opinionated take. If the goal is to be unopinionated, we could say something like

> It is valid to not close an `<li>`, but the closing `</ul>` is mandatory.

src/site/content/en/learn/html/overview/index.md
```diff
@@ -46,7 +46,7 @@ HTML is very, very forgiving. For example, if we omit the closing `</li>` tags,
 </ul>  
 ```
 
-Although it is valid to not close an `<li>`, it isn't good practice.  The closing `</ul>` is mandatory. If the unordered list's closing tag is omitted, the browser will try to determine where your list and list items end, but you might not agree with the decision.
+It is valid to not close an `<li>`, and some style guides (such as [Google’s](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags)) even recommend doing so. The closing `</ul>` is mandatory. If the unordered list's closing tag is omitted, the browser will try to determine where your list and list items end, but you might not agree with the decision.
```
mathiasbynens, line 49:
```suggestion
It is valid to not close an `<li>`, but the closing `</ul>` is mandatory. If the unordered list's closing tag is omitted, the browser will try to determine where your list and list items end, but you might not agree with the decision.
```

mathiasbynens:
> The current statement (“it isn't good practice”) is an opinionated take. If the goal is to be unopinionated, we could say something like
> 
> > It is valid to not close an `<li>`, but the closing `</ul>` is mandatory.

Done + PR title changed.

rachelandrew:
The goal isn't to be unopinionated, it's just it has a different opinion to that in the style guide you linked to. Having helped people fix issues over the years leading from badly nested markup (that usually shows up as a CSS problem) I think in general this is a good suggestion to follow as it's much clearer to see what's going on if everything is closed.

malchata:
I agree with the assertion @rachelandrew is making here. We should be encouraging well-formed HTML. I can't think of a good reason to not close list item elements. I think @estelle is making a good opinionated stance here.

rachelandrew:
I think we're happy to leave this as is.

PR #8741, opened 2022-09-28
Fix broken hyperlink to MDN

ericlaw1979:

Replace broken
  https://developer.mozilla.org/en-US/docs/docs/Learn/CSS/Building_blocks/Cascade_layers
with working
  https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_layers


malchata (approved):
LGTM.

PR #8782, opened 2022-10-03
Minor edits of Priority Hints article

tunetheweb:
Some minor edits to the [Priority Hints](https://web.dev/priority-hints/) article.

- Standardize capitalisation as "Priority Hints" (currently has a mixture of "Priority Hints", "Priority hints" and "priority hints"
- Fix alt text to not use quotes as breaks HTML
- Replace`<td>&nbsp;</td>` with `<td></td>` in tables (don't see need for `&nbsp;`).
- Fix table foot notes (See screenshot below)
- Add extra explanation that it raises/lowers priority rather than explicitly setting them to "High" or "Low"
- Switch to em-dashes
- Fix some code-markup errors
- Add extra explanation as to why still useful for browsers even if CDN doesn't support HTTP/2 and HTTP/3 prioritization.
- Standardize on "High"/"Low" (with quotes) when talking about priority it gets and `high`/`low` (with code formatting) when talking about the `fetchpriority` value.
- Add extra explanation on "Experimental"

Old footnotes:

image

New footnotes:

image

FYI: @pmeenan @addyosmani

### DIFF
--- a/src/site/content/en/fast/priority-hints/index.md
+++ b/src/site/content/en/fast/priority-hints/index.md
@@ -562,10 +562,12 @@ With the fix in Chrome 95 and the enhancement for Priority Hints, we hope that d
 - Image preloads will have a "Low" priority (without Priority Hints) and should be ordered relative to async scripts and other low or lowest priority tags.
 
 ## History
+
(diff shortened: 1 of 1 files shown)


malchata (approved):
LGTM.

PR #8784, opened 2022-10-03
ServiceWorkerRegistration.active and friends return null

jespertheend:

The `installing`, `waiting` and `active` getters on a ServiceWorkerRegistration return `null` now. I reckon this is outdated.

- [`ServiceWorkerRegistration.active` on mdn](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/active)
- [ServiceWorkerRegistration interface spec](https://w3c.github.io/ServiceWorker/#serviceworkerregistration-interface)

image

### DIFF
--- a/src/site/content/en/reliable/service-worker-lifecycle/index.md
+++ b/src/site/content/en/reliable/service-worker-lifecycle/index.md
@@ -342,9 +342,9 @@ So, to enable as many patterns as we can, the whole update cycle is observable:
 
 ```js
 navigator.serviceWorker.register('/sw.js').then(reg => {
-  reg.installing; // the installing worker, or undefined
-  reg.waiting; // the waiting worker, or undefined
-  reg.active; // the active worker, or undefined
+  reg.installing; // the installing worker, or null
+  reg.waiting; // the waiting worker, or null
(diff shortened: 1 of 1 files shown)


malchata:
Once a service worker is installed, those values should be null, because the service worker has been activated after the next page refresh. Try the following in the console to check if a service worker is activated post-install:

```js
const reg = await navigator.serviceWorker.getRegistration("/js/sw.js");
console.log(reg.active.state);
```

Also beware that if a service worker is not located at the root, its scope will be restricted to whatever folder it's in. You can specify a `Service-Worker-Allowed` header to get around this.

cc: @jakearchibald

malchata:
Ah, I see what you were referring to here.

I think this change makes sense, but didn't fully understand where you were coming from at first. Still, it'd be a good idea if @jakearchibald took a look at this, since this is a guide he wrote.

jespertheend:
Ah sorry my bad, I should have been more clear there.

So just to be clear: I think an older version of the spec said that these values would be `undefined` when non existent, but I think this changed at some point, because I'm observing that they are `null` now and the spec and mdn also state that they are initially `null`.

PR #8796, opened 2022-10-04
aria-label text change starts with same text that's visually present - Fixes #8795

### LINKED ISSUE #8795: content: ARIA in HTML aria-label Do example seems to be problematic
`<a aria-label="Some awesome article title">Read More</a>` seems to be problematic for some users and we should recommend better practice (if not directly an WCAG 2.5.3 Label in Name fail).

I would suggest starting with "Read more" also in aria-label, for example: `<a aria-label="Read more about some awesome article title">Read More</a>`.

References;

- [Knowbility WCAG 2.5.3](https://knowbility.org/blog/2018/WCAG21-253LabelInName/#example)
- [WCAG, Understanding Success Criterion 2.5.3: Label in Name](https://www.w3.org/WAI/WCAG22/Understanding/label-in-name.html#examples)
- [Sara Soueidan, Accessible Text Labels For All](https://www.sarasoueidan.com/blog/accessible-text-labels/#improving-the-experience-for-screen-reader-users-whilst-keeping-it-accessible-to-speech-input-users)

https://web.dev/learn/accessibility/aria-html/

`<a aria-label="Some awesome article title">Read More</a>` seems to be problematic for some users (for example screen-reader users that can see the screen and voice control users).

### What's the deadline?

As soon as possible I guess.

BogdanCerovac:
My suggestion for aria-label that would work better than current, Fixes #8795

Fixes #8795

- just a minor aria-label text change that would in my opinion work better for some users

### DIFF
--- a/src/site/content/en/learn/accessibility/aria-html/index.md
+++ b/src/site/content/en/learn/accessibility/aria-html/index.md
@@ -293,7 +293,7 @@ Assigned the wrong role.
 
 {% Compare 'better' %}
 ```html
-<a aria-label="Some awesome article title">Read More</a>
+<a aria-label="Read more about some awesome article title">Read More</a>
 ```
 
 {% CompareCaption %}


rachelandrew (approved):
Thanks!

PR #8797, opened 2022-10-04
Misc edits to font best practices document

tunetheweb:

- Restructures headings to allow the topics to show in the index at the side of the page
- Makes it clear inlining font-face will not cause earlier load unless all CSS is also inlined (this was raised on WebPerf slack and was original intent of this PR but then I did a load of other edits while in here)
- Adds comment on inlining CSS being a more advanced technique.
- Tones down anti-preload advice. It's still the best way to avoid font loading issues IMHO. We use it on web.dev for example!
- Expands section on font sub-setting
- Notes we're ready to move to a WOFF2 only world
- Adds comment on `font-display: block` as many (myself included!) hate the jarring shifts caused by `font-display: swap`, even if they help with LCP/CLS
- Adds short not on `size-adjust`.
- Other misc formats and edits

Now people (especially me!) get very opinionated about font-loading (I forever quote [this tweet](https://twitter.com/jaffathecake/status/1293250032632844294)) but tried to be objective and balanced here too. Let me know if you disagree with any of the changes!

### DIFF
--- a/src/site/content/en/fast/font-best-practices/index.md
+++ b/src/site/content/en/fast/font-best-practices/index.md
@@ -1,8 +1,7 @@
 ---
 layout: post
 title: Best practices for fonts
-subhead: |
-  Optimize web fonts for Core Web Vitals.
+subhead: Optimize web fonts for Core Web Vitals.
 authors:
   - katiehempenius
   - tunetheweb
@@ -10,42 +9,25 @@ date: 2021-06-03
 updated: 2022-10-04
 hero: image/j2RDdG43oidUy6AL6LovThjeX9c2/T4kc8pjG9mvfZwynORjM.jpg
 alt: A photo of letterpress type.
-description: |
-  Learn about how to optimize web fonts for Core Web Vitals.
+description: Learn about how to optimize web fonts for Core Web Vitals.
 tags:
   - blog
   - performance
   - web-vitals
 ---
 
-This article discusses performance best practices for fonts. There are a variety
-of ways in which web fonts impact performance:
+This article discusses performance best practices for fonts. There are a variety of ways in which web fonts impact performance:
 
-*   **Delayed text rendering:** If a web font has not loaded, browsers typically
-    delay text rendering. In many situations, this delays [First Contentful
-    Paint (FCP)](/fcp). In some situations, this delays [Largest
-    Contentful Paint (LCP)](/lcp/).
-*   **Layout shifts:** The practice of font swapping has the potential to [cause
-    layout
-    shifts](/debug-layout-shifts/#identifying-the-cause-of-a-layout-shift).
-    These layout shifts occur when a web font and its fallback font take up
-    different amounts of space on the page.
+- **Delayed text rendering:** If a web font has not loaded, browsers typically delay text rendering. In many situations, this delays [First Contentful Paint (FCP)](/fcp). In some situations, this delays [Largest Contentful Paint (LCP)](/lcp/).
+- **Layout shifts:** The practice of font swapping has the potential to [cause layout shifts](/debug-layout-shifts/#identifying-the-cause-of-a-layout-shift) and so impact [Cumulative Layout Shift (CLS)](/cls/). These layout shifts occur when a web font and its fallback font take up different amounts of space on the page.
 
-This article is broken down into three sections: font loading, font delivery,
-and font rendering. Each section explains how that particular aspect of the font
-lifecycle works and provides corresponding best practices.
+This article is broken down into three sections: **font loading**, **font delivery**, and **font rendering**. Each section explains how that particular aspect of the font lifecycle works and provides corresponding best practices.
 
 ## Font loading
 
-Fonts are typically important resources, as without them the user might be
-unable to view page content. Thus, best practices for font loading generally focus
-on making sure that fonts get loaded as early as possible. Particular care
-should be given to fonts loaded from third-party sites as downloading these font
-files requires separate connection setups.
+Fonts are typically important resources, as without them the user might be unable to view page content. Thus, best practices for font loading generally focus on making sure that fonts get loaded as early as possible. Particular care should be given to fonts loaded from third-party sites as downloading these font files requires separate connection setups.
 
-If you're unsure if your page's fonts are being requested in time, check the
-**Timing** tab within the **Network** panel in Chrome DevTools for more
-information.
+If you're unsure if your page's fonts are being requested in time, check the **Timing** tab within the **Network** panel in Chrome DevTools for more information.
 
 {% Img
     src="image/j2RDdG43oidUy6AL6LovThjeX9c2/hFVGVzDQHymbC5aIAtD9.png",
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -135,41 +84,34 @@ stylesheet to download.
...
 ```
 
 {% Aside %}
-    Inlining the font files themselves is not recommended. Inlining large
-    resources like fonts is likely to delay the delivery of the main document,
-    and with it, the discovery of other resources.
-
+    Note that if only some of the CSS is inlined, then the browser will still need to wait for all the CSS to be loaded, before it can discover if fonts are needed.<br>
+    <br>
+    Also note that inlining the font files themselves is not recommended. Inlining large resources like fonts is likely to delay the delivery of the main document, and with it, the discovery of other resources.
```
malchata, line 98: Ope, nevermind. I see you mentioned it here.

src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -135,41 +84,34 @@ stylesheet to download.
...
-resource hint to establish early connection(s) with the third-party origin.
-Resource hints should be placed in the `<head>` of the document. The resource
-hint below sets up a connection for loading the font stylesheet.
+### Preconnect to critical third-party origins
 
+If your site loads fonts from a third-party site, it is highly recommended that you use the [`preconnect`](https://developer.mozilla.org/docs/Web/HTML/Link_types/preconnect) resource hint to establish early connection(s) with the third-party origin. Resource hints should be placed in the `<head>` of the document. The resource hint below sets up a connection for loading the font stylesheet.
 
 ```html
 <head>
   <link rel="preconnect" href="https://fonts.com">
```
malchata, line 110: What if we use the Google Fonts snippet here, so we have a more concrete example?

src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -178,9 +120,7 @@ connection](https://developer.mozilla.org/docs/Web/HTTP/CORS#what_requests_use_c
 </head>
 ```
 
-When using the `preconnect` resource hint, keep in mind that a font provider may
-serve stylesheets and fonts from separate origins. For example, this is how the
-`preconnect` resource hint would be used for Google Fonts.
+When using the `preconnect` resource hint, keep in mind that a font provider may serve stylesheets and fonts from separate origins. For example, this is how the `preconnect` resource hint would be used for Google Fonts.
```
malchata, line 123: Ah, I see you do show the Google fonts snippet.

There's a lot of examples for `preconnect` here. Would it be possible to condense these and explain what it does after the code snippet? Might help for readability.

src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -189,90 +129,59 @@ serve stylesheets and fonts from separate origins. For example, this is how the
...
-a more effective approach. These adjustments come closer to addressing the root
-cause of late-discovered fonts - rather than just providing a workaround.
-
-In addition, using `preload` as a font-loading strategy should also be used
-carefully as it bypasses some of the browser's built-in content negotiation
-strategies. For example, `preload` ignores `unicode-range` declarations, and if
-used prudently, should only be used to load a single font format.
+Inlining font declarations and adjusting stylesheets may be a more effective approach. These adjustments come closer to addressing the root cause of late-discovered fonts—rather than just providing a workaround.
 
+In addition, using `preload` as a font-loading strategy should also be used carefully as it bypasses some of the browser's built-in content negotiation strategies. For example, `preload` ignores `unicode-range` declarations, and if used prudently, should only be used to load a single font format.
```
malchata, line 143: This feels like it should be in an `'important'` aside. I didn't know this!

src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -189,90 +129,59 @@ serve stylesheets and fonts from separate origins. For example, this is how the
...
 
+### Using self-hosted fonts
 
-### Best practices
+On paper, using a self-hosted font should deliver better performance as it eliminates a third-party connection setup. However, in practice, the performance differences between these two options is less clear cut: for example, the [Web Almanac](https://almanac.httparchive.org/en/2020/fonts#fig-7) found that sites using third-party fonts had a faster render than fonts that used first-party fonts.
 
+If you are considering using self-hosted fonts, confirm that your site is using a [Content Delivery Network (CDN)](/content-delivery-networks/) and [HTTP/2](/content-delivery-networks/#http2-and-http3). Without use of these technologies, it is much less likely that self-hosted fonts will deliver better performance. For more information, see [Content Delivery Networks](/content-delivery-networks/).
 
-#### Using self-hosted fonts
+{% Aside %}
```
malchata, line 157:
```suggestion
{% Aside 'caution' %}
```

src/site/content/en/fast/font-best-practices/index.md
```diff
@@ -189,90 +129,59 @@ serve stylesheets and fonts from separate origins. For example, this is how the
...
-If you are considering using self-hosted fonts, confirm that your site is using
-a [Content Delivery Network (CDN)](/content-delivery-networks/)
-and [HTTP/2](/content-delivery-networks/#http2-and-http3).
-Without use of these technologies, it is much less likely that self-hosted fonts
-will deliver better performance. For more information, see [Content Delivery
-Networks](/content-delivery-networks/).
+### Use WOFF2
 
-{% Aside %}
+Of the modern font fonts, [WOFF2](https://www.w3.org/TR/WOFF2/) is the newest, has the widest browser support, and offers the best compression. Because it uses Brotli, WOFF2 compresses 30% better than WOFF, leading to less data to download and therefore faster performance.
```
malchata, line 165: I'd add that WOFF is entirely unnecessary if you don't need Internet Explorer support. WOFF2 alone is good enough most of the time!

malchata (approved):
Little nits, but overall LGTM.

src/site/content/en/fast/font-best-practices/index.md
tunetheweb: If the document is streaming, and it's a large chunk of BASE64, then it delays the rest of the doc arriving, which means the preload scanner can't get stuck on it.

Plus it says "is likely to delay". Likely being the key here :-)

src/site/content/en/fast/font-best-practices/index.md
tunetheweb: I'm not sure how much it matters really. You would preload the most likely used font (which can be a subsetted font, later referenced in a font-face). I considered removing this completely because of that. So don't think we should add an `'important'`.

src/site/content/en/fast/font-best-practices/index.md
tunetheweb: I think `'caution'` is too much here.

image

Considered `'important'`, but even then it's debatable which is better. Self-hosting a slow font can still be better than third-partying a faster font.

Will leave as is.

PR #8820, opened 2022-10-07
Updated Critical resource color to blue from green as used in the latest chrome updates

DharmenderRajput:
critical resources are now displayed with Blue and not green. Please also update the images used in article that displays the old green color for critical resources.


malchata (approved):
LGTM.

PR #8849, opened 2022-10-11
Learn HTML: Tweak text for doctypes

### LINKED ISSUE #8751: Learn HTML: Incorrect information about the doctype
**Describe the bug**

The doctype section says

> The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's a document type declaration or DTD. 

This is incorrect. DTD stands for Document Type *Definition*, which is the thing the document type declaration references (in SGML and XML). HTML these days doesn't have a DTD.

> The DTD tells the browser you're using HTML.

No, it doesn't. The MIME type `text/html` tells the browser you're using HTML. It will be HTML regardless of the doctype.

> If omitted, browsers may use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the DOCTYPE helps prevent quirks mode.

Yes. This could be more precise and say "will" instead of "may", as there's no uncertainty.

Exception:  in `<iframe srcdoc="...">` the doctype is optional and the document will be in no-quirks mode.

**To Reproduce**
1. Go to https://web.dev/learn/html/document-structure/#lessdoctype-htmlgreater

zcorpan:
Fixes #8751

- Remove reference of DTD in favor of doctype.

### DIFF
--- a/src/site/content/en/learn/html/document-structure/index.md
+++ b/src/site/content/en/learn/html/document-structure/index.md
@@ -18,7 +18,7 @@ There are several features that should be considered essential for any and every
 
 ### `<!DOCTYPE html>`
 
-The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's  a special kind of token called "doctype". The doctype tells the browser to use standards mode. If omitted, browsers will use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the doctype helps prevent quirks mode.
+The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's  a special kind of node called "doctype". The doctype tells the browser to use standards mode. If omitted, browsers will use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the doctype helps prevent quirks mode.
 
 ### `<html>`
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/learn/html/document-structure/index.md
```diff
@@ -18,11 +18,11 @@ There are several features that should be considered essential for any and every
 
 ### `<!DOCTYPE html>`
 
-The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's a document type definition or DTD. The DTD tells the browser to use standards mode. If omitted, browsers will use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the DOCTYPE helps prevent quirks mode.
+The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's  a special kind of token called "doctype". The doctype tells the browser to use standards mode. If omitted, browsers will use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the doctype helps prevent quirks mode.
```
zcorpan, line 21: Maybe "node" instead of "token"?
```suggestion
The first thing in any HTML document is the preamble. For HTML, all you need is `<!DOCTYPE html>`. This may look like an HTML element, but it isn't. It's  a special kind of node called "doctype". The doctype tells the browser to use standards mode. If omitted, browsers will use a different rendering mode known as [quirks mode](https://developer.mozilla.org/docs/Web/HTML/Quirks_Mode_and_Standards_Mode). Including the doctype helps prevent quirks mode.
```

rachelandrew (approved):

PR #8871, opened 2022-10-14
Adding chrometober banner

tropicadri:
Adding the chrometober banner to direct people to the content.


rachelandrew (changes requested):
I'm not happy with the container queries working across all browsers wording, because they don't. That's a confusing message.

una:
+1, we also removed the CQ dependency in Chrometober

tropicadri:
Thank you @rachelandrew and @una sending it to Harleen to update it!

rachelandrew (approved):

PR #8874, opened 2022-10-17
change skip link z-index

estelle:
.course web-header has a z-index of 10, and so does the skip link. As the skip link is first in the source code order, it shows up behind the header on narrow screens. This fixes that.

Fixes #SOME_ISSUE_NUMBER

-
-
-


matthiasrohmer (approved):
Good find, thanks for fixing!

PR #8877, opened 2022-10-17
Respect raw permalink property for algolia index

### LINKED ISSUE #8872: Fantom/draft articles in the search results
**Describe the bug**

There is a fantom (draft?) article in the search results.

For example https://web.dev/en/spaces/2022-10-13-chrometober/

There might be more.

**To Reproduce**

1. Go to [web.dev](https://web.dev/)
2. Search for “Chrometober”
3. Click “How we built Chrometober” in the search results

**Expected behavior**

- No results in the search if it’s still a draft
- Working article if it’s published

**Screenshots**

image

matthiasrohmer:
Fixes #8872. Happy for suggestions to fix this more elegant.

The problem is that Spaces announcements which are only used to render as list items on the spaces pages are also indexed in search, while actually not having pages on their own.

As 11ty mangles the permalink property looking it up on the unprocessed frontmatter is the only option. `permalink` is undefined for both `post` and `post.data` in case of the spaces pages.

### DIFF
--- a/src/site/_filters/algolia-indexable.js
+++ b/src/site/_filters/algolia-indexable.js
@@ -27,6 +27,7 @@ function algoliaIndexable(posts) {
     (post) =>
       post.url &&
       post.data.title &&
+      !(post.template.frontMatter.data.permalink === false) &&
       !(post.data.noindex === true || post.data.noindex === 'true'),
   );
 }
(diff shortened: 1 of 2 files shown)


devnook (changes requested):
The problem is that permalink = false might come from other source than frontmatter (e.g. highter in the data cascade).

In DCC, we prepare indexable pages in a collection called algolia, and we take permalink into account there: https://github.com/djm/remark-shortcodes/blob/master/index.js

Maybe if you use a custom collection instead of a filter permalink would be accessible?

matthiasrohmer:
Thought of that but somehow for the spaces pages the permalink property did not end up in the cascaded data. Will debug further and see if I can fix it more thoroughly. Or the root cause even as that's probably also the reason for https://github.com/GoogleChrome/web.dev/issues/8882.

devnook:
Ping on this

PR #8883, opened 2022-10-17
feat: Chrometober Week 2

jh3y:
- Chrometober Week Two! ✨


matthiasrohmer (approved):

matthiasrohmer (changes requested):
Netlify preview has errors: 
image

matthiasrohmer (approved):
Netlify error likely related to the fact that they rewrite image sources and while doing so break the mark-up.

PR #8922, opened 2022-10-24
Clarifies INP calculation logic per WEC.

malchata:

- WEC had some feedback regarding the INP doc, and that some concepts could be clearer. This PR clarifies the INP calculation logic per WEC's request.


src/site/content/en/metrics/inp/index.md
```diff
@@ -44,18 +44,72 @@ This article explains how INP works, how to measure it, and offers advice for im
...
+
+The duration for every interaction is captured by the [Event Timing API](https://caniuse.com/mdn-api_performanceeventtiming). In cases where there are few interactions made with a page&mdash;less than 50 to be exact&mdash;the interaction with the highest latency of all observed interactions is chosen as the page's INP when the user is done with the page.
+
+However, in cases where there are many interactions with a page&mdash;50 or more to be specific&mdash;the interaction with the highest latency isn't the page's INP, but rather a high-latency interaction that's close to the worst. The table below provides more information on how INP is calculated based on the total number of interactions made with a page.
```
tunetheweb, line 51: This is confusing to me. Why would be exclude events when we have them? I still don't get why we take 98% when we can.

Can we explain the reasoning/thought process behind this, instead of just giving the table to show _how_ we do this calc, but not _why_ we do this at all?

src/site/content/en/metrics/inp/index.md
```diff
@@ -196,7 +250,7 @@ As with other methods exported by `web-vitals`, `onINP` accepts a function as an
 See the [`onINP()`](https://github.com/GoogleChrome/web-vitals#oninp) reference documentation for additional usage instructions.
 
 {% Aside 'warning' %}
-Gathering INP metrics in the field will only work on browsers that [fully support the Event Timing API](https://caniuse.com/mdn-api_performanceeventtiming), including its `interactionId` property.
+Gathering INP metrics in the field will only work on browsers that fully support the Event Timing API, including its `interactionId` property.
```
tunetheweb, line 253:
```suggestion
Gathering INP metrics in the field will only work on browsers that [fully support the Event Timing API, including its `interactionId` property](https://developer.mozilla.org/docs/Web/API/PerformanceEventTiming/interactionId#browser_compatibility).
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -14,7 +14,7 @@ tags:
 ---
 
 {% Aside %}
-Interaction to Next Paint (INP) is an experimental metric that assesses [responsiveness](/user-centric-performance-metrics/#types-of-metrics). When an interaction causes a page to become unresponsive, that is a poor user experience. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below. A low INP means the page was consistently able to respond quickly to all&mdash;or the vast majority&mdash;of user interactions.
+Interaction to Next Paint (INP) is a metric that assesses [responsiveness](/user-centric-performance-metrics/#types-of-metrics). When an interaction causes a page to become unresponsive, that is a poor user experience. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below. A low INP means the page was consistently able to respond quickly to all&mdash;or the vast majority&mdash;of user interactions.
```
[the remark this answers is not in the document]
malchata: One of the bits of feedback WEC has given is that the "experimental" moniker has led partners to believe major changes are still coming to the metric (which isn't necessarily true per @mmocny).

src/site/content/en/metrics/inp/index.md
malchata: I removed this link since it was linked earlier, but can commit if you feel strongly about this.

tunetheweb: I changed the link specifically to show which browsers support the `interactionId` property. So not same link as previously. That's also why I expended the link text.

src/site/content/en/metrics/inp/index.md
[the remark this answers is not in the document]
malchata: Fair. It is a bold step, but I'd like to get @anniesullie and @mmocny's thoughts on this change. If it allays suspicions that major changes are coming to the metric, I think it's worth removing, as INP has not changed (to my knowledge) for some time.

src/site/content/en/metrics/inp/index.md
tunetheweb: Why do we take the 98th percentile* for INP? That to me is not explained. This change explains how the 98th percentile is calced, but not _why_ we bother doing this?

Is it something to do with excluding the extremes? If so why 98th percentile? Why not 90th? or 75th?

`*` And btw I do agree with removing "98th percentile" phrase itself - too confusing with 75th percentile. But still should explain WHY we do this.

Something like:

> For pages with many interactions, we often saw outlier INP values that were unrepresentative of how responsive the page felt to users overall. For this reason we exclude the very worst interaction values for these pages to get a more accurate representation of responsiveness on these pages. The INP value used is logged in the table below.

Or something like that.

WDYT @malchata / @anniesullie / @mmocny ?

src/site/content/en/metrics/inp/index.md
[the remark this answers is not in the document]
philipwalton: Let's keep it for now to be consistent with the framing used in other tools and other parts of the site. If we're going to remove it we'll need to remove it in several places at the same time.

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,18 +44,72 @@ This article explains how INP works, how to measure it, and offers advice for im
 
 ## What is INP?
 
-INP aims to represent a page's overall responsiveness by measuring all click, tap, and keyboard interactions made with a page. The longest of those observed interactions&mdash;with some exceptions noted below&mdash;is chosen as the page's INP value when the user is done with the page.
+INP aims to represent a page's overall responsiveness. It does so by measuring every click, tap, and keyboard interaction made with a page. An _interaction_ is one or more event handlers that fire during the same logical user gesture. For example, "tap" interactions on a touchscreen device include multiple events, such as `pointerup`, `pointerdown`, and `click`. An interaction can be driven by JavaScript, CSS, built-in browser controls (such as form elements), or a combination thereof.
```
philipwalton, line 47: Nit: an interaction consists of one or more user action events (as opposed to event handlers, which may or may not be present).

philipwalton (changes requested):
Thanks for getting this started @malchata. Overall, I think the "What is INP?" section needs to go into a bit more detail and define some core concepts of INP that aren't covered here.

Some specific concepts that I think need to be explained/defined:

- What an interaction is and how interaction latency is determined
- What an event is, how it relates to an interaction, and how event duration is defined
- How the specific interaction used for INP is determined

Also, I think all of this should happen before introducing the Event Timing API. I don't think that API is relevant before the "How to measure INP" section.

malchata:
@philipwalton and @tunetheweb, I updated the doc per your suggestions. Let me know if this works.

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
 
 ## What is INP?
 
-INP aims to represent a page's overall responsiveness by measuring all click, tap, and keyboard interactions made with a page. The longest of those observed interactions&mdash;with some exceptions noted below&mdash;is chosen as the page's INP value when the user is done with the page.
+INP aims to represent a page's overall responsiveness. It does so by measuring every click, tap, and keyboard interaction made with a page.
```
tunetheweb, line 47: Nit. "With" reads weird to me.

```suggestion
INP aims to represent a page's overall responsiveness. It does so by measuring every click, tap, and keyboard interaction made on a page.
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
...
+
+### What is an interaction?
+
+An interaction is one or more discrete user input actions. More than one interaction may consist of what's considered a "logical user gesture". For a key board interaction, this would consist of both a `keydown` _and_ a `keyup` event.
+
+### What is an event?
+
+Interactions are driven by events. Events fire callbacks during the same logical user gesture. For example, "tap" interactions on a touchscreen device include multiple events, such as `pointerup`, `pointerdown`, and `click`. While JavaScript is a common driver of interactivity, interactions can also be driven by CSS and built-in browser controls (such as `<input>` elements, for example).
```
tunetheweb, line 55: "Driver of interactivity" sounds like initiator, rather than handler.

WDYT about this instead?

```suggestion
Interactions are driven by events. Events fire callbacks during the same logical user gesture. For example, "tap" interactions on a touchscreen device include multiple events, such as `pointerup`, `pointerdown`, and `click`. While JavaScript is a common handler of interactivity, interactions can also be handled by CSS and built-in browser controls (such as `<input>` elements, for example).
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
...
 
-INP aims to represent a page's overall responsiveness by measuring all click, tap, and keyboard interactions made with a page. The longest of those observed interactions&mdash;with some exceptions noted below&mdash;is chosen as the page's INP value when the user is done with the page.
+INP aims to represent a page's overall responsiveness. It does so by measuring every click, tap, and keyboard interaction made with a page.
+
+### What is an interaction?
+
+An interaction is one or more discrete user input actions. More than one interaction may consist of what's considered a "logical user gesture". For a key board interaction, this would consist of both a `keydown` _and_ a `keyup` event.
+
+### What is an event?
+
```
tunetheweb, line 54: I dunno if you need this sub-heading.

```suggestion
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
...
+Interactions are driven by events. Events fire callbacks during the same logical user gesture. For example, "tap" interactions on a touchscreen device include multiple events, such as `pointerup`, `pointerdown`, and `click`. While JavaScript is a common driver of interactivity, interactions can also be driven by CSS and built-in browser controls (such as `<input>` elements, for example).
+
+The time it takes to queue up an interaction, including the time it takes for the event callbacks for the interaction to run, and the time it takes to present the next frame that shows the result of the interaction is known as the _event duration_.
+
+### How is INP calculated?
+
+Event duration for every interaction is recorded throughout the page lifecycle. In cases where there are few interactions made with a page&mdash;less than 50 to be exact&mdash;the interaction with the highest latency is chosen as the page's INP when the user is done with the page.
```
tunetheweb, line 61: Do we think we should recognise there is an INP per interaction? And a total page-level INP?

I note you're calling it "_event duration_" below, but it is the duration of the interaction until the next paint.

Or is that just confusing the situation and should we reserve INP for the metric?

Some more context from @mmocny here: https://twitter.com/mmocny/status/1578083571440537609?s=20&t=L_xMLmiuB804rpzLAU8RLw

```suggestion
### How is INP calculated for each event?

The time it takes to queue up an interaction, including the time it takes for the event callbacks for the interaction to run, and the time it takes to present the next frame that shows the result of the interaction is known as the _event duration_.

### How is INP calculated for the page?

Event duration for every interaction is recorded throughout the page lifecycle. In cases where there are few interactions made with a page&mdash;less than 50 to be exact&mdash;the interaction with the highest latency is chosen as the page's INP when the user is done with the page.

File: src/site/content/en/metrics/inp/index.md [hunk 9]
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
...
+
+However, in cases where there are many interactions with a page&mdash;50 or more to be specific&mdash;the interaction with the highest latency _isn't_ the page's INP, but rather an interaction that's close to the worst. The reasoning for this is that if the user has made many interactions with a page, the worst interaction may be an outlier that would unfairly penalize a page that was otherwise interacted with numerous times.
```
tunetheweb, line 63: I'd swap this around and give the reason first, then what that means. Also didn't like the "close to the worst" wording - feels a bit unspecific.

```suggestion
In cases where there are many interactions with a page, the worst interaction may be an outlier that would unfairly penalize a page that was otherwise interacted with quickly numerous times. The more interactions, the more likely an outlier interaction will be experienced. For this reason, where there are many interactions with a page&mdash;50 or more to be specific&mdash;we exclude an increasing number of the the worst interactions, to look at an INP that is more representative of how responsive the page was overall for the user.
```

src/site/content/en/metrics/inp/index.md
malchata: I personally think this would add to confusion. INP—as I understand it—is a single value that's representative of overall page responsiveness by picking a single interaction from all observed interactions.

tunetheweb: Yeah I can see that. It's more obvious with CLS where there are individual "layout shifts" events and one "cumulative layout shift" metric. And for LCP it's the "LCP element" and the LCP metric.

For INP the full wording "Interaction to Next Paint" means it _could_ apply to both. But agree that would be confusing so maybe better to reserve this for the metric.

@philipwalton might be worth adding this to your INP terms doc, to state we use "INP" to refer to the metric, rather than the individual events?

tunetheweb (approved):

malchata:
Thanks for the approval here, @tunetheweb, but I do think I will wait for @philipwalton to approve before merging.

malchata:
Pinging @tunetheweb and @philipwalton to re-review. I added a commit yesterday that adds a note about INP and damage.

src/site/content/en/metrics/inp/index.md
```diff
@@ -44,20 +44,88 @@ This article explains how INP works, how to measure it, and offers advice for im
...
+An interaction's latency consists of the single longest [duration](https://w3c.github.io/event-timing/#ref-for-dom-performanceentry-duration%E2%91%A1:~:text=The%20Event%20Timing%20API%20exposes%20a%20duration%20value%2C%20which%20is%20meant%20to%20be%20the%20time%20from%20when%20user%20interaction%20occurs%20(estimated%20via%20the%20Event%27s%20timeStamp)%20to%20the%20next%20time%20the%20rendering%20of%20the%20Event%27s%20relevant%20global%20object%27s%20associated%20Document%E2%80%99s%20is%20updated) of a group of event callbacks that drives the interaction, from the time the user begins the interaction to the moment the next frame is presented with visual feedback.
```
tunetheweb, line 128: I find these long fragment search URLs quite brittle btw. They break as soon as any text is tweaked on the page. Should it just link to the duration link that page links to: https://w3c.github.io/performance-timeline/#dom-performanceentry-duration

tunetheweb (commented):
I've a good few comments still. Sorry, I know this PR is dragging on, but think it's an important one to get right.

src/site/content/en/metrics/inp/index.md
```diff
@@ -152,6 +223,12 @@ It's possible that a page can return no INP value. This can happen for a number
...
 - The page is being accessed by a bot such as a search crawler or headless browser that has not been scripted to interact with the page.
 
+## What if an interaction causes no layout changes?
+
+If you have an interaction with an event callback that causes no change to the layout in an event handler, that interaction can still be an INP candidate&mdash;**even if your event handler does nothing at all**. The reason for this is that the Event Timing API&mdash;which is what supplies the data needed to calculate the page's INP&mdash;will still add entries to the performance entry buffer in such cases.
+
+In these cases, you should be thinking about what sort of visual feedback you can provide for the user that lets them know that something is happening as a result of the interaction. One of the primary goals of INP is to recognize when you have an opportunity to present the next frame to the user&mdash;and that frame should show the result of that interaction, even if it's a small visual change that communicates whether some work is in progress.
```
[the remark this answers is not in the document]
malchata: I'll go play-by-play here:

1) It is _not_ a `0` INP. I've observed values as high as 30 for no-op event handlers.
2) I've removed the bit about event timing.
3) I think this is splitting hairs a bit, but I have removed that paragraph and pared things down a bit. Let me know what you think.

tunetheweb (approved):

src/site/content/en/metrics/inp/index.md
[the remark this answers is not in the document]
malchata: I updated the snippet @tunetheweb. Let me know what you think.

cc: @philipwalton for his review and approval.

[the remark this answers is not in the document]
tunetheweb: LGTM!

philipwalton (changes requested):
Sorry, I submitted these comments earlier, but now that I'm on GitHub and looking at this PR again, it looks like they're still "pending" for some reason. Re-submitting now.

src/site/content/en/metrics/inp/index.md
```diff
@@ -152,6 +223,10 @@ It's possible that a page can return no INP value. This can happen for a number
 - The page was loaded, but the user interacted with the page using gestures that did not involve clicking, tapping, or using the keyboard. For example, scrolling or hovering over elements does not factor into how INP is calculated.
 - The page is being accessed by a bot such as a search crawler or headless browser that has not been scripted to interact with the page.
 
+## What if an interaction causes no layout changes?
+
+If you have an interaction with an event callback that causes no change to the layout in an event handler, that interaction can still be a candidate for INP. In these cases, you should be thinking about what sort of visual feedback you _could_ provide for the user that lets them know that something is happening as a result of the interaction. The intent of INP is to measure the delay before any potential paint can complete, and that involves recognizing _when_ you have an opportunity to present the next frame to the user. In the case of asynchronous work especially, that frame should communicate that work is in progress as soon as possible.
```
philipwalton, line 228: This paragraph talks a lot about event callbacks that don't do anything, but the much more common case is that there aren't any event callbacks at all.

For example, if I tap on the body of a page, that will cause the browser to dispatch a number of events which the page may or may not have registered listeners for. But if the main thread is busy and takes a while to respond, that interaction could still become the INP candidate even if there were not event listeners registered for those particular events.

See the FID doc for how this is discussed there: https://web.dev/fid/#what-if-an-interaction-doesnt-have-an-event-listener

malchata:
Bumping this for @philipwalton. Does this guidance seem OK?

PR #8924, opened 2022-10-24
Update Optimize CLS article

tunetheweb:

- Remove Emojis from article headings. They're fun, but lead to weird URLs. And the Lighthouse links don't work because of them (which was the original small thing I came to fix, but as ever couldn't hold myself back from doing other changes while in here!).
- Update `aspect/ratio` info following @jakearchibald 's [recent post on this](https://jakearchibald.com/2022/img-aspect-ratio/).
- Add info on `<source>` support for art direction which has been added since this article was published.
- Add width and height to videos (should really practice what we preach!), though they all use 16/9 so have correct space reserved anyway.
- Remove references to Lighthouse 6 and Chrome 84 as those really are quite old by now. Think it's safe to just say Lighthouse and Chrome now.
- Add reference to Performance Insight's panel.

### DIFF
--- a/src/site/content/en/blog/optimize-cls/index.md
+++ b/src/site/content/en/blog/optimize-cls/index.md
@@ -389,7 +389,7 @@ To learn more about what CSS properties trigger layout, see [CSS Triggers](https
 
 The back/forward cache, or bfcache keeps pages in browsers memory for a short period so if you return to them, then they will be restored exactly as you left them. This means the fully loaded page is instantly available—without any shifts which may be normally seen during load to to any of the reason above. When this was rolled out to Chrome, we saw [noticeable improvements in CLS](https://twitter.com/anniesullie/status/1491399685961293828?s=20&t=Qj_nzSRZD0_c-HaAnfr98Q).
 
-However, some sites are inelgible for the bfcache. Ensuring your site is eligible is a sure fire way to keep CLS scores down. Read [the bfcache article](/bfcache/) for more details on how to test and identify any issues preventing bfcache usage.
+Some sites are ineligible for the bfcache due to a variety of reasons. Ensuring your site is eligible is a sure fire way to keep CLS scores down. Read [the bfcache article](/bfcache/) for more details on how to test and identify any issues preventing bfcache usage.
 
 ### Developer Tools
 
--- a/src/site/content/en/blog/optimize-cls/index.md
+++ b/src/site/content/en/blog/optimize-cls/index.md
@@ -131,7 +131,7 @@ img[Attributes Style] {
 
 Safari behaves similarly by using a `HTML Attributes` style source. Firefox does not currently display this calculated `aspect-ratio` at all in it's Inspector panel, but does use it for layout.
 
-The `auto` part of the above code, is important and causes the `640 / 360` to be overriden with the image dimensions once the image is downloaded. If the image dimensions are different this will still cause some layout shift after the impage loads, but this ensures the image aspect-ratio is still used ultimately when it becomes available—as it was in the past—in case the HTML is incorrect. Plus, the shift is likely to be a lot smaller than the 0x0 default image size when dimentions are not provided!
+The `auto` part of the above code is important as it causes the `640 / 360` to be overriden with the image dimensions once the image is downloaded. If the image dimensions are different this will still cause some layout shift after the image loads, but this ensures the image aspect ratio is still used ultimately when it becomes available—as it was in the past—in case the HTML is incorrect. Plus, the shift is likely to be a lot smaller than the 0x0 default image size when dimensions are not provided!
 
 Tip: If you're having a hard time understanding aspect ratio, a handy [calculator](https://aspectratiocalculator.com/16-9.html) is available to help.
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -112,19 +117,23 @@ Modern browsers now set the default aspect ratio of images based on an image's w
...
 ```css
-img {
-  aspect-ratio: attr(width) / attr(height);
+img[Attributes Style] {
+  aspect-ratio: auto 640 / 360;
 }
 ```
 
-This calculates an aspect ratio based on the `width` and `height` attributes before the image has loaded. It provides this information at the very start of layout calculation. As soon as an image is told to be a certain width (for example `width: 100%`), the aspect ratio is used to calculate the height.
+Safari does similarly with a `HTML Attributes` style source. Firefox does not currently display this calculated `aspect-ratio` at all in it's Inspector panel, but does use it for layout.
```
malchata, line 132:
```suggestion
Safari behaves similarly by using a `HTML Attributes` style source. Firefox does not currently display this calculated `aspect-ratio` at all in it's Inspector panel, but does use it for layout.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -112,19 +117,23 @@ Modern browsers now set the default aspect ratio of images based on an image's w
...
 
-Tip: If you're having a hard time understanding aspect ratio, a handy [calculator](https://aspectratiocalculator.com/16-9.html) is available to help.
+The `auto` part of the above code, is important and causes the `640 / 360` to be overriden with the image dimensions once the image is downloaded. If the image dimensions are different this will still cause some layout shift after the impage loads, but this ensures the image aspect-ratio is still used ultimately when it becomes available—as it was in the past—in case the HTML is incorrect. Plus, the shift is likely to be a lot smaller than the 0x0 default image size when dimentions are not provided!
```
malchata, line 134:
```suggestion
The `auto` part of the above code is important as it causes the `640 / 360` to be overriden with the image dimensions once the image is downloaded. If the image dimensions are different this will still cause some layout shift after the image loads, but this ensures the image aspect ratio is still used ultimately when it becomes available—as it was in the past—in case the HTML is incorrect. Plus, the shift is likely to be a lot smaller than the 0x0 default image size when dimensions are not provided!
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -315,41 +367,53 @@ Downloading and rendering web fonts can cause layout shifts in two ways:
 
 The following tools can help you minimize this:
 
-- <code>[font-display](/font-display/)</code> allows you to modify the rendering behavior of custom fonts with values such as <code>auto</code>, <code>swap</code>, <code>block</code>, <code>fallback</code> and <code>optional</code>. Unfortunately, all of these values (except [optional](http://crrev.com/749080)) can cause a re-layout in one of the above ways.
-- The [Font Loading API](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization#the_font_loading_api) can reduce the time it takes to get necessary fonts.
+- [`font-display`](/font-display/) allows you to modify the rendering behavior of custom fonts with values such as `auto`, `swap`, `block`, `fallback` and `optional`. Unfortunately, all of these values (except [`optional`](http://crrev.com/749080)) can cause a re-layout in one of the above ways.
+- The [Font Loading API](./optimize-webfont-loading/#the-font-loading-api) can reduce the time it takes to get necessary fonts.
 
-As of Chrome 83, I can recommend the following too:
+We recommend the following too to reduce CLS for fonts:
```
malchata, line 373:
```suggestion
More recommendations for reducing font-related CLS:
```

malchata: Try to avoid first person pronouns! :)

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -315,41 +367,53 @@ Downloading and rendering web fonts can cause layout shifts in two ways:
...
 
 - Using `<link rel=preload>` on the key web fonts: a preloaded font will have a higher chance to meet the first paint, in which case there's no layout shifting.
 - Combining `<link rel=preload>` and `font-display: optional`
 
-Read [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](/preload-optional-fonts/) for more details.
+Read [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](/preload-optional-fonts/) for more details, and [Best practices for fonts](./font-best-practices/) for other font best practices.
```
malchata, line 378:
```suggestion
Read [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](/preload-optional-fonts/) for more details, and [Best practices for fonts](/font-best-practices/) for other font best practices.
```

tunetheweb:
Thanks @malchata , Gilberto asked for an extra image to this article today and I said I was editing it as we speak. So might be one further update still to come.

PR #8926, opened 2022-10-24
Add Cache-conrol: no-store section to bfcache article

tunetheweb:

- Add section on `Cache-conrol: no-store` to bfcache article noting that when this is set on the page resource, it makes the page inelgible for the bfcache.

### DIFF
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -303,7 +303,15 @@ onAllChangesSaved(() => {
 {% endCompare %}
 
 
-### Avoid window.opener references
+### Do not serve pages with `Cache-control: no-store`
+
+`Cache-control: no-store` is an HTTP header web servers can set on responses to ensure the response is not stored in any cache. When this is set on the page itself (as opposed to any subresource), this means that page will not be included the bfcache. There is [a proposal to change this](https://bugs.chromium.org/p/chromium/issues/detail?id=1228611) but at present any pages sending this are ineligible to use the bfcache.
+
+For sites that simply want to ensure a reasonably fresh page, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page being used without being revalidated but, unlike `Cache-control: no-store`, still allows the bfcache to be used. For bfcache navigations a page refresh may not be expected or desirable for the user.
+
+For best performance, it is recommended that pages are cacheable, even if only for a short time. This will improve performance in general, but also avoid your site becoming ineligible for the bfcache and therefore not benefiting from those instant page restores. Where this is not possible `Cache-Control: no-cache`, `Cache-Control: max-age=0` should be used unless there are good security reasons for using `Cache-control: no-store`.
+
+### Avoid `window.opener` references
 
 In some browsers (including Chromium-based browsers) if a page was opened using
 <code>[window.open()](https://developer.mozilla.org/docs/Web/API/Window/open)</code>


src/site/content/en/blog/bfcache/index.md
```diff
@@ -303,7 +303,15 @@ onAllChangesSaved(() => {
 {% endCompare %}
 
 
-### Avoid window.opener references
+### Do not serve pages with `Cache-control: no-store`
+
+`Cache-control: no-store` is an HTTP header web servers can set on responses to ensure the response is not stored in any cache. When this is set on the page itself (as opposed to any subresource), this means that page will not be included the bfcache. There is [a proposal to change this](https://bugs.chromium.org/p/chromium/issues/detail?id=1228611) but at present any pages sending this are ineligible to use the bfcache.
```
fergald, line 308: Better link

https://github.com/fergald/explainer-bfcache-ccns/blob/main/README.md

fergald (commented):
LGTM with 1 comment

src/site/content/en/blog/bfcache/index.md
philipwalton, line 308:
```suggestion
`Cache-Control: no-store` is an HTTP header web servers can set on responses that instructs the browser not to store the response in any cache. When this is set on the page itself (as opposed to any subresource), it means that page will not be eligible for bfcache. There is [a proposal to change this behavior](https://github.com/fergald/explainer-bfcache-ccns/blob/main/README.md) but at present any pages using `Cache-control: no-store` are not eligible for bfcache.
```

Modified the wording a bit for clarity. Also, FYI, Responses with `no-store` can be stored in the [Cache Storage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage), so I wouldn't use "ensures" here.

src/site/content/en/blog/bfcache/index.md
```diff
@@ -303,7 +303,15 @@ onAllChangesSaved(() => {
...
+
+For sites that simply want to ensure an up-to-date page is used each time, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page being used without being revalidated but, unlike `Cache-control: no-store`, still allows the bfcache to be used. For bfcache navigations a page refresh may not be expected or desirable for the user.
```
philipwalton, line 310:
```suggestion
For sites that simply want to ensure an up-to-date page is used each time, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page from being used without revalidation first. And unlike `Cache-Control: no-store`, these directives do not disqualify a page from being eligible for bfcache.
```

I'm not sure that last sentence is necessary, as the rest of the post makes the argument why bfcache is a better user experience. I don't think you need to restate it here.

src/site/content/en/blog/bfcache/index.md
```diff
@@ -303,7 +303,15 @@ onAllChangesSaved(() => {
 {% endCompare %}
 
 
-### Avoid window.opener references
+### Do not serve pages with `Cache-control: no-store`
```
philipwalton, line 306:
```suggestion
### Avoid `Cache-Control: no-store`
```

Are there really no good reasons to use `no-store`? If we really do want to always discourage it then we should probably give some examples of how you can prevent surprising users in situations like if the user logs out and then clicks the back button (see other comments).

src/site/content/en/blog/bfcache/index.md
```diff
@@ -303,7 +303,15 @@ onAllChangesSaved(() => {
...
+
+For sites that simply want to ensure an up-to-date page is used each time, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page being used without being revalidated but, unlike `Cache-control: no-store`, still allows the bfcache to be used. For bfcache navigations a page refresh may not be expected or desirable for the user.
+
+For best performance, it is recommended that pages are cacheable, even if only for a short time. This will improve performance in general, but also avoid your site becoming ineligible for the bfcache and therefore not benefiting from those instant page restores. Where this is not possible `Cache-Control: no-cache` or `Cache-Control: max-age=0` should be used unless there are good security reasons for using `Cache-control: no-store`.
```
philipwalton, line 312: I'm not sure this paragraph is necessary. It mostly just restates the previous two paragraphs.

Instead, it might be good to end with a code example explaining how a developer could check to ensure the user is still logged in a `pageshow` event callback before allowing a page to be restored from bfcache, e.g.:

```js
addEventListener('pageshow', (event) => {
  if (
    event.persisted &&
    // Check to see if the user is still logged in....
    document.cookie.match(/.../)
  ) {
    // Force reload if the user is not logged in...
    location.reload();
  }
});
```

philipwalton (changes requested):

src/site/content/en/blog/bfcache/index.md
philipwalton: Hmmm, apparently I forgotten that [I already added this code](https://web.dev/bfcache/#never-use-the-unload-event:~:text=The%20following%20code%20checks%20for%20the%20presence%20of%20a%20site%2Dspecific%20cookie%20in%20the%20pageshow%20event%20and%20reloads%20if%20the%20cookie%20is%20not%20found%3A).

src/site/content/en/blog/bfcache/index.md
fergald: I think ensuring that logged-in-only information is not re-presented is a legit use. pageshow handlers could do something close but reliably hiding sensitive information there is tricky (without just deleting the whole DOM) and maybe defeatable with devtools. Basically it's the simplest and most reliable method.

src/site/content/en/blog/bfcache/index.md
tunetheweb: I specifically added the last sentence to answer the question of "why should back/forward navigations be special?". For example, if I'm a news publisher I might prefer the latest page is used, but I argue on back/forward navigations (specifically on "back") a fresh page might not be expected rather than going back to the page you were just at.

WDYT? Still not worth including that last sentence?

The rest of your edits are find btw so will commit them with or without the last sentence as you see fit.

src/site/content/en/blog/bfcache/index.md
tunetheweb: Yeah that's why I had "should be used unless there are good security reasons for using `Cache-control: no-store`." below. Maybe should expand upon that?

However, Chrome is also planning on clearing the bfcache once cookies change before enabling the `no-store` use case (I presume that is still the case @fergald ?), which should cover this case too. So I'm kinda reluctant to recommend that's `no-store` is used for this case, then in the (near?) future we override that. Hence why I left it a little vague with that line.

src/site/content/en/blog/bfcache/index.md
tunetheweb: Gave reasons for the two paragraphs above so late me know if you still feel the same.

Agree, it's probably a good idea to link to the https://web.dev/bfcache/#update-stale-or-sensitive-data-after-bfcache-restore section so will add that once we decide if keeping these or not.

src/site/content/en/blog/bfcache/index.md
fergald: Our intention there is to stop over-blocking and that has to be done in a safe way, so anyone using CCNS to protect logged-in-only info is correct to do so and should continue doing so.

tunetheweb: Gotcha. So guidance is: `no-store` for sensitive information and `no-cache`/`max-age=0` for avoiding stale information without revalidated and we just think the former is being over used at present for reasons covered by the latter. And we think that often not revalidating on bfcache restore is actually OK.

In that case think we should make that clear. But will see where we land on the current comments before tackling that.

fergald: It's not that it's being overused (well it also is but assuming it's actually being used appropriately) it's that it's too blunt an instrument. We never cache, just in case we cache after logout. We know that's 90% of the time, it would have been fine to cache. The proposal is for us to start figuring that out and to let pages tell us directly.

tunetheweb: Right but, since a web site can’t control this the advice is still `no-store` on sensitive information. For example, they are at /private-page and then navigate to /logout to prevent restoring on back they should use `no-store` on /private-page. At some point we (Chrome) might handle that better by allowing certain “backs” despite `no-store` but advice for developer wouldn’t change even then - use `no-store` and let us worry about that (with extra checks via JavaScript on restore if really necessary).

But pages without private info, but with “fresh” preference (e.g. news publisher) then `no-store` is overkill and should not be used as restoring back to that page is more likely what the user expects.

That about sum up the situation? Will think of how to express this is an easy to understand way for the post!

fergald: Yes. Thanks.

malchata (approved):
Not a bfcache expert, but the additions read fine. LGTM!

src/site/content/en/blog/bfcache/index.md
```diff
@@ -302,27 +297,66 @@ onAllChangesSaved(() => {
 {% endCompareCaption %}
 {% endCompare %}
 
+### Minimize use of `Cache-Control: no-store`
 
-### Avoid window.opener references
+`Cache-Control: no-store` is an HTTP header web servers can set on responses that instructs the browser not to store the response in any HTTP cache. This should be used for resources containing sensitive user information, for example pages behind a login.
+
+Bfcache is not a HTTP cache but, historically, when `Cache-Control: no-store` is set on the page itself (as opposed to any subresource), browsers have not be included the page in the bfcache. There is [work currently underway to change this behavior for Chrome](https://github.com/fergald/explainer-bfcache-ccns/blob/main/README.md) in a privacy-preserving manner, but at present any pages using `Cache-Control: no-store` will not get the performance benefits of the bfcache.
```
tunetheweb, line 304:
```suggestion
Bfcache is not a HTTP cache but, historically, when `Cache-Control: no-store` is set on the page itself (as opposed to any subresource), browsers have not included the page in the bfcache. There is [work currently underway to change this behavior for Chrome](https://github.com/fergald/explainer-bfcache-ccns/blob/main/README.md) in a privacy-preserving manner, but at present any pages using `Cache-Control: no-store` will not get the performance benefits of the bfcache.
```

src/site/content/en/blog/bfcache/index.md
```diff
@@ -302,27 +297,66 @@ onAllChangesSaved(() => {
...
+
+For best performance, it is recommended that pages that do not contain sensitive user information are cacheable, even if only for a short time. This will improve performance in general, but also avoid your site becoming ineligible for the bfcache and therefore not benefiting from those instant page restores.
+
+For sites that do not contain sensitive user information but still wish to ensure stale pages are not served, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page from being used without being revalidated first. However, unlike `Cache-Control: no-store`, these directives do not prevent pages from being included in—and getting the performance benefits of—the bfcache. For back/forward navigations a page refresh may not be expected nor desirable for the user.
+
+In the next section we show how you can ensure stale data is not served if you do want to update the page on bfcache restores but still retain the performance benefits of using the bfcache.
```
philipwalton, line 310:
```suggestion
Note that when a page is restored from bfcache, it is restored from memory, not from the HTTP cache. As a result, directives like `Cache-Control: no-cache` or `Cache-Control: max-age=0` are not taken into account, and no revalidation occurs before the content is displayed to the user.

This is still likely a better user experience, however, as bfcache restores are instant and—since pages do not stay in the bfcache for very long—it's unlikely that the content is out of date. However, if your content does change minute-by-minute, you can fetch any updates using the `pageshow` event, as outlined in the next section.
```

Add some content explaining that cache control directives do not apply to bfcache restores (which I think is important to clarify).

Also, note that our style guide discourages the use of "we", see: https://developers.google.com/style/person

philipwalton (changes requested):

src/site/content/en/blog/bfcache/index.md
```diff
@@ -302,27 +297,66 @@ onAllChangesSaved(() => {
...
 {% endCompare %}
 
+### Minimize use of `Cache-Control: no-store`
 
-### Avoid window.opener references
+`Cache-Control: no-store` is an HTTP header web servers can set on responses that instructs the browser not to store the response in any HTTP cache. This should be used for resources containing sensitive user information, for example pages behind a login.
+
+Bfcache is not a HTTP cache but, historically, when `Cache-Control: no-store` is set on the page itself (as opposed to any subresource), browsers have not included the page in the bfcache. There is [work currently underway to change this behavior for Chrome](https://github.com/fergald/explainer-bfcache-ccns/blob/main/README.md) in a privacy-preserving manner, but at present any pages using `Cache-Control: no-store` will not get the performance benefits of the bfcache.
+
+For best performance, it is recommended that pages that do not contain sensitive user information are cacheable, even if only for a short time. This will improve performance in general, but also avoid your site becoming ineligible for the bfcache and therefore not benefiting from those instant page restores.
```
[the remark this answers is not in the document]
tunetheweb:
```suggestion
```

src/site/content/en/blog/bfcache/index.md
```diff
@@ -302,27 +297,66 @@ onAllChangesSaved(() => {
...
+
+For sites that do not contain sensitive user information but still wish to ensure stale pages are not served, using `Cache-Control: no-cache` or `Cache-Control: max-age=0` prevents the page from being used without being revalidated first. However, unlike `Cache-Control: no-store`, these directives do not prevent pages from being included in—and getting the performance benefits of—the bfcache. For back/forward navigations a page refresh may not be expected nor desirable for the user.
```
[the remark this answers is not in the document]
tunetheweb: Looks good thanks but think we should remove "truly" form "it should only be set on pages that contain truly sensitive information where caching of any sort is never appropriate". Data is either sensitive or not. Edited your suggestion to remove that word.

src/site/content/en/blog/bfcache/index.md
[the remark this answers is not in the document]
tunetheweb: OK removed.

philipwalton (approved):

PR #8931, opened 2022-10-25
Tweak service worker reference. Fixes #4978

### LINKED ISSUE #4978: The Offline Cookbook: On user interaction section is misleading
**Describe the bug**
The [On user interaction](https://web.dev/offline-cookbook/#on-user-interaction) section of The Offline Cookbook is misleading.  Particularly this last part "...meaning you don't need to involve the service worker to add things to the cache."

While you *can* add things to the cache from the page, those things are still downloaded through the service worker. So the only way to not "..involve the service worker..." is to make sure the service worker does not handle the fetch events for the urls being added to the cache by the page.  In other words the scenario described in that section is more complicated than it is made out to be.

**To Reproduce**
I tested this locally. Unfortunately I do not have an example site where you can verify it yourself. 

This behavior is logical based on how `cache.add` is described here: https://developer.mozilla.org/en-US/docs/Web/API/Cache/add

jakearchibald:
Fixes #4978

### DIFF
--- a/src/site/content/en/reliable/offline-cookbook/index.md
+++ b/src/site/content/en/reliable/offline-cookbook/index.md
@@ -178,8 +178,8 @@ document.querySelector('.cache-article').addEventListener('click', function (eve
 });
 ```
 
-The [caches API][caches_api] is available from pages as well as service workers, meaning you don't
-need to involve the service worker to add things to the cache.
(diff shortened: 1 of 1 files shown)


AaronForinton (approved):

PANHAX855K (commented):
Facebook password connect Gmail accoun panhax855@gmail.com with me

PR #8932, opened 2022-10-25
Replace < and > with entity names in a caption

samimaat:
Seems like ´<tool-tip>´ breaks the caption, because it gets treated as an element. This causes the caption to leak over to the remaining text.

I don't know why nothing breaks in the previous inclusion of `<tool-tip>`.


AaronForinton:
Looks like this was fixed somewhere else and then merged into this PR. It's showing 0 files changed now. Since you didn't get a chance to sign the CLA I'm going to assume your change was picked up by one of our writers and integrated into the docs.

PR #8960, opened 2022-10-30
Update index.md

seneca:
change infeasible to unfeasible

Fixes #SOME_ISSUE_NUMBER

-
-
-


rachelandrew:
Thanks for your PR, however infeasible is correct, as per Garner's Modern English Usage and [Merriam Webster](https://www.merriam-webster.com/dictionary/infeasible).

(I think both are actually correct, though infeasible is more commonly used in American English, which we use on this site.)

PR #8973, opened 2022-11-01
[Eyedropper article] Add spec draft link

tomayac:

- Add spec draft link


rachelandrew:
@tomayac this page is on the move https://github.com/GoogleChrome/developer.chrome.com/pull/4166 so I've added your change into the migrated version.

PR #8991, opened 2022-11-02
Upgrade web-vitals to 3.0.4

tunetheweb:

- Prerender2 is coming, and web-vitals 3.0.4 contains some small bug fixes for that so let's upgrade to it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28096,9 +28096,9 @@
       }
     },
     "node_modules/web-vitals": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/web-vitals/-/web-vitals-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


philipwalton:
This is fine with me, but before merging we should confirm with @devnook that the build process doesn't automatically install the latest semver-matching dependency already. If so then we don't need to do these types of updates.

philipwalton (approved):
LGTM.

Probably also worth confirming with @devnook that the build doesn't automatically install the latest semver-matching dependency. I assume it uses package-lock, but worth confirming...

matthiasrohmer:
Yes, that's needed if we want to update. We are gradually trying to specify exact versions in package.json as well to have consistent dependencies across npm/yarn/with or without lock file/what not, but for now this looks good 🙂

PR #8999, opened 2022-11-03
Handle prerender page views

tunetheweb:

- Stops logging pageviews for prerendered pages until they are activated, in preparation for Prerender2 coming soon.

~@philipwalton I've kept this simple for now and not merged the inline analytics code for now. Seems like that should be a separate PR if we want to clean that up and may want to keep at least part of that (the bit that defines `window.ga`) as some pages send separate events.~ Refactored to move this as discussed below.


src/lib/analytics.js
```diff
@@ -157,6 +157,28 @@ window.addEventListener(
   },
 );
 
+// Handle prerenders that have already activated
+if (
+  performance &&
+  performance.getEntriesByType &&
+  performance.getEntriesByType('navigation')[0].activationStart > 0
```
philipwalton, line 164: I think there's a race condition here. Since the code in this file could run later than the code in the template, it's possible that for prerendered pages a pageview hit will be sent twice.

Consider this case:

1. The page is prerendered but activation happens before the template code has a chance to run
2. When the template code runs the `!document.prerendering` conditional, it returns `true` because the page is no longer prerendering, and a pageview hit is sent.
3. The code in this file runs later and the `activationStart > 0` conditional is also `true`, so another pageview hit is sent.

src/site/_includes/partials/analytics.njk
```diff
@@ -10,12 +10,16 @@ ga('set', 'transport', 'beacon');
...
-try {
-  ga('set', '{{ site.analytics.dimensions.NAVIGATION_TYPE }}', performance.getEntriesByType('navigation')[0].type.replace(/_/g, '-'));
-} catch (error) {
-  ga('set', '{{ site.analytics.dimensions.NAVIGATION_TYPE }}', '(not set)');
-}
 ga('set', '{{ site.analytics.dimensions.COLOR_SCHEME_PREFERENCE }}', window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light');
-ga('send', 'pageview');
+// Do not send a pageview on pages which are prerendering
+// These will be handled later once the page becomes active
+if (!document.prerendering) {
```
philipwalton, line 16: As I mentioned in the other comment, it's possible that this code will run after the page has already been activated, which means this conditional will return true and a normal pageview will be sent without the `prerender` type set.

I don't think there's any good way to do this without either:

* Doing all the prerender checking in the template code, or
* Moving all of the pageview sending logic from the template into the analytics file (which is ultimately a better place for it)

philipwalton (changes requested):

src/lib/analytics.js
tunetheweb: Good spot. Fixed.

src/site/_includes/partials/analytics.njk
```diff
@@ -10,12 +10,6 @@ ga('set', 'transport', 'beacon');
 ga('set', 'anonymizeIp', true);
 ga('set', 'page', window.location.pathname);
 ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version}}');
-try {
-  ga('set', '{{ site.analytics.dimensions.NAVIGATION_TYPE }}', performance.getEntriesByType('navigation')[0].type.replace(/_/g, '-'));
-} catch (error) {
-  ga('set', '{{ site.analytics.dimensions.NAVIGATION_TYPE }}', '(not set)');
-}
 ga('set', '{{ site.analytics.dimensions.COLOR_SCHEME_PREFERENCE }}', window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light');
```
philipwalton, line 13: I think it probably makes sense to move this line into the `analytics.js` file as well.

src/lib/analytics.js
```diff
@@ -138,25 +138,64 @@ store.subscribe(({isSignedIn}) => {
...
+    performance.getEntriesByType &&
+    performance.getEntriesByType('navigation')[0].activationStart > 0
+  ) {
+    ga('set', dimensions.NAVIGATION_TYPE, 'prerender');
+    ga('send', 'pageview');
+    return;
+  }
+
+  // Handle prerenders that activate in the future
+  if (document.prerendering) {
```
philipwalton, line 173: I think it probably makes sense to defer everything in this file until after the page activates, and I think that'll make the code much easier to follow as well.

Perhaps something like this:

```js
const whenActivated = new Promise((resolve) => {
  if (document.prerendering) {
    document.addEventListener('prerenderingchange', () => resolve());
  } else {
    resolve();
  }
});
```

...and then you can create an async `initAnalytics` function that awaits that promise as the first thing it does and then sends the pageview and initializes everything else afterward:

```js
async function initAnalytics() {
  await whenActivated;

  // 1) Set all custom dimensions...

  // 2) Then send the pageview 

  // 3) Set up pageshow listener for bfcache pageviews...

  // 4) Register web vitals functions...
}

initAnalytics();
```

src/lib/analytics.js
```diff
@@ -138,25 +138,64 @@ store.subscribe(({isSignedIn}) => {
...
+        ga('set', dimensions.NAVIGATION_TYPE, 'prerender');
+        ga('send', 'pageview');
+      },
+      {once: true},
+    );
+    return;
+  }
+
+  // Handle normal page views
+  try {
```
philipwalton, line 186: This try/catch was added in the template to handle cases where `performance.getEntriesByType()` may not be supported by the current browser. If you use the logic to determine navigation type I suggest above, then a try/catch is probably not necessary.

philipwalton (changes requested):

src/lib/analytics.js
```diff
@@ -138,25 +138,64 @@ store.subscribe(({isSignedIn}) => {
...
+    (e) => {
+      if (e.persisted) {
+        ga('set', dimensions.NAVIGATION_TYPE, 'back-forward-cache');
+        ga('send', 'pageview');
+      }
+    },
+  );
+
+  // Handle prerenders that have already activated
+  if (
```
philipwalton, line 162: I'd extract all the initial navigation type logic into a single function. Something like:

```js
function getNavigationType() {
  if (document.wasDiscarded) {
    return 'restore';
  }

  const navEntry = self.performance &&
      performance.getEntriesByType &&
      performance.getEntriesByType('navigation')[0];

  if (navEntry) {
    if (navEntry.activationStart > 0) {
      return 'prerender';
    } else {
      return navEntry.type.replace(/_/, '-')
    }
  }
  return '(not set)';
}
```

Then you can do something like this, and keep all your custom dimension setting code together int the logic flow.

```js
ga('set', dimensions.NAVIGATION_TYPE, getNavigationType());
```

src/site/_includes/partials/analytics.njk
tunetheweb: Yeah I went back and forth on this one myself. Eventually, as it was simple I left it here just in case any other event (that may run before analytics.js ran) needed it. will move.

philipwalton (approved):

philipwalton:
@matthiasrohmer these changes LGTM but since this affects most pages on the site I wanted to give you or @devnook a chance to take a look before we merge.

tunetheweb:
Yeah that one confused me too for a bit, but eventually came to same conclusion as you. It kind of makes sense as it's in an iframe, and the owning page is still firing analytics, but still a bit confusing. And there's some events in the JS that aren't being logged because of this too.

There's also a couple of events [this one](https://github.com/GoogleChrome/web.dev/blob/8c6f90464bbd036ec7ee1ac8f7449d69cbb6b451/src/lib/components/AssessmentQuestion/index.js#L91-L97) and [this one](https://github.com/GoogleChrome/web.dev/blob/8c6f90464bbd036ec7ee1ac8f7449d69cbb6b451/src/lib/components/CourseLinks/index.js#L118-L125) that can fire outside of analytics.js, hence why I left the main analytics code initialisation in the template, and only moved page views to the JavaScript file. Surprised there weren't more to be honest!

philipwalton:
> There's also a couple of events [this one](https://github.com/GoogleChrome/web.dev/blob/8c6f90464bbd036ec7ee1ac8f7449d69cbb6b451/src/lib/components/AssessmentQuestion/index.js#L91-L97) and [this one](https://github.com/GoogleChrome/web.dev/blob/8c6f90464bbd036ec7ee1ac8f7449d69cbb6b451/src/lib/components/CourseLinks/index.js#L118-L125) that can fire outside of analytics.js, hence why I left the main analytics code initialisation in the template, and only moved page views to the JavaScript file.

Yeah, I think, when we upgrade to GA4, we should rethink how all of this is done. IMO we should have a single `analytics.js` module that exports a function that all other modules can use to report events. I don't think we should rely on the global `ga()` or `gtag()` functions as then it's too easy to have your analytics get out of sync. (But that's for a future PR.)

tunetheweb:
Nice thing about ga is it just reads off the `windows.ga` object, which is easy enough to set up inline in the template with minimal effort. A function adds loading order dependencies (unless we do similar).

The `ga` global is also quite a nice way to see the history of analytics events sent:

image

But anyway, we're getting off topic...

tunetheweb:
@devnook @matthiasrohmer any concerns here? Keen to get this merged so I can look at adding some prerender to our pages as samples for Chrome 108

devnook (approved):
Thanks @tunetheweb and @philipwalton !

PR #9009, opened 2022-11-05
API docs migration 2

AaronForinton:
Aligns with https://github.com/GoogleChrome/developer.chrome.com/pull/4210


redirects.yaml
```diff
@@ -339,6 +339,12 @@ redirects:
     destination: https://developer.chrome.com/articles/periodic-background-sync/
   - source: /eyedropper
     destination: https://developer.chrome.com/articles/eyedropper/
+  - source: /web-share-target
+    destination: https://developer.chrome.com/articles/web-share-target/
+  - source: /local-fonts
+    destination: https://developer.chrome.com/articles/web-share-target/
+  - source: /notification-triggers
+    destination: https://developer.chrome.com/docs/web-platform/notification-triggers/
```
sofiayem, line 347: As I can see this article has translations but the destination doesn't have links to translated versions. Or how does locale selection work on dcc?

sofiayem (changes requested):

redirects.yaml
AaronForinton: I think this covers the translated versions since it's so high up in the directory

rachelandrew (approved):

PR #9019, opened 2022-11-07
Fixes build on windows

johnrich85:

- Replaces hard coded path separator with path.sep so that the build works on Windows.


matthiasrohmer (approved):
Awesome, thanks! We potentially have a bunch of errors like this, really glad someone with a fitting environment is now here to fix them 🙂

PR #9020, opened 2022-11-07
Consolidation of /measure and PSI

### LINKED ISSUE #8852: Consolidation of /measure and PSI
So far the difference between PSI and web.dev/measure were the additional categories that /measure was presenting and PSI did not. However, we will soon launch a new version of PSI where all the differences are removed. This creates the following problems:

- there will be ambiguity between the 2 tools (why there are 2? what's the difference?)
- the current FAQ on /measure becomes incorrect and misleading
- we need to maintain /measure with no obvious added value to the user

The solution we propose is as follows:

- Remove the PSI banner from the homepage of [web.dev](http://web.dev/)
- Remove the Url input tool from the /measure page and replace it with a static page
- Use said PSI banner as a hero banner on that static page
- Add a paragraph of text explaining why the tool was removed and advantages of tools consolidation.

johnrich85:
Fixes #8852

### DIFF
--- a/src/site/content/en/measure/index.njk
+++ b/src/site/content/en/measure/index.njk
@@ -53,7 +53,7 @@ pageSpeed:
     </p>
 
     <p>
-      Going forward, PageSpeed Insights will be what you need. It will give you detailed lab insights
+      Going forward, <a href="https://pagespeed.web.dev/">PageSpeed Insights</a> will be what you need. It will give you detailed lab insights
       into your website's performance, but also field data insights, which is important as it allows
       you to assess your website's user experience from the perspective of your website's actual
       users.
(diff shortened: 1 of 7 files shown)


src/site/content/en/measure/index.njk
```diff
@@ -6,92 +6,58 @@ description: |
...
+
+<section class="bg-mid-bg measure__section gap-top-size-1">
+  <div class="wrapper all-center flow">
+    <h2 class="no-link w-headline--two w-font-weight--medium">Please note</h2>
 
-  <section class="flow all-center wrapper gap-top-size-4" data-size="narrow">
-    <h2 class="no-link w-headline--two w-font-weight--medium">
-      How your site is measured
-    </h2>
     <p class="color-mid-text">
```
devnook, line 49: I think that both paragarphs should have same text color as per design.

devnook (changes requested):
Thanks @johnrich85 ! Almost there.

devnook (approved):

PR #9025, opened 2022-11-07
Fix Additional Resources links

extra808:
Two of the links had Notion URLs instead of what I think are the intended destinations.

Fixes #SOME_ISSUE_NUMBER

-
-
-


rachelandrew (approved):
Thank you!

PR #9052, opened 2022-11-10
Add note on unload handlers on mobile to bfcache article

### LINKED ISSUE #9044: content: Update bfcache article to clarify `unload` ineligibility on mobile
https://web.dev/bfcache/#never-use-the-unload-event says:

> Chrome and Firefox have chosen to make pages ineligible for bfcache if they add an unload listener, which is less risky but also disqualifies a lot of pages.

AFAIK this isn't necessarily true. At least in Chrome, a page that uses `unload` may still be eligible for bfcache if the user is on mobile. It'd be good to clarify this behavior because I don't think this exception is documented anywhere.

cc @philipwalton @tunetheweb

tunetheweb:
Fixes #9044

- Clarifies that unload handler can still allow bfcache on mobile, but that's still not a reason to use it

### DIFF
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -234,7 +234,9 @@ On desktop, Chrome and Firefox have chosen to make pages ineligible for bfcache 
 listener, which is less risky but also disqualifies _a lot_ of pages. Safari
 will attempt to cache some pages with an `unload` event listener, but to reduce
 potential breakage it will not run the `unload` event when a user is navigating
-away, which makes the event very unreliable. Mobile browsers will attempt to cache some pages with an `unload` event listener but the `unload` event is extremely unreliable on mobile as processes are killed to save resources.
+away, which makes the event very unreliable.
+
+On mobile, most browsers will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile.
 
 Instead of using the `unload` event, use the `pagehide` event. The `pagehide`
 event fires in all cases where the `unload` event currently fires, and it
@@ -531,6 +533,8 @@ the bfcache, including:
 - when the user quits the browser and starts it again
 - when the user duplicates a tab
 - when the user closes a tab and uncloses it
+
+Even without those exclusions the bfcache will be discarded after a period to conserve memory.
 
 So, website owners should not be expecting a 100% bfcache hit ratio for all
 `back_forward` navigations. However, measuring their ratio can be useful to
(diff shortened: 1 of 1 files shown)


philipwalton (changes requested):
What about reframing it like this:

- Firefox has chosen to make pages ineligible...
- Safari will attempt to cache some pages...
- Chrome has chosen to match Firefox's behavior on desktop and Safari's behavior on mobile...

WDYT?

tunetheweb:
Hmmm I think that's potentially more confusing.

To me that phrasing begs these questions:
- What about Firefox on mobile? Imagine it gets little usage, but I thought all mobile browsers ignored unload for bfcache?
- Is Safari the same on mobile and desktop?
- Why has Chrome decided to do a little of both as that seems very confusing.

It also doesn't point out that unload handler is completely unrealiable on mobile. Maybe slightly out of scope for this doc, but another reason not to us em.

philipwalton:
> * What about Firefox on mobile? Imagine it gets little usage, but I thought all mobile browsers ignored unload for bfcache?

I believe Firefox behaves the same on desktop and mobile. Do you have an android device? If so you could test it.

> * Is Safari the same on mobile and desktop?

Last time I tested, yes it was.

> * Why has Chrome decided to do a little of both as that seems very confusing.

What I wrote in the post originally was true at the time. Chrome's initiate implementation was to match Safari, but they changed because they observed breakage on Desktop (Chrome has much more enterprise desktop usage than Safari).

This is another reason that  I'm hesitant to be too specific. The primary point is that browsers implement this differently, so it's best to avoid using the event altogether.

tunetheweb:
Tweaked further after discussion with @philipwalton . Could you take another look?

philipwalton (approved):

src/site/content/en/blog/bfcache/index.md
```diff
@@ -230,12 +230,14 @@ time](https://developer.chrome.com/blog/page-lifecycle-api/#the-unload-event)).
...
 can improve the user experience—but might also risk breaking the page.
 
-Chrome and Firefox have chosen to make pages ineligible for bfcache if they add an `unload`
+On desktop, Chrome and Firefox have chosen to make pages ineligible for bfcache if they add an `unload`
 listener, which is less risky but also disqualifies _a lot_ of pages. Safari
 will attempt to cache some pages with an `unload` event listener, but to reduce
 potential breakage it will not run the `unload` event when a user is navigating
 away, which makes the event very unreliable.
 
+On mobile, most browsers will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile.
```
rviscomi, line 239: Can we be more specific that Chrome is one of these browsers? Part of the motivation for #9044 was to be able to cite this article as a source for explaining the CLS changes in CrUX, so explicitly connecting it back to Chrome would be helpful.

rviscomi: The previous paragraph specifically calls out Chrome and Firefox for making pages ineligible, so it seems ok to clarify known exceptions where that's not the case.

Agree MDN should be the canonical source for exact browser behavior.

tunetheweb: Works for me.

rviscomi: Cool I'll tee up a new PR and add you both as reviewers

PR #9065, opened 2022-11-13
Make navigation visible when focus is in it

estelle:
Right now, if the user tabs thru the navigation using a keyboard, the sidebar stays hidden. For 20 clicks. This forces the navigation to be visible when the focus is that sidebar.


estelle:
@rachelandrew @AaronForinton @matthiasrohmer Should I close this?

matthiasrohmer (approved):
No, sorry @estelle this just slipped! That's a good improvement, thanks for doing that.

PR #9077, opened 2022-11-15
provide additional info re: styling checkboxes/radios

scottaohara:
This PR adds some missing nuance to the custom styling of checkboxes and radios, alludes to the fact that radios and checkboxes are more styleable now than they used to be, and adds more reference materials that I've written (and the already linked to resources mentioned) but were not directly referenced via this article


malchata (approved):
This all checks out. LGTM!

PR #9100, opened 2022-11-21
Add prerender2 speculation rules

### LINKED ISSUE #9099: Inline scripts are loaded twice
**Describe the bug**

Inline scripts are loaded twice on web.dev

**To Reproduce**
1. Go to https://web.dev/learn/html/
2. Open Console
3. Note it states:  "Uncaught SyntaxError: Identifier 'initialCourse' has already been declared (at (index):1:4614)"
4. View the source and search for `initialCourse={pages:[],percent:"0"}` and you'll see the script is in there twice.

image

In the code we can see the head.njk is loading the inline scripts AND the script-loader.njk:

https://github.com/GoogleChrome/web.dev/blob/60987df07c07f6231f668e8637c1825e5ea8957c/src/site/_includes/partials/head.njk#L48-L56

And script-loader.njk is loading the inline script again:

https://github.com/GoogleChrome/web.dev/blob/60987df07c07f6231f668e8637c1825e5ea8957c/src/site/_includes/partials/script-loader.njk#L1-L6

So can understand why it's included twice.

**Expected behavior**

They should be loaded once.

**Screenshots**

tunetheweb:

- Adds a new Prerender component to inject speculation rules and GA event so we can see how many prerender attempts were made
- Adds example prerender to http://web.dev/what-are-pwas/ to prerender http://127.0.0.1:8080/progressive-web-apps/
- Upgrades web-vitals to latest version (unrelated, but has a `restore` nav type that might influence things so would rather have this).
- Removes duplicated inlineScripts I spotted while working on this. Fixes #9099
- Fixes missing `/` from slug.


tunetheweb:
Tested this and seems to work well. Think it's good to merge.

FYI, to see if a page was pre-rendered (as our site is pretty fast already so sometimes difficult to tell), you can enter this in Console, after the page is loaded:

```
performance.getEntriesByType('navigation')[0].activationStart;
```

If it's > 0 then it was prerendered. Note prerendering doesn't happen for long lived Chrome sessions where memory usage is above a certain value so may need to restart Chrome to get it working.

src/site/_includes/components/Prerender.js
```diff
@@ -0,0 +1,70 @@
...
+    url.toString().startsWith('https://')
+  ) {
+    return '';
+  }
+
+  // Only prerender if browser supports speculationrules
+  // and not in SaveData mode (Chrome should exclude SaveData
+  // browser anyway, but no harm double checking)
+  return html`
+    <script type="module">
```
philipwalton, line 40: Why create an inline script here? To me it seems simpler to dynamically generate the `<script type="speculationrules">` object itself rather than dynamically generating a script that will then create and insert the `<script type="speculationrules">` element. (Plus it has the benefit of not creating an additional inline script that requires its own CSP entry.)

Also, what if a particular page wants to prerender (or prefetch) more than 1 URL?

src/site/_includes/partials/head.njk
```diff
@@ -45,13 +45,6 @@
 {# Include the GA object so we can start queuing events. #}
 {% include 'partials/analytics.njk' %}
 
-{% for item in inlineScripts %}
```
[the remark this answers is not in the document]
tunetheweb: Yes it was see this comment in PR:
> Removes duplicated inlineScripts I spotted while working on this. Fixes #9099

src/site/_includes/components/Prerender.js
tunetheweb: > Why create an inline script here?

We need to create an inline script anyway to send the GA event further down the code.

I figured by adding this dynamically it would allow us to check some things before adding it, currently this:

```
  // Only prerender if browser supports speculationrules
  // and not in SaveData mode (Chrome should exclude SaveData
  // browser anyway, but no harm double checking)
```

With this check:

```js
      if (
        HTMLScriptElement.supports &&
        HTMLScriptElement.supports('speculationrules') &&
        !navigator?.connection?.saveData
      ) {
```

There's no harm inserting the speculationrules JSON for unsupporting browsers and in theory Chrome should prevent this doing anything in SaveData mode, but don't want to send the GA event in either case, and figured better to do same check for both since I needed an inline script anyway.

> Also, what if a particular page wants to prerender (or prefetch) more than 1 URL?

This can be done in one of two ways:
1. Have multiple URLs in the speculationrules JSON
2. Have multiple speculationrules JSON

2 is easier when different parts of the page might inject this. We could support 1 in the Prerender.js , but it complicates it a little and not sure it will be used that much, so not sure it's worth it. WDYT?

tunetheweb: The other thing we could do is have a `data-prerender` attribute (or class) on `<a href ...>` elements and have a script run through and pick up all those URLs, dedup them and then add a single speculationrules JSON (and ga events) for them.

That's similar to what we do with `gc-analytics-event` currently.

It would also potentially allow more future prerendering options (e.g. hover or mouse down on a link).

WDYT?

src/site/_includes/partials/head.njk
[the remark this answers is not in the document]
philipwalton: Ahh, sorry, I really should read the rull PR description...

src/site/_includes/components/Prerender.js
philipwalton: > We need to create an inline script anyway to send the GA event further down the code.

There are other ways to do that. Our `analytics.js` module could look for `script[type=speculationrules]` elements and send the events itself. That would be my preference as it keeps all analytics code in one place.

For example, currently, if someone were to move the `partials/analytics.njk` include to the the foot of the page (rather than in the `<head>` like it is now), then the code you have here would silently break. Without any sort of test to ensure this code runs as expected, this is risky.

tunetheweb: Oh I quite like that. Will change to that.

tunetheweb: Only thing with that is that it will only work with speculation rules available when analytics.js runs (basically static speculation rules in the source HTML).

Probably OK, but if we start to dynamically inject them later (e.g. on link hover, or mouse down) then need to remember to track analytics separately.

src/lib/analytics.js
```diff
@@ -172,6 +172,49 @@ function getNavigationType() {
...
+     */
+    const prerenderURLs = new Set(
+      [...document.querySelectorAll('script[type="speculationrules"]')]
+        .map((script) => JSON.parse(script.innerHTML))
+        .map((json) => json.prerender)
+        .filter((rules) => rules.map((rule) => rule.source === 'list'))
+        .map((rules) => rules.map((rule) => rule.urls))
+        .flat()
+        .flat(),
+    );
```
tunetheweb, line 200: Is there any better way of writing this?

Basically want to extract all URLs for multiple instances of this:

```html
      <script type="speculationrules">
        {
          "prerender": [
            {
              "source": "list",
             "urls": ["url1", "url2"]
            }
          ],
        }
      </script>
```

philipwalton: I'd add a helper function that gets the full list of `prerender` rules. I tested this one out and it seems to work well:

```js
function getPrerenderRules() {
  return [...document.querySelectorAll('script[type=speculationrules]')]
    .map((s) => {
      try {
        return JSON.parse(s.textContent).prerender;
      } catch {
        // Ignore parse errors.
      }
    })
    .flat()
    .filter((rule) => rule && rule.source === 'list');
}
```

Then in your `logPrerenders()` function you could further flatted to get the list of URLs (note this uses `Set()` to remove duplicates):

```js
new Set(getPrerenderRules().map((r) => r.urls).flat())
```

src/site/_includes/components/Prerender.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Prerender by adding Speculation Rules JSON
+ * Also adds a GA event so can see how many prerender attempts were made
+ * versus prerender navigations.
+ * @param {string} url The url to prerender
+ * @returns {string}
+ */
+module.exports = (url) => {
+  // Only prerender internal URLs
```
philipwalton, line 27: Note that this logic assumes we never fully write out same-origin URLs.

I'm pretty sure that's true throughout the site, but I'm not sure there's anything enforcing it, so this is relying on us following a convention—and if you're going to do that you could just as easily follow a convention to not prerender cross-origin URLs.

(I.e. I'm not sure how much value this logic is adding.)

philipwalton: If you really wanted to check this, a more robust way would be something like this:


```js
const ORIGIN = 'https://web.dev';

// ...

if (new URL(url, ORIGIN).origin !== ORIGIN) {
  return '';
}
```

tunetheweb: Yeah I thought about that but there is a linting rule enforcing this.

And adding logic to check the domain matches (especially for local dev and staging URLs is more complicated than it's worth IMHO.

tunetheweb: OK added that code. Thanks.

src/lib/analytics.js
tunetheweb: Great - thanks. Added.

src/lib/analytics.js
```diff
@@ -213,6 +245,8 @@ async function initAnalytics() {
   onINP(sendToGoogleAnalytics);
   onLCP(sendToGoogleAnalytics);
   onTTFB(sendToGoogleAnalytics);
+
+  logPrerenders();
```
philipwalton, line 249: You previously had some code here to feature-detect `speculationrules` support. I think you want to add that back, no? (Either here or inside the `logPrerenders()` function).

src/lib/analytics.js
```diff
@@ -172,6 +172,38 @@ function getNavigationType() {
...
+  const prerenderURLs = new Set(
+    getPrerenderRules()
+      .map((r) => r.urls)
+      .flat(),
+  );
+
+  prerenderURLs.forEach((prerenderURL) => {
+    ga('send', 'event', {
+      eventCategory: 'Site-Wide Custom Events',
+      eventAction: 'Prerender attempt',
```
philipwalton, line 198:
```suggestion
      eventAction: 'prerender_attempt',
```

Eventually this is going to switch to GA4, so I think it makes sense to align to the conventions that system uses. See: https://support.google.com/analytics/answer/9234069

philipwalton (changes requested):

src/site/_includes/components/Prerender.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Prerender by adding Speculation Rules JSON
+ * Also adds a GA event so can see how many prerender attempts were made
+ * versus prerender navigations.
+ * @param {string} url The url to prerender
+ * @returns {string}
+ */
+module.exports = (url) => {
+  const ORIGIN = 'https://web.dev';
```
philipwalton, line 27: Actually, rather than defining this here, you should probably import this from the site data. e.g.

```js
const site = require('../../_data/site');

// ...

  if (!url || new URL(url, site.url).origin !== site.url) {
    return '';
  }
```

philipwalton: Last thought: rather than returning `''`, maybe we should throw an error instead. e.g.:

```js
const {html, oneLine} = require('common-tags');
const site = require('../../_data/site');

// ...

  if (!url || new URL(url, site.url).origin !== site.url) {
    throw new Error(oneLine`
      Cannot create prerender rule for URL: "${url}",
      only same-origin prerendering is supported at this time.
    `);
  }
```

I think throwing a build error is better than silently failing (where it may not be obvious why the shortcode didn't work).

src/lib/analytics.js
tunetheweb: Added back in.

src/site/_includes/components/Prerender.js
tunetheweb: Done, but I wonder if this will ever cause problems? For example, we currently prerender the first /blog/ article and next course page. What if they are ever external links? In that case the build will fail. I don't _think_ they ever should be, but if we roll this out more, we may start to see that so one to keep an eye on.

devnook (approved):

PR #9102, opened 2022-11-21
Compat widget implementation title change

AaronForinton:


AaronForinton:
@rachelandrew I've removed one of the widgets because it's mobile browsers only, and as the implementation stands this is showing as four red crosses. I think this will give the impression the API is completely unsupported and so removing it is the best way to go

rachelandrew (approved):

PR #9105, opened 2022-11-22
Fix deadline updates logic after yield

zuoaoyuan:
Deadline was extended by 50ms from the original timestamp before yielding, it puts us under the risk of yielding twice without get any of our task run if the other main thread works we yielded to occupied more than 50ms.
This fix gives us a brand new 50ms deadline every time we got resumed.


malchata (approved):
LGTM!

PR #9110, opened 2022-11-22
Log 404 pages in Google Analytics

tunetheweb:

- Add a 404 event to make it easier to see 404 pages.

You can see these by searching for Page Title of "404" but can't then see referrer. Having as an event is easier.


philipwalton:
I talked with @tunetheweb about this offline, but I think it probably makes sense to close this PR given the changes in #9168.

Also, as Barry points out, you can already get this data if you build a report that filters for title == "404", so no additional events are needed.

PR #9116, opened 2022-11-24
Rename and redirect native URLs

jecfish:
@rachelandrew how can I update the WebFu redirect urls, I am not able to push the changes because the project status is archived. https://github.com/google/WebFundamentals/blob/main/src/content/en/fundamentals/_redirects.yaml


tomayac (approved):
LGTM

Issue #7435, opened 2022-03-03
Request for a new "Update" aside.


malchata:
**Is your feature request related to a problem? Please describe.**
web.dev currently has a number of usual variations of the [aside component](https://web.dev/handbook/web-dev-components/#asides). Asides are a useful way for writers to interrupt the main flow of an article to call out some side note, pitfall, important note, and so on.

Currently, we lack an "update" variation on the aside component, which would be useful to use in select cases where updates concerning a developing situation around a certain API or metric can be effectively communicated. For example, we are currently using an "important" aside at this top of [this post](https://web.dev/responsiveness/) when an "update" aside would have been more appropriate.

**Describe the solution you'd like**
Asides that use a variation different than the default use a string to denote the variation type (for example, `'important'` or `'caution'`). An update aside should be something like this:

```
{% Aside 'update' %}
This unsolved mystery has new information. Check out this new video from Robert Stack to learn more.
{% endAside %}
```

cc: @philipwalton

heyawhite:
I've made a request about examples as well: https://github.com/GoogleChrome/developer.chrome.com/issues/2170

mihajlija:
+1 for the need for some sort of Updates component, especially on dcc as we are hosting things that are under active development there.

mihajlija:
Not stale

matthiasrohmer:
Resolved by https://github.com/GoogleChrome/web.dev/pull/9119.

heyawhite:
Note of availability / design in the docs: https://web.dev/handbook/web-dev-components/#update-asides

PR #9119, opened 2022-11-25
feat: add an update variation on the aside component

mamieorine:
Feature: [#7435](https://github.com/GoogleChrome/web.dev/issues/7435)

- Add an "update" variation on the aside component

**Note:** The translations of the title ("update") in other languages have been done by Google Translate.


src/site/_includes/components/Aside.js
```diff
@@ -119,6 +119,12 @@ function Aside(content, type = 'note') {
       icon = 'celebration.svg';
       title = i18n(`i18n.common.${type}`, locale);
       break;
+
+    case 'update':
+      icon = 'update.svg';
+      title = i18n(`i18n.common.${type}`, locale);
+      utilities.main = 'bg-primary-box-bg color-primary-box-text';
```
devnook, line 126: I wonder if tying the color of this aside to "primary" color of the site is a good idea here? Let's say we update site's theme and now "primary" color was red, would we  still want this type of aside to be red? @tokyodanj wdyt?

mamieorine: I created new variables for update aside theme named `bg-state-update-bg` and `color-state-update-text`.

mamieorine:
@devnook I created a new variable color for a new "Update" aside. Please help review this PR again, thank you 😊

src/styles/settings/_colors.scss
```diff
@@ -238,6 +238,9 @@ $WEB_EXTENDED_SECONDARY_COLOR: #e51661;
 $WEB_EXTENDED_TERTIARY_COLOR: #6001ff;
 $WEB_VITALS_COLOR: #2962ff;
 
+// ASIDE
+$UPDATE_ASIDE_COLOR: $WEB_PRIMARY_COLOR_600;
```
devnook, line 242: If you tie this variable to $WEB_PRIMARY_COLOR_600, then there's no point in adding additional variable anymore. In your solution, if WEB_PRIMARY_COLOR changes, so does the Aside color (and we want to avoid this). Please copy the # color code instead.

devnook (changes requested):
Aside is still tied to the primary color var.

mamieorine:
@devnook I have fixed this by using the HEX color code instead of primary color var.

devnook (approved):

PR #9131, opened 2022-11-28
Clarify browser behavior in bfcache post

### LINKED ISSUE #9044: content: Update bfcache article to clarify `unload` ineligibility on mobile

rviscomi:
Closes #9044 

Clarifies the mobile browser behavior related to `unload` listeners. This PR builds on the changes in #9052.

### DIFF
--- a/src/site/content/en/blog/bfcache/index.md
+++ b/src/site/content/en/blog/bfcache/index.md
@@ -236,7 +236,7 @@ will attempt to cache some pages with an `unload` event listener, but to reduce
 potential breakage it will not run the `unload` event when a user is navigating
 away, which makes the event very unreliable.
 
-On mobile, Chrome and Safari will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile.
+On mobile, Chrome and Safari will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile. Firefox treats pages that use `unload` as ineligible for the bfcache, except on iOS, which requires all browsers to use the WebKit rendering engine, and so it behaves like Safari.
 
 Instead of using the `unload` event, use the `pagehide` event. The `pagehide`
 event fires in all cases where the `unload` event currently fires, and it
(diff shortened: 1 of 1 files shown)


tunetheweb (approved):

src/site/content/en/blog/bfcache/index.md
```diff
@@ -236,7 +236,7 @@ will attempt to cache some pages with an `unload` event listener, but to reduce
 potential breakage it will not run the `unload` event when a user is navigating
 away, which makes the event very unreliable.
 
-On mobile, most browsers will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile.
+On mobile, Chrome and Safari will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile.
```
philipwalton, line 239: The previous paragraph outlines the behavior of all three major browser engines, so I think if we're going to be specific about Chrome and Safari on mobile then we should also be specific about Firefox on mobile.

philipwalton: I'm not sure about the last phrase: "except on iOS, where it behaves like Safari."

This makes it sound like Firefox is mimicking the Safari behavior, when what's actually happening is it's using the webkit rendering engine.

I think we should either:

- Drop the last phrase (since I think it's somewhat well known that all browsers on iOS are actually using webkit?)
- Rephrase it to: "except on iOS, which requires all browsers to use the WebKit rendering engine".

tunetheweb: Yeah I had similar thoughts.

> since I think it's somewhat well known that all browsers on iOS are actually using webkit?

You'd be surprised!! I was out with a senior web developer (that used to work for me!) and found out he didn't know this.

My vote would be for:

```suggestion
On mobile, Chrome and Safari will attempt to cache pages with an `unload` event listener since the risk of breakage is lower due to the fact that the `unload` event has always been extremely unreliable on mobile. Firefox treats pages that use `unload` as ineligible for the bfcache, except on iOS, which requires all browsers to use the WebKit rendering engine", and so it behaves like Safari.
```

philipwalton: LGTM

philipwalton (approved):

PR #9138, opened 2022-11-29
Amends the prefetching guide to include information about core web vi…

malchata:

- Updates the [link prefetching guide](https://web.dev/link-prefetch/) in various places to reflect effects of prefetching on various metrics (including Core Web Vitals).
- Minor edits for code style, and removal of content that may not be necessary anymore.

### DIFF
--- a/src/site/content/en/blog/link-prefetch/index.md
+++ b/src/site/content/en/blog/link-prefetch/index.md
@@ -47,7 +47,7 @@ For example, in a product listing page, you can prefetch the page for the most p
 eBay implemented prefetching for the first five results on a search page to speed up future pages loads and saw a positive impact on conversion rates.
 {% endAside %}
 
-Prefetching entire pages—while costly in terms of bandwidth—can greatly improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be much lower, any other subsequent time-based metrics will often be much faster, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
+While prefetching resources does use additional bandwidth, it can greatly improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be much lower, any other subsequent time-based metrics will often be much faster, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
 
 ### Prefetching static assets
 
--- a/src/site/content/en/blog/link-prefetch/index.md
+++ b/src/site/content/en/blog/link-prefetch/index.md
@@ -47,7 +47,7 @@ For example, in a product listing page, you can prefetch the page for the most p
 eBay implemented prefetching for the first five results on a search page to speed up future pages loads and saw a positive impact on conversion rates.
 {% endAside %}
 
-While prefetching resources does use additional bandwidth, it can greatly improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be much lower, any other subsequent time-based metrics will often be much faster, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
+While prefetching resources does use additional bandwidth, it can improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be lower, subsequent time-based metrics will often be lower as well, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
 
 ### Prefetching static assets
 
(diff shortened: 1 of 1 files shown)


src/site/content/en/blog/link-prefetch/index.md
```diff
@@ -45,6 +47,10 @@ For example, in a product listing page, you can prefetch the page for the most p
 eBay implemented prefetching for the first five results on a search page to speed up future pages loads and saw a positive impact on conversion rates.
 {% endAside %}
 
+Prefetching entire pages—while costly in terms of bandwidth—can greatly improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be much lower, any other subsequent time-based metrics will often be much faster, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
```
tunetheweb, line 50: I'm not sure about the wording "entire pages". `<link rel="prefetch">` only fetches the resource requested (e.g. the document, or a subresource) not the entire page. `<link rel="prerender">` currently fetches a page and all its sub resources, but that's changing, is less well supported, and is out of scope for this article.

Also "while costly in terms of bandwidth" is a bit drastic.

Maybe something like this:

```suggestion
While prefetching resources does use additional bandwidth, it can greatly improve most performance metrics. [Time to First Byte (TTFB)](/ttfb/) will often be much lower, as the document request results in a cache hit. Because TTFB will be much lower, any other subsequent time-based metrics will often be much faster, including [Largest Contentful Paint (LCP)](/lcp/) and [First Contentful Paint (FCP)](/fcp/).
```

src/site/content/en/blog/link-prefetch/index.md
```diff
@@ -45,6 +47,10 @@ For example, in a product listing page, you can prefetch the page for the most p
...
+
+[Cumulative Layout Shift (CLS)](/cls/) can also see signficiant improvements, especially in cases where web fonts are necessary to render text. Because the web fonts used by the page will be prefetched along with the HTML document, any font-related layout shifts will be eliminated, as the fonts responsible for the shift won't be network-bound, and will load instantly.
```
tunetheweb, line 52: How often is it the case that a page will use different fonts than the calling page? In my experience that's pretty rare, so not sure this is a great example. So I'd probably cut this line to be honest, if there isn't a better example.

```suggestion
```

src/site/content/en/blog/link-prefetch/index.md
```diff
@@ -55,6 +61,13 @@ For example, Netflix takes advantage of the time users spend on logged-out pages
...
 {% endAside %}
 
+The effect of prefetching static assets on performance metrics depends on the resource being prefetched:
+
+- Prefetching web fonts can eliminate layout shifts. In cases where [`font-display: swap;` is used](/font-display/), the swap period for the font is eliminated, resulting in faster rendering of the text and eliminating layout shifts. If a future page uses the prefetched font and the page's LCP element is a block of text using a web font, LCP for that element will also be faster.
```
tunetheweb, line 66: Again I'd probably drop the font example, but maybe it's OK to leave here for those (rare IMHO) occasions where this is the case?

src/site/content/en/blog/link-prefetch/index.md
```diff
@@ -129,10 +145,6 @@ You can also implement smarter prefetching with libraries that use `prefetch` un
 
 Both quicklink and Guess.js use the [Network Information API](https://developer.mozilla.org/docs/Web/API/Network_Information_API) to avoid prefetching if a user is on a slow network or has [`Save-Data`](https://developer.mozilla.org/docs/Web/HTTP/Headers/Save-Data) turned on.
 
-{% Aside %}
-A wine company Jabong implemented prefetching with quicklink and achieved 2.7 s faster Time To Interactive on future pages.
-{% endAside %}
-
```
tunetheweb, line 135: Why was this removed out of interest? Cause it's not backed up with a link?

tunetheweb (changes requested):

src/site/content/en/blog/link-prefetch/index.md
```diff
@@ -55,6 +59,12 @@ For example, Netflix takes advantage of the time users spend on logged-out pages
...
 {% endAside %}
 
+The effect of prefetching static assets on performance metrics depends on the resource being prefetched:
+
+- Prefetching images can significantly lower LCP times for LCP image elements.
+- Prefetching stylesheets can improve both FCP and LCP, as the network time to download the stylesheet will be eliminated. Since stylesheets are render blocking, they can reduce LCP when prefetched. In cases where the subsequent page's LCP element is a CSS background image requested via the `background-image` property, the image will also be prefetched as a dependent resource of the prefetched stylesheet.
+- Prefetching JavaScript will allow the processing of the prefetched script to occur much sooner than if it were required to be fetched by the network first during navigation. This can have an effect on responsiveness metrics such as [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/). In cases where markup is rendered on the client via JavaScript, LCP can be improved through reduced resource load delays, and client-side rendering of markup containing a page's LCP element can occur sooner.
```
tunetheweb, line 66: I've had a change of heart here so think we should add it back in (but maybe not as top one):

```suggestion
- Prefetching JavaScript will allow the processing of the prefetched script to occur much sooner than if it were required to be fetched by the network first during navigation. This can have an effect on responsiveness metrics such as [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/). In cases where markup is rendered on the client via JavaScript, LCP can be improved through reduced resource load delays, and client-side rendering of markup containing a page's LCP element can occur sooner.
- Prefetching web fonts that are not already used by the current page, can eliminate layout shifts. In cases where [`font-display: swap;` is used](/font-display/), the swap period for the font is eliminated, resulting in faster rendering of the text and eliminating layout shifts. If a future page uses the prefetched font and the page's LCP element is a block of text using a web font, LCP for that element will also be faster.
```

tunetheweb (approved):

PR #9148, opened 2022-12-01
[lighthouse] redirect perf content to d.c.c

brendankenny:

- adds redirects to migrated perf content on d.c.c
- deletes web.dev lighthouse-performance infrastructure


rachelandrew:
@AaronForinton could you take a look at this epic set of redirects? This is to finish the process of moving some docs to d.c.c. they were moved a while ago but the redirects and final tidying up hadn't happened.

This is also relevant because there may well be old issues raised about these docs on web.dev, which can be closed once the content isn't here any more.

malchata (approved):
LGTM.

PR #9149, opened 2022-12-01
[lighthouse] redirect best practices content to d.c.c

brendankenny:
Companion to #9148

~Currently based on top of that PR to reduce conflicts, but I can rebase on main once that PR lands.~ Updated!

- adds redirects to migrated best practices content on d.c.c


brendankenny:
@AaronForinton is it ok if I assign the rest of these to you as well? Two more after this, but these are all much smaller than #9148

test/unit/src/site/_filters/getPaths.js
```diff
@@ -34,9 +34,9 @@ describe('getPaths', function () {
 
   it('returns all path names the post belongs to if more than 1', function () {
     const post = {
-      fileSlug: 'is-on-https',
+      fileSlug: 'lcp',
     };
-    const expected = ['lighthouse-best-practices', 'lighthouse-pwa'];
```
brendankenny, line 39: note the test change, but I just found a different article in multiple paths to replace it. Seems to work?

AaronForinton (approved):

PR #9153, opened 2022-12-02
Fix prerender breaking JSON by monitoring it

tunetheweb:

- Fix issue broken in #9100 

TIL you can’t run a JS minifier on JSON!

And also that we don’t run minifier on staging 😔


malchata (approved):
LGTM!

PR #9155, opened 2022-12-02
[lighthouse] redirect SEO content to d.c.c

brendankenny:
Next step after #9149

- adds redirects to migrated SEO content on d.c.c


AaronForinton (approved):
LGTM

PR #9161, opened 2022-12-02
Update index.md

gunayanil:
If an object contains functions, then the `structuredClone` API actually throws a `DataCloneError` exception error, so functions are not quietly discarded. It might also be correct to say that "API attemps to throw a `DataCloneError` exception`.

Fixes #SOME_ISSUE_NUMBER

- Content update: How `structuredClone` deals with functions inside an object.


gunayanil:
Any update?

gunayanil:
Any update?

rachelandrew (approved):

PR #9168, opened 2022-12-04
Add initial GA4 implementation

philipwalton:
This PR switches web.dev from using the deprecated `analytics.js` library to the newer `gtag.js` library. It keeps our existing analytics implementation largely the same, but instead of just sending data to a single analytics account (UA-126406676-2) it also send data to a new GA4 account (G-18JR3Q8PJ8).

Sending data to both accounts is temporary until Universal Analytics is [officially turned down](https://support.google.com/analytics/answer/11583528) in July 2023, at which point we can remove the line that adds the config for UA-126406676-2, which will be the only change required at that point.

While the UA configuration is largely unchanged from what it does today, there are a couple notable changes to the code as well as the GA4 implementation worth mentioning:

**1) Adding author, tag, and learning path data.**

Three new `<meta>` tags have been added to every generated page to indicate the authors, tags, or learn paths for this page. For example, the `/why-speed-matters/` page contains the following:

```html
<meta name="authors" content="ansteychris,bojanpavic,jlwagner" />
<meta name="paths" content="fast" />
<meta name="tags" content="performance" />
```

The analytics code then looks for these tags and adds custom event parameters to capture this information if found. This will allow authors to more easily see how many pageviews they're receiving for all of their articles or for tag/path owners to see how many total pageviews each tag/path gets (cc: @samdutton).

_Note: I'm not sure if we need both "paths" and "tags", but it did seem like some of the paths weren't fully covered by tag, so it seems to make sense to include both._

**2) Logging analytics events to the console on localhost**

When not in production mode, the global `gtag` function is assigned to `console.debug()`, which will make it easier for people to see what analytics events are being logged, and it will also mean we don't send data to Google Analytics while developing locally in most cases. Also, by using `console.debug()` it will not show up by default so it won't add a lot of unnecessary noise to people's consoles. However, when DevTools is configured to show debug logs, it'll look like this:

Screen Shot 2022-12-04 at 12 38 20 PM

**3) Update other modules to not use `ga()`/`gtag()` functions directly**

The `analytics.js` module has always exported function to send event and exception data to Google Analytics, and some code used those functions and some code called the underlying functions directly.

In this PR, all code has been updated to use the functions exported by the `analytics.js` module, which is important because the `analytics.js` module runs the initialization code first, which is needed to ensure the author, tag, learn path, etc. data is added to all events sent to Google Analytics.

**4) Removing the analytics partial from pattern pages**

The `analytics.njk` partial include has been removed from all of the patterns pages that included it. Since those pages were not loading the library from Google Analytics, they were not actually sending any data to Google Analytics, so removing this partial doesn't change what data is being captured.

If we decide that we _do_ want to capture analytics data for these pattern pages when we should probably create a single pattern layout template rather than individually including the `analytics.njk` partial (which has clearly been error-prone).

**5) Updating the language used in the code**

Usage of the verb "track" has been replaced with "log" to better convey what's actually happening. Since this analytics implementation is not tracking individual users, this terminology is more accurate. Note the entire industry has been moving away from the word "track" for analytics use cases unrelated to remarketing or ad targeting.


src/site/_includes/partials/analytics.njk
```diff
@@ -1,14 +1,11 @@
...
-ga('set', 'page', window.location.pathname);
-ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version}}');
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
+loadScript('https://www.googletagmanager.com/gtag/js?id={{ site.analytics.ids.ga4 }}', null);
+{% if (site.env != 'prod') %}
+  {# In development, log debug events to the console instead of sending them to GA #}
+  window.gtag = console.debug;
+{% endif %}
```
tunetheweb, line 9: Should we only load the script if in prod mode?

And is it worth noting that googletagmanager call will actually load both scripts (based on the config) and not just the ga4 scripts? that confused me initially.

And is there any risk that the googletagmanager script will override window.gtag after it loads asynchronously? It doesn't seem to but just wondering? Maybe another reason to only load that if needed?


```suggestion
{% if (site.env === 'prod') %}
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());
  {# Note we only need one call to GTM and it will load all the analytics from config object #}
  loadScript('https://www.googletagmanager.com/gtag/js?id={{ site.analytics.ids.ga4 }}', null);
{% else %}
  {# In development, log debug events to the console instead of sending them to GA #}
  window.gtag = console.debug;
{% endif %}
```

tunetheweb (commented):
This change looks good. I'm particularly pleased about the console.log functionality in non-prod!

Not entirely sure how feasible it is to remove direct calls to gtag and always go throgh `logEvent` if a particular page wants to send an event. How would you suggest we implement #9110 for example? And would you like to include that in this PR? 😁

samdutton:
>
> Three new <meta> tags have been added to every generated page to indicate
> the authors, tags, or learn paths for this page.


Nice work. Thank you — especially for this!

src/lib/components/CourseLinks/index.js
```diff
@@ -116,11 +118,10 @@ class CourseLinks extends HTMLElement {
...
     // Fire analytics if there's been a 10%+ bucket jump.
-    if (ga && newPercent - oldPercent >= 10) {
-      ga('send', 'event', {
-        eventCategory: 'Course Events',
-        eventAction: `course: ${courseKey} progress`,
-        eventLabel: newPercent.toString(),
+    if (newPercent - oldPercent >= 10) {
+      logEvent(`course: ${courseKey} progress`, {
+        event_category: 'Course Events',
+        event_label: newPercent.toString(),
```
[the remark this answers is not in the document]
philipwalton: Hmmm, this pattern of inlining some scripts makes this migration a lot more complex. The easiest option for this PR is probably to switch this back go using `gtag()` but I really don't like that as a long term solution.

The only good justification for inlining scripts is to ensure critical code can run as fast as possible. However, analytics reporting is clearly not critical code, so I'd argue this should be removed and added to the `src/lib/pages/course.js` script, which is bundled by rollup so it can include analytics exports.

Also, looking at the logic in the file, course progress is stored in `localStorage`, so it should be relatively easy for code in the `src/lib/pages/course.js` file to inspect `localStorage` and report progress.

I'll update this with a TODO.

/cc @devnook

src/lib/components/CourseLinks/index.js
```diff
@@ -22,6 +22,8 @@
  * it needs to read from localstorage and then modify its children.
  */
 
+import {logEvent} from '../../analytics';
+
```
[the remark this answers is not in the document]
philipwalton: Removed (see other comment below).

src/site/_includes/partials/analytics.njk
philipwalton: I'll move the loading of the script into the prod conditional in my next commit. We do, however, need to ensure the `gtag()` function is always defined since other scripts call it.

src/site/_includes/partials/analytics.njk
```diff
@@ -1,14 +1,12 @@
...
-ga('create', '{{ site.analytics.ids.prod }}');
-ga('set', 'transport', 'beacon');
-ga('set', 'anonymizeIp', true);
-ga('set', 'page', window.location.pathname);
-ga('set', '{{ site.analytics.dimensions.TRACKING_VERSION }}', '{{ site.analytics.version}}');
+window.dataLayer = window.dataLayer || [];
+function gtag(){dataLayer.push(arguments);}
+gtag('js', new Date());
+{% if (site.env != 'prod') %}
+  loadScript('https://www.googletagmanager.com/gtag/js?id={{ site.analytics.ids.ga4 }}', null);
```
philipwalton, line 6: @devnook FYI, in your cookie baner changes PR, loading this script should be gated on users having given consent. That seems like the easiest way to ensure no data is sent for users who haven't consented.

philipwalton: Also (sorry if this is obvious) but FYI loading the script doesn't need to happen here, it could happen at any point after the three lines above have run.

```js
window.dataLayer = window.dataLayer || [];
window.ga ||	function gtag(){dataLayer.push(arguments);}
function () {	gtag('js', new Date());
```

src/lib/analytics.js
```diff
@@ -6,152 +6,202 @@ import {
...
-    label,
-    value,
-  };
-}
+// A promise that settles once all analytics has been initialized.
+// Internally this assigned the `resolve()` function to the module-level
+// `markAnalyticsInitialized` variable.
+const whenAnalyticsInitialized = new Promise((resolve) => {
+  markAnalyticsInitialized = resolve;
+});
```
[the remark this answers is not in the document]
philipwalton: This is to ensure all custom dimensions have been set before any data is sent. If external code can call `gtag('event', ...)` before this code runs, we'll have data sent to GA that doesn't include all of the dimensions we care about.

src/lib/analytics.js
```diff
@@ -6,152 +6,202 @@ import {
   onLCP,
   onTTFB,
 } from 'web-vitals/attribution';
-import {dimensions} from 'webdev_analytics';
-import {store} from './store';
```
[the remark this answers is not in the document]
philipwalton: I'm not sure. @devnook would be the best person to answer that.

src/lib/analytics.js
```diff
@@ -259,6 +341,12 @@ async function initAnalytics() {
   onTTFB(sendToGoogleAnalytics);
 
   logPrerenders();
+
+  markAnalyticsInitialized();
```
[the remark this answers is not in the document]
philipwalton: Addressed in my other comment.

src/lib/components/CourseLinks/index.js
```diff
@@ -116,11 +116,16 @@ class CourseLinks extends HTMLElement {
...
-      ga('send', 'event', {
-        eventCategory: 'Course Events',
-        eventAction: `course: ${courseKey} progress`,
-        eventLabel: newPercent.toString(),
+    if (newPercent - oldPercent >= 10) {
+      // TODO: ideally, no code would directly call the `gtag()` function,
+      // instead, code should import functions from the analytics.js module
+      // so it can ensure everything is properly initialized before any
+      // data is sent to GA. Also, it doesn't make sense to include
+      // non-critical logic in a script that is inlined, so this should be
```
matthiasrohmer, line 124: Now tracked in https://github.com/GoogleChrome/web.dev/issues/9277.

src/site/content/en/third_party/chrometober-2022/chrometober.njk
```diff
@@ -11,8 +11,7 @@
 		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="">
 		<link href="https://fonts.googleapis.com/css2?family=Germania+One&amp;family=Nanum+Gothic:wght@400;700&amp;display=swap" rel="stylesheet">
 		<!-- <script is:inline type="module" src="https://flackr.github.io/scroll-timeline/dist/scroll-timeline.js"></script> -->
-	<link rel="stylesheet" href="/third_party/chrometober-2022/assets/index.9ed1d485.css">
-{% include 'partials/analytics.njk' %}
+  	<link rel="stylesheet" href="/third_party/chrometober-2022/assets/index.9ed1d485.css">
```
matthiasrohmer, line 14: As Chrometober doesn't use the main JavaScript, tracking will stop for this page.

src/site/_includes/partials/analytics.njk
```diff
@@ -1,12 +0,0 @@
-{% set analyticsScript %}
-window.dataLayer = window.dataLayer || [];
-function gtag(){dataLayer.push(arguments);}
```
philipwalton: We actually still need this file since some of the code references the global `gtag()` function (see #9277). Also, if/when we switch to GTM then we will need the global `dataLayer` variable, so I think it makes sense to keep everything but the `loadScript()` call here in this partial.

I'll add a commit with these changes.

src/site/content/en/third_party/chrometober-2022/chrometober.njk
philipwalton: Analytics tracking was actually never happening on this page. If you look at what the `analytics.njk` partial was doing (prior to my changes) it was just initializing some variables and not loading the Google Analytics script.

 If you visit the page itself you can verify that nothing is being loaded from or sent to `https://www.google-analytics.com`: https://web.dev/third_party/chrometober-2022/chrometober.html

I think the  fact that this went unnoticed further highlights the need for a central analytics implementation rather than individual pages including (or not including) a partial.

philipwalton:
@devnook I tweaked your changes a bit and also updated the cookie bar functionality to add the ability to decline cookies, which I believe is a requirement for us now. PTAL.

src/site/content/en/third_party/chrometober-2022/chrometober.njk
devnook: This is the directcion this PR takes - to move analytics init to the entrypoint app.js file instead of partial. Another thing is to make all pages use the templates consistently.

src/lib/store.js
```diff
@@ -49,7 +49,8 @@ const initialState = {
   // cookie policy.
   // We automatically accept cookies in dev and test environments so the cookie
   // banner doesn't interfere with tests.
-  userAcceptsCookies: !isProd,
+  // cookiePreference: isProd ? null : 'accepts',
```
devnook, line 52: I think we need the isProd check for visual test to be consistent.

philipwalton: Oops, had commented this out to test and forgot to add it back. Good catch! Fixed in https://github.com/GoogleChrome/web.dev/pull/9168/commits/c7174162afacc5198aa1345b2375775b03637fef.

devnook (approved):

PR #9200, opened 2022-12-08
Fix: fallback for upcoming src syntax

dpschen:
## Changes proposed in this pull request

Add a second `src` definition for the fallback instead of defining them inline.

## Reason:
Original solution didn't work. 'Example 22' in the [linked working draft](https://www.w3.org/TR/css-fonts-4/#font-face-src-parsing) also suggests using two definitions.


dpschen:
Sry, closed this by accident.
I think this is still a valid change.

rachelandrew (approved):

PR #9211, opened 2022-12-09
Async function article fixup

### LINKED ISSUE #9192: [async functions] "Avoid going too sequential" recommends bad practice
The [*Careful! Avoid going too sequential* section of the *Async functions: making promises friendly* article](https://web.dev/async-functions/#careful-avoid-going-too-sequential) has two examples:
```js
async function parallel() {
  const wait1 = wait(500); // Start a 500ms timer asynchronously…
  const wait2 = wait(500); // …meaning this timer happens in parallel.
  await wait1; // Wait 500ms for the first timer…
  await wait2; // …by which time this timer has already finished.
  return "done!";
}
```
and
```js
async function logInOrder(urls) {
  // fetch all the URLs in parallel
  const textPromises = urls.map(async url => {
    const response = await fetch(url);
    return response.text();
  });

  // log them in sequence
  for (const textPromise of textPromises) {
    console.log(await textPromise);
  }
}
```
These (and also the original `reduce`-based `logInOrder` implementation) are prone to causing unhandled promise rejections. See
https://stackoverflow.com/questions/45285129/any-difference-between-await-promise-all-and-multiple-await and https://stackoverflow.com/questions/46889290/waiting-for-more-than-one-concurrent-await-operation for details.

The solution for the first is simple - just use `Promise.all`:
```js
async function parallel() {
  const wait1 = wait(500); // Start a 500ms timer asynchronously…
  const wait2 = wait(500); // …meaning this timer happens in parallel.
  await Promise.all([wait1, wait2]); // Wait until both timers have finished (but throw if any of them rejects prematurely)
  return "done!";
}
```
The solution for the second is a bit more complicated if you want the logs both in order and as soon as possible. The simple approach
```js
async function logInOrder(urls) {
  // fetch all the URLs in parallel
  const results = await Promise.all(urls.map(async url => {
    const response = await fetch(url);
    return response.text();
  }));

  // log them afterwards, in sequence
  for (const result of results) {
    console.log(results);
  }
}
```
does log all the results in the end, and doesn't log any if there's an error (not unusual with fetch, even when [ignoring http errors](http://blog.niftysnippets.org/2018/06/common-fetch-errors.html)). To get the desired result and not cause unhandled rejections, you'd need to explicitly ignore the errors at first:
```js
async function logInOrder(urls) {
  // fetch all the URLs in parallel
  const textPromises = urls.map(async url => {
    const response = await fetch(url);
    if (!response.ok) throw new Error(response.statusText);
    return response.text();
  });
  // ignore rejections that may only be handled later
  for (const textPromise of textPromises) {
    textPromise.catch(() => {});
  }
  // log them in sequence
  for (const textPromise of textPromises) {
    console.log(await textPromise);
  }
}
```
…and furthermore you might actually want to wait until all responses are received, log all successful ones and then throw an `AggregateError` for the failures (if any). Not sure if this would go beyond the scope of the article though - but concurrency is hard, there's no excuse imo.

jakearchibald:
Fixes #9192

- Remove duplicate article
- Update formatting
- Add fix for unhandled promise rejections (#9192)
- Add note about `for await…of`


src/site/content/en/blog/async-functions/index.md
```diff
@@ -316,27 +318,61 @@ performs the fetches in parallel. Thankfully there's an ideal middle-ground.
...
 {% CompareCaption %}
 In this example, the URLs are fetched and read in parallel, but the "smart"
 `reduce` bit is replaced with a standard, boring, readable for-loop.
 {% endCompareCaption %}
 
 {% endCompare %}
 
+{% Aside 'important' %}
+`markHandled` is used to avoid "unhandled promise rejections". If a promise
+rejects, and it was never given a rejection handler (e.g. via
```
rachelandrew, line 351:
```suggestion
rejects, and it was never given a rejection handler (for example, via
```

rachelandrew (approved):
Just a tiny style guide nit.

src/site/content/en/blog/async-functions/index.md
```diff
@@ -316,27 +318,61 @@ performs the fetches in parallel. Thankfully there's an ideal middle-ground.
 {% Compare 'better', 'Recomm ended - nice and parallel' %}
```
jakearchibald, line 318:
```suggestion
{% Compare 'better', 'Recommended - nice and parallel' %}
```

Hah, I wonder when that happened

PR #9225, opened 2022-12-11
fix code in /learn/html/links

myakura:
there's a code assigning a string to `a.href` in https://web.dev/learn/html/links/#links-and-javascript ; however the string misses an opening quote.

Screenshot: part of learn/html/links where there's an error in sytax

Fixes #SOME_ISSUE_NUMBER

- fix code in example


rachelandrew (approved):
Thanks!

PR #9229, opened 2022-12-11
Style element is supposed to be interactive

estelle:
there is a style block that is supposed to be interactive and enable styling itself. If this is not allowed, https://codepen.io/estelle/pen/LYroGqK can be included as a codepen


AaronForinton (approved):
Thank you, gonna merge this

PR #9302, opened 2022-12-27
Fix incorrect title for WebWorker article

RuyiLi:
It appears that in the batch H5R migration done in #8036, the title for the WebSocket article was mistakingly used for the WebWorker article as well. This PR restores the title from the original H5R post.

- Fix the incorrect title in the following article: https://web.dev/workers-basics/


AaronForinton (approved):
Thank you. Approving now and I'll be merging this after the end of our content freeze on the 3rd Jan.

PR #9322, opened 2023-01-01
update legacy go get syntax, add note

gaycodegal:
web bundle files linked did not work on Android, but one i built myself works fine. Example web bundles will need to be updated

- `go get` no longer works outside of a module. thus, I used `go install` which `go get` recommends.
- web bundles linked no longer work on android, so i mentioned that in article


gaycodegal:
ah a chrome license agreement. i'm not signing an agreement to contribute a quick fix, feel free to duplicate this work yourself then

PR #9335, opened 2023-01-04
Stop the web.dev homepage LCP image being lazy loaded

tunetheweb:

- Currently the LCP image on web.dev home page is the second image, as the video doesn't count. The second image is lazy loaded, affecting our LCP :-( This PR fixes that.

### DIFF
--- a/src/site/_includes/partials/picked-case-study.njk
+++ b/src/site/_includes/partials/picked-case-study.njk
@@ -11,7 +11,7 @@
   {% if featureCard and featureCard.video %}
   {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background", fetchpriority="high"%}
   {% else %}
-  {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background" %}
+  {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background", loading="eager" %}
   {% endif %}
 </a>
 {% endif %}
--- a/src/site/_includes/partials/picked-case-study.njk
+++ b/src/site/_includes/partials/picked-case-study.njk
@@ -8,6 +8,10 @@
 <a class="feature-card" href="{{ pickedItem.url }}" data-treatment="bg-image">
   <span class="feature-card__eyebrow">Case study</span>
   <h3 class="feature-card__title">{{ pickedItem.data.title }}</h3>
+  {% if featureCard and featureCard.video %}
+  {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background", fetchpriority="high"%}
+  {% else %}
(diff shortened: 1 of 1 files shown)


src/site/_includes/partials/picked-case-study.njk
```diff
@@ -8,6 +8,10 @@
 <a class="feature-card" href="{{ pickedItem.url }}" data-treatment="bg-image">
   <span class="feature-card__eyebrow">Case study</span>
   <h3 class="feature-card__title">{{ pickedItem.data.title }}</h3>
+  {% if featureCard and featureCard.video %}
+  {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background", fetchpriority="high"%}
+  {% else %}
   {% Img src=pickedItem.data.hero, alt="", width="570", height="330", class="feature-card__background" %}
+  {% endif %}
```
tunetheweb, line 15: WDYT @philipwalton ? Is this complexity worth it? Or should we just add fetchpriority to this image regardless as it's an above the fold image on desktop anyway even when it's not LCP?

philipwalton: Yeah, it's tricky. On mobile that image is never the LCP image, so I'd lean toward just turning off `loading=lazy` rather than setting `fetchpriority=high`.

tunetheweb: But that's the same with the first image in right hand column on mobile where we DO set fetchpriority="high" regardless.

tunetheweb: Given it will be just below the viewport, probably should remove `loading="lazy"` off the second image ANYWAY as it should never really be lazy loaded. Will add that to this PR.

But I'd vote for still explicitly adding `fetchpriority="high"` in cases where the above image is a video. The mobile LCP image still gets a high fetchpriority as well anyway and think the small hit to that on mobile (due to any bandwidth contention) is worth the bigger gain on desktop when this is LCP.

philipwalton (approved):

MichealThabbet:
Thanks a lot for this e-mail . But really i am not mean anything although i need for me to look about small functions and i am not found it after a lot of time of my research i found it at google dev so i am sorry about it and if i need help can I contact with you ? Thanks a lot Sent from my iPhoneOn 4 Jan 2023, at 10:21 PM, Barry Pollard ***@***.***> wrote:﻿
Merged #9335 into main.

MichealThabbet:
Sent from my iPhoneOn 5 Jan 2023, at 1:17 AM, Mesho Thabet ***@***.***> wrote:﻿Thanks a lot for this e-mail . But really i am not mean anything although i need for me to look about small functions and i am not found it after a lot of time of my research i found it at google dev so i am sorry about it and if i need help can I contact with you ? Thanks a lot Sent from my iPhoneOn 4 Jan 2023, at 10:21 PM, Barry Pollard ***@***.***> wrote:﻿
Merged #9335 into main.

PR #9336, opened 2023-01-04
Adds a note on responsiveness thresholds and INP in the RAIL documentation.

malchata:

- The [RAIL documentation](https://web.dev/rail/) currently advises developers to respond to user inputs within 50 milliseconds, but this conflicts with the [INP thresholds](https://web.dev/inp/#whats-a-good-inp-value) which state that 200 milliseconds or less is fine. This PR adds an aside to clarify this.

### DIFF
--- a/src/site/content/en/fast/rail/index.md
+++ b/src/site/content/en/fast/rail/index.md
@@ -123,6 +123,10 @@ processing time:
   <figcaption>How idle tasks affect input response budget.</figcaption>
 </figure>
 
+{% Aside 'important' %}
+The 50 millisecond threshold for responding to interactions is great when conditions are ideal. For users on hardware with less memory and processing power, this threshold can be very difficult to achieve. This is why the [Interaction to Next Paint (INP)](/inp/) metric's "good" threshold is set to 200 milliseconds or less.
+{% endAside %}
+
 ## Animation: produce a frame in 10&nbsp;ms
 
 **Goals**:


tunetheweb (commented):
Happy to approve this change, but I wonder if we shouldn't just delete this article? It seems to be somewhat at odds with our Core Web Vitals initiative and so is confusing to have them both.

Or maybe we should leave it there for historical reasons (it's linked elsewhere) but add a note at the top to say that CWV is the recommended performance measurement method now over RAIL?

Any opinions here @philipwalton ?

tunetheweb:
Also if you are keeping it, could you update this line to point to PSI since /measure/ is on it's way out?:

https://github.com/GoogleChrome/web.dev/blob/c7cd93cbc67f234ad45b51368def0da6ba801694/src/site/content/en/fast/rail/index.md#L306

philipwalton:
> Or maybe we should leave it there for historical reasons (it's linked elsewhere) but add a note at the top to say that CWV is the recommended performance measurement method now over RAIL?

Yeah, I think doing this probably makes sense. The page is linked to from enough places that I don't think we want to remove it, but I do think we should point people to Core Web Vitals in a note at the top.

tunetheweb:
I had a look at what's involved and opened #9352 which I think can supersede this PR. Tagged you both as reviewers.

PR #9343, opened 2023-01-05
Slight tweaks to bring INP doc more in line with the other Web Vitals

tunetheweb:
The INP table of contents looks quite different to the others:

image

This changes it to be more in line:

image

- Change "good" to just good, to be consistent.
- Remove caveat about how it's hard to quantify - it is for all metrics but we don't feel the need to explain the rational for them in this doc (though the CWVs are documented in [Defining the Core Web Vitals metrics thresholds](https://web.dev/defining-core-web-vitals-thresholds/) and we should add to that if INP becomes one).
- Indent some headers to match other Web Vitals TOCs.

### DIFF
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -63,7 +63,7 @@ An interaction's latency consists of the single longest [duration](https://w3c.g
 For more details on how INP is measured, read the ["What's in an interaction?"](#whats-in-an-interaction) section.
 {% endAside %}
 
-### What's a good INP value?
+### What is a good INP score?
 
 Pinning labels such as "good" or "poor" on a responsiveness metric is difficult. On one hand, you want to encourage development practices that prioritize good responsiveness. On the other hand, you must account for the fact that there's considerable variability in the capabilities of devices people use to set achievable development expectations.
 
--- a/src/site/content/en/metrics/inp/index.md
+++ b/src/site/content/en/metrics/inp/index.md
@@ -65,6 +65,8 @@ For more details on how INP is measured, read the ["What's in an interaction?"](
 
 ### What's a good INP value?
 
+Pinning labels such as "good" or "poor" on a responsiveness metric is difficult. On one hand, you want to encourage development practices that prioritize good responsiveness. On the other hand, you must account for the fact that there's considerable variability in the capabilities of devices people use to set achievable development expectations.
+
 To ensure you're delivering user experiences with good responsiveness, a good threshold to measure is the **75th percentile** of page loads recorded in the field, segmented across mobile and desktop devices:
 
 - An INP below or at **200 milliseconds** means that your page has **good responsiveness**.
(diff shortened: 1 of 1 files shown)


src/site/content/en/metrics/inp/index.md
```diff
@@ -63,9 +63,7 @@ An interaction's latency consists of the single longest [duration](https://w3c.g
 For more details on how INP is measured, read the ["What's in an interaction?"](#whats-in-an-interaction) section.
 {% endAside %}
 
-## What's a "good" INP value?
```
malchata, line 66: I have to reject this change, because it's important that people know that the reason why 200 milliseconds is the target is because of the varying capabilities of devices out there. Otherwise we will get questions like "why not 100 milliseconds?" while citing research that supports that conclusion.

src/site/content/en/metrics/inp/index.md
```diff
@@ -101,7 +99,7 @@ To ensure you're delivering user experiences with good responsiveness, a good th
 Since INP is experimental, the guidance around thresholds may change over time as it is fine-tuned. [The CHANGELOG at the end of this article](#changelog) will be updated to reflect any changes.
 {% endAside %}
 
-## What's in an interaction?
+### What's in an interaction?
```
malchata, line 102: Why are these headings being brought in one more level?

src/site/content/en/metrics/inp/index.md
tunetheweb: Isn't that the same [as for LCP](https://web.dev/defining-core-web-vitals-thresholds/#largest-contentful-paint) where the ideal time would be 1 second, but for achievability reasons we stretch this out to 2.5 seconds?

tunetheweb: Though maybe we should keep the caveat paragraph until such time that INP becomes a CWV and so can be added to that same note?

I still think we should remove the quotes around "good" for consistency.

src/site/content/en/metrics/inp/index.md
tunetheweb: See above screenshots in first comment.

The other metrics all have this H2 outline, but INP is the exception - hence this PR:

- What is XXX?
- XXX in details (optional - it's there for CLS and FID, though LCP does not have this, so I skipped it here too to make the change easier rather than having to rearrange the text or add a new paragraph to avoid a stacked heading).
- How to measure
- How to improve
- Additional resources (optional - it's there for CLS and LCP, though FID does not have this so skipped it for now but do think we'll want to add it later)
- CHANGELOG

src/site/content/en/metrics/inp/index.md
tunetheweb: OK I added that paragraph back in for now.

src/site/content/en/metrics/inp/index.md
```diff
@@ -63,7 +63,7 @@ An interaction's latency consists of the single longest [duration](https://w3c.g
 For more details on how INP is measured, read the ["What's in an interaction?"](#whats-in-an-interaction) section.
 {% endAside %}
 
-## What's a "good" INP value?
+### What's a good INP value?
```
philipwalton, line 66:
```suggestion
### What is a good INP score?
```

For full consistency. Since these heading links get used a lot it'd be good to make them all the same.

malchata (approved):
LGTM!

PR #9351, opened 2023-01-06
Remove course analytics gtags

### LINKED ISSUE #9277: Clean up CourseLinks gtag usage
https://github.com/GoogleChrome/web.dev/pull/9168 introduces a generic `logEvent` function that adds all relevant information to analytics events. 

src/lib/components/CourseLinks/index.js until now is an exemption, as it is inlined and hence uses a direct `gtag` call which should be avoided.

tunetheweb:
Fixes #9277

- Remove gtag calls as currently not working and turns out they are not actually used!
- I've left the percentage calculation in there, just in case we want to surface it in the front end in future.

### DIFF
--- a/src/lib/components/CourseLinks/index.js
+++ b/src/lib/components/CourseLinks/index.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/lib/components/CourseLinks/index.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Google LLC
+ * Copyright 2023 Google LLC
```
matthiasrohmer, line 2: From what I am used to I wouldn't update the year in the license header, as I _think_ it reads like 'from that point on'. But might well be I am missing out on some kind of policy or guideline.

matthiasrohmer (approved):
Thanks Barry!

src/lib/components/CourseLinks/index.js
tunetheweb: Looks like you're right! Reverted it.

PR #9358, opened 2023-01-07
chore(content): clarify meaning of LightWallet

schalkneethling:
When reading the article and clicking on the link to LightWallet, my first reaction was that the link was incorrectly marked up.

In its current state, it links to the article the reader is reading. This was not what I was expecting. After reading the article, I realized that LightWallet is the name used to define and test against a predefined performance budget with the Lighthouse CLI.

The update here attempts to clarify this upfront and removes the link to avoid confusion.


AaronForinton (approved):
I've got no idea why this article was linking back to itself! Thanks for spotting this.

PR #9383, opened 2023-01-11
Percy revamp.

johnrich85:
First steps towards: #9369 

- Adds new pages to handbook/content-types
- Uses the new pages for visual testing.

### DIFF
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - main
-    paths: &paths
+    paths:
       - '**.scss'
       - 'package.json'
       - 'src/component-library/**'
@@ -20,7 +20,19 @@ on:
       - 'src/site/content/en/handbook/content-types/**'
 
   pull_request:
-    paths: *paths
+    paths:
+      - '.github/workflows/percy-workflow.yml'
+      - '**.scss'
+      - 'package.json'
+      - 'src/component-library/**'
+      - 'src/fonts/**'
+      - 'src/build/**.js'
+      - 'src/**.js'
+      - '!src/site/_collections/**.js'
+      - '!src/site/_data/**.js'
+      - '!src/site/content/**.js'
+      - 'shared/**.js'
+      - 'src/site/content/en/handbook/content-types/**'
 
 jobs:
   percy:
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - main
-    paths:
+    paths: &paths
       - '**.scss'
       - 'package.json'
       - 'src/component-library/**'
@@ -20,19 +20,7 @@ on:
       - 'src/site/content/en/handbook/content-types/**'
 
   pull_request:
-    paths:
-      - '.github/workflows/percy-workflow.yml'
-      - '**.scss'
-      - 'package.json'
-      - 'src/component-library/**'
-      - 'src/fonts/**'
-      - 'src/build/**.js'
-      - 'src/**.js'
-      - '!src/site/_collections/**.js'
-      - '!src/site/_data/**.js'
-      - '!src/site/content/**.js'
-      - 'shared/**.js'
-      - 'src/site/content/en/handbook/content-types/**'
+    paths: *paths
 
 jobs:
   percy:
--- a/.github/workflows/percy-workflow.yml
+++ b/.github/workflows/percy-workflow.yml
@@ -7,21 +7,32 @@ on:
     branches:
       - main
     paths:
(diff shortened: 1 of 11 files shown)


.github/workflows/percy-workflow.yml
```diff
@@ -7,21 +7,32 @@ on:
...
       - 'src/**.js'
+      - '!src/site/_collections/**.js'
+      - '!src/site/_data/**.js'
+      - '!src/site/content/**.js'
       - 'shared/**.js'
-      - '!src/site/_data/countries.js'
-      - 'package.json'
-      - '**.njk'
-      - '**.scss'
+      - 'src/site/content/en/handbook/content-types/**'
```
matthiasrohmer, line 35: Can you try using a YAML reference for the two `paths` keys? They seem to be the same and would make things easier to see and maintain.

redirects.yaml
```diff
@@ -23,6 +23,11 @@ redirects:
   - source: /subscribe
     destination: /newsletter
 
+  # Test post
+
+  - source: /test-post
+    destination: /handbook/content-types/example-post
```
matthiasrohmer, line 29: What's that redirect for?

src/site/_collections/hooks/utils.js
```diff
@@ -68,15 +68,20 @@ const index = (items, href, testItems) => {
...
  */
-const individual = (items, lang) => {
+const individual = (items, lang, indexedOnly = false) => {
   let paginated = [];
 
   for (const item of items) {
-    if (item.elements.length > 0) {
+    const elements = indexedOnly
+      ? item.elements.filter((element) => element.data.noindex !== true)
+      : item.elements;
```
matthiasrohmer, line 80:
```suggestion
    let elements = item.elements;
    if (indexedOnly) {
      elements = elements.filter((element) => element.data.noindex !== true);
    }
```

With the .filter call I think not using a ternary operator here is easier to read 🙂

src/site/content/en/handbook/content-types/example-landing-page/index.njk
```diff
@@ -0,0 +1,147 @@
...
+        title: >-
+          A post without a thumbnail
+        subhead: >-
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+          In et eros volutpat, commodo neque quis, egestas tortor
+        url: /handbook/content-types/example-landing-page/
+        thumbnail:
+        alt: Lorem Ipsum
+        authors:
+          - razvancaliman
```
matthiasrohmer, line 20: It might be better to use an example author here, following https://github.com/GoogleChrome/developer.chrome.com/pull/4791 - that's also true for the other example posts; though I know it's not ideal to test multiple authors...

src/site/content/en/handbook/content-types/example-post/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 layout: post
+permalink: /handbook/content-types/example-post
```
matthiasrohmer, line 3: Isn't that the default permalink which is inflected from the file path? Why does it need to be specifically defined here?

redirects.yaml
johnrich85: So the example post already existed. I've just moved it to the content-types directory (checked with Ewa beforehand) and thought it'd be better to redirect rather than leave a dead link. Might be unnecessary with it being a sample post though...

src/site/content/en/handbook/content-types/example-post/index.md
johnrich85: So, from memory, there's some logic that flattens the URLs of posts. Without that permalink it ends up in the root directory of dist. I thought it'd be better kept within the content-types path for consistency

redirects.yaml
matthiasrohmer: I'd vote to remove 🙂 as it isn't publicly facing it shouldn't hurt and prevents someone from guessing why it's redirected in the first place

src/site/content/en/handbook/content-types/example-post/index.md
matthiasrohmer: Flatting only happens for `/blog/` (I think 😅). Tried removing it and the page is still available at http://localhost:8080/handbook/content-types/example-post, if that also works for you I'd vote to remove it.

matthiasrohmer (approved):
Looks good to me 🙂

src/site/content/en/handbook/content-types/example-post/index.md
johnrich85: Huh - possibly something windows related but just tested again and got same result:
image

matthiasrohmer:
@devnook, can't merge myself as content files have been changed 🙂

Unionarmy7:
www.mint.com

PR #9393, opened 2023-01-12
Added core web vitals guidance to the compression and minification guide.

malchata:

- The [minify and compress network payloads guide](https://web.dev/reduce-network-payloads-using-text-compression/) doesn't have any guidance on the effect of Core Web Vitals. This PR adds in that guidance.

### DIFF
--- a/src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
+++ b/src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
@@ -3,7 +3,9 @@ layout: post
 title: Minify and compress network payloads
 authors:
   - houssein
+  - jlwagner
 date: 2018-11-05
+updated: 2023-01-13
 description: |
   There are two useful techniques that can be used to improve the performance of
   your web page, minification and data compression. Incorporating both of these
@@ -118,7 +120,7 @@ Use express.js to compress assets with [gzip](/codelab-text-compression) and [Br
 
 ## Static compression
 
-Static compression  involves compressing and saving assets ahead of time. This
+Static compression involves compressing and saving assets ahead of time. This
 can make the build process take longer, especially if high compression levels
 are used, but ensures that no delays happen when the browser fetches the
 compressed resource.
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
 
 app.use(express.static('public'));
 </pre>
+
+## Effects on Core Web Vitals
+
+Depending the type of resource, both compression and minification can have a variety of effects on Core Web Vitals. Since both minification and compression reduce the payload size of the delivered resource, it's important to understand what Core Web Vitals are affected by these performance optimizations.
```
tunetheweb, line 164: This makes it sound like it has good and bad effects. Surely it's pretty much always good? So should it be more "can positively impact Core Web Vitals" or something like that.

Also I think there should be a generic comment in this section about the fact smaller payloads, reduces network contention, which improves loading performance.

src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
...
+
+### CSS
+
+When CSS resources are compressed and minified, a number of performance improvements may be observed in your Core Web Vitals metrics.
+
+- Where [First Contentful Paint (FCP)](/fcp/) is concerned, making your CSS as small as it can possibly be means that the browser will take less time to download the resource. This means that parsing of CSS can occur sooner, which should move up the rendering step.
+- Where [Largest Contentful Paint (LCP)](/lcp/) is concerned, it largely depends. If your LCP resource is specified in a CSS `background-image` property, making your CSS smaller means that the potential LCP resource can be discovered more quickly. However, LCP candidates that come from a CSS `background-image` property are not always optimal for LCP, as external CSS resources [don't benefit from the browser preload scanner](/preload-scanner/#css-background-images). If you can't move an LCP image out of CSS and into your HTML, then you should be preloading that image resource.
```
tunetheweb, line 171: Again "it largely depends". Does that mean you'd suggest NOT compressing or minifying for some cases? No.

I'd also merge this, concentrate less on the `background-image` example and concentrate instead on the render-blocking aspect. Something like this:

```suggestion
CSS is render-blocking so the quicker it can be downloaded and parsed, the quicker these renders can happen which can help both [First Contentful Paint (FCP)](/fcp/) and [Largest Contentful Paint (LCP)](/lcp/).
```

Yeah an image may still need to be fetch for LCP but not for text ones. Plus images (without `fetchpriority="high"`) are not even fetched until the browser gets through the critical resources. So I think above shorter version if fine.

src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
...
+
+### JavaScript
+
+JavaScript is a bit tricky, as it depends on what your JavaScript resources are providing for your website.
+
+- If your JavaScript is providing interactivity for your website, minification and compression will ensure that your website's [First Input Delay (FID)](/fid/) and [Interaction to Next Paint (INP)](/inp/) metrics may be improved during startup. However, it's also important that you're only serving the JavaScript that a given page needs, otherwise you may risk delays in responding to user inputs during that crucial startup period.
```
tunetheweb, line 177: Is this true? Until the JavaScript is run, the event handlers aren't attached so they have no impact on FID and INP.

I'm not sure compression and minification causes that big a difference to FID and INP to be honest. It will be uncompressed after downloading. And as it's parsed and compiled it's basically minified anyway. Yeah parsing minified code is likely a bit more efficient but not sure it's noticeable.

And on that note, doesn't JS always provide interactivity? And if it doesn't you don't care about FID and INP so think first clause is a bit redundant?

What about this:

```suggestion
Compressing and minifying JavaScript can have positive impacts on your website:

- Making the JavaScript resources smaller allows the JavaScript to download faster, allowing it to be to be processed earlier, allowing your page to become interactive earlier. However, it's also important that you're only serving the JavaScript that a given page needs, otherwise you may risk delays in responding to user inputs during that crucial startup period.
```

On this one I don't think there's an actual FID/INP benefit so not mentioning the metrics specifically.

src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
...
+- If your JavaScript is providing markup for a page, then the situation changes. As always, smaller payloads will download faster. In the case of JavaScript, reducing resource load time means you're going to arrive at the parsing, compiling, and execution steps sooner. This may result in faster LCP times if the client-rendered markup contains the page's LCP element. Beyond that, you should try to avoid rendering markup on the client, as [it defeats the preload scanner](/preload-scanner/#rendering-markup-with-client-side-javascript). If you can't avoid client-rendering of markup, then you should preload that render-critical piece of JavaScript so that it can arrive to the browser as fast as it possibly can.
```
tunetheweb, line 178:
```suggestion
- If your JavaScript is providing markup for a page, then it is especially critical to allow this to happen as fast as possible. As always, smaller payloads will download faster. In the case of JavaScript, reducing resource load time means you're going to arrive at the parsing, compiling, and execution steps sooner. This may result in faster LCP times if the client-rendered markup contains the page's LCP element. Beyond that, you should try to avoid rendering markup on the client, as [it defeats the preload scanner](/preload-scanner/#rendering-markup-with-client-side-javascript).
```

I removed the last sentence - why do you think you need to preload the JS? It'll be loaded anyway. Or do you mean preload the resources in that JS so they can be discovered in the markup?

src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
...
+
+### SVG images
+
+Raster image types are inherently compressed, be it lossy or lessless formats. However, SVG—being a vector image format and text-based—can benefit from compression and minification in the same way that other text-based resources can.
```
tunetheweb, line 182: Anyway to emphasis this more? Text-based compression has a dramatic impact on file sizes.

What about:

```suggestion
Raster image types are inherently compressed, be it lossy or lessless formats. However, SVG—being a vector image format and text-based—benefits greatly from compression and minification in the same way that other text-based resources can.
```

src/site/content/en/fast/reduce-network-payloads-using-text-compression/index.md
```diff
@@ -156,3 +158,35 @@ const app = express();
...
+
+For static assets such as CSS, JavaScript, and images, you should try to avoid dynamic compression. Static compression is an alternative compression method where the resource is compressed ahead of time and stored on disk, which means that there's no server time involved in compressing the resource at the time of the request. It's the best of both worlds: you get reduced resource payloads, without any of the delays that come with dynamic compression.
+
+The only exception to this rule is with dynamically generated HTML. This is HTML that changes based on the user, such as HTML payloads generated for authenticated users. You can't statically compress this type of markup. However, if you're running a website that doesn't perform personalization of markup on the server, you can safely compress it statically.
```
tunetheweb, line 192: I think this is too strong. Dynamic-based compression is fine and barely noticeable at mid-level settings, unless your server is severely under powered.

A low-level compression settings it pointless.
At high-level compression settings it can cause a performance issue.

Current wording is aimed at the dangers of high-level compression, but makes it looks like it applies to all.

I'd reword this to emphasis the dangers of too high a dynamic compression level (and also the pointless-ness of too low a compression level) instead, and that mid-level is usually the good trade off.

And that static compression allows the best of both worlds as you say.

Could also be good to reference the recommended settings in this table: https://almanac.httparchive.org/en/2021/compression#fig-10

PR #9406, opened 2023-01-13
Disable autoplay when using prefers reduced motion

### LINKED ISSUE #9388: Address a11y issue with homepage animation
The year in review homepage animation loops constantly and can't be paused. 

We could probably just stop it looping, and add controls to it.

tunetheweb:
Fixes #9388

- Adds a bit of JavaScript to display `autoplay` on videos (and ensures `control` attribute is also set) when `prefers-reduced-motion` is enabled.

Note: `autoplay` is discouraged, and should only be used for short clips without considerable motion (if even then).

### DIFF
--- a/src/lib/app.js
+++ b/src/lib/app.js
@@ -13,6 +13,14 @@ import removeServiceWorkers from './utils/sw-remove';
 // See .unsupported-notice in _page-header.scss
 document.body.classList.remove('unresolved');
 
+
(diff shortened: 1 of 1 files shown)


matthiasrohmer (commented):
I played around with something similar yesterday but was worried that doesn't meet the bar for web.dev 😄  Hence forwarding to Engineer in Chief, @devnook. Thanks though, Barry!

devnook (approved):

PR #9413, opened 2023-01-16
Adds pattern-set and pattern to visual test suite

johnrich85:
Further steps towards: https://github.com/GoogleChrome/web.dev/issues/9369

- Adds pattern-set and pattern samples to handbook/content-types
- Uses the new pages for visual testing.


src/lib/patterns.js
```diff
@@ -24,10 +24,19 @@ const glob = require('fast-glob');
...
+const contentRoot = path.join(__dirname, '../../src/site/content/en');
+const basePath = path.join(contentRoot, 'patterns');
+const globBasePath = path.join(
+  contentRoot,
+  '{patterns,handbook/content-types/example-pattern}',
+);
+
+const files = glob.sync(
+  forceForwardSlash(path.join(globBasePath, '**', 'index.md')),
+);
```
matthiasrohmer, line 39: I find it hard to grasp what's happening here. I think there are two things specifically that I stumble over: 

`forceForwardSlash` is needed for fast-glob to correctly handle Windows paths right? I wonder if that's still needed if path formats wouldn't be mixed, but instead all built by `path.join`.

Would that work with the following approach?

```js
const contentRoot = path.join(__dirname, '..', '..', 'src', 'site', 'content', 'en');
const basePath = path.join(contentRoot, 'patterns');

const patternPaths = [
  path.join(contentRoot, 'patterns', '**', 'index.md'),
  path.join(contentRoot, 'handbook', 'content-types', 'example-pattern', '**', 'index.md'),
]

const patterns = glob.sync(patternPaths);
```

I didn't test this, but the idea would be to let path.join generate full paths, instead of joining partials - if that makes sense...?

src/lib/patterns.js
```diff
@@ -58,7 +67,10 @@ const allPatterns = files.reduce((patterns, file) => {
     return patterns;
   }
 
-  const assetsPaths = glob.sync(path.join(path.dirname(file), 'assets', '*'));
+  const assetsPaths = glob.sync(
+    forceForwardSlash(path.join(path.dirname(file), 'assets', '*')),
+  );
+
```
matthiasrohmer, line 73: At this point the path should be OS universal, no?

src/lib/utils/path.js
```diff
@@ -0,0 +1,15 @@
+/**
```
matthiasrohmer, line 1: If we keep this, then a license header is missing, but following the comment above I hope we can remove this.

src/site/content/en/handbook/content-types/example-pattern/demo.njk
```diff
@@ -0,0 +1,56 @@
...
+        padding: 4px;
+        border-radius: 8px;
+        line-break: anywhere
+      }
+    </style>
+  </head>
+  <body>
+    <h1>Sample demo</h1>
+
+    <img src="assets/fugu.svg" alt="Fugu fish." width="128" height="128">
```
matthiasrohmer, line 39: Can we rename `fugu.svg` to something that's non-specific? Maybe just `example.svg` - then it's clear this does not need to be updated if something about Fugu changes.

johnrich85:
Hey @rachelandrew . Would you mind giving the new sample content a once over, please?

src/site/content/en/handbook/content-types/example-pattern/demo.njk
```diff
@@ -0,0 +1,56 @@
...
+        line-break: anywhere
+      }
+    </style>
+  </head>
+  <body>
+    <h1>Sample demo</h1>
+
+    <img src="assets/example.svg" alt="Example" width="128" height="128">
+
+    <p>To create a demo like this, simply add a demo.njk file to the pattern's root directory</p>
```
rachelandrew, line 41:
```suggestion
    <p>To create a demo like this, add a demo.njk file to the pattern's root directory</p>
```

src/site/content/en/handbook/content-types/example-pattern/demo.njk
```diff
@@ -0,0 +1,56 @@
...
+
+    <p>You can add JS to the demo as demonstrated below.</p>
```
rachelandrew, line 43:
```suggestion
    <p>You can add JavaScript to the demo as demonstrated below.</p>
```

rachelandrew (approved):
As with the other, please merge the small edits.

PR #9414, opened 2023-01-17
Fix "Using WebTransport" article

queengooborg:
This PR aims to fix issues with the "Using WebTransport" article that had arisen due to spec changes after the article was originally written.  In this PR, the following changes have been performed:

- Update all interface names to retain parity with spec
- Fix all examples to use the correct method names
- Fix spec links to point to current specs (or correct parts of specs)
- Correct formatting in select locations
- Update the "last updated" date (has not been changed within the last few commits)

See https://github.com/mdn/browser-compat-data/pull/18638 for the inspiration and research behind these changes.


AaronForinton (approved):
Thank you for helping us keep this article up to speed. Approved

PR #9425, opened 2023-01-18
Adds the optimize time to first byte guide.

malchata:

- Adds the Optimize Time to First Byte guide.

### DIFF
--- a/src/site/content/en/blog/optimize-ttfb/index.md
+++ b/src/site/content/en/blog/optimize-ttfb/index.md
@@ -1,0 +1,229 @@
+---
+title: Optimize Time to First Byte
+subhead: |
+  Learn how to optimize for the Time to First Byte metric.
+authors:
+  - jlwagner
+date: 2023-01-18
+hero: image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/IuvYBfNBVKH7qU3872qg.jpg
+thumbnail: image/jL3OLOhcWUQDnR4XjewLBx4e3PC3/OBITO4G8MUJMILC7IFE0.jpg
+alt: A picture of several racks of servers, with white text superimposed on it that reads "Optimize Time to First Byte".
+description: |
+  Learn how to optimize for the Time to First Byte metric.
+tags:
+  - blog
+  - performance
+  - web-vitals
+---
+
+[Time to First Byte (TTFB)](/ttfb/) is a foundational web performance metric that precedes every other meaningful user experience metric such as [First Contentful Paint (FCP)](/fcp/) and [Largest Contentful Paint (LCP)](/lcp/). This means that high TTFB values add time to the metrics that follow it.
+
+It's recommended that your server responds to navigation requests quickly enough so that the **75th percentile** of users experience an FCP [within the "good" threshold](/fcp/#what-is-a-good-fcp-score). As a rough guide, most sites should strive to have a TTFB of **0.8 seconds or less**.
+
+<figure>
+  <picture>
+    <source
+      srcset="{{ "image/W3z1f5ZkBJSgL1V1IfloTIctbIF3/ILJ1xKjzVisqOPPyHYVA.svg" | imgix }}"
+      media="(min-width: 640px)"
+      width="800"
+      height="200">
+    {%
+      Img
+        src="image/W3z1f5ZkBJSgL1V1IfloTIctbIF3/EcKicxW5ErYYhf8RvpeO.svg",
+        alt="Good TTFB values are 0.8 seconds or less, poor values are greater than 1.8 seconds, and anything in between needs improvement",
+        width="640",
+        height="480"
+    %}
+  </picture>
+</figure>
+
+{% Aside 'important' %}
+TTFB is not a Core Web Vitals metric, so it's not absolutely necessary that sites meet the "good" TTFB threshold, provided that it doesn't impede their ability to score well on the metrics that matter.
+
+Websites vary in how they deliver content. A low TTFB is crucial for getting markup out to the client as soon as possible. However, if a website delivers the initial markup quickly, but that markup then requires JavaScript to populate it with meaningful content—as is the the case with Single Page Applications (SPAs)—then achieving the lowest possible TTFB is especially important so that the client-rendering of markup can occur sooner.
+
+Conversely, a server-rendered site that does not require as much client-side work may have a slightly lower TTFB, but good FCP and LCP values. This is why the TTFB thresholds are a “rough guide”, and will need to be weighed against how your site delivers its core content.
+{% endAside %}
+
+## How to Measure TTFB
+
+Before you can optimize TTFB, you need to observe how it affects your website's users. You should rely on [field data](/lab-and-field-data-differences/#field-data) as a primary source of observing not just TTFB, but every meaningful performance metric. To find out how to measure TTFB in both the lab and the field, [consult the TTFB metric page](/ttfb/#how-to-measure-ttfb).
+
+### Understanding high TTFB with `Server-Timing`
+
+The [`Server-Timing` response header](https://developer.mozilla.org/docs/Web/HTTP/Headers/Server-Timing) can be used in your application back end to measure distinct back end processes that could contribute to high latency. The header value's structure is flexible, accepting, at minimum, a handle that you define. Optional values include a duration value (via `dur`), as well as an optional human-readable description (via `desc`).
+
+{% Aside %}
+If measuring back end latency with `Server-Timing` is not feasible, then a suitable alternative may be to rely on [Application Performance Monitoring (APM)](https://en.wikipedia.org/wiki/Application_performance_management) to detect and diagnose back end performance problems.
+{% endAside %}
+
+`Serving-Timing` can be used to measure many application back end processes, but there are some that you may want to pay special attention to:
+
+- Database queries
+- Server-side rendering time, if applicable
+- Disk seeks
+- Edge server cache hits/misses (if using a CDN)
+
+All parts of a `Server-Timing` entry are colon-separated, and multiple entries can be separated by a comma:
+
+```http
+// Two metrics with descriptions and values
+Server-Timing: db;desc="Database";dur=121.3, ssr;desc="Server-side Rendering";dur=212.2
+```
+
+The header can be set using your application back end's language of choice. In PHP, for example, you could set the header like so:
+
+```php
+<?php
+// Get a high-resolution timestamp before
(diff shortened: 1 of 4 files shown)


src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+  </picture>
+</figure>
+
+{% Aside 'important' %}
+TTFB is not a Core Web Vitals metric, so it's not absolutely necessary that sites meet the "good" TTFB threshold, provided that it doesn't impede their ability to score well on the metrics that matter.
+
+Websites vary in how they deliver content. A low TTFB is crucial for getting markup out to the client as soon as possible. However, if a website delivers the initial markup quickly, but that markup then requires JavaScript to populate it with meaningful content—as is the the case with Single Page Applications (SPAs)—then achieving the lowest possible TTFB is especially important so that the client-rendering of markup can occur sooner.
+
+Conversely, a server-rendered site that does not require as much client-side work may have a slightly lower TTFB, but good FCP and LCP values. This is why the TTFB thresholds are a “rough guide”, and will need to be weighed against how your site delivers its core content.
+{% endAside %}
```
tunetheweb, line 46: Looks like you need to use `<p>` tags in an `{% Aside %}` to break up paragraphs:

```suggestion
{% Aside 'important' %}
<p>TTFB is not a Core Web Vitals metric, so it's not absolutely necessary that sites meet the "good" TTFB threshold, provided that it doesn't impede their ability to score well on the metrics that matter.<p>

<p>Websites vary in how they deliver content. A low TTFB is crucial for getting markup out to the client as soon as possible. However, if a website delivers the initial markup quickly, but that markup then requires JavaScript to populate it with meaningful content—as is the the case with Single Page Applications (SPAs)—then achieving the lowest possible TTFB is especially important so that the client-rendering of markup can occur sooner.<p>

<p>Conversely, a server-rendered site that does not require as much client-side work may have a slightly lower TTFB, but good FCP and LCP values. This is why the TTFB thresholds are a “rough guide”, and will need to be weighed against how your site delivers its core content.<p>
{% endAside %}
```

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+## How to Measure TTFB
+
+Before you can optimize TTFB, you need to observe how it affects your website's users. You should rely on [field data](/lab-and-field-data-differences/#field-data) as a primary source of observing not just TTFB, but every meaningful performance metric. To find out how to measure TTFB in both the lab and the field, [consult the TTFB metric page](/ttfb/#how-to-measure-ttfb).
```
tunetheweb, line 50:
```suggestion
Before you can optimize TTFB, you need to observe how it affects your website's users. You should rely on [field data](/lab-and-field-data-differences/#field-data) as a primary source of observing TTFB as it affected by redirects, whereas lab-based tools are often measured using the final URL therefore missing this extra delay. To find out how to measure TTFB in both the lab and the field, [consult the TTFB metric page](/ttfb/#how-to-measure-ttfb).
```

src/site/content/en/blog/optimize-ttfb/index.md
tunetheweb, line 46: I also think the expanded contents of this aside should be copied to /ttfb/ as part of this PR as it's good info to put in there too and the rest of this section is the same.

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+### Avoid multiple page redirects
+
+One potential contributor to a high TTFB is [redirects](https://developer.mozilla.org/docs/Web/HTTP/Redirections). Redirects occur when a navigation request for a document receives a response that informs the browser that the resource exists at another location. One redirect can certainly add unwanted latency to a navigation request, but it can certainly get worse if that redirect points to another resource that results in _another_ redirect—and so on.
+
+There are two types of redirects:
+
+- **Same-origin redirects**, where the redirect occurs entirely on your website.
+- **Cross-origin redirects**, where the redirect occurs initially on another origin—such as from a social media URL shortening service, for example—before arriving at your website.
+
+You want to focus on eliminating same-origin redirects, as this is something you will have direct control over. This would involve checking links on your website to see if any of them result in a 302 or 301 response code. When it comes to cross-origin redirects, there's nothing you can really do to fix that source of added latency.
```
tunetheweb, line 179:
```suggestion
You want to focus on eliminating same-origin redirects, as this is something you will have direct control over. This would involve checking links on your website to see if any of them result in a 302 or 301 response code.

Cross-origin redirects are trickier as these are often outside of your control, but try to avoid multiple redirects where possible - for example, by using multiple link shorteners when sharing links.
```

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+Once you have a good HSTS policy in place, you can speed things up on the first visit to an origin [by adding your site to the HSTS preload list](https://developer.mozilla.org/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security).
+
+{% Aside 'caution' %}
+Be **very careful** when implementing HSTS, as setting too aggressive a policy without sufficient testing [can break your website](/bbc-hsts/#deploying-hsts).
+{% endAside %}
+
+### Stream markup to the browser
+
+Browsers are optimized to deliver markup efficiently through streaming, meaning that markup is handled in chunks as it arrives from the server. This is crucial where large markup payloads are concerned, as it means the browser can parse that large chunk of markup incrementally, as opposed to waiting for the entire response to arrive before parsing can begin.
```
tunetheweb, line 191: Browses don't deliver markup! 😁

```suggestion
Browsers are optimized to process markup efficiently when it is streamed, meaning that markup is handled in chunks as it arrives from the server. This is crucial where large markup payloads are concerned, as it means the browser can parse that the chunks of markup incrementally, as opposed to waiting for the entire response to arrive before parsing can begin.
```

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+Though browsers are great at streaming markup, it's crucial to do all that you can to keep that stream flowing so those initial bits of markup are on their way as soon as possible. If the back end is holding things up, that's a problem. Because back end stacks are numerous, it would be beyond the scope of this guide to cover every single stack and the issues that could arise in each specific one.
```
tunetheweb, line 193:
```suggestion
Though browsers are great at handling streaming markup, it's crucial to do all that you can to keep that stream flowing so those initial bits of markup are on their way as soon as possible. If the backend is holding things up, that's a problem. Because backend stacks are numerous, it would be beyond the scope of this guide to cover every single stack and the issues that could arise in each specific one.
```

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+{% Aside %}
+Not every JavaScript runtime can take advantage of streaming server-side rendering. Newer JavaScript runtimes such as [Deno](https://deno.land/) support this out of the box, but older versions of [Node.js](https://nodejs.org/) may not support it. Check to see if this is the case for you, and see what you can do to upgrade or switch your runtime for better server-side rendering performance.
+{% endAside %}
+
+Another way to ensure markup is streamed to the browser is to rely on [static rendering](/rendering-on-the-web/#static-rendering) which generates HTML files during build time. While this approach isn't suitable for every page on every website—such as those requiring a dynamic response as part of the user experience—it can be beneficial for those pages that don't require markup to be personalized to a specific user.
```
tunetheweb, line 201: Is this really streaming? Word-smithed to make it more like it is:

```suggestion
Another way to ensure markup is streamed to the browser quickly is to rely on [static rendering](/rendering-on-the-web/#static-rendering) which generates HTML files during build time. With the full file available immediately, web servers can start sending the file immediately and the inherit nature of HTTP will result in streaming markup. While this approach isn't suitable for every page on every website—such as those requiring a dynamic response as part of the user experience—it can be beneficial for those pages that don't require markup to be personalized to a specific user.
```

src/site/content/en/blog/optimize-ttfb/index.md
```diff
@@ -0,0 +1,229 @@
...
+
+[The `103 Early Hints` header](https://developer.chrome.com/blog/early-hints/) is an early response code that the server can send to the browser while the backend is busy preparing markup. This header can be used to hint to the browser that there are render-critical resources the page should begin downloading while the markup is being prepared. For [supporting browsers](https://caniuse.com/mdn-http_status_103), the effect can be faster document rendering (CSS) and quicker availability of core page functionality (JavaScript).
+
+{% Aside %}
+If your website doesn't do a lot of processing on the back end to prepare markup—for example, static sites—then `103 Early Hints` probably won't help much. It tends to work best for sites that involve considerable backend time before markup can be sent to the user.
+{% endAside %}
+
+## Conclusion
+
+Since there are so many combinations of back-end application stacks, there's no one article that can encapsulate _everything_ you can do to lower your website's TTFB. However, these are some options you can explore to try and get things going just a little bit faster on the server side of things.
```
tunetheweb, line 225:
```suggestion
Since there are so many combinations of backend application stacks, there's no one article that can encapsulate _everything_ you can do to lower your website's TTFB. However, these are some options you can explore to try and get things going just a little bit faster on the server side of things.
```

tunetheweb (changes requested):
We should link this guide in the following places:

- /ttfb/
- fast.json
- learn-core-web-vitals.json

malchata:
@tunetheweb, I've added links to the new guide in `fast.json` and the [TTFB metric page](/ttfb/), but not really sure if it belongs in the `learn-core-web-vitals.json` file, since TTFB is not strictly a Core Web Vital. Thoughts?

tunetheweb:
> @tunetheweb, I've added links to the new guide in `fast.json` and the [TTFB metric page](/ttfb/), but not really sure if it belongs in the `learn-core-web-vitals.json` file, since TTFB is not strictly a Core Web Vital. Thoughts?

Yeah was struggling with that myself. It's a good diagnostic metric for CWV though. Maybe should have a "Diagnostic metrics" section on that page?

I also wonder if we shouldn't just retired learn-core-web-vitals and redirect it to the more comprehensive /fast/ page? Seems a lot of overlap.

Anyway, let's merge this for now.

tunetheweb (approved):

PR #9458, opened 2023-01-26
Syncs TTFB aside

tunetheweb:
Ports changes in #9429 to the other reference of this.


malchata (approved):
LGTM! Thanks for doing this, @tunetheweb!

PR #9490, opened 2023-02-01
Link directly to Optimize TTFB

rviscomi:
The Optimize LCP page has a section about reducing TTFB and it links to the TTFB metric page. Now that we have a dedicated Optimize TTFB page, this PR updates the LCP guide to point to the TTFB guide.


malchata (approved):
Good catch! LGTM!

PR #9502, opened 2023-02-02
Remove the word "simply"

adactio:
I know the intention is to convey this step isn't hard, but using words like "just" or "simply" can end up having the uninteded effect of making the reader feel like they should know more.

Fixes #SOME_ISSUE_NUMBER

-
-
-


AaronForinton (approved):

AaronForinton:
Agreed. Thank you.

PR #9509, opened 2023-02-03
Fix mispelled word in same-site / same-origin article

recvfrom:
- Replace 'schemelses' with 'schemeless'


rachelandrew (approved):
Thank you!

PR #9532, opened 2023-02-08
Document speculation rules experiment with prefetch

tunetheweb:

- Add automatic prefetching on mouse down of internal links, via [this origin trial](https://developer.chrome.com/origintrials/#/view_trial/705939241590325249). We want to test out this feature on our own site and collect feedback for a future blog post on this feature.
- Adds a simple `minifyJSON` function which strips whitespace (you can't use `minifyJS` as they are slightly different and using `minifiyJS` creates [invalid JSON](https://twitter.com/tunetheweb/status/1598697968374202369?s=20&t=OZd5R4Rt3aUqImKHQ-YRAA)).


src/lib/analytics.js
```diff
@@ -69,7 +69,14 @@ function sendToGoogleAnalytics({
     metric_value: value,
     // This should already by set globally, but to ensure it's consistent
     // with the web-vitals library, set it again.
-    navigation_type: navigationType,
+    // Override for 'navigational-prefetch' for the prefetch origin trial
+    // experiment (https://github.com/GoogleChrome/web.dev/pull/9532)
+    navigation_type:
+      navigationType === 'navigate' &&
+      performance.getEntriesByType('navigation')[0].deliveryType ===
```
jeremyroman, line 76: Do you want to also handle the cases of browsers that don't support the performance timeline API, like getNavigationType does?

(It could probably be done more concisely than it is there with the nullish coalescing syntax, assuming this goes through a build step to elide that for older UAs.)

src/site/_includes/macros/prerender.njk
```diff
@@ -3,7 +3,18 @@
...
+    {
+      "source":"list",
+      "urls":[
+        "{{ url }}"
+      ]
+    }
+  ]
+}
+{% endset %}
+<script type="speculationRules">{{ speculationrules | minifyJSON | cspHash | safe }}</script>
```
matthiasrohmer, line 18: The type is probably not case-sensitive, the variable name probably is...? As you are setting `speculation*R*ules` above.

matthiasrohmer (approved):
Looks good, one nit. Created https://github.com/GoogleChrome/web.dev/issues/9559 for follow-up.

src/lib/analytics.js
tunetheweb: I did think about that but it's loaded as a module and every browser that supports that _should_ support this?

But added it anyway.

src/site/_includes/macros/prerender.njk
tunetheweb: Oh good spot! Changed that right at the end of course!

PR #9558, opened 2023-02-10
Setup dependabot

matthiasrohmer:
Follow-up to https://github.com/GoogleChrome/developer.chrome.com/pull/5237. Configures dependabot to have PRs to update @mdn/browser-compat-data and 11ty and webdev-infra.


andreban (approved):
LGTM

PR #9592, opened 2023-02-15
feat: display latest post from a specific tag on the homepage

### LINKED ISSUE #9590: Implement latest post for homepage grid.
Implementation follow-up issue for #9404.

mamieorine:
Fixes #9590

- Update the card component to display latest post from specific tags
- Update the homepage grid to display latest post from specific tags on each section
- Update the document on [/handbook/markup-cards-collection/](https://web.dev/handbook/markup-cards-collection/)


jimjiminyjimjim (commented):
LGTM

matthiasrohmer (commented):
I only quickly glanced at it, two questions: does that cover the case `featured == latest` and setting both a featured and a latest post?

mamieorine:
> I only quickly glanced at it, two questions: does that cover the case `featured == latest` and setting both a featured and a latest post?

Do you mean the case that the author setting up to show the featured post and the latest post, in which both the featured and latest one refer to the same post? 

Actually, there is no frontmatter key for displaying featured posts yet. There are 3 ways of showing the post on the homepage grid.
- The author needs to define the title, description, eyebrow, etc, by themselves.
- Find by URL. The card will automatically pull post data when setting `url: /[my-post]/`.
- Display the latest post when set `isLatestPost: true`.

So I think the author can avoid displaying the duplicate post in each tag by the available card setup that I mentioned above.

Also, if they want to display a specific post as the featured post, they can set up the eyebrow like this code sample below:

```
---
cards:
  - url: /terra-dark-mode/
    eyebrow:
      icon: 'featured'
      text: 'Featured'
    cardLayout: 'vertical'
---
```

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @fileoverview Returns posts filtering by specific tags.
+ */
+
+/**
+ * @param {object} post An eleventy post object.
```
matthiasrohmer, line 22: We have `EleventyPage` as a type 🙂

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param {array} tags Array of tag name.
```
matthiasrohmer, line 23: `string[]`.

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,50 @@
...
+ * @return {EleventyCollectionObject} Posts filtered by tags.
+ */
+function filterByTag(posts, tags) {
+  // @ts-ignore
+  const recentPosts = posts.sort((postA, postB) => {
+    if (!postB.date || !postA.date) return 0;
+
+    // @ts-ignore
+    return new Date(postB.date) - new Date(postA.date);
+  });
```
matthiasrohmer, line 45: Two things: actually sorting after filtering would be quicker, because you can assume that you have less elements to sort after filtering. But I wouldn't expect the method to sort at all, that's not what the name says.

src/site/_includes/partials/cards-collection.njk
```diff
@@ -1,10 +1,24 @@
...
+    {% if card.isLatestPost %}
+      {% set latestPost = collections.blogPosts %}
+      {% if card.tags %}
+        {% set latestPost = collections.all | filterByTag(card.tags) %}
+      {% endif %}
+      {% set latestPostInTag = latestPost[0] %}
+      {% set featured = latestPostInTag.url | findByUrl %}
+
+      {% set eyebrowIcon = 'latest' %}
+      {% set eyebrowText = 'Latest' %}
```
matthiasrohmer, line 13: This would need to go through the `i18` filter.

src/site/content/en/handbook/markup-cards-collection/index.md
```diff
@@ -19,6 +19,8 @@ YAML properties available for displaying a layout of cards:
 - `title` - the title of the card.
 - `description` - the description of the card.
 - `thumbnail` - the thumbnail of the card.
+- `isLatestPost` - display the latest post.
+- `tags` - display the latest post in specific tags.
```
matthiasrohmer, line 23: What do you think of combining those two keys? So if you set `tags` it's simply the latest one?

src/site/content/en/index.md
```diff
@@ -89,8 +90,8 @@ themes:
     cards:
       - url: /top-cwv-2023/
         eyebrow:
-          icon: featured
-          text: Featured
+          icon: latest
+          text: Latest
```
matthiasrohmer, line 94: Was this a test? 🙂

src/site/_includes/partials/cards-collection.njk
```diff
@@ -1,10 +1,24 @@
 {% if cards %}
   {% for card in cards %}
     {% set featured = null %}
+    {% if card.isLatestPost %}
+      {% set latestPost = collections.blogPosts %}
+      {% if card.tags %}
+        {% set latestPost = collections.all | filterByTag(card.tags) %}
+      {% endif %}
+      {% set latestPostInTag = latestPost[0] %}
+      {% set featured = latestPostInTag.url | findByUrl %}
```
matthiasrohmer, line 10: I think I would move most of this logic to a filter named `latestPostByTags`, which uses `this.ctx.collections[tag]` to access the already created, memoized and sorted collections.

matthiasrohmer (commented):
Left some comments. The essential bit is: we already got collections per tag and shouldn't recreate them.

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {string[]} tags Array of tag name.
+ * @return {EleventyCollectionItem} Latest post filter by tag.
+ */
+
+function latestPostByTags(_, tags) {
+  let posts = [];
+  tags.forEach((tag) => {
+    // @ts-ignore
+    posts = [...posts, this.ctx.collections[tag].pop()];
+  });
```
matthiasrohmer, line 34: These are wild three lines 😅  I think what you are looking for is `Array.prototype.map`, since you are creating an array from another array.

L33 is problematic in two ways: you are destructuring an array (`...posts`) just to add one element, while you could instead add to the existing array (`Array.prototype.push`). But the other problematic bit is `.pop()`, as it alters the original array, which we do not own in this context - so that's evil.

I think this can be rewritten as:

```
const posts = tags.map((tag) => {
  const tagPosts = this.ctx.collections[tag];
  return tagPosts[tagPosts.length - 1];
}
```

But you can actually improve this even more and get rid of the sorting, if you use an intermediate variable to store current newest post, like this:

```js
let newestPost = null;
for (const tag of tags) {
  const tagPosts = this.ctx.collections[tag];
  const candidatePost = tagPosts[tagPosts.length - 1];

  // Accounts for the first run of the loop
  if (!newestPost) { 
    newestPost = candidatePost; 
    return;
  }

  // Check if the current candidate post is newer then the newest
  // post that is currently stored
  if (new Date(newestPost.date) > new Date(candidatePost.date)) {
    newestPost = candidatePost;
  }
}
```

That's slightly more complicated to read, but does the job in one loop (while `map` and `sort` would be two). Basically a fun interview question...

src/site/content/en/index.md
matthiasrohmer: Why I am asking: `url` is hardcoded to `/top-cwv-2023/`.

src/site/_filters/filter-by-tag.js
mamieorine: I have refactored this logic. Thank you for your suggestion.

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,55 @@
...
+function latestPostByTags(_, tags) {
+  let latestPost = null;
+  for (const tag of tags) {
+    // @ts-ignore
+    const tagPosts = this.ctx.collections[tag];
+    const candidatePost = tagPosts[tagPosts.length - 1];
+
+    if (!latestPost) {
+      latestPost = candidatePost;
+      return;
```
matthiasrohmer, line 38: Never trust foreign code without testing 😄 my suggestion had an error that renders all "Latest" tiles empty. This needs to be `continue;`.

src/site/_filters/filter-by-tag.js
```diff
@@ -0,0 +1,55 @@
+/*
```
matthiasrohmer, line 1: The main export of this file now is `latestPostByTags`, the file should hence be named `latest-post-by-tags.js`.

matthiasrohmer (changes requested):
Almost there. Please reset all changes to index.md, except the CSS and UI category, as in the other categories there is now some overlap between Featured and Latest that I the content team is better suited to sort out. CSS/UI on the other hand seems to be safe and is good to demonstrate how this works. Thanks!

src/site/_filters/filter-by-tag.js
mamieorine: my bad, I did not notice this bug.

src/site/_filters/filter-by-tag.js
mamieorine: I have renamed this file from `filter-by-tag.js` to `latest-post-by-tags.js`

matthiasrohmer (approved):
This works now. Forwarding to @rachelandrew for visibility and approval of making the CSS & UI theme use the latest post.

rachelandrew (approved):
LGTM - thanks! Once this is landed I'll look at updating the others.

PR #9594, opened 2023-02-15
2 fact checks for TTFB advice

tigt:

- One of Node.js’s founding goals was to support gzipped streaming HTTP responses — it’s always supported it
- While the optional chaining operator _is_ useful, it doesn’t add anything to the shown code sample


tigt:
Just signed the CLA

albajarazeth:
@tigt It says that CLA still missing, so you know how you go to "Details" to create the signature right, then you have to go to the very bottom of the page to where it says:

🔁 New Contributors: Update this check after signing the CLA by clicking [here]

Then once you click there it'll update and you'll get the little check mark.

I hope this helps : )

malchata (approved):
LGTM!

Stephen0531:
And  u say that there's 4 checks that's has passed  cause I haven't collected not the first one o of them yet and don't understand what I need to do or how to keep from losing them to collect them the safest way to gain withdraw deposits of bonus really could use it really badly right now honestly

PR #9611, opened 2023-02-17
feat: load recaptcha when newsletter form is in view

### LINKED ISSUE #9528: Load and show recaptcha on homepage on-demand
Extracted from https://github.com/GoogleChrome/web.dev/pull/9409

On the homepage, recaptcha also shows up in the trace of these interactions, tracking attribute changes (like inert) across the whole page with mutation observers. It's not a huge amount of work, but again seems completely unnecessary for this interaction. Should recaptcha be loaded more on demand, maybe once a user starts interacting with the form at the bottom of the page?

We could also load it only when the subscribe form get scrolled into the view.

Blazzike:
Fixes #9528

- Loads reCAPTCHA .js script when the newsletter form element comes into view (detected via a single use IntersectionObserver)
- Shows a tiny loading animation while the reCAPTCHA script is being fetched/executed
- Tested and working on various browsers

Loading animation placeholder:
output

Recorded using a throttled connection.

### DIFF
--- a/src/lib/components/Subscribe/index.js
+++ b/src/lib/components/Subscribe/index.js
@@ -33,31 +33,48 @@ class Subscribe extends BaseElement {
     window['recaptchaSuccess'] = this.captchaCheck.bind(this);
   }
 
-  connectedCallback() {
-    super.connectedCallback();
-    /** @type {HTMLFormElement} */
-    this.form = this.querySelector('form');
-    const intersectionObserver = new IntersectionObserver(([entry]) => {
-      if (!entry.isIntersecting) {
-        return;
+  /**
+   * Called by the IntersectionObserver defined in this#connectedCallback.
+   *
+   * @param {IntersectionObserverEntry[]} entries
+   */
+  onIntersection(entries) {
+    const entry = entries[0];
+    if (!entry.isIntersecting) {
+      return;
+    }
+
+    // can safely assume this.intersectionObserver is defined, since onIntersection is being called
+    this.intersectionObserver.disconnect();
+    window.recaptchaLoadCallback = () => {
+      /** @type {HTMLDivElement} */
+      const recaptchaContainerEl = this.querySelector('.g-recaptcha');
+      // fix for percy re-executing JavaScript with pre-rendered DOM
+      if (!this.hasInitialised) {
+        recaptchaContainerEl.children[0].remove();
+        this.hasInitialised = true;
       }
 
-      intersectionObserver.disconnect();
-      window.recaptchaLoadCallback = () => {
-        /** @type {HTMLDivElement} */
-        const recaptchaContainerEl = this.querySelector('.g-recaptcha');
-        recaptchaContainerEl.children[0].remove();
-        window.grecaptcha.render(recaptchaContainerEl, {
-          sitekey: recaptchaContainerEl.dataset.sitekey,
-        });
-      };
+      window.grecaptcha.render(recaptchaContainerEl, {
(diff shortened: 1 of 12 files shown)


src/lib/components/Subscribe/index.js
```diff
@@ -37,6 +37,28 @@ class Subscribe extends BaseElement {
...
+    const intersectionObserver = new IntersectionObserver(([entry]) => {
+      if (!entry.isIntersecting) {
+        return;
+      }
+
+      intersectionObserver.disconnect();
+      window.recaptchaLoadCallback = () => {
+        /** @type {HTMLDivElement} */
+        const recaptchaContainerEl = this.querySelector('.g-recaptcha');
+        recaptchaContainerEl.children[0].remove();
```
matthiasrohmer, line 49: I wonder how this would work in Percy 🤔 we had a problem with double-initing components on d.c.c. - maybe it makes sense to guard all this by checking for the `unresolved` class the BaseElement adds?

src/scss/web-components/_web-subscribe.scss
```diff
@@ -25,5 +25,48 @@ web-subscribe {
 
   .g-recaptcha {
     height: 78px;
+
+    &-- {
+      &placeholder {
+        width: 100%;
```
matthiasrohmer, line 31: Maybe my knowledge is outdated but that feels like a wild interpretation of BEM for me 😄 I know, the CSS is a mess already but I would probably go for `.g-recaptcha__placeholder` and `.g-recaptcha__loader-container` and `.g-recaptcha__loader` and probably drop the `g-re` bit and just use `captcha` as the parent name.

src/scss/web-components/_web-subscribe.scss
```diff
@@ -25,5 +25,48 @@ web-subscribe {
...
+          }
+
+          50% {
+            transform: translateX(72px);
+          }
+
+          100% {
+            transform: translateX(0);
+          }
+        }
```
matthiasrohmer, line 62: Love the initiative, but I think it doesn't really fit the web.dev aesthetic. Can you give me a simple greyish or blue open circle spinner instead? 😅  Sorry! Might be good to even extract this to `scss/blocks` to have it reusable

types/utils/global.d.ts
```diff
@@ -24,6 +24,8 @@ interface WebComponentsType {
 declare global {
   interface Window extends Window {
     inert: boolean;
+    recaptchaLoadCallback: () => void;
+    loadScript: (url: string, type?: string) => void;
```
matthiasrohmer, line 28: Techncially, as both are set at runtime they are `recaptchaLoadCallback?` and `loadScript?`, no?

matthiasrohmer (commented):
Left some comments, mostly style stuff, the implemented functionality looks good to me 🙂

src/lib/components/Subscribe/index.js
```diff
@@ -33,10 +33,49 @@ class Subscribe extends BaseElement {
...
+    };
+
+    window.loadScript(
+      'https://www.google.com/recaptcha/api.js?onload=recaptchaLoadCallback&render=explicit',
+    );
+  }
+
   connectedCallback() {
     super.connectedCallback();
+    this.hasInitialised = !this.classList.contains('unresolved');
```
matthiasrohmer, line 70: I am wondering, can we not exit early here if the thing is already inited? So

```
if (!this.classList.contains('unresolved')) {
  return;
}
```

PR #9623, opened 2023-02-20
feat: show parent title under search result page title

### LINKED ISSUE #9507: Improve search results
**Difficult to differenciate search results**
I find it very difficult to choose the right page in search results. This is clearly illustrated in the screenshot below, with several items having the exact same title. 🤭😬 

Screenshot of search results for the phrase 'images'. The results show four identical titles

I think some additional metadata for each item could improve it a lot. Here is a quick mockup to get the creative juices flowing 🤓 Google and Algolia are pretty good at search stuff.
 
image

Blazzike:
Fixes #9507

- Pulls the parent page (by path)'s title and shows it in the search algolia results

**Note** Although local tests have been added and pass, I've been unable to test this functionality until algolia indexes the site - not really sure how to test, might be best to create a second index for a separate deployment?

Here's a screenshot of how it would look:
Screenshot 2023-02-20 at 12 12 55


matthiasrohmer (commented):
You can use the index `test_web_dev` as a playground. You'll need to build the site (`npm run build`), add a .env file to the root with the following tokens:

```
ALGOLIA_APP_ID=
ALGOLIA_API_KEY=
```

and then run `npm run algolia`. I'll send you the credentials via chat. Updating the test screenshots then would be great 🙂  thanks!

matthiasrohmer:
The changes of this PR are now testable via the `test_web_dev` index.

test/unit/src/site/_filters/algolia-item.js
```diff
@@ -5,18 +5,34 @@ const {
...
+/**
+ * Adds parent to memorized posts for following tests
+ */
+const initParent = () => {
+  const mockParent = createPost();
+  mockParent.data.title = 'Parent Title';
+  mockParent.url = '/parent/';
+  memoize([mockParent]);
+};
+
```
matthiasrohmer, line 32: I have a hard time explaining what feels off with this test. Here are the bits: the `findByUrl` module has its own test suite actually, importing another module than the tested one into a test suite is not the best practice I'd say, but things should rather be mocked instead.

Mocking though is what happens above in `createPost` and as the `algolia-item` filter has barely any logic besides prop-getting it is barely testing anything.

I'd vote to extract the 

```js
const parentUrl = `${post.data.page.filePathStem
    .split('/')
    .slice(0, -2)
    .join('/')}/`;
``` 

bit to a separate function and test this instead. And simply leave this test suite here alone. Thanks for thinking about updating it though!

src/site/_filters/algolia-item.js
```diff
@@ -19,13 +19,19 @@ const md5 = require('./md5');
...
  * @param {EleventyCollectionItem} post
  * @return {AlgoliaItem}
  */
 function algoliaItem(post) {
   const data = post.data;
+  const parentUrl = `${post.data.page.filePathStem
+    .split('/')
+    .slice(0, -2)
+    .join('/')}/`;
+
```
matthiasrohmer, line 34: As explained in the other comment, extract to it's own method, add a proper JSDoc comment and maybe add a test case for it, for known paths.

matthiasrohmer (commented):
Looks good, but the test case is off - not really on you though.

test/unit/src/site/_filters/algolia-item.js
```diff
@@ -5,18 +5,35 @@ const {
...
+ * Adds parent to memorized posts for following tests, this is required as get-post-parent-url
+ * is called by algoliaItem(post)
+ */
+const initParent = () => {
+  const mockParent = createPost();
+  mockParent.data.title = 'Parent Title';
+  mockParent.url = '/parent/';
+  memoize([mockParent]);
+};
+
```
matthiasrohmer, line 33: Can we revert this? 🙂 We shouldn't require `memoize` here.

Blazzike: ah this was just because getPostParentURL now requires a parent to be memorised, but I'll just change the function code so it doesn't instead

matthiasrohmer (approved):
Thanks 🎉

PR #9635, opened 2023-02-22
Add a note for Icons in Safari

luciomartinez:

- Highlight that the workaround to make icons work in Safari is no longer needed


AaronForinton:
Hi @luciomartinez thanks for highlighting that this section is outdated. I think we'll just remove it entirely.

AaronForinton:
Confirming for reference with the MDN [compat table](https://developer.mozilla.org/en-US/docs/Web/Manifest#browser_compatibility).

AaronForinton (approved):

PR #9613, opened 2023-02-17
Draft: web.dev/url [DO NOT PUBLISH BEFORE TECHNICAL REVIEW]

samdutton:
An article explaining the names of URL parts.

@heyawhite — as discussed, I've requested you as a reviewer, but feel free to pass this on.

Once this has had tech writer review, I'll get a technical review done.

**[DO NOT PUBLISH BEFORE TECHNICAL REVIEW]**

PR #9643, opened 2023-02-23
Make Widget shortcode more versatile

devnook:
At the moment, {% Widget %} shortcode only accepted patternId as demo source. This meant that an inline demo would need to be created as a code pattern, not a standalone demo, and therefore incur high overhead cost of creating a CodePattern, including it in the patterns collection etc.

This PR allows widget to receive a url for any standalone or web.dev demo instead.

It also checks in a first inline demo under demos/url-parts, which is intended to be used in #9613 article. Code in demos/url-parts/index.html and third_party/ is temporary and intended as a starting point for code that would be checked in together with the article.


src/site/content/en/patterns/example-set/image-demo/assets/sample.html
```diff
@@ -1,2 +0,0 @@
-<p>This sample gets displayed in the code samples area,
```
devnook, line 1: This file was a duplicate and was causing build issues.

src/site/_includes/components/Widget.js
```diff
@@ -1,27 +1,72 @@
...
+  let url;
+  try {
+    // Check if it's external url.
+    const externalUrl = new URL(demoUrl);
+    url = externalUrl.href;
+  } catch (e) {
+    // Otherwise, check if internal url exists.
+    url = path.join('/', demoUrl, '/');
+    if (!findByUrl(url)) {
+      return '';
```
matthiasrohmer, line 39: Been wondering this with other shortcodes too: for amp.dev we rendered an aside in those cases during development saying something like "You're shortcode is not showing anything because of reasons". A simple `console.warn` would also work for me, just so that authors know they gave input that doesn't work for that shortcode.

matthiasrohmer (commented):
Looks good to me, just one author UX comment.

src/site/_data/site.js
```diff
@@ -20,6 +20,7 @@ const {PAGINATION_COUNT} = require('../_utils/constants');
 
 module.exports = {
   env: process.env.ELEVENTY_ENV || 'dev',
+  isProd: () => process.env.ELEVENTY_ENV === 'prod',
```
matthiasrohmer, line 23: 🎉 ☺️

matthiasrohmer (approved):

PR #9661, opened 2023-02-26
Improve learn images CMS/WordPress section

adamsilverstein:
* update intro indicating API has improved over time
* update quality section to `wp_editor_set_quality` which controls quality for all mime types (also raise quality to 70 as 65 is quite low and likely to show degrdation - the default is 82 for JPEG and 86 for WebP currently)
* add a section explaining how to output WebP images using the `image_editor_output_format` filter
* move section explaing how uploads are handled above filter descriptions

### DIFF
--- a/src/site/content/en/learn/images/cms/index.md
+++ b/src/site/content/en/learn/images/cms/index.md
@@ -119,18 +119,29 @@ automated responsive images.
 
 ## Content Management Systems
 
-WordPress was one of the earliest adopters of native responsive images markup, and the API has gone largely unchanged since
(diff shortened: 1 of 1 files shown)


adamsilverstein:
Thanks for this great article. I have offered some slight improvements in this PR for your review.

One other comment is that the call out to `Jetpack's Site Accelerator` feel a little bit like you are favoring one plugin. I would suggest keeping the gist of the paragraph without mentioning a specific image CDN solution. Something like:

"
Perhaps the best way to serve images from WordPress is to connect it to an image CDN which can provide server-side negotiation for encodings (etc EXISTING TEXT)... plugins exist to connect with nearly every image CDN service.
"

rachelandrew (approved):

PR #9727, opened 2023-03-07
replace pseudo-class with pseudo-element

petardotjs:
Just a small mistake in the quiz section - ::backdrop is considered a pseudo-element rather than a pseudo-class.


rachelandrew (approved):
Thank you!

PR #9729, opened 2023-03-08
Make it installable codelab: Add aside to tell the reader to use a standalone tab

### LINKED ISSUE #7146: "Make it installable" PWA codelab doesn't work
The "Make it installable" PWA codelab doesn't work. When you follow the instructions, you don't get an Install button.

**To Reproduce**
1. Go to https://web.dev/codelab-make-installable/
2. Click "Remix to edit"
3. Copy the sample code into line 5, `/* Put code here */`
4. Click "View app" to see it in action.

Here's my sample fork. https://glitch.com/edit/#!/ludicrous-fork-cacao

**Expected behavior**
The "Install" button should appear at the bottom of the screen.

**Actual behavior**
The "Install" button doesn't appear.

**Explanation**
The problem is that the codelab is listening for the `onbeforeinstallprompt` event, but that event can't fire, for three reasons:

1. **The `onbeforeinstallprompt` never fires on `<iframe>`d sites.** By default, web.dev attempts to run the PWA in an `<iframe>` (on a custom glitch.com domain, e.g. https://ludicrous-fork-cacao.glitch.me) inside another `<iframe>` (on glitch.com). To eliminate the top-level `<iframe>`, the codelab would have to link to https://glitch.com/edit/#!/make-it-installable directly, (i.e. launching Glitch in "full screen" mode,) rather than embedding it in the codelab `<iframe>` layout. 
2. **There are two layers of `<iframe>`.** Launching Glitch in "full screen" mode only eliminates _one_ layer of `<iframe>`. The user would then need to click the "..." menu in the upper-right corner of the app preview to "Open in new window" to launch the app in a top-level tab. (I think it would also be wise to include a button in the code lab to pop the app out of its `<iframe>`, like the `requireHTTPS` link that's already there.)
3. **A branch new `glitch.me` domain won't meet the user-engagement heuristic at first.** Even if we do launch the app in a top-level tab, even _then_ the Install button won't show up, because the install criteria for a PWA are that the app "meets a user engagement heuristic." As of Jan 2022, to activate the `onbeforeinstallprompt` event, the user has to click at least once on the page, and remain on the page for at least 30 seconds. So the documentation needs to be updated to tell the codelab user to click somewhere on the page, and wait 30 seconds for the button to show up.

**Screenshots**
image

tomayac:
Fixes #7146.

- Add aside to tell the reader to use a standalone tab.

### DIFF
--- a/src/site/content/en/progressive-web-apps/codelab-make-installable/index.md
+++ b/src/site/content/en/progressive-web-apps/codelab-make-installable/index.md
@@ -54,6 +54,12 @@ To show the install prompt, call `prompt()` on the saved `beforeinstallprompt`
 event. Calling `prompt()` is done in the install button click handler because
 `prompt()` must be called from a user gesture.
 
+{% Aside 'gotcha' %}
+You need to open your app in its own browser tab. The side-by-side preview pane
+in Glitch doesn't work for this codelab, since it will be run in an iframe.
+Choose Glitch's "Preview in a new window" mode instead.
+{% endAside %}
+
 1. Add a click event handler for the install button.
 1. Call `prompt()` on the saved `beforeinstallprompt` event.
 1. Log the results of the prompt.


src/site/content/en/progressive-web-apps/codelab-make-installable/index.md
```diff
@@ -54,6 +54,12 @@ To show the install prompt, call `prompt()` on the saved `beforeinstallprompt`
 event. Calling `prompt()` is done in the install button click handler because
 `prompt()` must be called from a user gesture.
 
+{% Aside 'gotcha' %}
+You need to open your app in its own browser tab. The side-by-side preview pane
+in Glitch doesn't work for this codelab, since it will be run in an iframe.
+Choose Glitch's "Preview in a new window" mode instead.
```
dfabulich, line 60: I think we need more documentation than this, and, specifically, I claim that the codelab itself should be updated.

1. The "Preview in a new window" mode is only visible after you click the "Edit Fullscreen" button. (And if you click "Got It," the words "Edit Fullscreen" go away, and the button is an icon-only button with this icon. https://glitch.com/edit/edit/images/background-light/icons/fullscreen.svg It's likely impossible to find if you don't know what you're looking for or why you're looking for it.
2. Even when you do open in fullscreen mode, the "Preview in a new window" option is only visible after you click the "Preview" button.
3. It's so easy to make a mistake here, that I think the codelab itself should include code to detect whether it's being loaded in an iframe, and provide feedback to the user to explain the problem.

dfabulich:
Even if this merges, this only addresses two of the problems I identified in #7146. The other problem is that the user has to click on the page and wait 30 seconds in order for `onbeforeinstallprompt` to fire.

https://ludicrous-fork-cacao.glitch.me/ doesn't have the `onbeforeinstallprompt` event fire until it meets the [user-engagement heuristics](https://web.dev/install-criteria/), i.e. the user has to click somewhere on the page and wait 30 seconds. (It's very difficult to test this, but the easiest way is to create a fresh new Chrome profile and visit the page in that profile. The install button won't appear until you've waited 30 seconds and clicked somewhere on the page.)

Since this codelab is intended to teach developers how to use `onbeforeinstallprompt`, the codelab code sample should be updated to make it clearer when/why the user-engagement heuristics haven't been met.

src/site/content/en/progressive-web-apps/codelab-make-installable/index.md
tomayac: One problem I see is that describing accurately how it looks on Glitch today may get outdated when they update their UI tomorrow.

dfabulich: I claim the thing to do is to edit the code of the codelab, in the same style as the current requireHTTPS code. It should show a message like this:

```html
    <p id="requireTopLevel" class="hidden">
      <b>STOP!</b> This page <b>must not</b> be served in a frame.
      Please <a target="_top">click here to load the page as the top-level window</a>.
    </p>
```

tomayac:
@beaufortfrancois It looks like you have edit access to https://glitch.com/~make-it-installable. Could I ask you to review the discussion here, and then make the required changes to the codelab? The gist is: the codelab only works as intended if run from its own browser tab, but not if run side-by-side the web.dev article in an iframe. Thanks!

@rachelandrew, as an immediate mitigation, the present PR should work, but I fully agree with @dfabulich that the codelab should reflect the requirements, too, hence looping in François.

rachelandrew (approved):

dfabulich:
@beaufortfrancois If you were to set up the Glitch as a Github project (which I think is possible?) I'd be happy to file a PR.

beaufortfrancois:
I've eventually edited codelab as I wasn't able to export it to GitHub ;(

I've added code below:

```html
    <p id="requireTopLevel" class="hidden">
      <b>STOP!</b> This page <b>must not</b> be served in a frame. Please
      <a target="_top">click here</a> to load the page as the top-level window.
    </p>
```
```js
/**
 * Warn the page must not be served in an iframe.
 */
if (window.self !== window.top) {
  const requireTopLevel = document.getElementById("requireTopLevel");
  const link = requireTopLevel.querySelector("a");
  link.href = window.location.href;
  requireTopLevel.classList.remove("hidden");
}
```

image

tomayac:
Thank you very much, @beaufortfrancois! Also thanks to @dfabulich for reporting the issue in the first place and reviewing the steps! Great team work! `<3`

PR #9731, opened 2023-03-08
Update priority hints browser compat info

tunetheweb:

Updates the [Priority Hints](https://web.dev/priority-hints/) article to:
- Adds Browser Compat data (at the top and also in the Browser Compat section)
- Adds note that these are now part of HTML and Fetch standard even if still marked as experimental on MDN (@pmeenan will this stay that way until another browser implements this, or should this now be removed from MDN?)


malchata (approved):

tunetheweb:
OK cool. Presume you're happy with this PR in the meantime then @pmeenan ?

PR #9745, opened 2023-03-09
feat: clean up all `isDesignSystemContext` conditions

### LINKED ISSUE #9701: Remove `isDesignSystemContext` conditional for rendering old design system markup
Implementation follow-up issue for https://github.com/GoogleChrome/web.dev/issues/9650.

The  `isDesignSystemContext` condition is to detect if the current page was in the design system and if so, the new markup is rendered. 

**List of the components which support both new design and the existing system**

- /_includes/components/Banner.js
- /_includes/components/Compare.js
- /_includes/components/CompareCaption.js

When all of these components have been cleaned up, the following can be safely removed:
- Code block which supports the old system
- The `isDesignSystemContext` conditional and code in that block should run as normal.
- The code from the `/// DELETE THIS WHEN ROLLOUT COMPLETE` comment downwards.
- [Custom collections](https://github.com/GoogleChrome/web.dev/blob/main/.eleventy.js#L158.).
- [Helper](https://github.com/GoogleChrome/web.dev/blob/main/src/lib/utils/is-design-system-context.js).

mamieorine:
Fixes #9701 

- Clean up all `isDesignSystemContext` conditions
- Remove the helper function `is-design-system-context`


jimjiminyjimjim (commented):
lgtm

matthiasrohmer (approved):
Tested comparing build outputs for `main` and this PR, no differences.

PR #9757, opened 2023-03-11
Fix minor typo

rahulrao0209:
Fix minor typos by adding a question mark.

- Add a question mark.


jvilders:
There was no typo. Adding the question mark would make the sentence incorrect for two reasons:

- The sentence does not _ask_ how something can be done, it instructs how to do it, so a question mark is inappropriate for the intent
- If it were a question, the structure of the sentence would be wrong: "How to validate in real-time?" is not grammatically correct English for a question. It would need to be something like "How [can/do/should] I validate in real-time?" or "How [can/do/should] validation be done in real-time?" instead.

rahulrao0209:
Even though I believe "How to validate in real-time?" to be grammatically fine, thanks for the clarification. 
Closing the pull request as this is not considered to be a typo.

PR #9828, opened 2023-03-28
Optimizing CLS updates

tunetheweb:

Updates the [Optimize Cumulative Layout Shift](https://web.dev/optimize-cls/) to refresh it in line with our current recommendations and bring it more inline with [Optimize Largest Contentful Paint](https://web.dev/optimize-lcp/) and [Optimize Interaction to Next Paint](https://web.dev/optimize-inp/) by making the following changes:

- Adds section on "Understanding where your shifts are coming from" (includes the [Developer Tools section](https://web.dev/optimize-cls/#developer-tools) from bottom of old version).
- Moves most of the previous content under "Common causes of CLS" section.
- Merges Ads, Embeds and Dynamic content into one section as all have basically the optimization techniques
- Removes two of the video examples ([this one](https://web.dev/optimize-cls/#ads,-embeds-and-iframes-without-dimensions:~:text=Ads%20without%20sufficient%20space%20reserved.) and [this one](https://web.dev/optimize-cls/#ads,-embeds-and-iframes-without-dimensions:~:text=Embed%20without%20space%20reserved.)), as take up a lot of space without adding huge amount compared to other videos and diagrams that already exist (sorry!)
- Trims some of the Ad/Embed/Dynamic content text that I felt was not necessary for understanding how to optimize CLS.
- Refreshes the fonts section
- Moves BFcache to it's own section as a little different than the other techniques.
- Moves myself to primary author/contact person (keeping @addyosmani as secondary as still lots and lots of his content in this) and removes "thanks" line from bottom (sorry!) as content has changed a fair bit so don't think their initial reviews still stand.


src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+{% Img src="image/W3z1f5ZkBJSgL1V1IfloTIctbIF3/2C3v6dGwPx2yFyYpVdue.png", alt="Lighthouse Screenshot showing the CLS audits providing more information to help you identify and address CLS issues", width="800", height="544" %}
+
+{% Aside important %}
+  Lighthouse will identify the elements that were shifted, but often these are the ones _impacted_ rather than the elements _causing_ the CLS. For example, if a new element is inserted into the DOM the elements that are beneath it will show in this audit, but the root cause is the addition of the new element above. Hopefully, the shifted element will be sufficient to help you identify and resolve the root cause.
+{% endAside %}
+
+### Identifying non-load CLS issues
+
+When the CrUX and Lighthouse CLS scores of PageSpeed Insights are not inline, then this likely indicates post-load CLS. To replicate that, you can browse your web page while [recording CLS using a JavaScript snippet](/cls/#measure-cls-in-javascript) pasted into the console.
```
malchata, line 94:
```suggestion
When the CrUX and Lighthouse CLS scores of PageSpeed Insights are not in line, then this likely indicates post-load CLS. To replicate that, you can browse your web page while [recording CLS using a JavaScript snippet](/cls/#measure-cls-in-javascript) pasted into the console.
```

malchata: Not sure if you want to cover this, but the snippets feature in DevTools is a good use for this and avoids having to repeatedly copying/pasting into the console.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+Alternatively, the Web Vitals Chrome [extension](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma) can be used to monitor CLS as you interact with a page, either in a heads up display, or in the console.
```
malchata, line 96:
```suggestion
Alternatively, the [Web Vitals Chrome extension](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma) can be used to monitor CLS as you interact with a page, either in a heads up display, or in the console.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+The [Performance panel](https://developer.chrome.com/docs/devtools/evaluate-performance/) in DevTools highlights layout shifts in the **Experience** section. The **Summary** view for a `Layout Shift` record includes the cumulative layout shift score as well as a rectangle overlay showing the affected regions.
+
+<figure>
+  {% Img src="image/admin/ApDKifKCRNGWI2SXSR1g.jpg", alt="Layout Shift records being displayed in the Chrome DevTools performance panel when expanding the Experience section", width="800", height="438" %}
+  <figcaption>After recording a new trace in the Performance panel, the <b>Experience</b> section of the results is populated with a red-tinted bar displaying a <code>Layout Shift</code> record. Clicking the record allows you to drill down into impacted elements (e.g. note the moved from/to entries).</figcaption>
+</figure>
+
+The newer [Performance Insights panel](https://developer.chrome.com/docs/devtools/performance-insights/) also includes full support for identifying shifts.
```
malchata, line 105: Might not be worth calling out the perf insights panel, as it's my understanding that a replacement/substantial upgrade for the performance panel is in the works and may end up with the removal of the performance insights panel.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+### Measuring CLS elements in the field
+
+It is also possible to measure both the CLS, and perhaps more importantly, the elements impacting your CLS score in the field and feed them back to your analytics service.
+
+This allows the users seeing the impact of CLS to provide the data needed to identify these shifts. This can help point you in the right direction of where the issue is, and also rank the issues in order of important based on most frequently experienced. However, like Lighthouse, this will measure the elements that shifted, rather than the root causes of those shifts.
```
malchata, line 111: > This allows the users seeing the impact of CLS to provide the data needed to identify these shifts.

Can this sentence be rewritten? I had to reread this a few times before it made sense to me.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+Read our [Debug performance in the field](/debug-performance-in-the-field/) post for more information on how to do this.
+
+{% Aside %}
+  RUM solutions that measure CLS in the field may show differences that CrUX data as explained in the [Why is CrUX data different from my RUM data?](crux-and-rum-differences/) post. In particular CLS that happens in iframes is not measurable from Web APIs but is visible to the user, so is included in CrUX. So while field data can be invaluable for identifying CLS issues, be aware that it may be incomplete for certain scenarios.
```
malchata, line 116:
```suggestion
  RUM solutions that measure CLS in the field may show differences that CrUX data as explained in the [Why is CrUX data different from my RUM data?](crux-and-rum-differences/) post. In particular, CLS that happens in iframes is not measurable from Web APIs but is visible to the user, and is therefore included in CrUX. So while field data can be invaluable for identifying CLS issues, be aware that it may be incomplete in certain scenarios.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -74,12 +155,12 @@ The most common causes of a poor CLS are:
...
+#### History
 
 In the early days of the web, developers would add `width` and `height` attributes to their `<img>` tags to ensure sufficient space was allocated on the page before the browser started fetching images. This would minimize reflow and re-layout.
 
 ```html
-<img src="puppy.jpg" width="640" height="360" alt="Puppy with balloons" />
+<img src="puppy.jpg" width="640" height="360" alt="Puppy with balloons">
 ```
 
 You may notice `width` and `height` above do not include units. These "pixel" dimensions would ensure a 640x360 area would be reserved. The image would stretch to fit this space, regardless of whether the true dimensions matched or not.
```
malchata, line 166: Might be worth clarifying this and refer to the concept of aspect ratio, as that's what the unitless values provide.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -104,13 +185,13 @@ This means if we know one of the dimensions, the other can be determined. For a
 
 Knowing the aspect ratio allows the browser to calculate and reserve sufficient space for the height and associated area.
 
-### Modern best practice
+#### Modern best practice
 
 Modern browsers now set the default aspect ratio of images based on an image's width and height attributes so it's valuable to set them to prevent layout shifts. Thanks to the CSS Working Group, developers just need to set `width` and `height` as normal:
```
malchata, line 190: Kind of odd to mention CSS WG here. I'd drop the mention of the WG and focus on the solution itself.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -180,138 +259,46 @@ Chrome, Firefox and Safari now support setting `width` and `height` on the `sour
 </picture>
 ```
 
-## Ads, embeds and iframes without dimensions
+### Ads, embeds, and other late-loaded content
 
-As we have seen images have special considerations, but these are not the only content that can cause layout shifts.
-
-### Advertisements
+As we have seen images have special considerations, but these are not the only content that can cause layout shifts. Ads, embeds, iframes, and other dynamically injected content can all cause content after these to shift down, increasing your CLS.
```
malchata, line 264:
```suggestion
As we have seen, images have special considerations. However, images are not the only type of content that can cause layout shifts. Ads, embeds, iframes, and other dynamically injected content can all cause content after these to shift down, increasing your CLS.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -180,138 +259,46 @@ Chrome, Firefox and Safari now support setting `width` and `height` on the `sour
...
-    Lighthouse impact of reserving space for this banner on CLS
-  </figcaption>
-</figure>
+The techniques for dealing with these are all similar. The major differences are how much control you have over the content that will be inserted. If this is inserted by a third-party like an ad partner you may not know the exact size of content that will be inserted, nor be able to control any CLS happening within those embeds.
 
-#### Statically reserve space for the ad slot
+#### Statically reserve space for late-loading content
 
-Statically style slot DOM elements with the same sizes passed to your tag library. This can help ensure the library doesn't introduce layout shifts when it loads. If you don't do this, the library may change the size of the slot element after page layout.
+If placing late-loading content in the content flow, ensure shifts are eliminated by reserving the space for them in the initial layout.
```
malchata, line 274: "Eliminate" is strong language here, as you're not entirely eliminating the layout shift. My vote is for "reduce".

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -180,138 +259,46 @@ Chrome, Firefox and Safari now support setting `width` and `height` on the `sour
...
 
-This can be as simple as adding a `min-height` styling to reserve space or, for responsive ad sizings using the new [`aspect-ratio`](/aspect-ratio/) CSS property in a similar manner to the way browsers automatically use this for images with dimensions provided.
+This can be as simple as adding a `min-height` styling to reserve space or, for responsive content such as ads, using the new [`aspect-ratio`](/aspect-ratio/) CSS property in a similar manner to the way browsers automatically use this for images with dimensions provided.
```
malchata, line 276:
```suggestion
This can be done by adding a `min-height` styling to reserve space or, for responsive content such as ads, using the [`aspect-ratio`](/aspect-ratio/) CSS property.
```

malchata: Removed trivializing language, also I was hitching trying to read the end of the sentence, and it didn't seem necessary IMO.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -180,138 +259,46 @@ Chrome, Firefox and Safari now support setting `width` and `height` on the `sour
...
   <figcaption>
     Reserving space for ads can prevent layout shifts
   </figcaption>
 </figure>
 
-Also consider the sizes of smaller ad serves. If a smaller ad is served, a publisher can style the (larger) container to avoid layout shifts. The downside to this approach is that it will increase the amount of blank space, so keep in mind the trade-off here.
+You may need to account for subtle differences in ad/placeholder sizes between different form factors using media queries.
 
-Alternatively, set the initial size to the smallest ad size, and accept some level of shift for larger ads. This will not fully eliminate CLS, but will hopefully reduce the impact of it to a more managable level.
+For content that may not have a fixed height—like ads you may not be able to reserve the exact correct amount. If a smaller ad is served, a publisher can style the (larger) container to avoid layout shifts. Or choose the most likely size for the ad slot based on historical data. The downside to this approach is that it will increase the amount of blank space, so keep in mind the trade-off here.
```
malchata, line 287:
```suggestion
For content that may not have a fixed height—like ads—you may not be able to reserve the exact amount of space needed to eliminate the layout shift entirely. If a smaller ad is served, a publisher can style the (larger) container to avoid layout shifts, or choose the most likely size for the ad slot based on historical data. The downside to this approach is that it will increase the amount of blank space, so keep in mind the trade-off here.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -333,14 +320,13 @@ You've probably experienced layout shifts due to UI that pops-in at the top or b
     </figcaption>
   </figure>
 
-If you need to display these types of UI affordances, reserve sufficient space in the viewport for it in advance (for example, using a placeholder or skeleton UI) so that when it loads, it does not cause content in the page to surprisingly shift around.
+If you need to display these types of UI affordances, reserve sufficient space in the viewport for it in advance (for example, using a placeholder or skeleton UI) so that when it loads, it does not cause content in the page to surprisingly shift around. Alternatively, ensure the element is not part of the document flow by overlaying the content where this makes sense. See our [Best practices for cookie notices](/cookie-notice-best-practices/) post for more recommendations on these types of components.
```
malchata, line 323:
```suggestion
If you need to display these types of UI affordances, reserve sufficient space in the viewport for it in advance (for example, using a placeholder or skeleton UI) so that when it loads, it does not cause content in the page to surprisingly shift around. Alternatively, ensure the element is not part of the document flow by overlaying the content where this makes sense. See the [Best practices for cookie notices](/cookie-notice-best-practices/) post for more recommendations on these types of components.
```

malchata: Style guide says to avoid first person.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+Changes to CSS property values can require the browser to react to these changes. A number of values trigger re-layout, paint and composite such as `box-shadow` and `box-sizing`. Try to avoid animating these.
 
-Read [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](/preload-optional-fonts/) for more details, and [Best practices for fonts](/font-best-practices/) for other font best practices.
+A number of CSS properties can be changed in a much more performant manner. For example, `transform` animations can be used to `translate`, `scale`, `rotate`, or `skew` without triggering a re-layout and so completely avoiding CLS.
 
-### Animations
+When animations are instead done but changing `top` and `left` instead of using `transform: translate` it causes CLS. This happens **even when the element being moved is in it's own layer and so does not cause shifts to other elements**. The net effect to the user may be the same, but one is counted as CLS, and one is not. This may seem like CLS should be changed to treat these the same, but there are other considerable performance benefits of using non-composited animations.
```
malchata, line 348:
```suggestion
When animations are instead done by changing `top` and `left` CSS propertes instead of using `transform: translate`, layout shiftrs occur. This happens **even when the element being moved is in it's own layer and so does not cause shifts to other elements**. The net effect to the user may be the same, but one is counted as CLS, and one is not. This may seem like CLS should be changed to treat these the same, but there are other considerable performance benefits of using non-composited animations.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+Downloading and rendering web fonts is typically handled in one of two ways before the web font is downloaded:
 
-### bfcache
+- The fallback font is swapped with web font (FOUT—flash of unstyled text)
+- "Invisible" text is displayed using the fallback font until a web font is available the text is made visible (FOIT—flash of invisible text)
 
-The back/forward cache, or bfcache keeps pages in browsers memory for a short period so if you return to them, then they will be restored exactly as you left them. This means the fully loaded page is instantly available—without any shifts which may be normally seen during load to to any of the reason above. When this was rolled out to Chrome, we saw [noticeable improvements in CLS](https://twitter.com/anniesullie/status/1491399685961293828?s=20&t=Qj_nzSRZD0_c-HaAnfr98Q).
+It is important to understand that **both of these cause CLS**. Even though the invisible is not visible to the user, it is layed out using the fallback font and so the text block of the font, and the surrounding content shifts when the web font loads—in the exact same way as for the visible font for FOUT.
```
malchata, line 359:
```suggestion
It is important to understand that **both of these cause CLS**. Even though the text is invisible, it is laid out using the fallback font, so the text block of the font and the surrounding content shifts when the web font loads—in the exact same way as for the visible font for FOUT.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
 
-Some sites are ineligible for the bfcache due to a variety of reasons. Ensuring your site is eligible is a sure fire way to keep CLS scores down. Read [the bfcache article](/bfcache/) for more details on how to test and identify any issues preventing bfcache usage.
+The following tools can help you minimize this:
 
-### Developer Tools
+- `font-display: optional` can avoid a re-layout as the web font is only used if it is available by the time of initial layout.
```
malchata, line 363:
```suggestion
- `font-display: optional` can avoid a layout invalidation as the web font is only used if it is available by the time of initial layout.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+- Ensuring the appropriate fallback font is used. For example, using `font-family: "Google Sans",sans-serif;` will ensure the browsers `sans-serif` fallback font is used while `"Google Sans"` is loaded. Not specifying a fallback font using just `font-family: "Google Sans"` will mean the default font is used, which on Chrome is "Times"—a serif font which is a worse match than the default `san-serif` font.
```
malchata, line 364:
```suggestion
- Ensure the appropriate fallback font is used. For example, using `font-family: "Google Sans", sans-serif;` will ensure the browser's `sans-serif` fallback font is used while `"Google Sans"` is loaded. Not specifying a fallback font using just `font-family: "Google Sans"` will mean the default font is used, which on Chrome is "Times"—a serif font which is a worse match than the default `sans-serif` font.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+- Using `<link rel=preload>` on the key web fonts: a preloaded font will have a higher chance to meet the first paint, in which case there's no layout shifting.
 
-There are a number of tools available to measure and debug Cumulative Layout Shift (CLS).
+Read [Best practices for fonts](/font-best-practices/) for other font best practices.
 
-[Lighthouse](https://developer.chrome.com/docs/lighthouse/overview/) includes support for measuring CLS in a lab setting. It will also highlight the nodes that cause the most layout shifting.
+## Ensure pages are eligible for bfcache
 
-{% Img src="image/tcFciHGuF3MxnTr1y5ue01OGLBn2/J11KOGFVAOjRMdihwX5t.jpg", alt="Lighthouse includes support for measuring CLS in the metrics section", width="800", height="309" %}
+One sure fire way to keep CLS scores down is by ensuring your web pages are eligible for the back/forward cache (bfcache).
```
malchata, line 373:
```suggestion
A highly effective technique for keeping CLS scores low is to ensure your web pages are eligible for the back/forward cache (bfcache).
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+The bfcache keeps pages in browsers memory for a short period after you navigate away so if you return to them, then they will be restored exactly as you left them. This means the fully loaded page is instantly available—without any shifts which may be normally seen during load due to any of the reasons above.
 
-<figure>
-  {% Img src="image/admin/ApDKifKCRNGWI2SXSR1g.jpg", alt="Layout Shift records being displayed in the Chrome DevTools performance panel when expanding the Experience section", width="800", height="438" %}
-  <figcaption>After recording a new trace in the Performance panel, the <b>Experience</b> section of the results is populated with a red-tinted bar displaying a <code>Layout Shift</code> record. Clicking the record allows you to drill down into impacted elements (e.g. note the moved from/to entries).</figcaption>
-</figure>
+While this does potentially still mean the initial page load encounters CLS, when a user goes back through pages they are not seeing the same CLS again and again. Back and forward navigations are pretty common on many sites. For example, returning to a contents page, or a category page, or search results.
```
malchata, line 377:
```suggestion
While this does potentially still mean the initial page load encounters CLS, when a user goes back through pages they are not seeing the same layout shifts repeatedly. Back and forward navigations are common on many sites. For example, returning to a contents page, or a category page, or search results.
```

malchata (changes requested):

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+- Web Fonts
+
+{% Aside %}
+  For a visual overview of some of the context presented in this guide, see this section of the Optimize for Core Web Vitals from Google I/O 2020:
+  {% YouTube id="AQqFZ5t8uNc", startTime='88', width=2482, height=1396 %}
+{% endAside %}
+
+## Understanding where your shifts are coming from
+
+Before we start looking at solutions to common CLS issues, it's always important to understand your CLS score and where the shifts are coming from. A big part of the problem is understanding your CLS score—the fix afterwards if often the easier part!
```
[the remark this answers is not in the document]
tunetheweb: Sentence ends abruptly without that. I also think saying "is often" is not trivializing. Saying "the fix it easy" would be I agree, but the "the fix is often the easier part" doesn't come across as trivializing (just cause something is easier doesn't mean it's easy!).

Would you mind having another read and see if you still think it's something we need to change @malchata ?

Though did notice a typo:

```suggestion
Before we start looking at solutions to common CLS issues, it's always important to understand your CLS score and where the shifts are coming from. A big part of the problem is understanding your CLS score—the fix afterwards is often the easier part!
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+### CLS in lab tools versus field
+
+It is quite common to hear developers think the CLS measured by the [Chrome User Experience Report (CrUX)](https://developer.chrome.com/docs/crux/) is incorrect as it does not match the CLS they measure using Chrome DevTools or lab-based testing tools. Many lab-based web performance tools like Lighthouse may not show the full CLS of a page, as they do a simple load of the page to measure some web performance metrics and make some guidance.
 
-## Images without dimensions
+CrUX is the official dataset of the Web Vitals program and for that CLS is measured throughout the full life of the page and not just during the inital page load that these tools typically measure.
```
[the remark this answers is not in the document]
tunetheweb: At the minute [the CrUX docs](https://developer.chrome.com/docs/crux/about/) calls it the Web Vitals program:

> CrUX is the official dataset of the [Web Vitals](https://web.dev/vitals/) program. All user-centric Core Web Vitals metrics will be represented in the dataset.

@rviscomi what's your take on this, given the discussions you've been having recently? Happy to change this here to the "Core Web Vitals program" as long as that's the agreed direction and the other docs will follow at some point.

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: I thought about this, but let's leave it in for now until we get more clarity as to where that's going.

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: Me too! I don't think we even need this sentence.

```suggestion
This can help point you in the right direction of where the issue is, and also rank the issues in order of important based on most frequently experienced. However, like Lighthouse, this will measure the elements that shifted, rather than the root causes of those shifts.
```

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: It's coming in a few paragraphs. I left this part of the doc largely untouched and am comfortable with that.

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: Was existing text but OK to change this:

```suggestion
Modern browsers now set the default aspect ratio of images based on an image's `width` and `height` attributes so developers just need to set these, and include the above CSS, to prevent layout shifts:
```

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: Let's go with "can be avoided":

```suggestion
If placing late-loading content in the content flow, layout shifts can be avoided by reserving the space for them in the initial layout.
```

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
rviscomi: Yeah, "Web Vitals" is how we refer to the program/initiative.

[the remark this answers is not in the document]
tunetheweb: OK let's commit the other changes:

```suggestion
CrUX is the official dataset of the Web Vitals program, and for that, CLS is measured throughout the full life of the page and not just during the initial page load that lab tools typically measure.
```

src/site/content/en/blog/optimize-cls/index.md
malchata: Your call. If you feel strongly about it, then go with it.

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: Let's keep this. We use "re-layout" in a few places in this doc (and did in the old version) and "invalidation" sounds kinda technical.

malchata (approved):
LGTM!

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+  {% YouTube id="AQqFZ5t8uNc", startTime='88', width=2482, height=1396 %}
+{% endAside %}
+
+## Understanding where your shifts are coming from
+
+Before we start looking at solutions to common CLS issues, it's always important to understand your CLS score and where the shifts are coming from. A big part of the problem is understanding your CLS score—the fix afterwards is often the easier part!
+
+### CLS in lab tools versus field
+
+It is quite common to hear developers think the CLS measured by the [Chrome User Experience Report (CrUX)](https://developer.chrome.com/docs/crux/) is incorrect as it does not match the CLS they measure using Chrome DevTools or other lab tools. Web performance lab tools like Lighthouse may not show the full CLS of a page, as they do a simple load of the page to measure some web performance metrics and provide some guidance.
```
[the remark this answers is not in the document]
tunetheweb: Changed to:

"Web performance lab tools like Lighthouse may not show the full CLS of a page as they typically do a simple load of the page to measure some web performance metrics and provide some guidance (though [Lighthouse user flows](/lighthouse-user-flows/) do allow you measure beyond the default page load audit)."

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+The [Performance panel](https://developer.chrome.com/docs/devtools/evaluate-performance/) in DevTools highlights layout shifts in the **Experience** section. The **Summary** view for a `Layout Shift` record includes the cumulative layout shift score as well as a rectangle overlay showing the affected regions.
+
+<figure>
+  {% Img src="image/admin/ApDKifKCRNGWI2SXSR1g.jpg", alt="Layout Shift records being displayed in the Chrome DevTools performance panel when expanding the Experience section", width="800", height="438" %}
+  <figcaption>After recording a new trace in the Performance panel, the <b>Experience</b> section of the results is populated with a red-tinted bar displaying a <code>Layout Shift</code> record. Clicking the record allows you to drill down into impacted elements (e.g. note the moved from/to entries).</figcaption>
+</figure>
+
+The newer [Performance Insights panel](https://developer.chrome.com/docs/devtools/performance-insights/) also includes full support for identifying shifts.
```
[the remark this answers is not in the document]
tunetheweb: I'd lead with the Web Vitals extension (I originally wanted to de-emphasis that) but, as discussed, I still believe web snippets are an important tool for those that do not wish to use this tool and believe we should be support developers that are comfortable with these (particularly more advanced developers), rather than forcing them to install an extension.

I've replaced the Performance Insights panel section with a section on Lighthouse user flows.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+### Measuring CLS elements in the field
```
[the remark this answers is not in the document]
tunetheweb: Expanded on the middle section a bit:

>The [attribution functionality of the web-vitals library](https://github.com/GoogleChrome/web-vitals#send-attribution-data) allows this additional information to be collected. Read our [Debug performance in the field](/debug-performance-in-the-field/) post for more information on how to do this. Other RUM providers have also started collecting and presenting this data similarly.

I think the flow of "here's how to identify, here's how to get your users to identify" works better. Also is the pattern we should really be encouraging to CLS issues are caught before making it to the field.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -180,138 +259,46 @@ Chrome, Firefox and Safari now support setting `width` and `height` on the `sour
...
 
-- HTML fallback and a JavaScript tag transforming the fallback into a fancy embed
-- Inline HTML snippet
-- iframe embed
+Conversely, dynamically injected content near the middle of the viewport may not shift as many elements as the content above it is less likely to move, but will still cause some CLS. Even content injected at the bottom of the screen will cause CLS as the content it replace is moved off-screen.
 
-These embeds often aren't aware in advance just how large an embed will be (for example, in the case of a social media post - does it have an embedded image? video? multiple rows of text?). As a result, platforms offering embeds do not always reserve enough space for their embeds and can cause layout shifts when they finally load.
+The ideal scenario is not to shift any other content so reserving the appropriate space is preferred, but where this is not possible, minimizing the shifts can at least reduce the impact—both to your users and your CLS scores.
```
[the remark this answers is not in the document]
tunetheweb: Like the split of the sentence but going to drop the "But".

> The ideal scenario is not to shift any other content so reserving the appropriate space is preferred. Where this is not possible, minimizing the shifts can at least reduce the impact—both to your users and your CLS scores.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+Changes to CSS property values can require the browser to react to these changes. A number of values trigger re-layout, paint and composite such as `box-shadow` and `box-sizing`. Try to avoid animating these.
 
-Read [Prevent layout shifting and flashes of invisible text (FOIT) by preloading optional fonts](/preload-optional-fonts/) for more details, and [Best practices for fonts](/font-best-practices/) for other font best practices.
+A number of CSS properties can be changed in a much more performant manner. For example, `transform` animations can be used to `translate`, `scale`, `rotate`, or `skew` without triggering a re-layout and so completely avoiding CLS.
 
-### Animations
+When animations are instead done by changing `top` and `left` CSS propertes instead of using `transform: translate`, layout shifts occur. This happens **even when the element being moved is in it's own layer and so does not cause shifts to other elements**. The net effect to the user may be the same, but one is counted as CLS, and one is not. This may seem like CLS should be changed to treat these the same, but there are other considerable performance benefits of using non-composited animations.
```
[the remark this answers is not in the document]
tunetheweb: Changed it to this:

> When animations are instead done by changing `top` and `left` CSS properties instead of using `translate`, layout shifts occur. This happens **even when the element being moved is in it's own layer and so does not cause shifts to other elements**. Composited animations via `translate` are exempt from CLS as they cannot impact other elements. There are also other considerable performance benefits of using non-composited animations since they do no cause re-layout and therefore are much less work for the browser.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
 
-**Summary:** Prefer `transform` animations to animations of properties that trigger layout changes.
+To learn more about what CSS properties trigger layout see [High-performance animations](/animations-guide/).
```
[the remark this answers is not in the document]
tunetheweb: They don't necessarily result in changes, but they do trigger the expensive layout processing, so would prefer to keep as is (this is also from the original article.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+It is important to understand that **both of these cause CLS**. Even though the text is invisible, it is laid out using the fallback font. This means the text block using the font, and the surrounding content, shifts when the web font loads—in the exact same way as for the visible font for FOUT.
 
-Some sites are ineligible for the bfcache due to a variety of reasons. Ensuring your site is eligible is a sure fire way to keep CLS scores down. Read [the bfcache article](/bfcache/) for more details on how to test and identify any issues preventing bfcache usage.
+The following tools can help you minimize this:
 
-### Developer Tools
+- `font-display: optional` can avoid a re-layout as the web font is only used if it is available by the time of initial layout.
+- Ensure the appropriate fallback font is used. For example, using `font-family: "Google Sans", sans-serif;` will ensure the browser's `sans-serif` fallback font is used while `"Google Sans"` is loaded. Not specifying a fallback font using just `font-family: "Google Sans"` will mean the default font is used, which on Chrome is "Times"—a serif font which is a worse match than the default `sans-serif` font.
```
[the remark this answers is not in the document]
tunetheweb: Example here where it did cause noticeable CLS: https://groups.google.com/a/chromium.org/g/chrome-ux-report/c/Te3fs6zhoc0/m/3861Cb37CAAJ

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -349,62 +335,53 @@ In some cases adding content dynamically is an important part of user experience
...
+- Closing the differences between the fallback font and the web font using the new `size-adjust`, `ascent-override`, `descent-override`, and `line-gap-override` APIs as detailed in the [Improved font fallbacks](https://developer.chrome.com/blog/font-fallbacks/) post.
```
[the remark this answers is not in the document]
tunetheweb: It's not necessarily about make the fonts geometrically similar, but more about making the text blocks similar, so prefer my wording.

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb: > The biggest structural change seems to be the two new CLS phases.

Well I think the biggest structural change is the identifying section, rather than this being a list of CLS patterns. But yes I definitely did want to emphasis those two types as it's a very common source of misunderstanding.

> Each one is referenced a couple of different ways throughout the post, so worth a closer look to make it all consistent

Cheers - fixed.

> Also, I think it'd be useful to define these phases in the intro section and what the boundary is between the two (soonest of onload or user interaction?).

Well that whole "CLS in lab tools versus field" section was supposed to be that introduction. And I'm basically differentiating them as whether a lab-based tool like Lighthouse or PSI identifies them, or whether a developer needs to do some more digging themselves.

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
rviscomi: My suggested change was to clarify what "differences" are closed by the APIs and to make it a more complete sentence. Could you address those things in your own wording?

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,114 @@ tags:
...
+
+{% Aside important %}
+  Lighthouse will identify the elements that were shifted, but often these are the ones _impacted_ rather than the elements _causing_ the CLS. For example, if a new element is inserted into the DOM, the elements that are beneath it will show in this audit, but the root cause is the addition of the new element above. Hopefully, the shifted element will be sufficient to help you identify and resolve the root cause.
+{% endAside %}
+
+### Identifying post-load CLS issues
+
+When the CrUX and Lighthouse CLS scores of PageSpeed Insights are not in line, then this likely indicates post-load CLS. To replicate that, the [Web Vitals Chrome extension](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma) can be used to monitor CLS as you interact with a page, either in a heads up display, or in the console.
+
+Alternatively, you can browse your web page while [recording CLS using a JavaScript snippet](/cls/#measure-cls-in-javascript) pasted into the console.
```
rviscomi, line 97: WDYT about linking "monitor CLS" in the previous paragraph to the snippet source? That addresses my concerns while I think addressing yours about supporting advanced use cases.

tunetheweb: Sorry I don’t understand what you  mean by linking to the snippet source? Do you mean link to the web-vitals extension source instead of the CLS docs?

rviscomi, line 97:
```suggestion
When the CrUX and Lighthouse CLS scores of PageSpeed Insights are not in line, then this likely indicates post-load CLS. To replicate that, the [Web Vitals Chrome extension](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma) can be used to [monitor CLS](/cls/#measure-cls-in-javascript) as you interact with a page, either in a heads up display, or in the console.
```

rviscomi: Suggested the change

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb:
```suggestion
- Minimize the size differences between the fallback font and the web font using the new `size-adjust`, `ascent-override`, `descent-override`, and `line-gap-override` APIs as detailed in the [Improved font fallbacks](https://developer.chrome.com/blog/font-fallbacks/) post.
```

src/site/content/en/blog/optimize-cls/index.md
tunetheweb: Would prefer to keep this separate for those that cannot install the extension, and to allow those to monitor all layout shifts (and not just those that result in a larger CLS as the extension does.).

tunetheweb: I think that's more confusing. I think we've made it clear here the preferred way is the extension but think we should offer the alternative for those that cannot use it ([eg1](https://twitter.com/DanShappir/status/1642430405813870592?s=20), [eg2](https://webperformance.slack.com/archives/C04BK7K1X/p1678197673330799?thread_ts=1678197553.113529&cid=C04BK7K1X)), or do not want to use the extension.

tunetheweb: In #9845 we've removed the full, complex, CLS snippet and refer readers to the web-vitals source code instead for that. We have, however, added a simpler, layout shifts snippet to that code and this PR refers to that.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
 
-**Summary:** Always include `width` and `height` size attributes on your images and video elements. Alternatively, reserve the required space with [CSS aspect ratio boxes](https://css-tricks.com/aspect-ratio-boxes/). This approach ensures that the browser can allocate the correct amount of space in the document while the image is loading.
+CLS is very common during page load, as all the necessary resources are fetched to initially render the page, but CLS can also happen after the initial load. Many post-load shifts may occur [as the result of a user interaction](/cls/#user-initiated-layout-shifts) and therefore will be excluded from the CLS score as they are expected shifts—as long as they occur within 500 milliseconds of that interaction.
+
+However, other post-load shifts that are unexpected by the user may be included where there was no qualifying interaction. For example, as you scroll down the page and lazy loaded content is loaded. Other common causes of post-load CLS are on interactions of transitions—for example on Single Page Apps—which take longer than the 500 millisecond grace period.
```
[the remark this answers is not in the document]
rviscomi: Lazy-load sentence is still incomplete

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
rviscomi: Unresolving to continue the offline discussion about the emphasis on JS snippets.

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+  <figcaption>After recording a new trace in the Performance panel, the <b>Experience</b> section of the results is populated with a red-tinted bar displaying a <code>Layout Shift</code> record. Clicking the record allows you to drill down into impacted elements (e.g. note the moved from/to entries).</figcaption>
+</figure>
+
+The newer [Performance Insights panel](https://developer.chrome.com/docs/devtools/performance-insights/) also includes full support for identifying shifts.
+
+### Measuring CLS elements in the field
+
+It is also possible to measure both the CLS, and perhaps more importantly, the elements impacting your CLS score in the field and feed them back to your analytics service.
+
+This can help point you in the right direction of where the issue is, and also rank the issues in order of important based on most frequently experienced. However, like Lighthouse, this will measure the elements that shifted, rather than the root causes of those shifts.
```
rviscomi, line 112:
```suggestion
This can help point you in the right direction of where the issue is, and also rank the issues in order of importance based on most frequently experienced. However, like Lighthouse, this will measure the elements that shifted, rather than the root causes of those shifts.
```

src/site/content/en/blog/optimize-cls/index.md
```diff
@@ -16,32 +16,113 @@ tags:
...
+
+Read our [Debug performance in the field](/debug-performance-in-the-field/) post for more information on how to do this.
+
+{% Aside %}
+  RUM solutions that measure CLS in the field may show differences that CrUX data as explained in the [Why is CrUX data different from my RUM data?](crux-and-rum-differences/) post. In particular, CLS that happens in iframes is not measurable from Web APIs but is visible to the user, and is therefore included in CrUX. So while field data can be invaluable for identifying CLS issues, be aware that it may be incomplete in certain scenarios.
```
rviscomi:
```suggestion
  RUM solutions that measure CLS in the field may show differences from CrUX data as explained in the [Why is CrUX data different from my RUM data?](crux-and-rum-differences/) post. In particular, CLS that happens in iframes is not measurable from web APIs but is visible to the user, and is therefore included in CrUX. So while field data can be invaluable for identifying CLS issues, be aware that it may be incomplete in certain scenarios.
```

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
rviscomi: The preceding section on post-load CLS lists all the different ways to diagnose post-load issues, but it doesn't make clear which tool(s) to use for the job or when. And given that this section follows it, a reader wouldn't yet know what they're supposed to be doing in these tools to find the post-load issues.

So I still think it'd be helpful to rearrange the sections, so that we set the stage with collecting CLS attribution info in the field, then talk about using that info to reproduce layout shifts experienced by real users. That way it's a bit more methodical than interacting with the page to see what happens.

rviscomi (dismissed):

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb: I disagree. I think users are unlikely to measure in the field before they know they have a problem. And once they have a problem they will be unlikely to wait while sufficient field data is collected to wait to fix it.

To me, adding field data is you moving to a more advanced level of managing your CLS. But most users are likely to start without it. So the progression from trying to identify it first yourself, seeing how painful that is, and then measuring in field as a solution to that seems a natural order.

Those users who already are collecting the field attribution data, have likely already been through this, or know to jump straight to this data as they know they are collecting it.

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb: Fixed.

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb: > The preceding section on post-load CLS lists all the different ways to diagnose post-load issues, but it doesn't make clear which tool(s) to use for the job or when. And given that this section follows it, a reader wouldn't yet know what they're supposed to be doing in these tools to find the post-load issues.

I've change this a bit more to remove the Performance Panel tracing section (I've moved it to the previous Load CLS issue section as believe it's more useful there) and also added a section on common post-load CLS causes.

I still believe the RUM section follows more naturally in the next section, as a solution to the "guesswork" nature of this. Especially as, like I said above, anyone that's already collecting this in RUM likely knows to go there (at least until this attribution data becomes a lot more common in RUM solutions).

src/site/content/en/blog/optimize-cls/index.md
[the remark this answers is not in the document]
tunetheweb: Going to go with this. It follows [previous guidance](https://web.dev/debug-layout-shifts/) and is presented as a secondary option, and the snippet is not even included in this page. I think that makes the recommendation of the Extension clear.

PR #9833, opened 2023-03-29
fix: space missing between two HTML tags on the aside component

### LINKED ISSUE #9774: Space missing between consecutive `a` and `em` elements
Under https://web.dev/better-responsiveness-metric/#interaction-types, the aside reads:

> The [current metric](https://web.dev/inp/)_only_ considers keyboard, mouse, and touch events, and does _not_ consider hover or scrolling when calculating INP.

There's a space missing between the `current metric` link and the emphasized `only` text.

However, there is a space in the [source code](https://github.com/GoogleChrome/web.dev/blob/main/src/site/content/en/blog/better-responsiveness-metric/index.md?plain=1):

```md
The [current metric](/inp/) _only_ considers keyboard, mouse, and touch events, and does _not_ consider hover or scrolling when calculating INP.
```

So it's not clear if this is a bug in the way the markdown gets rendered to HTML or if it's more of a presentation issue in CSS.

**Expected:**

image

**Actual:**

image

mamieorine:
Fixes #9774 

The current minifyHTML transformation cause the bug of missing space between two HTML tags on the aside on the built page: https://web.dev/better-responsiveness-metric/#interaction-types

`The <a href=/inp/>current metric</a><em>only</em> considers keyboard, mouse, and touch events, and does <em>not</em> consider hover or scrolling when calculating INP.</div>`

> The [current metric](https://web.dev/inp/)only considers keyboard, mouse, and touch events, and does not consider hover or scrolling when calculating INP.

We can spot that there is a space missing between the `current metric` link and the emphasized `only` text.

---

- Replaces the current minifyHTML transformation with a shared minifyHTML from webdev-infra as we use on d.c.c. 

The result after running `npm run build` with `ELEVENTY_ENV=prod` is shown below:

`The <a href=/inp/>current metric</a> <em>only</em> considers keyboard, mouse, and touch events, and does <em>not</em> consider hover or scrolling when calculating INP.</div>`

> The [current metric](https://web.dev/inp/) only considers keyboard, mouse, and touch events, and does not consider hover or scrolling when calculating INP.


mamieorine:
> 

The diff result of the build output of this page: https://web.dev/better-responsiveness-metric/ 

- Old minifyHTML - 74,071 bytes
- New minifyHTML  - 73,914 bytes

I ran `diff -wEB old.html new.html` and I found that the differences between the 2 files are the indentation, whitespace, and blank line. Many places contain extra whitespace in the output file with old minifyHTML.

Screenshot 2023-03-30 at 09 43 55

Old minifyHTML result shown in the top section, while the old one shown in the bottom section.

Apart from the spacing, I also spotted some differences of the SVG structure, as shown below:

Old minifyHTML `<svg viewbox="0 0 24 24" height=24 width=24 xmlns=http://www.w3.org/2000/svg>`
New minifyHTML `<svg width=24 height=24 viewBox="0 0 24 24">`

I tested on both structures, both can perfectly render the SVG.

PR #9845, opened 2023-03-30
CWV Metric doc tweaks

tunetheweb:

- Moves the BrowserCompat widgets to the top of the docs similar to other APIs. Previously they were under the "How to measure in JavaScript" section, but the compat applies to the whole doc and not just the JavaScript APIs.
- Adds code to measure individual layout shifts to CLS doc, as an explainer into how the metric is calculated.
- Adds code to measure individual interactions to INP doc, as an explainer into how the metric is calculated.
- Added a section on LCP heuristics for non-contentful exclusions (shared with @anniesullie who's happy with it).
- Removed a lot of the optimizing advice and refer to guide instead. No point in duplicating it and risking it getting out of sync.


src/site/content/en/metrics/inp/index.md
```diff
@@ -176,11 +178,24 @@ The best way to measure your website's INP is by gathering metrics from actual u
...
+      console.log('Interaction:', delay, entry);
+    }
+  }
+}).observe({type: 'event', buffered: true});
+```
+
+### Measure INP In JavaScript
 
-Writing your own [`PerformanceObserver`](https://developer.mozilla.org/docs/Web/API/PerformanceObserver) to measure INP can be difficult. To measure INP in JavaScript, it's advised that you use the [`web-vitals` JavaScript library](https://github.com/GoogleChrome/web-vitals), which exports an `onINP` function to do this work for you. You can then get a page's INP by passing a function to the `onINP` method:
+To measure INP in JavaScript, it's advised that you use the [`web-vitals` JavaScript library](https://github.com/GoogleChrome/web-vitals), which exports an `onINP` function to handle all the aggregation of the individual interactions. You can then get a page's INP by passing a function to the `onINP` method:
```
rviscomi, line 198:
```suggestion
To measure INP in JavaScript, it's advised that you use the [`web-vitals` JavaScript library](https://github.com/GoogleChrome/web-vitals), which exports an `onINP` function to handle the aggregation of each individual interaction. You can then get a page's INP by passing a function to the `onINP` method:
```

src/site/content/en/metrics/cls/index.md
```diff
@@ -332,16 +334,23 @@ available in the following tools:
...
+
+### Measure CLS in JavaScript
 
-To measure CLS in JavaScript, you can use the [Layout Instability
-API](https://github.com/WICG/layout-instability). The following example shows
-how to create a
-[`PerformanceObserver`](https://developer.mozilla.org/docs/Web/API/PerformanceObserver)
-that listens for unexpected `layout-shift` entries, groups them into sessions,
-and logs the maximum session value any time it changes.
+To measure CLS in JavaScript, you need to group these unexpected `layout-shift` entries into sessions, and calculate the maximum session value. The following example shows how to log CLS to the console any time it changes.
```
rviscomi, line 353: Should we recommend web-vitals.js here instead, similar to the INP doc?

cc @malchata

tunetheweb: We do further down (after a lot of caveats!) but yes I think there’s an argument that, like INP, the complexity since adding windowing has outgrown the usefulness of having an embedded snippet in the docs and we should recommend web-vitals.js for most users (and perhaps link to the source for more advanced users who DO want to implant themselves).

if we do that then I wonder if it should be super light on details (like INP is) or pretty detailed (like CLS currently is) or somewhere in the middle?

My gut feel is somewhere in the middle. INP feels too light and we’re just asking people to take out word for it because “it’s complicated”. CLS however maybe has the opposite problem and perhaps tries to explain too much (plus we’d need to add extra explanation to replace what is currently in the code we’d be removing, for the rest of that to still make sense).

WDYT?

rviscomi: Sounds good in theory but not sure I understand what the change would actually look like. Curious to also get Jeremy's feedback before picking a direction.

malchata: My take is that if web-vitals.js makes it easier, we should encourage it over homegrown solutions. I'd drop the perf observer code snippet, IMO and recommend what we've already done in-house.

malchata (changes requested):

src/site/content/en/metrics/inp/index.md
```diff
@@ -13,6 +13,8 @@ tags:
   - web-vitals
 ---
 
+{% BrowserCompat 'api.PerformanceEventTiming.interactionId' %}
```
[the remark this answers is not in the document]
tunetheweb: My thinking here was that the whole metric is only supported in Chromium-based browsers. At present we leave this very late in the page to the "Measure INP In JavaScript" section, but that seems very late to me and suggests it's only the JavaScript API that isn't supported - but it's not just the JavaScript API - it's not reported by DevTools or any other way.

For most other posts on web.dev we put the Browser Compat widget near the top of the post. Maybe I've gone too far putting it right to the top here, but do think having it WAY down at the bottom, by the JavaScript API, is very late.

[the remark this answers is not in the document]
tunetheweb: As discussed offline your concern was more showing an EventTiming compat widget (as there is no INP one) so added a comment in https://github.com/GoogleChrome/web.dev/pull/9845/commits/091c9de6cc3ea40e8b0d6bf15d81e42650177213 to make it clearer.

src/site/content/en/metrics/inp/index.md
```diff
@@ -176,11 +178,24 @@ The best way to measure your website's INP is by gathering metrics from actual u
 - [Lighthouse User Flows](/lighthouse-user-flows/).
 - [Web Vitals extension for Chrome](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma).
 
-### Measure INP In JavaScript
+### Measure interactions In JavaScript
 
-{% BrowserCompat 'api.PerformanceEventTiming.interactionId' %}
+INP is calculated based on interation events. The following example shows how to create a [`PerformanceObserver`](https://developer.mozilla.org/docs/Web/API/PerformanceObserver) to log interaction entries to the console:
```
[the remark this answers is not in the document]
tunetheweb: Corrected the `input delay` issue in https://github.com/GoogleChrome/web.dev/pull/9845/commits/091c9de6cc3ea40e8b0d6bf15d81e42650177213

As discussed we may move this snippet to somewhere else in future, but as a quick fix we'll add it here.

src/site/content/en/metrics/cls/index.md
tunetheweb: And it's gone in https://github.com/GoogleChrome/web.dev/pull/9845/commits/ef2f05dee7f552af4349175dd4005f97b56d216a

src/site/content/en/metrics/inp/index.md
```diff
@@ -13,8 +13,10 @@ tags:
...
+
 {% Aside %}
-Interaction to Next Paint (INP) is an experimental metric that assesses [responsiveness](/user-centric-performance-metrics/#types-of-metrics). When an interaction causes a page to become unresponsive, that is a poor user experience. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below. A low INP means the page was consistently able to respond quickly to all&mdash;or the vast majority&mdash;of user interactions.
+Interaction to Next Paint (INP) is an experimental metric that uses data from the [`Event Timing API`](https://developer.mozilla.org/docs/Web/API/PerformanceEventTiming) API, that assesses [responsiveness](/user-centric-performance-metrics/#types-of-metrics). When an interaction causes a page to become unresponsive, that is a poor user experience. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below. A low INP means the page was consistently able to respond quickly to all&mdash;or the vast majority&mdash;of user interactions.
```
malchata, line 19:
```suggestion
Interaction to Next Paint (INP) is an experimental metric that uses data from the [`Event Timing API`](https://developer.mozilla.org/docs/Web/API/PerformanceEventTiming) API. It assesses [responsiveness](/user-centric-performance-metrics/#types-of-metrics). When an interaction causes a page to become unresponsive, that is a poor user experience. INP observes the latency of all interactions a user has made with the page, and reports a single value which all (or nearly all) interactions were below. A low INP means the page was consistently able to respond quickly to all&mdash;or the vast majority&mdash;of user interactions.
```

malchata (approved):
Minor nit, otherwise LGTM.

Issue #9792, opened 2023-03-20
Unable to close the local navigation


ionStici:
**Describe the bug**
The local navigation of the learning courses reopens each time you click the 'Close' icon.

**To Reproduce**
1. Go to a learning course: '[https://web.dev/learn/html/](https://web.dev/learn/html/)'
2. Click on the 'hamburger icon' to open the local navigation
3. Try closing the local navigation by clicking the 'Close' icon

**Expected behavior**
Expected local navigation bar to close

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/109459805/226284882-0f61fe46-245d-4af9-a19f-0e8e04970f40.png)

Issue #9639, opened 2023-02-23
<web-navigation-drawer> needs two interactions to be closed when not using its close button


TonySpegel:
**Describe the bug**
Whenever the mobile side-nav is visible a second interaction with the background element is needed to close it

**To Reproduce**
1. Go to [https://web.dev/learn/](https://web.dev/learn/) or almost any other page on web.dev
2. Set your viewport to be smaller than `50em` 
3. Open the side-nav by pressing the hamburger menu button
4. Wait for the side-nav to open, then tap/click the background

**Expected behavior**
I expect to close the side-nav by using only one tap/click on the page body

**Screenshots**
![side nave opened](https://user-images.githubusercontent.com/1145514/220795673-da45431d-3d78-4a36-bed2-9a9557fdb258.png)
Fig. 1: side nav is opened, no interaction with page body

![image](https://user-images.githubusercontent.com/1145514/220795850-2782b4c5-6a79-434b-bff0-510e0222a985.png)
Fig. 2: side-nav is opened, first interaction with page body - the background overlay has disappeared. Another interaction is needed to close the side-nav

rahulrao0209:
@AaronForinton I tried cloning and running the project locally to identify a fix for this issue, however, I am running into the following problem with the shortcode being used in the nunjucks templates. Tried ignoring all content using .env file but it didn't seem to help. Let me know if anything can be done to run the project locally.
Thanks! 
![image](https://user-images.githubusercontent.com/63695122/222884916-b7e29d3b-56af-46ce-9dbd-7dd9b75d133f.png)

matthiasrohmer:
@rahulrao0209, is there more to your error output? If it has been a while since you tried this `npm install` might already help, as the webdev-infra package updates regularly.

rahulrao0209:
@matthiasrohmer, the same issue persists after `npm install`. I had tried `npm install` and `npm ci` before as well.
Attaching a similar screenshot for reference.

![image](https://user-images.githubusercontent.com/63695122/225514518-743f7235-3b84-4513-8f02-277d11ef0b52.png)

matthiasrohmer:
@rahulrao0209, sorry didn't notice it in your first screenshot. The project unfortunately doesn't support Windows. As translations are discovered by path (which works differently on Windows) the translation process fails.

You can try running the project using the [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/install) though.

rahulrao0209:
[...]

PR #9848, opened 2023-03-31
fix: the navigation drawer cannot be closed

mamieorine:
Fixes https://github.com/GoogleChrome/web.dev/issues/9639, fixes https://github.com/GoogleChrome/web.dev/issues/9792

For https://github.com/GoogleChrome/web.dev/issues/9639, the navigation drawer on the mobile view needs two clicks on the outside of the drawer container to close itself. It should close when clicking just once, which is the same behaviour as when clicking on the close button.

For https://github.com/GoogleChrome/web.dev/issues/9792, the local navigation of the learning courses reopens each time when click the 'Close' icon.

It happens because the focus event is still on the drawer element so it cannot be completely closed.

Thus, the change proposed in this pull request is removing the focus event from the navigation drawer after clicking the close button or clicking outside the drawer container, using `HTMLElement.blur()`.

### DIFF
--- a/src/lib/components/NavigationDrawer/index.js
+++ b/src/lib/components/NavigationDrawer/index.js
@@ -83,7 +83,7 @@ export class NavigationDrawer extends BaseStateElement {
   addEventListeners() {
     this.drawerContainer.addEventListener('click', this.onBlockClicks);
     this.closeBtn.addEventListener('click', closeNavigationDrawer);
-    this.addEventListener('click', closeNavigationDrawer);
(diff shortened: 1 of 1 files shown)


mamieorine:
I tested it on Safari and Chrome on my mobile device (iOS).

https://user-images.githubusercontent.com/45324326/229210620-36a8baca-2303-44bc-b5e2-a908abc155b4.MP4

matthiasrohmer (approved):

matthiasrohmer (approved):

PR #9857, opened 2023-03-31
`/import-maps-in-all-modern-browsers/`: Fix examples by replacing `lodash` with `d3-array`.

### LINKED ISSUE #9856: The example in the import maps blog post is broken.
The example snippets on <https://web.dev/import-maps-in-all-modern-browsers/> don't work when used together. The `lodash` script referenced in the import map has no export statements, so the import statement fails.

To reproduce, serve this file somewhere:
```html
<!DOCTYPE html>
<script type="importmap">
  {
    "imports": {
      "lodash": "https://unpkg.com/lodash@4.17.21/lodash.js"
    }
  }
</script>
<script type="module">
  import _ from 'lodash';

  console.log(_.map([1, 2, 3], (n) => n * 2));
</script>
```

This error will be shown in the console:

```
Uncaught SyntaxError: The requested module 'lodash' does not provide an export named 'default' (at (index):10:10)
```

(I have a PR ready to go for this that switches the example to `d3-array` instead.)

bicknellr:
The snippets in this article don't work together because the URL for `lodash` used points to a file that's not a module and contains no exports. Simply replacing `lodash` with `lodash-es` causes the example to work, but `lodash`'s main file transitively imports over 600 internal modules - which seems like a bad example to set. Even importing `lodash-es/map.js` directly still transitively imports over 100 modules. To both fix the example and avoid loading an unnecessarily large dependency tree for such a simple example, I've replaced `lodash` with `browser-fs-access`, which is a module and has no transitive imports.

Fixes #9856

- Replace use of `lodash` in the `/import-maps-in-all-modern-browsers/` article with `browser-fs-access`.

Edit: `d3-array` was replaced with `browser-fs-access`.

### DIFF
--- a/src/site/content/en/blog/import-maps-in-all-modern-browsers/index.md
+++ b/src/site/content/en/blog/import-maps-in-all-modern-browsers/index.md
@@ -31,7 +31,7 @@ corresponding URLs. For example:
 <script type="importmap">
   {
     "imports": {
-      "lodash": "https://unpkg.com/lodash@4.17.21/lodash.js"
+      "d3-array/": "https://unpkg.com/d3-array@3.2.3/src/"
     }
   }
 </script>
@@ -44,9 +44,9 @@ include the lodash library in your code. Note that the `import` keyword is only 
 
 ```html
 <script type="module">
-  import _ from 'lodash';
+  import map from 'd3-array/map.js';
 
-  console.log(_.map([1, 2, 3], (n) => n * 2));
+  console.log(map([1, 2, 3], (n) => n * 2));
 </script>
 ```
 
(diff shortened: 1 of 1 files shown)


tomayac:
Thanks so much for the PR. Our readers might know https://unpkg.com/browser-fs-access@0.33.0/dist/index.modern.js, it was the topic of several posts. You could import `fileOpen` from it?

bicknellr:
Updated; I also fixed the nearby prose referring to lodash that I missed before.

tomayac:
LGTM, thanks for the fix. Over to @rachelandrew for final check and merging.

rachelandrew (approved):

PR #9888, opened 2023-04-12
use-imagemin-to-compress-images: normalize plugin case

jzern:
Imagemin-svgo -> imagemin-svgo


jzern:
@khempenius @Meggin please have a look. Just a small cosmetic change.

AaronForinton (approved):

PR #9889, opened 2023-04-12
use-imagemin-to-compress-images: add imagemin-webp to lossless

jzern:
imagemin-webp supports both lossy and lossless encodes since version
2.0.0


jzern:
@khempenius @Meggin. The structure of the table confused me at first. I had to look at the plugin's site to determine lossless webp was supported.

jzern:
I might also put SVG under the lossless column rather than lossy unless that plugin has some preprocessing options.

AaronForinton (approved):

PR #9899, opened 2023-04-12
[Fetch API] Network status error code sample update

### LINKED ISSUE #9519: Rework code sample in "Implement error handling when using the Fetch API"

https://web.dev/fetch-api-error-handling/#when-the-network-status-code-represents-an-error

The code sample here could be improved to avoid introducing potentially unfamiliar concepts to readers (e.g., the optional chaining operator), to keep the code short, and to address the problem more idiomatically:

```js
let response;

try {
  response = await fetch('https://httpbin.org/status/429');
} catch (error) {
  console.log('There was an error', error);
}

// Uses the 'optional chaining' operator
if (response?.ok) {
  console.log('Use the response here!');
} else {
  console.log(`HTTP Response Code: ${response?.status}`)
}
```

Rather than conditionally assigning a response, we can throw an error if `Response.ok` is falsy:

```js
try {
  const response = await fetch('https://httpbin.org/status/429');
  // network error in the 4xx–5xx range
  if (!response.ok) {
    throw new Error(`${response.status} ${response.statusText}`);
  }
  // use response here if we didn't throw above
  doSomethingWith(response);
} catch (error) {
  console.log(error);
}
```

Happy to contribute an edit for this upon approval.

AaronForinton:
Fixes #9519

### DIFF
--- a/src/site/content/en/web-dev-basics-one/fetch-api-error-handling/index.md
+++ b/src/site/content/en/web-dev-basics-one/fetch-api-error-handling/index.md
@@ -152,19 +152,16 @@ To check that the HTTP status code was successful, you can use any of the follow
 -  Use any other metadata, such as [`Response.headers`](https://developer.mozilla.org/docs/Web/API/Response/headers), to assess whether the response was successful.
 
 ```js
-let response;
-
 try {
-  response = await fetch('https://httpbin.org/status/429');
+  const response = await fetch('https://httpbin.org/status/429');
+  // network error in the 4xx–5xx range
+  if (!response.ok) {
+    throw new Error(`${response.status} ${response.statusText}`);
+  }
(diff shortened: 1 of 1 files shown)


AaronForinton:
cc @AleksandrHovhannisyan

jakearchibald (changes requested):
The content prior to the code example needs updating, as it refers to elements in the old code example that aren't in the new example.

Also, I think it's worth stating why the fetch API doesn't throw in these cases: you still have a response.

For example, a 404 response can still be an image, and an `` will render it. A 403 response can still parse as JSON, and provide further details about the error than can be included in the status text etc etc.

AaronForinton:
@jakearchibald Do you have any time to contribute a few edit suggestions?

rachelandrew (approved):

PR #9920, opened 2023-04-18
Only send search requests for queries longer than n characters

ShayPunter:
Fixes [#76](https://github.com/GoogleChrome/webdev-infra/issues/76)

- Send requests to Algolia if search query is 4 or more characters
- - number calculated based on hits from popular search queries and length of keyword
- Add a message to search results to inform the user that they need to enter 4 characters for search suggestions

n has been worked out in the following manner:
1. Sort the keyword spreadsheet by count and average the length of the top 10 keywords (output: 5 (rounded))
2. Sort the keyword spreadsheet by nbhits, exclude queries with just one char, average the length of the top 10 keywords (output: 2 (rounded))
3. Average both the count and nbhits keywords, which produced 4 (rounded)
4. Take into account that there are a lot of 3 letter abbreviations and then we end up with 3 as our final number
5. 
## How to test
To test this change, spin up a local development environment, checkout `76_only_send_seach_queries_that_are_longer_than_n_characters`, build and run the application and then go to the search bar and begin searching, if two characters have been searched, it will display a message saying to type at least 3 characters for search results to return


jimjiminyjimjim (commented):
lgtm

PR #9996, opened 2023-05-04
Update workbox usage statistics

aleksandr-kiliushin:

- Update workbox usage statistics.
- Minor fixes in code formatting.


malchata (approved):
This is a good catch. Ideally, we'd want to dynamically pull this data so it self-updates, but there's no good mechanism for doing so currently. So will accept and merge this so that it's up to date (until the next web almanac, anyway).

PR #9998, opened 2023-05-04
feat: replace pre-push hook with lintstaged pre-commit hook

### LINKED ISSUE #80: Notes inside of a code block
"And import it within the event listener that fires when the button is pressed:" is in between two bits of code.

http://crobdodson.c.googlers.com:8080/path/fast/reduce-javascript-payloads-with-code-splitting/codelab-code-splitting?tenant=web

Blazzike:
Fixes #80

- Uses lintstaged package to run linting on modified files only 
- Lints on pre-commit rather than pre-push


.lintstagedrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "*.scss": "npm run lint:scss",
+  "*.md": "npm run lint:md",
+  "*.js": "npm run lint:js"
```
matthiasrohmer, line 4: Haven't tested it yet, but from looking at the commands I think this is not yet only linting files that changes but if one js file changed for example still the full JS codebase would be linted.

Blazzike: It actually does magically work correctly! lint-staged automatically passes each filename as a parameter to the commands above. :-)

matthiasrohmer (approved):

PR #10010, opened 2023-05-08
Fix redirects without slash

tunetheweb:
Fixes https://twitter.com/programmingart/status/1655483178147094529?s=20

- Allow slash-less redirects to work. For example, https://web.dev/priority-hints will now work as well as https://web.dev/priority-hints/ and both will redirect to https://web.dev/fetch-priority/

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -202,7 +202,7 @@ redirects:
   - source: /progressive-web-apps/#installable
     destination: /progressive-web-apps/#make-it-installable
 
-  - source: /push-notifications/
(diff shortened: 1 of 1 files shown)


matthiasrohmer:
Yeah unfortunately Netlify doesn't care about the Firebase redirects.yaml. Did a Firebase staging deploy at https://web-dev-production-1--staging-9kwej4js.web.app/priority-hints - seems to do the trick 🙂  Thanks for fixing this!

matthiasrohmer (approved):

PR #10035, opened 2023-05-11
Links and banners

rachelandrew:
Post-IO updates.


matthiasrohmer (approved):
Added `https://developer.mozilla.org/en-US/blog/baseline-unified-view-stable-web-features/` with `git commit --no-verify`, will create a linter fix on a different branch to allow `en-US/blog` links.

PR #10056, opened 2023-05-14
Update index.md

Yatin-kathuria:
Update the Wrong reference in paragraph

Fixes #SOME_ISSUE_NUMBER

-
-
-


malchata:
The content as-is is correct. If you specify an MP4 first, then WebM second, the browser will use the MP4 source _even if_ it supports WebM.

PR #10075, opened 2023-05-19
Updates INP metric doc.

malchata:

- Updates INP metric doc and aligns some copy with it and the Optimize INP guide.

cc: @mmocny @anniesullie


src/site/content/en/metrics/inp/index.md
```diff
@@ -54,7 +54,7 @@ INP is a metric that assesses a page's overall responsiveness to user interactio
 A note on how INP is calculated
 {% endDetailsSummary %}
 
-As stated above, INP is calculated by observing all the interactions made with a page. The chosen value is then a [percentile](https://en.wikipedia.org/wiki/Percentile) of those interactions. A formula is then used to choose a high percentile value of those interactions. For pages with few interactions, the interaction with the worst latency (the 100th percentile) is chosen. For pages with many interactions, the 99th or 98th percentile is chosen.
+As stated above, INP is calculated by observing all the interactions made with a page. The resulting value is determined by a formula that calculates a high [percentile](https://en.wikipedia.org/wiki/Percentile) value of all the observed interactions. The interaction with the worst latency is chosen, dropping the worst of every 50 interactions.
```
anniesullie, line 57: Would be awesome if we just didn't say "percentile" at all. cc @mmocny for thoughts.

anniesullie (commented):
LGTM with one nit

src/site/content/en/metrics/inp/index.md
mmocny: Agreed!

I think if you just drop the line completely and leave the last one it still works.  However, maybe we should expand what we mean by dropping. Perhaps something like:

"The interaction with the worst latency is reported as INP, however, one worst interaction is ignored for every 50 interactions.  The vast majority of page experiences do not have over 50 interactions."

src/site/content/en/metrics/inp/index.md
```diff
@@ -54,7 +54,7 @@ INP is a metric that assesses a page's overall responsiveness to user interactio
 A note on how INP is calculated
 {% endDetailsSummary %}
 
-As stated above, INP is calculated by observing all the interactions made with a page. The chosen value is then a [percentile](https://en.wikipedia.org/wiki/Percentile) of those interactions. A formula is then used to choose a high percentile value of those interactions. For pages with few interactions, the interaction with the worst latency (the 100th percentile) is chosen. For pages with many interactions, the 99th or 98th percentile is chosen.
+As stated above, INP is calculated by observing all the interactions made with a page. The interaction with the worst latency is reported as INP, however, one worst interaction is ignored for every 50 interactions. The vast majority of page experiences do not have over 50 interactions.
```
tunetheweb, line 57: I still think we should explain this more. WDYT about this:

```suggestion
As stated above, INP is calculated by observing all the interactions made with a page. For most sites the interaction with the worst latency is reported as INP. However, for pages with large numbers of interactions, random hiccups can result in an unusually high interaction on an otherwise responsive site. The more interactions, the more likely this is to happen. To counter this, and give a better measure of the actual responsiveness for those types of pages, we ignore one highest interaction for every 50 interactions. The vast majority of page experiences do not have over 50 interactions so will report the worst interaction. The 75th percentile of all the page views is then reported as usual, which further removes outliers to give a value that the vast majority of users experience or better.
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -158,56 +157,22 @@ INP can be measured both in [the field](/lab-and-field-data-differences/#field-d
...
-- [`web-vitals` JavaScript library](https://github.com/GoogleChrome/web-vitals).
-
-### Lab tools
+### In the field
 
-- Lighthouse Panel in DevTools, available in "Timespan Mode".
-- [Lighthouse npm module](https://www.npmjs.com/package/lighthouse).
-- [Lighthouse User Flows](/lighthouse-user-flows/).
-- [Web Vitals extension for Chrome](https://chrome.google.com/webstore/detail/web-vitals/ahfhijdlegdabablpippeagghigmibma).
+Ideally, your journey in optimizing INP will start with field data. At its best, field data from a Real User Monitoring (RUM) provider will give you not only a page's INP value, but also contextual data that highlights what specific interaction was responsible for the INP value itself, whether the interaction occurred during or after page load, the type of interaction (click, keypress, or tap), and other valuable information.
```
tunetheweb, line 162: IMHO web-vitals is one of the best for this (and the pioneer!) so let's drop the word "provider":

```suggestion
Ideally, your journey in optimizing INP will start with field data. At its best, field data from Real User Monitoring (RUM) will give you not only a page's INP value, but also contextual data that highlights what specific interaction was responsible for the INP value itself, whether the interaction occurred during or after page load, the type of interaction (click, keypress, or tap), and other valuable information.
```

src/site/content/en/metrics/inp/index.md
```diff
@@ -158,56 +157,22 @@ INP can be measured both in [the field](/lab-and-field-data-differences/#field-d
...
+{% Aside 'objective' %}
+**Read to learn more:**&nbsp;[Find slow interactions in the field](/find-slow-interactions-in-the-field/).
+{% endAside %}
 
-onINP(({value}) => {
-  // Log the value to the console, or send it to your analytics provider.
-  console.log(value);
-});
-```
+If you're not relying on a RUM provider to get field data, the [INP field data guide](/find-slow-interactions-in-the-field/) advises [using the Chrome User Experience Report (CrUX) via PageSpeed Insights](/find-slow-interactions-in-the-field/#get-field-data-quickly-with-crux) to help fill in the gaps. CrUX is the official dataset of the Core Web Vitals program and provides a high-level summary of metrics for millions of websites, including INP. However, CrUX often does not provide the contextual data you'd get from a RUM provider to help you to analyze issues. Because of this, we still recommend that sites use a RUM provider when possible, or implement their own RUM solution to supplement what is available in CrUX.
```
tunetheweb, line 168: I'm not loving this framing.

I think we should say:
- CrUX is best for the score. To me, as the official dataset of the CWV initiative that should be the starting point. It's the most accurate number IMHO given the limitations on Web APIs used by RUM and given that NONE of the RUM providers measure INP throughout the full page lifecycle (many stop after first interaction).
- RUM allows you to get more detailed info than CrUX can provide - including attribution details (e.g. element clicked for the back INP).
- Lab should then be used to replicate (ideally with RUM as a clue as to which pages, and while elements, are causing the slowness).

tunetheweb (changes requested):

tunetheweb (changes requested):

PR #10077, opened 2023-05-19
docs: fix misleading error in code snippet

idorenyinudoh:
Change proposed in this pull request:

- fix misleading error in code snippet

not sure what the `<strong>` element was intended for


malchata (approved):
Good catch. This was likely because this article existed on WebFundamentals, and the intent was to highlight this in bold type, but that doesn't work for our current syntax highlighting solution.

PR #10118, opened 2023-06-01
Fix link to scroll-driven animations article

bramus:
It’s hosted on DCC, not web.dev


rachelandrew (approved):
Thank you—I was darting between the beta post and this and obviously forgot which site I was on!

PR #10121, opened 2023-06-02
updating make it installable code lab to remove SW as a requirement

tropicadri:
updating make it installable code lab to remove SW as a requirement for installation.


malchata (approved):
LGTM!

PR #10123, opened 2023-06-02
Improve wording in lazy loading article

tunetheweb:
Noticed some language would could potentially be misread so tweaked it a bit. Also removed some old info that's no longer necessary and updated others.

I also removed some new lines (so review ignoring whitespace changes in GitHub to make it easier).

### DIFF
--- a/src/site/content/en/fast/browser-level-image-lazy-loading/index.md
+++ b/src/site/content/en/fast/browser-level-image-lazy-loading/index.md
@@ -8,7 +8,7 @@ authors:
   - mathiasbynens
   - tunetheweb
 date: 2019-08-06
-updated: 2022-10-18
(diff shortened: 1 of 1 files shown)


src/site/content/en/fast/browser-level-image-lazy-loading/index.md
```diff
@@ -8,7 +8,7 @@ authors:
   - mathiasbynens
   - tunetheweb
 date: 2019-08-06
-updated: 2022-10-18
+updated: 2023-06-02
```
malchata, line 11:
```suggestion
updated: 2023-06-05
```

malchata (approved):
Small change to update the `updated` field, otherwise LGTM.

PR #10139, opened 2023-06-10
Learn HTML, JS explanation

junaga:
Shouldn't it be like this?


malchata:
@estelle I'm not sure if what's highlighted here is correct (regardless of the proposed change)—the browser has a [preload scanner](https://web.dev/preload-scanner/) that can discover resources in raw markup payloads [even if rendering of the page and parsing of the DOM is blocked](https://web.dev/preload-scanner/#how-to-tell-when-the-preload-scanner-is-working). This may need an update.

junaga:
please do fix this guys <3

estelle (approved):

malchata (approved):

PR #10152, opened 2023-06-11
Fix syntax error in origin-private-file-system example

Mr0grog:
This just adds a missing quote to close a string in one of the code examples in the Origin Private File System blog post.

I just happened to notice this one while reading the post, but it seems like you could probably add a Remark plugin to flag these errors automatically. If it would be helpful, I’d be happy to take a look at setting up [remark-lint-code](https://github.com/Qard/remark-lint-code) and [remark-lint-code-eslint](https://github.com/Qard/remark-lint-code-eslint) for that (or something custom if you don’t like those or they don’t turn out to work well).


Mr0grog:
Still seems relevant?

rachelandrew (approved):

PR #10177, opened 2023-06-20
Learn HTML, typo

junaga:
`links` is a JS _property_ of the `document` object

### DIFF
--- a/src/site/content/en/learn/html/links/index.md
+++ b/src/site/content/en/learn/html/links/index.md
@@ -181,7 +181,7 @@ or `<label>` either. While the HTML page will still render, nesting focusable an
 
 ## Links and JavaScript
 
-The `links` attribute returns an `HTMLCollection` matching `a` and `area` elements that have an `href` attribute.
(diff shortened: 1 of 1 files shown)


src/site/content/en/learn/html/links/index.md
```diff
@@ -181,7 +181,7 @@ or `<label>` either. While the HTML page will still render, nesting focusable an
 
 ## Links and JavaScript
 
-The `links` attribute returns an `HTMLCollection` matching `a` and `area` elements that have an `href` attribute.
+The `links` property returns an `HTMLCollection` matching `a` and `area` elements that have an `href` attribute.
```
malchata, line 184: Does this seem like the correct nomenclature, @rachelandrew? I think it is from a JavaScript perspective, but not entirely sure since there's overlap with HTML concepts here cc: @estelle

rachelandrew: yeah that looks correct.

malchata (approved):

PR #10190, opened 2023-06-26
Add frontmatter fields to the export

devnook:


src/site/_includes/components/Export.js
```diff
@@ -64,6 +64,18 @@ async function Export() {
...
   }
 
+  const authors = JSON.stringify(page?.template?.frontMatter?.data.authors);
+
+  const frontMatter = `project_path: /_project.yaml
+book_path: /_book.yaml
+
+{% import "/_macros.html" as macros %}
+
+# ${page?.template?.frontMatter?.data.title}
```
matthiasrohmer, line 74: Does this get the computed title? Haven't checked if there are cases where we use one.

src/site/_includes/components/Export.js
```diff
@@ -64,6 +64,18 @@ async function Export() {
     return '';
   }
 
+  const authors = JSON.stringify(page?.template?.frontMatter?.data.authors);
+
+  const frontMatter = `project_path: /_project.yaml
+book_path: /_book.yaml
+
```
matthiasrohmer, line 71: There is also `author_name`, used in dynamic content cards.

devnook: true, we can add it

src/site/_includes/components/Export.js
devnook: It does not. We don't use computed titles for posts though.

PR #10191, opened 2023-06-26
Update window controls overlay incorrect function name

jadjoubran:
Change `handleOrientationChange ` to `handleDisplayModeChange`

Fixes incorrect function name


jadjoubran:
@tomayac

tomayac (approved):
LGTM

Thank you, good catch.

tomayac:
Fixed on the new backend.

PR #10194, opened 2023-06-26
Add script to create _playlist.yaml files for courses

devnook:


.eleventy.js
```diff
@@ -279,6 +280,45 @@ module.exports = function (config) {
     }
   }
 
+  if (process.env.ALT_BUILD) {
+    config.addCollection('courses', (collections) => {
+      const coursesGlob = '**/en/learn/*/*.md';
+      collections.getFilteredByGlob(coursesGlob).forEach(async (course) => {
```
matthiasrohmer, line 286: This could lead to unfulfiled promises, respectively unwritten files as it's not awaited higher up I think.

devnook: I added a warning comment

matthiasrohmer (approved):

PR #10196, opened 2023-06-26
Update index.md

theneekz:
I think the statement stands on it's own without needing to reference 'world' ranking

- removes "even in first-world countries" from [this section](https://web.dev/optimizing-content-efficiency-javascript-startup-optimization/#network)


malchata (approved):
Good catch. Will merge ASAP when workflows finish.

PR #10273, opened 2023-07-23
Fix preload as=style

rviscomi:
The correct `as` attribute value for preloading stylesheets should be `style` rather than `stylesheet`.


tunetheweb (approved):
Change is correct. @rachelandrew / @devnook  not sure when we can apply this correction?

PR #10274, opened 2023-07-23
Update index.md

LordMoMA:
Update browser support based on https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/postTask

Fixes #SOME_ISSUE_NUMBER

-Update browser compatibilities docs based on [https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/postTask](https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/postTask)
-
-


rachelandrew:
Closing as we will use compat data here on the migrated site.

PR #10290, opened 2023-07-26
Fix cells’ order.

SpikyC:


SpikyC:
@jecfish

SpikyC:
Fixed by #10303.

PR #10340, opened 2023-08-16
Speculation rules: match any hash fragment.

jeremyroman:
This expands the scope slightly, but also we're contemplating a change which would make this unnecessary -- but while developer.chrome.com is using this syntax, it's more difficult to tell at a glance how common other uses are.

### DIFF
--- a/src/site/_includes/partials/head.njk
+++ b/src/site/_includes/partials/head.njk
@@ -52,11 +52,11 @@
       "where": {
         "and": [
           {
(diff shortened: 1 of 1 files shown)


jeremyroman:
See also https://github.com/GoogleChrome/developer.chrome.com/pull/7089.

jeremyroman:
@tunetheweb

tunetheweb (approved):
LGTM. I hate that we can't just use `*` :-(

Need @matthiasrohmer to merge though.

PR #10424, opened 2023-09-29
Fix broken RSS feeds

tunetheweb:

- Fixes broken RSS feed for feedly. as reported by several users.
 
It looks like they don't accept `type="text/html"` anymore. The [spec](https://www.ietf.org/rfc/rfc4287.html#section-3.1.1) specifies we should be using `"html"` anyway. Doesn't look like we changed anything here, so maybe something changed at Feedly's end?

FYI, this was originally changed too `text/html` in #5873 but with no explanation why. D.C.C still uses `"html"`

[Prod](https://web.dev/feed.xml) (and recent other staging URLs) has this on Feedly:

image

This [branch's preview](https://deploy-preview-10424--web-dev-staging.netlify.app/feed.xml) has this on Feedly:

image

### DIFF
--- a/src/site/_includes/feed.njk
+++ b/src/site/_includes/feed.njk
@@ -20,7 +20,7 @@
       <link href="{{ absolutePostUrl }}"/>
       <updated>{{ post.date | rssDate }}</updated>
       <id>{{ absolutePostUrl }}</id>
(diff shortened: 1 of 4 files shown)


tomayac:
The `mode=escaped ` thing doesn't really seem defined. Maybe move to `<![CDATA[…]]>` now that you touch the code? Also: `3>` for doing this in the first place.

tunetheweb:
Agreed, but kinda keeping changes to a minimal for... well reasons I'm sure you understand :-)

matthiasrohmer (approved):

PR #10435, opened 2023-10-04
Update firebase-tools

devnook:


matthiasrohmer (approved):
Related https://github.com/firebase/firebase-tools-ui/issues/939.

